PR #7, opened 2024-04-17
Cratify workspace

idky137:
Cratifying zingo-proxy to improve usability and separate functionality.


idky137:
May be better to use the underlying feature branch for review as almost everything has moved so looking at changes may be confusing. The only new code is in zingo-proxyd::grpc, this is a mirror of zingolib::zingo_netutils and has been moved here to reduce dependencies.

idky137:
CI currently does not work as lightwalletd/zebrad need to be launched manually, meaning tests cannot be run in CI. This will be fixed in a separate pr.

AloeareV (approved):

AloeareV (changes requested):
I'd like to keep the single source of zingo-netutils. Other than that, the cratifaction is great!

AloeareV (approved):

PR #15, opened 2024-05-03
Add test framework

idky137:
Adds test code to allow integration tests to be run without a manual launch of zcashd and lightwalletd (required for adding integration tests to CI). Restructures Nym Code and removes feature gate.

- uses Zingolib's RegtestManager to launch Lightwalletd and Zcashd in regtest mode for integration tests.
- uses Zingolib's ClientBuilder to create a test wallet.
- adds proxy launcher for use in testing. Upgrades zingoproxyd using this.
- adds test manager to launch RegtestManager and Zingo-Proxy test server.
- simplifies feature use.
- (moves Nym enhanced CompactTxStreamerServer service RPC code to zingo-rpc::rpc::nymservice)

Closes #11 

Note: CI will be added/fixed in a separate PR.

PR #26, opened 2024-05-24
Add jsonrpc

### LINKED ISSUE #27: Add JsonRPC Client
Add a JsonRPC Client for communication with full node (ZebraD / ZcashD).
Required for the implementation of the LightWalletD service RPCs.
Implement get_lightd_info service RPC to test JsonRPC client.

idky137:
Adds jsonRPC client for communication with full node (zebrad/zcashd) and implements lwd service RPCs. 

Removes Zingo-Proxy's dependency on LightWalletD.
- Add JsonRpcConnector for communication with node along with structs for input / output types.
- Implements lightwallet service RPCs required by zingolib.
- Adds Connection checker to establish connection with node on startup.
- Adds new startup message.
- Adds full block and transaction deserialization implementations required for the blockcache (currently used by get_block_range RPC).
- Adds mempool state monitoring functionality required by the blockcache state engine (currently used by get_mempool_stream RPC).

Closes #27, #29


idky137:
built on top of #15, merge that pr before reviewing here.

zingo-proxyd/src/proxy.rs
```diff
@@ -7,33 +10,36 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
+) {
     let mut handles = vec![];
     let nym_addr_out: Option<String>;
 
-    println!("Loading Zing-Proxy..");
+    // startup_message();
+    println!("@zingoproxyd: Launching Zingo-Proxy..\n@zingoproxyd: Launching gRPC Server..");
+    let proxy_handle = spawn_server(proxy_port, lwd_port, zebrad_port, online.clone()).await;
+    handles.push(proxy_handle);
+    tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
```
AloeareV: How much work would it take to make a more nuanced way of detecting when the server's ready, other than waiting for hopefully-long-enough?

zingo-proxyd/src/proxy.rs
```diff
@@ -7,33 +10,36 @@ use std::sync::atomic::{AtomicBool, Ordering};
...
     #[cfg(not(feature = "nym_poc"))]
     {
+        println!("@zingoproxyd[nym]: Launching Nym Server..");
         let nym_server: NymServer = NymServer(NymClient::nym_spawn(nym_conf_path).await);
         nym_addr_out = Some(nym_server.0 .0.nym_address().to_string());
 
-        let nym_proxy_handle = nym_server.serve(online.clone()).await;
+        let nym_proxy_handle = nym_server.serve(online).await;
         handles.push(nym_proxy_handle);
+        tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
```
AloeareV, line 51: As above.

zingo-rpc/src/utils.rs
```diff
@@ -35,3 +35,13 @@ macro_rules! define_grpc_passthrough {
         }
     };
 }
+
+/// Returns build info for Zingo-Proxy.
+pub fn get_build_info() -> (String, String, String, String, String) {
```
AloeareV: Not blocking, but a return value of a 5-string tuple is hard to look at from the calling scope and understand what's what..this would be a good use for a struct, I think, purely so that we can work with named fields on the return type.

zingo-proxyd/src/proxy.rs
idky137: I agree definitely needs fixing before production, i think I can create a server health check RPC and poll the output of that until the server is ready. The use below isn't as easy to fix and may need prior work to removing but i'll sort this one in this PR.

zingo-proxyd/src/proxy.rs
idky137: This may need more work, the nym server currently only handles send_transaction commands and needs to be updated to handle variable command types, then we can implement a similar health check command. There may also be functionality within the Nym-Rust code base but as the server needs re-implementing I think this is OK here for now?

zingo-rpc/src/utils.rs
idky137: Added BuildInfo struct and updated get_build_info in [d56984e]. Currently fields are public for direct access but im happy to make them private and add getter functions.

AloeareV: I think the use-case for it is straight-forward enough that direct field access is the simplest way of doing things...all the struct is for is naming return values, so adding getters/setters is a lot of ceremony.

zingo-proxyd/src/proxy.rs
idky137: added in [bd0c4c0]. polls result of get_lightd_info, currently shuts down zingoproxyd if connection cannot be established after 3 attempts, this could be increased or changed to a timeout?

zingo-rpc/src/blockcache/utils.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl std::fmt::Display for ParseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ParseError::Io(err) => write!(f, "IO Error: {}", err),
+            ParseError::InvalidData(msg) => write!(f, "Invalid Data Error: {}", msg),
+        }
+    }
+}
```
AloeareV: I would enthusiastically recommend the crate [thiserror](https://docs.rs/thiserror/latest/thiserror/).
I don't know if there's an easy way to do the from JsonConnector error as a string like that, but the entire chunk I have selected could become
```suggestion
/// Parser Error Type.
#[derive(Debug, thiserror::Error)]
pub enum ParseError {
    #[error("IO Error: {0}")]
    Io(#[from] std::io::Error),
    #[error("Invalid Data Error: {0}")]
    InvalidData(String),
}
```

AloeareV: We're currently in the process of switching to thiserror in zingolib. It's not on all our error types, but we've a bunch of it going on for example [here](https://github.com/zingolabs/zingolib/blob/dev/zingolib/src/lightclient/send.rs#L166)

zingo-rpc/src/blockcache/block.rs
```diff
@@ -0,0 +1,426 @@
...
+                }) => {
+                    return Err(ParseError::InvalidData(
+                        "Received object block type, this should not be possible here.".to_string(),
+                    ));
+                }
+                Ok(GetBlockResponse::Raw(block_hex)) => {
+                    Ok(FullBlock::parse_to_compact(
+                        block_hex.as_ref(),
+                        Some(tx.into_iter().map(|s| s.into_bytes()).collect()),
+                        0, //trees.sapling as u32,
```
AloeareV: I don't understand where these come from.

zingo-rpc/src/blockcache/transaction.rs
```diff
@@ -0,0 +1,737 @@
...
+    pub cmu: Vec<u8>,
+    /// Ephemeral public key for Diffie-Hellman key exchange.
+    ///
+    /// Size[bytes]: 32
+    pub ephemeral_key: Vec<u8>,
+    /// Encrypted transaction details including value transferred and an optional memo.
+    ///
+    /// Size[bytes]: 580
+    pub enc_ciphertext: Vec<u8>,
+    // OutCiphertext [IGNORED] - Size[bytes]: 80
```
AloeareV, line 188: Is this an Output, or a CompactOutput? ... Now that I think about it, it has to be a CompactOutput with the ignored fields...and there's no situation where we'd actually return a full output, only as part of a rawtransaction...so I have answered my own question, but these ignored fields are a bit confusing.

zingo-rpc/src/jsonrpc/primitives.rs
AloeareV, line 1: I like the serde::renames in order to actually underscore names properly!

zingo-rpc/src/blockcache/block.rs
AloeareV: By 'these' I mean the 0 and 2 that are commented sapling and orchard

zingo-rpc/src/blockcache/utils.rs
idky137: Ah yes oscar showed me this just after i coded these errors. ive left them in for now while im chasing bugs but will refactor ParseError and JsonRpcConnectorError using ThisError before marking as ready for review.

zingo-rpc/src/blockcache/transaction.rs
idky137: This is actually an Output in a Full Transaction, however some fields are ignored as they are not required by Zingo-Proxy as we only create CompactTx from this data. This mirrors functionality in LightWalletD. Ignored fields have been included as comments to make deserialization more understandable and to allow inclusion of any of these fields if required by Zingolib in the future (by changing skip_bytes() to read_bytes() in the relevant parse_from_slice impl. On line 683 of this file there is a FullTransaction method, to_compact(), that converts a Zingo-Proxy FullTransaction into a librustzcash CompactTx.

zingo-rpc/src/blockcache/block.rs
idky137: sorry these were placeholders as the struct i was originally using for the GetBlockResponse::trees field was from zebra and had private fields, making it impossible to retrieve the required values. I have since added a struct in Zingo-Proxy to hold this data and have removed the placeholder values.

zingo-rpc/src/blockcache/utils.rs
idky137: Ive converted all inetrnal error types to use thiserror in [4e91fff].

idky137:
All lightwallet RPC's zingolib use are now implemented. get_block_range and get_mempool_stream should be reimplemented once the internal blockcache and state engine are implemented but are working, just slow. There are a few unwraps i want to handle and a few clippy errors but apart from a some small fixes this PR is now code complete. I know its got quite big so if needed I can split it into a few separate PRs.

Oscar-Pepper (approved):

PR #36, opened 2024-07-18
Bump safely

idky137:
This PR is preliminary work towards updating dependency versions. The desired version have been added to Cargo.tomls but the actual update will be done in a separate PR. 

The majority of this pr is removing dependencies. In some places code has been added where required:
- to remove LibRustZcash the CompactTcStreamer is now built locally.
- to remove zebra primitive structs have been added to zingo-rpc::primitives.
- zingo-testutils has been removed apart from code under the nym_poc feature flag, which is to be removed in full once a Nym enhanced zingolib build is complete.

Other work in this PR:
- The Cargo.tomls have been restructured for readability.
- errors have been moved into sub mods and NymError was created to handle Nym related error.
- connect_to_node_get_info integration test was rewritten to use a zingo lightclient, simplifying dependency tree.

Two bugs in test code ware also fixed. 
- Adding a panic to the custom test panic handler fixed a bug where, if an auxiliary thread panicked, the whole test was shut down and passed. This now results in a test failure, showing the underlying panic info.
- Due to the block cache not yet being implemented, get_block_range would sometimes timeout. The timeout duration was increased for get_block_range to handle this, however this should be fixed with the block cache.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8480,7 +8480,6 @@ dependencies = [
  "tonic",
  "tower",
  "zcash_client_backend",
- "zcash_encoding 0.2.0 (git+https://github.com/zingolabs/librustzcash.git?branch=nym_integration)",
  "zebra-chain",
  "zebra-rpc",
  "zebra-state",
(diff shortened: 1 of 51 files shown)


zancas (approved):
My review is superficial.  This is a major refactor.

Issue #48, opened 2024-09-10
1.1 - Update Zaino-Testutils for ZIP317 and ZIP320


idky137:
Zaino's Zingolib and liobrustzcash dependencies need to be updated for ZIP317 and ZIP320 before integration with Zebra. There are also several other bugs that have been fixed in lightwalletd in the last few months that has caused changed in the expected behaviour by light wallets.

### Tasks
 - [x] Task 1: Temporarily remove Nym code to allow development on Zaino while the dependency conflict is being fixed.
 - [x] Task 2: Update dependency versions  and fix and error in Zaino-testutils from update.
 - [x] Task 3: Fix https://github.com/zcash/lightwalletd/issues/493 in Zaino.

- Blocked on https://github.com/zingolabs/zingolib/pull/1386.

idky137:
Merging #74, #75, #76 and #77 will complete this deliverable.

PR #75, opened 2024-10-08
Remove nym

idky137:
This PR comments out all nym code in Zaino. This is required so integration with zebra can happen while the dependency conflict problem is tackled separately.

- This PR Completes Task 1 of #48

PR #74, opened 2024-10-08
Replace hyper with reqwest

idky137:
Much of hyper's high level API was removed in 1.0 and for our current use (Zaino-Fetch::jsonrpc::connector::JsonRpcConnector) we do not require the low level functionality now offered by hyper. For this reason the use of hyper here has been replaced with reqwest.

This is the first work towards updating the zingolib and librustzcash versions used by zaino.

- This PR is the first work towards #48

PR #76, opened 2024-10-08
Bump deps

idky137:
This PR updates the zingolib, librustzcash and rustc versions used by Zaino.
It also fixes errors in Zaino-testutils and integration-tests caused by the update.

 -The fix for https://github.com/zcash/lightwalletd/issues/493 (that currently breaks the get_mempool_steam RPC) will will be completed in a separate PR.

- This will complete Task 2 of #48.

### DIFF
--- a/zaino-testutils/src/lib.rs
+++ b/zaino-testutils/src/lib.rs
@@ -314,34 +314,3 @@ pub mod zingo_lightclient {
         tokio::time::sleep(std::time::Duration::from_secs(5)).await;
     }
 }
-
-// /// Zingo-Indexer wrapper for Zingolib's Pool Enum.
-// #[derive(Debug, Clone, Copy, PartialEq, Eq)]
-// pub enum Pool {
-//     /// Orchard pool.
-//     Orchard,
-//     /// Sapling pool
-//     Sapling,
-//     /// Transparent poool.
-//     Transparent,
-// }
-
-// impl From<Pool> for zingolib::wallet::Pool {
-//     fn from(test_pool: Pool) -> Self {
-//         match test_pool {
-//             Pool::Orchard => zingolib::wallet::Pool::Orchard,
-//             Pool::Sapling => zingolib::wallet::Pool::Sapling,
-//             Pool::Transparent => zingolib::wallet::Pool::Transparent,
-//         }
-//     }
-// }
-
-// impl From<zingolib::wallet::Pool> for Pool {
-//     fn from(pool: zingolib::wallet::Pool) -> Self {
-//         match pool {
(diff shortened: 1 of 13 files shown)


idky137:
built atop of #74 and #75

idky137:
monitor_unverified_mempool test fails due to changes in mempool transaction height. this will be fixed in a separate PR.

zaino-testutils/src/lib.rs
```diff
@@ -275,57 +276,41 @@ fn create_temp_conf_files(
...
+    /// Sends funds to address given, handles proposals internally.
+    ///
+    /// recievers should be in the form vec![Address, Amount, Option<Memo>]
+    pub async fn quick_send(
+        zingo_client: &zingolib::lightclient::LightClient,
+        receivers: Vec<(&str, u64, Option<&str>)>,
+    ) -> Result<String, zingolib::lightclient::send::send_with_proposal::QuickSendError> {
+        zingolib::testutils::lightclient::from_inputs::quick_send(zingo_client, receivers)
+            .await
+            .map(|tx_ids| tx_ids.into_iter().next().unwrap().to_string())
```
Oscar-Pepper, line 298: why not just import and use `from_inputs::quick_send` directly? then in the tests you can call `.first()` if you want to ignore the multiple txids returned

zaino-testutils/src/lib.rs
```diff
@@ -275,57 +276,41 @@ fn create_temp_conf_files(
...
+        let zingo_client_saved = zingo_client.export_save_buffer_async().await.unwrap();
+        let zingo_client_loaded = std::sync::Arc::new(
+            zingolib::lightclient::LightClient::read_wallet_from_buffer_async(
+                zingo_client.config(),
+                &zingo_client_saved[..],
+            )
+            .await
+            .unwrap(),
+        );
+        zingolib::lightclient::LightClient::start_mempool_monitor(zingo_client_loaded.clone());
```
Oscar-Pepper, line 312: hmmm.. im not sure this will work as the mempool monitor is attached to a clone of the lightclient that wants to sync from mempool. I think we want to put the lightclient in an Arc in the code where we want to start the mempool and then call the zingolib version of `start_mempool_monitor` directly. then any other time the lightclient is called in the code you can use the lightclient in the arc but still pass it to mempool

Oscar-Pepper: this is my blocking comment btw, i think this needs to change

Oscar-Pepper (changes requested):
i dont think the mempool will work correctly if we clone the LC. see my comments

Oscar-Pepper (approved):
on the grounds that the mempool is working and these changes are best made on top of a later PR, i am merging this as is.

Issue #56, opened 2024-09-10
2.4 - Provide interface to / provide internal representation of mempool from Zebrad


idky137:
Due to requirements of the get_mempool_stream gRPC service it is more efficient to keep an internal mempool rather than serve mempool data directly. Depending on functionality the IndexerMempool will either use the ReadStateService or Zaino-Fetch as its backend. If possible it should be configurable to use either.

- Task 2.4.3 Blocked on #54 
- Task 2.4.4 Blocked on #122
- Task 2.4.5 Blocked by #112 
- Task 2.4.6 Blocked on #96

### Tasks
 - [x] [Task 2.4.1: Implement efficient Broadcast stream for high concurrency use case.](https://github.com/zingolabs/zaino/issues/116)
 - [x] [Task 2.4.2: Add `Mempool` trait.](https://github.com/zingolabs/zaino/issues/117)
 - [x] [Task 2.4.3: Implement Mempool for `StateService`.](https://github.com/zingolabs/zaino/issues/118)
 - [x] [Task 2.4.4: Implement Mempool for `FetchService`.](https://github.com/zingolabs/zaino/issues/119)
 - [x] [Task 2.4.5: Add `Mempool` trait to `Indexer` trait.](https://github.com/zingolabs/zaino/issues/120)
 - [x] [Task 2.4.6: Add test set that comprehensively tests mempool functionality.](https://github.com/zingolabs/zaino/issues/121)

PR #77, opened 2024-10-10
Fix mempool tx

idky137:
Updates get_mempool_stream to return tx height as mempool height instead of -1.

- Completes Task 3 of #48


idky137:
Built atop of #74, #75 and #76

idky137:
This is not the most elegant implementation, the mempool should hold the block height,  but the mempool is being rewritten and moved in #56 so this is sufficient until then.

Oscar-Pepper (approved):
given your comment I agree this is fine for now.

Issue #50, opened 2024-09-10
1.3 Convert Indexer service to use zebra’s primitives.


idky137:
Currently Zaino uses zcash primitives defined in Zaino-Fetch. Zaino-Serve should use Zebra's primitives directly to ease integration with Zebra's ReadStateService. For this either Zaino-Fetch should be converted to use Zebra's primitives directly or conversion impls should be implemented.

- Blocked on #48

### Tasks
- [ ] Task 1: *[Work in Zebra] Expose primitives as pub or provide getters / setters and derive Serde::deserialize.
- [ ] Task 2: Convert Zaino-Fetch to use Zebra's primitives directly or provide conversion impls.
- [ ] Task 3: Convert Zaino-Serve to use Zebra's primitives.

idky137:
Merging #82 and https://github.com/ZcashFoundation/zebra/pull/8964 will complete this deliverable.

PR #82, opened 2024-10-23
Zebra primitives

idky137:
This PR removed the primitives mod from zaino-fetch, using zebra's structs directly possible and adding wrapper structs in zaino-fetch::jsonrpc::response.rs where required.

- This PR will complete Tasks 2 and 3 of #50.

*NOTE: There are currently several patched libraries in Zebra v2.0.0rc that are causing conflicts. For this reason we are currently using v1.9.0 and will update to 2.0.0 once these patches are removed.


Cargo.toml
```diff
@@ -25,10 +25,19 @@ license = "Apache-2.0"
...
+# NOTE: This may have to move to Zaino-testutils once Nym code is reintroduced and Zaino-Wallet becomes a dependency of zingolib.
 zingolib = { git = "https://github.com/zingolabs/zingolib.git", tag = "zaino_dep_002_091024_95e5b0d8f9d5ee0485c6141533da2f727aeafae2", features = ["zaino-test"] }
 
-# Temporarily removed until dependency conflic can be resolved.
-# NymSdk
+# Librustzcash
+zcash_protocol = { git = "https://github.com/zingolabs/librustzcash.git", tag = "zcash_client_sqlite-0.11.2_plus_zingolabs_changes-1-g7ad60b5d5-2-g121371a08" }
+
+# Zebra
+zebra-chain = { git = "https://github.com/idky137/zebra.git", branch = "zaino_dep_changes" }
```
Oscar-Pepper, line 36: non-blocking. a tag might be better practice here. i like using branches for development but then i create a tag when i send a PR for review

zaino-fetch/src/jsonrpc/response.rs
```diff
@@ -75,11 +64,78 @@ pub struct SendTransactionResponse(#[serde(with = "hex")] pub TransactionHash);
 /// Also see the notes for the [`Rpc::get_best_block_hash`] and `get_block_hash` methods.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, serde::Deserialize, serde::Serialize)]
 #[serde(transparent)]
-pub struct GetBlockHash(#[serde(with = "hex")] pub BlockHash);
+pub struct GetBlockHash(#[serde(with = "hex")] pub zebra_chain::block::Hash);
```
Oscar-Pepper, line 67: non-blocking. structs shouldn't be action verbs generally. I think `getblockhash` should return a BlockHash struct

Oscar-Pepper: or even, there is a merit to naming the same as the struct its wrapping `Hash`

Oscar-Pepper (approved):
left non-blocking comments

Issue #52, opened 2024-09-10
1.5 - Implement lightwalletd gRPC functionality in Rust using JSONRPC interface for backwards compatibility.


idky137:
Currently Zaino only has the subset of lightwallet gRPC services that Zingolib uses implemented. The full set of services should be implemented in Zaino-Fetch (backed by Zebra's JsonRPC Service) for backwards compatibility. The deprecated services can then use these implementations until they are fully removed.

### Tasks
 - [ ] Task 1: implement remaining lightwallet gRPC services (shown in #40).

NOTE 1: Currently compact blocks are fetched as full blocks are parsed to compact blocks before being served to the client. This is inefficient and compact block should be served directly from disc / memory. This work will be completed in #55.

NOTE 2: Currently Lightwalletd does not return the correct error types from its gRPC services [https://github.com/zcash/lightwalletd/issues/497]. We will implement Zaino to use the correct tonic statuses [https://github.com/grpc/grpc/blob/master/doc/statuscodes.md#status-codes-and-their-use-in-grpc] rather than implement functionality as it is in Lightwalletd.

idky137:
Local mempool to be implemented in #56.

idky137:
- Merging linked open PRs and completing https://github.com/zingolabs/zaino/issues/95, https://github.com/zingolabs/zaino/issues/94 and https://github.com/zingolabs/zaino/issues/93 will complete this deliverable.

PR #83, opened 2024-10-28
Fix get latest block

idky137:
Fixes the response of get_latest_block. The block hash in this response should be in display order (reversed). This PR uses functionality in zebra for this.

- Fixes get_latest_block test in https://github.com/zingolabs/zcash-local-net.
- Work towards #52

- Built atop #82 as functionality exposed in that PR is used here. Merge that PR first.


Oscar-Pepper (approved):
nice!

PR #86, opened 2024-10-30
Implement Get block range nullifiers

idky137:
Implements the get_block_range_nullifers gRPC service.

Currently there are no tests for this in Zaino, these will be added when the last PRs for [https://github.com/zingolabs/zcash-local-net] land and we can integrate that repo into Zaino-testutils. There is a test in [https://github.com/zingolabs/zcash-local-net] that tests this function "integrations::get_block_range_nullifiers".

- Work towards https://github.com/zingolabs/zaino/issues/52

### DIFF
--- a/zaino-serve/src/rpc/service.rs
+++ b/zaino-serve/src/rpc/service.rs
@@ -273,15 +273,14 @@ impl CompactTxStreamer for GrpcClient {
 
     /// Server streaming response type for the GetBlockRangeNullifiers method.
     #[doc = " Server streaming response type for the GetBlockRangeNullifiers method."]
-    type GetBlockRangeNullifiersStream = tonic::Streaming<CompactBlock>;
+    type GetBlockRangeNullifiersStream = std::pin::Pin<Box<CompactBlockStream>>;
 
     /// Same as GetBlockRange except actions contain only nullifiers.
     ///
-    /// This RPC has not been implemented as it is not currently used by zingolib.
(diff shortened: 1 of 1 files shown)


zaino-serve/src/rpc/service.rs
```diff
@@ -301,8 +300,60 @@ impl CompactTxStreamer for GrpcClient {
...
     {
         println!("[TEST] Received call of get_block_range_nullifiers.");
-        Box::pin(async {
-            Err(tonic::Status::unimplemented("get_block_range_nullifiers not yet implemented. If you require this RPC please open an issue or PR at the Zingo-Indexer github (https://github.com/zingolabs/zingo-indexer)."))
+        let zebrad_uri = self.zebrad_uri.clone();
+        Box::pin(async move {
+            let blockrange = request.into_inner();
+            let mut start = blockrange
+                .start
+                .map(|s| s.height as u32)
```
Oscar-Pepper: `as` performs an unchecked cast. consider using from / into here

Oscar-Pepper (changes requested):
im concerned an unchecked cast from u64 to u32 could cause undefined overflow behaviour etc.

zaino-serve/src/rpc/service.rs
idky137: Added change in [fc22fc2]. Also fixed error messages.

Oscar-Pepper (approved):

PR #87, opened 2024-10-30
Fix get block range

idky137:
Fixes the order that compact blocks are returned in by the get_block_range and get_block_range_nullifiers RPCs. Before they were always returned in descending order, now they are returned in the order that they are requested in (start -> end). Error codes will be fixed in a separate PR

- There is a test in [https://github.com/zingolabs/zcash-local-net] that tests this function "integrations::get_block_range_nullifiers".

- Work towards https://github.com/zingolabs/zaino/issues/52

- Built atop #86

### DIFF
--- a/zaino-serve/src/rpc/service.rs
+++ b/zaino-serve/src/rpc/service.rs
@@ -223,15 +223,23 @@ impl CompactTxStreamer for GrpcClient {
                 .end
                 .map(|e| e.height as u32)
                 .ok_or(tonic::Status::invalid_argument("End block not specified"))?;
-            if start > end {
+            let rev_order = if start > end {
                 (start, end) = (end, start);
-            }
+                true
+            } else {
+                false
+            };
             println!("[TEST] Fetching blocks in range: {}-{}.", start, end);
             let (channel_tx, channel_rx) = tokio::sync::mpsc::channel(32);
             tokio::spawn(async move {
                 // NOTE: This timeout is so slow due to the blockcache not being implemented. This should be reduced to 30s once functionality is in place.
                 let timeout = timeout(std::time::Duration::from_secs(120), async {
-                    for height in (start..=end).rev() {
+                    for height in start..=end {
+                        let height = if rev_order {
+                            end - (height - start)
+                        } else {
+                            height
+                        };
                         println!("[TEST] Fetching block at height: {}.", height);
                         let compact_block = get_block_from_node(&zebrad_uri, &height).await;
                         match compact_block {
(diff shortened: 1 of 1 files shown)


zaino-serve/src/rpc/service.rs
```diff
@@ -226,15 +226,23 @@ impl CompactTxStreamer for GrpcClient {
...
                 .map(|e| e.height as u32)
                 .ok_or(tonic::Status::invalid_argument("End block not specified"))?;
-            if start > end {
+            let rev_order = if start > end {
                 (start, end) = (end, start);
-            }
+                true
+            } else {
+                false
+            };
```
Oscar-Pepper, line 317: im not sure i like the fact that this doesn't error when the start is higher than the end ... this would error if it was a true `Range` rust struct so im not sure why this is supported. is this how lwd works??

Oscar-Pepper: i will add this test to zcash-local-net to check

zaino-serve/src/rpc/service.rs
```diff
@@ -301,8 +308,68 @@ impl CompactTxStreamer for GrpcClient {
...
     {
         println!("[TEST] Received call of get_block_range_nullifiers.");
-        Box::pin(async {
-            Err(tonic::Status::unimplemented("get_block_range_nullifiers not yet implemented. If you require this RPC please open an issue or PR at the Zingo-Indexer github (https://github.com/zingolabs/zingo-indexer)."))
+        let zebrad_uri = self.zebrad_uri.clone();
+        Box::pin(async move {
+            let blockrange = request.into_inner();
+            let mut start = blockrange
+                .start
+                .map(|s| s.height as u32)
```
Oscar-Pepper: can we use from / into here for a checked cast from u64 to u32?

Oscar-Pepper (changes requested):

zaino-serve/src/rpc/service.rs
idky137: This is actually the correct behaviour. When the start is higher than the end the blocks are returned in descending order. This is how zingolib currently asks for blocks.

idky137: Lines 238:242 reverse the block order fetched if rev_order is true.

zaino-serve/src/rpc/service.rs
idky137: This was updated in a separate PR and added here in [2da2481]

zaino-serve/src/rpc/service.rs
Oscar-Pepper: oh ok i will merge!

Oscar-Pepper (approved):

PR #89, opened 2024-11-02
Fix error types

idky137:
Fixes the errors returned for get_block, get_block_range, get_nullifiers, get_nullifiers_range RPCs.

- Work towards #52


zaino-fetch/src/jsonrpc/connector.rs
```diff
@@ -126,15 +126,15 @@ impl JsonRpcConnector {
             if body_str.contains("Work queue depth exceeded") {
                 if attempts >= max_attempts {
                     return Err(JsonRpcConnectorError::new(
-                        "Work queue depth exceeded after multiple attempts",
+                        "Error: The node's rpc queue depth was exceeded after multiple attempts",
```
Oscar-Pepper, line 130: im not sure if we should consider this blocking but I think we should try to avoid repeated Error: Error: Error status xxxx in the format display of error types. Im not sure what the standard practice is but my preferred method is to not have "Error:" in the format display of error types (or this case strings which we should change to error types and implement fmt::display at some point soon). The idea here is all the different error types can navigate many different levels and enums without "collecting" more "Error:"s at the start. then finally, at the top level where it will be called directly by the user/consumer we put "Error: {}" in the final fmt:display. I have seen you have done this later in the PR but i havent checked this is the top level fmt::display.

Oscar-Pepper: the final error printed to user should preferably look like this:
"Error: Serialisation error. 2nd level error. 2nd level thing failed." for example. IMO

Oscar-Pepper: this would be made up of the top level error message for example in an eprintln:
match result {
Ok(_) => (),
Err(top_level_error_enum) =>  eprintln!"Error: {}", top_level_error_enum)
} 

then the top level error enum with thiserror macro:
pub enum TopLevelError {
#error("Serialisation error. {0}")
SerialisationFailed(SerialisationError),
#error("Some other error. {0}")
SomeOtherError,
}

then serialisation error enum:
enum SerialisationError {
#error("2nd level thing failed. {0}")
2ndLevelThingFailed(2ndLevelThing),
#error("2nd level other thing failed. {0}")
2ndLevelOtherThing,
}

where 2ndLevelthing might be some struct with useful info like if its a block height for example it would produce:
"Error: Serialisation error. 2nd level error. 2nd level thing failed. Blockheight: 10"

zaino-serve/src/rpc/service.rs
```diff
@@ -351,13 +349,23 @@ impl CompactTxStreamer for GrpcClient {
...
                             Err(e) => {
-                                // TODO: Hide server error from clients before release. Currently useful for dev purposes.
-                                if channel_tx
-                                    .send(Err(tonic::Status::internal(e.to_string())))
-                                    .await
-                                    .is_err()
-                                {
-                                    break;
+                                if height > chain_height {
+                                    let _ = channel_tx
```
Oscar-Pepper: sometimes the compliler asks to put "let _ = " but i think this is bad advice. sometimes its useful to use this to keep something in scope but i still prefer "let _unused_name_of_thing =" in stead of just underscore for that. in this case i believe you need to actually handle your result with a match and error handling or atleast an expect if it cant fail

zaino-serve/src/rpc/service.rs
```diff
@@ -483,13 +485,23 @@ impl CompactTxStreamer for GrpcClient {
...
```
Oscar-Pepper: see comment above about "let _ ="

Oscar-Pepper (changes requested):
im not sure if i consider the error stuff blocking but i think it could be improved. the "let _ =" i believe is a mistake here and i consider blocking, we should not have wrapped results lingering around IMO

zaino-fetch/src/jsonrpc/connector.rs
idky137: I think this would be best to do in a seperate PR in milestone 2. Ive added a todo in [724ef3f](https://github.com/zingolabs/zaino/pull/89/commits/724ef3f0544e68aa3b5aaffc81d8d8dc10badc3c)

zaino-serve/src/rpc/service.rs
idky137: change added in [724ef3f](https://github.com/zingolabs/zaino/pull/89/commits/724ef3f0544e68aa3b5aaffc81d8d8dc10badc3c)

zaino-serve/src/rpc/service.rs
idky137: change added in [724ef3f](https://github.com/zingolabs/zaino/pull/89/commits/724ef3f0544e68aa3b5aaffc81d8d8dc10badc3c)

Oscar-Pepper (approved):
Great!

PR #98, opened 2024-11-08
Implement Utxo rpcs

idky137:
Implements get_address_utxos and get_address_utoxs_stream rpcs.

- Build atop Fix error types https://github.com/zingolabs/zaino/pull/89, https://github.com/zingolabs/zaino/pull/90, #91,  #92 and #97

- Work towards 1.5 - Implement lightwalletd gRPC functionality in Rust using JSONRPC interface for backwards compatibility. https://github.com/zingolabs/zaino/issues/52

- Closes #94

PR #97, opened 2024-11-06
Implement Address balance rpcs

idky137:
Implements get_taddress_balance and get_taddress_balance_stream rpcs.

- Build atop Fix error types https://github.com/zingolabs/zaino/pull/89, https://github.com/zingolabs/zaino/pull/90, #91 and #92

- Work towards 1.5 - Implement lightwalletd gRPC functionality in Rust using JSONRPC interface for backwards compatibility. https://github.com/zingolabs/zaino/issues/52

- Closes #93

PR #92, opened 2024-11-05
Mempool rpcs

idky137:
Fixes get_mempool_stream and implements get_mempool_tx rpcs.

- Build atop Fix error types https://github.com/zingolabs/zaino/pull/89, https://github.com/zingolabs/zaino/pull/90 and #91

- Work towards 1.5 - Implement lightwalletd gRPC functionality in Rust using JSONRPC interface for backwards compatibility. https://github.com/zingolabs/zaino/issues/52

PR #91, opened 2024-11-02
Taddress txids

idky137:
Fixes error handling in the get_taddress_tzids RPC.

- Build atop Fix error types #89 and #90

- Work towards 1.5 - Implement lightwalletd gRPC functionality in Rust using JSONRPC interface for backwards compatibility. #52

PR #100, opened 2024-11-12
Subtree roots RPCs

### LINKED ISSUE #95: Implement subtree roots RPCs
Implement the get_subtree_roots gRPC service in zaino.

- Required for #52

idky137:
Implements get_subtree_roots RPC. Also adds a timeout to the get_address_utxos RPC and unifies timeout times.

- Build atop Fix error types https://github.com/zingolabs/zaino/pull/89, https://github.com/zingolabs/zaino/pull/90, #91,  #92, #97, #98 and #99.

- Work towards 1.5 - Implement lightwalletd gRPC functionality in Rust using JSONRPC interface for backwards compatibility. https://github.com/zingolabs/zaino/issues/52

- Closes #95

### DIFF
--- a/zaino-serve/src/rpc/service.rs
+++ b/zaino-serve/src/rpc/service.rs
@@ -18,10 +18,7 @@ use zaino_fetch::{
 use zaino_proto::proto::{
     compact_formats::{CompactBlock, CompactTx},
     service::{
-        compact_tx_streamer_server::CompactTxStreamer, Address, AddressList, Balance, BlockId,
-        BlockRange, ChainSpec, Duration, Empty, Exclude, GetAddressUtxosArg, GetAddressUtxosReply,
-        GetAddressUtxosReplyList, GetSubtreeRootsArg, LightdInfo, PingResponse, RawTransaction,
-        SendResponse, SubtreeRoot, TransparentAddressBlockFilter, TreeState, TxFilter,
+        compact_tx_streamer_server::CompactTxStreamer, Address, AddressList, Balance, BlockId, BlockRange, ChainSpec, Duration, Empty, Exclude, GetAddressUtxosArg, GetAddressUtxosReply, GetAddressUtxosReplyList, GetSubtreeRootsArg, LightdInfo, PingResponse, RawTransaction, SendResponse, ShieldedProtocol, SubtreeRoot, TransparentAddressBlockFilter, TreeState, TxFilter
     },
 };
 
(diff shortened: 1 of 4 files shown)


zaino-serve/src/rpc/service.rs
```diff
@@ -1491,8 +1529,135 @@ impl CompactTxStreamer for GrpcClient {
...
+            .await?;
+            let subtree_roots_args = request.into_inner();
+            let pool = match ShieldedProtocol::try_from(subtree_roots_args.shielded_protocol) {
+                Ok(protocol) => protocol.as_str_name(),
+                Err(_) => return Err(tonic::Status::invalid_argument("Error: Invalid shielded protocol value.")),
+            };
+            let start_index = match u16::try_from(subtree_roots_args.start_index) {
+                Ok(value) => value,
+                Err(_) => return Err(tonic::Status::invalid_argument("Error: start_index value exceeds u16 range.")),
+            };
```
Oscar-Pepper, line 1548: could consider let else here
```suggestion
            let Ok(start_index) = u16::try_from(subtree_roots_args.start_index) else {
                return Err(tonic::Status::invalid_argument("Error: start_index value exceeds u16 range."));
            };
```

Oscar-Pepper: untested suggestion, just showing the idea

zaino-serve/src/rpc/service.rs
```diff
@@ -1629,46 +1793,65 @@ impl CompactTxStreamer for GrpcClient {
...
-                                )))
-                                .await;
+                        entries += 1;
+                        if addr_args.max_entries > 0 && entries > addr_args.max_entries {
+                            break;
+                        }
+                        let checked_index = match i32::try_from(utxo.output_index) {
+                            Ok(index) => index,
+                            Err(_) => {
+                                let _ = channel_tx
```
Oscar-Pepper, line 1811: what is the purpose of this "let _" ? if its to silence warning about unwrapped result then this result should be handled?

zaino-serve/src/rpc/service.rs
```diff
@@ -1629,46 +1793,65 @@ impl CompactTxStreamer for GrpcClient {
...
+                                        "Error: Index out of range. Failed to convert to i32.",
+                                    )))
+                                    .await;
+                                return;
+                            }
+                        };
+                        let checked_satoshis = match i64::try_from(utxo.satoshis) {
+                            Ok(satoshis) => satoshis,
+                            Err(_) => {
+                                let _ = channel_tx
```
Oscar-Pepper, line 1822: and here

zaino-serve/src/rpc/service.rs
```diff
@@ -1629,46 +1793,65 @@ impl CompactTxStreamer for GrpcClient {
...
+                    Ok(_) => {
+                        return;
+                    }
+                    Err(_) => {
+                        channel_tx
+                            .send(Err(tonic::Status::deadline_exceeded(
+                                "Error: get_mempool_stream gRPC request timed out",
+                            )))
+                            .await
+                            .ok();
```
Oscar-Pepper, line 1854: here `.ok()` is used in a similar way to `let _ =`, is this because in all these calls we dont care if it errors? My opinion is we will always want to handle errors with either an error return or expect (if it cant fail). if an error is returned then the handler of this thread can decide to ignore the Err case

Oscar-Pepper (approved):
left a few non-blocking comments

PR #102, opened 2024-11-14
update zcash local net pt2

Oscar-Pepper:


idky137 (changes requested):
All looks good. I got to the same failure mentioned in the zcash-local-net PR. Just need to update the README with the new directory locations for chain_cache. It might be best to do this after the doc update merges as test instructions have moved to their own doc.

idky137 (approved):

PR #105, opened 2024-11-18
use autoinherit to set workspace manifest to single source of truth

zancas:


idky137:
I dont think we should be moving deps to workpace. This has caused a lot of problems in the past a took quite a bit of time of unravel last time. I am still in the process of doing the opposite to this..

idky137:
I think given that Zaino-State is intended to be used as a library we should make all dependencies not required but Zaino-State (Zaino-Serve, Zainod, Zaino-Testutils, Integration-Tests) Crate level to make the lib as easy / non restrictive as possible.

idky137 (changes requested):
I think given that Zaino-State is intended to be used as a library we should make all dependencies not required by Zaino-State (Zaino-Serve, Zainod, Zaino-Testutils, Integration-Tests) Crate level, to make the lib as easy to use / non restrictive as possible.

idky137 (approved):

Issue #58, opened 2024-09-10
3.1 - Implement remote access to Zebra's ReadStateService in Zaino


idky137:
Currently Zebra's ReadStateService only enables direct access to chain data (both Zebra and any process interfacing with the ReadStateService must be running on the same hardware). Zaino will extend this functionality, using a Hyper wrapper, to allow Zebra and Zaino (or software built using Zaino's IndexerStateService as its backend) to run on different hardware and should enable a much greater range of deployment strategies (eg. running validator, indexer or wallet processes on separate hardware). 

It should be noted that this will primarily be designed as a remote link between Zebra and Zaino and it is not intended for developers to directly interface with this service, but instead to use functionality exposed by the client library in Zaino (IndexerStateService).

Implementation details have not yet been finalised and need to be discussed with ECC / ZF. A final spec will be produced in milestone 2 (#57).

### Tasks
 - [ ] [Task 3.1.1: Implement `RemoteReadStateService` using a Hyper wrapper.](https://github.com/zingolabs/zaino/issues/133)
 - [ ] [Task 3.1.2: Implement configuration options in Zaino.](https://github.com/zingolabs/zaino/issues/134)

idky137:
This has been removed from the current roadmap. We may implement this in the future but it is not currently a priority.

PR #115, opened 2024-11-20
Allow connecting to Zebrad on any hostname

emersonian:
This patch allows for Zaino to connect to a Zebrad that is not running on localhost.

This is my first-ever Rust pull request so bear with me! I am unsure how to run the tests, am happy to fix anything that broke.

Thank you for your great work on Zaino!

### DIFF
--- a/zainod/src/config.rs
+++ b/zainod/src/config.rs
@@ -12,6 +12,8 @@ pub struct IndexerConfig {
     /// LightWalletD listen port [DEPRECATED].
     /// Used by zingo-testutils.
     pub lightwalletd_port: u16,
+    /// Full node / validator listen hostname.
+    pub zebrad_hostname: Option<String>,
     /// Full node / validator listen port.
     pub zebrad_port: u16,
     /// Full node Username.
@@ -52,6 +54,7 @@ impl Default for IndexerConfig {
             tcp_active: true,
             listen_port: Some(8080),
             lightwalletd_port: 9067,
+            zebrad_hostname: Some("127.0.0.1".to_string()),
             zebrad_port: 18232,
             node_user: Some("xxxxxx".to_string()),
             node_password: Some("xxxxxx".to_string()),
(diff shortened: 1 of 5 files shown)


daira:
I am opposed to merging this. We need to establish a strong security norm and expectation that the connection between Zaino and Zebrad is within a trust boundary. That is, we must be able to make the design assumption that Zaino can trust the Zebrad it is connected to, because it *necessarily* is run by the same party.

emersonian:
How can this be tested in a staging environment without the ability to connect to other hosts? I'm all for including warnings in the documentation with your concerns, but restricting to only allowing zebra to run on localhost is excessive and slows down integration testing across stakeholders.

Use case: Zec.rocks has around ten zebra instances synced and running around the world. I took two out of our public load balancers (one mainnet, one testnet) and am pointing Zaino at them for integration testing, all communicating within their clusters' internal networks. There is no security risk in this case, internal communications are encrypted by the Container Network Interface (CNI).

Restricting to localhost slows down the development of Docker Compose and Kubernetes configs for Zaino. Sure, add warnings, but don't force everyone to run Zebra on localhost.

idky137:
Could we query Zebra's `enable_cookie_auth` setting using the RPC server? We could then enforce `cookie_auth=true` when connecting to Zebra remotely? We could also add something similar for Zcashd using the user and password fields in its RPC server. Im not sure how disruptive this would be Zec.rocks though?

idky137:
Code looks good though!

idky137:
Hey sorry I haven't got back regarding this. 

I want to make sure everyone is happy with the decision we make regarding how we open up Zaino for remote use before merging anything into dev. 

We do have work planned for connecting to zebra remotely in Milestone 3 of our current grant (https://github.com/zingolabs/zaino/issues/58) but that is a few months away and will not be focusing on the RPC backend in Zaino.

I want to make sure Zaino can be used how the community wants though so think it would be good to work out a solution for the RPC backend separate to that.

emersonian:
I suggest that we start by matching the security modeling of lightwalletd rather than further restricting it. Lightwalletd allows connection to node RPCs running on any configurable host, today. If we want to improve upon that level of security, let's decide in a separate discussion.

Zaino must work in containers/Docker. Listening only on localhost blocks supporting containers (reasonably).

idky137:
I will raise this at one of the next arborist calls, originally I had planned to open up zaino for remote use around this time and agree we are not adding any new security weaknesses but want to check.. 

We do have time put aside in milestone 3 for enabling "safe" remote access to zebra but I would like to make it possible for you to start testing zaino before then. 

Maybe we can merge this functionality under an "unstable" feature flag during development.

EDIT: We are also close to upgrading the JsonRPC backend (https://github.com/zingolabs/zaino/pull/149 is the first of a few PR that will do this) which should give some decent speedups and would be good to start wider testing once this is in..

zancas:
I have heard a proposal to enable "testing_very_insecure" as a conditionally compiled feature.   That seems sensible to me.

I don't think a few months of testing until MileStone 3 is complete is too much time in test mode.

Indeed multiple months in test-mode is almost certainly correct, regardless.

idky137:
If we are happy with this I will have another look at this pr once #153 lands, there have been some structural changes and I think there will be a couple other places that this code will need to propaget to. I will also look at adding the feature flag.

emersonian:
Flagging this as a blocker, zec.rocks cannot run zaino until it works in cloud-native environments (Kubernetes or Docker).

This is the standard [across other projects](https://zebra.zfnd.org/user/docker.html) that zaino needs to meet:

```
cd zaino
docker compose up
# (Spins up zaino and zebra, works locally once synced)
```

Perhaps as a compromise we limit zaino to only connect to zebrad hosted on RFC1918 addresses, unless an insecure flag is passed? In-cluster communications are within a security boundary.

zainod/src/config.rs
```diff
@@ -65,22 +68,36 @@ impl Default for IndexerConfig {
 /// Attempts to load config data from a toml file at the specified path.
 pub fn load_config(file_path: &std::path::PathBuf) -> IndexerConfig {
     let mut config = IndexerConfig::default();
+    println!("Loading config: {:?}", file_path);
```
pacu, line 71: shall this be a call so some logger?

zainod/src/config.rs
```diff
@@ -65,22 +68,36 @@ impl Default for IndexerConfig {
...
-                lightwalletd_port: parsed_config.lightwalletd_port,
-                zebrad_port: parsed_config.zebrad_port,
-                node_user: parsed_config.node_user.or(config.node_user),
-                node_password: parsed_config.node_password.or(config.node_password),
-                max_queue_size: parsed_config.max_queue_size,
-                max_worker_pool_size: parsed_config.max_worker_pool_size,
-                idle_worker_pool_size: parsed_config.idle_worker_pool_size,
-            };
+    if !file_path.exists() {
+        eprintln!("File does not exist: {:?}", file_path);
```
pacu, line 74: Is this a debug call or some problem?

zainod/src/config.rs
```diff
@@ -65,22 +68,36 @@ impl Default for IndexerConfig {
 /// Attempts to load config data from a toml file at the specified path.
 pub fn load_config(file_path: &std::path::PathBuf) -> IndexerConfig {
```
pacu, line 69: this should be a fallible function that returns some error

zainod/src/config.rs
```diff
@@ -65,22 +68,36 @@ impl Default for IndexerConfig {
...
+                        node_password: parsed_config.node_password.or(config.node_password),
+                        max_queue_size: parsed_config.max_queue_size,
+                        max_worker_pool_size: parsed_config.max_worker_pool_size,
+                        idle_worker_pool_size: parsed_config.idle_worker_pool_size,
+                    };
+                }
+                Err(e) => eprintln!("Error parsing TOML: {}", e),
+            }
         }
+        Err(e) => eprintln!("Error reading file: {}", e),
```
pacu, line 98: Return Error

pacu (changes requested):
Setting aside discussions on  whether connecting to non-local nodes should be allowed these changes could introduce some improvements

idky137:
Hey, thanks for the review! Due to quite a lot of changes that have happened in zaino since this PR was opened this work will now be completed in #180

pacu:
according to @emersonian there was some agreement on how to settle this. What are the next steps for this PR? Closing, further modifications?

arya2:
SSH port forwarding may be a good alternative to this change, though adding a `testing-very-insecure` rust compilation config flag seems okay too.

PR #139, opened 2024-11-26
Remove lighwalletd

idky137:
Removes Lightwalletd from production code.

- Build atop #138

PR #138, opened 2024-11-26
Update Testutils

idky137:
Updates Testutils to use zcash-local-net and use zebra as our testing backend as well as zcashd.

- Closes #96

Issue #114, opened 2024-11-19
Task 2.1.5: Add tests for Zcash RPC functionality backed by  `StateService`


idky137:
Add tests for Zcash RPC functionality backed by  `StateService`.

These test should load a cached chain using Zebra regtest mode to query.

- Blocked on #96

Issue #112, opened 2024-11-19
Task 2.1.3: Add Indexer Trait


idky137:
Add Indexer trait to Zaino-state with the following methods:
- [getinfo](https://zcash.github.io/rpc/getinfo.html)
- [getblockchaininfo](https://zcash.github.io/rpc/getblockchaininfo.html)
- [getaddressbalance](https://zcash.github.io/rpc/getaddressbalance.html)
- [send_raw_transaction](https://zcash.github.io/rpc/sendrawtransaction.html)
- [getblock](https://zcash.github.io/rpc/getblock.html)
- [getrawmempool](https://zcash.github.io/rpc/getrawmempool.html)
- [z_gettreestate](https://zcash.github.io/rpc/z_gettreestate.html)
- [z_getsubtreesbyindex](https://zcash.github.io/rpc/z_getsubtreesbyindex.html)
- [get_raw_transaction](https://zcash.github.io/rpc/getrawtransaction.html)
- [getaddresstxids](https://zcash.github.io/rpc/getaddresstxids.html)
- [getaddressutxos](https://zcash.github.io/rpc/getaddressutxos.html)

PR #140, opened 2024-12-03
Read state service

### LINKED ISSUE #110: Task 2.1.1: Create StateService struct in Zaino-State, initially containing Zebra's ReadStateService, and implement sync method (using Zebra's TrustedChainSync) as a persistent background process.
Create StateService struct in Zaino-State, initially containing Zebra's ReadStateService, and implement sync method (using Zebra's TrustedChainSync) as a persistent background process.

idky137:
This is the initial work on integrating Zebras `ReadStateService` into Zaino.

- Adds `StateService` to zaino-state, implements spawn().
- Implements get_info, get_blockchain_info, get_block_header, get_block for StateService.
- Moves AtomicStatus to Zaino-state.

- built atop #138, #139

- Closes #110, #112
- Work towards #111, #114


idky137:
Current issues with zebra's regtest mode have been documented here [https://github.com/zingolabs/zaino/issues/146, https://github.com/zingolabs/zaino/issues/145, https://github.com/zingolabs/zaino/issues/144] and highlighted in code. This code should be merged to allow continued development in zaino-state and should be retested once zebra regtest is running more smoothly.

idky137:
There are several unwraps in zaino-state::state.rs that will be removed in https://github.com/zingolabs/zaino/pull/142.

zaino-proto/src/proto/service.rs
```diff
@@ -277,8 +277,8 @@ impl ShieldedProtocol {
 /// Generated client implementations.
 pub mod compact_tx_streamer_client {
     #![allow(unused_variables, dead_code, missing_docs, clippy::let_unit_value)]
-    use tonic::codegen::*;
     use tonic::codegen::http::Uri;
+    use tonic::codegen::*;
```
AloeareV, line 281: What's being used from codegen? Glob imports are a big hit to readability imo.

AloeareV: 1. I realize that this was a format run that reordered imports, and the glob import wasn't added in this PR.

zaino-state/src/state.rs
```diff
@@ -0,0 +1,1058 @@
...
+            read_state_service,
+            _latest_chain_tip: latest_chain_tip,
+            _chain_tip_change: chain_tip_change,
+            sync_task_handle,
+            _rpc_client: rpc_client,
+            data,
+            status: AtomicStatus::new(StatusType::Spawning.into()),
+        };
+
+        state_service.status.store(StatusType::Syncing.into());
```
AloeareV, line 97: As far as I can tell this line happens before anything has a chance to access the status, why set it to spawning in the first place?

zaino-state/src/state.rs
```diff
@@ -0,0 +1,1058 @@
...
+
+        poll_fn(|cx| state_service.read_state_service.poll_ready(cx)).await?;
+
+        state_service.status.store(StatusType::Ready.into());
```
AloeareV, line 101: Same as above, this status has just been created on line 87, and no other processes have access to it to read it. Why not just set the status as ready when state_service is created?

zaino-state/src/state.rs
```diff
@@ -0,0 +1,1058 @@
...
+                self.status.store(StatusType::Busy.into());
+                std::task::Poll::Pending
+            }
+        })
+        .await
+    }
+
+    /// Fetches the current status
+    pub fn status(&self) -> StatusType {
+        self.status.load().into()
```
AloeareV, line 144: I assume this function exists so that a best guess as to the status can be accessed from a synchronous context? In terms of correctness, should any status check offload to fetch_status_from_validator? Will that be run on a loop regularly enough that this field is a reasonable-enough guess?

zaino-state/src/state.rs
```diff
@@ -0,0 +1,1058 @@
...
+            .await?;
+        let zebra_state::ReadResponse::BlockHeader { header, .. } = block_header else {
+            return Err(StateServiceError::Custom(
+                "Unexpected response type for BlockHeader".into(),
+            ));
+        };
+        let tip_block_time = header.time;
+        let now = Utc::now();
+        let zebra_estimated_height =
+            NetworkChainTipHeightEstimator::new(tip_block_time, tip_height, &network)
```
AloeareV, line 227: I don't understand this TipHeightEstimation business. How would the tip block time be greater than now? I don't see how anything could be learned from that, other than whether or not the caller's system clock is functioning correctly.

AloeareV (approved):

PR #142, opened 2024-12-17
Get block range

idky137:
- Adds get_compact_block and get_block_range methods to StateService. 
 - Moves Stream response types from zaino-serve to zaino-state.
 - Work towards https://github.com/zingolabs/zaino/issues/111, https://github.com/zingolabs/zaino/issues/114.

- Built atop #139, #140

Issue #121, opened 2024-11-20
Task 2.4.6: Add test set that comprehensively tests mempool functionality


idky137:
- Blocked on #96

Issue #119, opened 2024-11-20
Task 2.4.4 Implement Mempool for `FetchService`


idky137:
- Blocked on #122

PR #152, opened 2025-01-03
Update mempool

### LINKED ISSUE #116: Task 2.4.1: Implement efficient Broadcast stream for high concurrency use case
Zaino's internal mempool will need to be able to serve a large number of clients concurrently. For this reason a simple broadcast may not be the best approach. We require lock free read access from an arbitrary number of client tasks and clone free update of the channel.

One possible approach is to use evmap (or possible dashmap) as a lock free map and either a broadcast or watch to alert listeners of updates.

This should be build so it can be used for any efficient broadcast streams required in Zaino (eg. the non-finalised state in #122)

idky137:
Adds the new mempool implementation in zaino-state and underlying "Broadcast" struct.

 - Adds "Broadcast" struct that uses dashmap for efficient state distribution.
 - Adds Mempool struct to zaino-state
 - (Adds Memepool trait)
 - Adds Mempool to FetchService and implements remaining grpc methods for FetchService.

 - Closes https://github.com/zingolabs/zaino/issues/116, https://github.com/zingolabs/zaino/issues/117, https://github.com/zingolabs/zaino/issues/119, https://github.com/zingolabs/zaino/issues/121

 - Build atop Read state service #140, Get block range #142, Add Fetch service to Zaino-State #149, #150. As uses functionality in those PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -931,6 +931,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "dashmap"
+version = "6.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5041cc499144891f3790297212f32a74fb938e5136a14943f338ef9e0ae276cf"
+dependencies = [
+ "cfg-if",
+ "crossbeam-utils",
+ "hashbrown 0.14.5",
+ "lock_api",
+ "once_cell",
+ "parking_lot_core",
+]
+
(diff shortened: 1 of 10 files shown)


idky137:
As FetchService and StateService will now use the same mempool implementation a separate mempool trais will not be used.

AloeareV (approved):

PR #153, opened 2025-01-07
Add indexer service

idky137:
This PR ties together work from the previous several PRs implementing `FetchService` and the mempool update by adding a generic `IndexerService` and updating zaino-serve to use this struct.

- Add generic IndexerService and IndexerServiceSubscriber structs to zaino-state::indexer.rs.
- Adds IndexerService<FetchService> to zainod::indexer::Indexer.
- Adds IndexerServiceSubscriber<FetchServiceSubscriber> to zaino-serve::{Server, Workerpool, Worker, GrpcClient}.
- Updates `impl CompactTxStreamerServer for GrpcServer` to use IndexerServiceSubscriber.
- Propagates changes throughout code.
- Removes outdated code from zaino-fetch and zaino-serve.

PR #156, opened 2025-01-09
Add macro helpers

AloeareV:
Builds atop #153

@idky137 this doesn't yet work for the few service methods that don't pass their argument in to the dispatched function call. This shouldn't be too hard to extend...but I am worried that this macro makes this file too hard to read/modify, and I might want to rewrite it from scratch with a focus on something more readable. What do you think?

### DIFF
--- a/zaino-serve/src/rpc/service.rs
+++ b/zaino-serve/src/rpc/service.rs
@@ -51,6 +51,17 @@ macro_rules! client_method_helper {
                 .await?,
         ))
     };
+    // for the no-input variant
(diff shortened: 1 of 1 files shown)


idky137:
This is brilliant! 

Are the methods that require more work the ones that take no input of the one that takes the streaming input? Could we update either the `LightwalletIndexer` trait to add empty inputs or update the stream types if it would be easier? 

Looks really good though! 

If you have a preference on how to get the last methods working Im happy either way..

idky137 (approved):

PR #158, opened 2025-01-14
Compact block cache pt1

idky137:
Adds the local compact block cache to Zaino-State.

 - Built atop #153


AloeareV (approved):
Reviewed via screenshare!

PR #163, opened 2025-01-20
Compact block cache pt2

idky137:
This is the second PR adding the internal compact block cache to the FetchService in zaino-state. 

- Updates Zaino's config files and example config (zindexer.toml) by adding the new options required for integrating the BlockCache into the FetchService.
- Removes old (unused) config options.
- Updates Zaino-Testutils TestManager to handle new config options.
- Adds several review changes to the BlockCache.

The actual BlockCache integration will be done in the following PR.

### DIFF
--- a/zaino-state/src/fetch.rs
+++ b/zaino-state/src/fetch.rs
@@ -393,11 +393,7 @@ impl ZcashIndexer for FetchServiceSubscriber {
     ) -> Result<GetSubtrees, Self::Error> {
         Ok(self
             .fetcher
-            .get_subtrees_by_index(
-                pool,
-                start_index.0,
-                limit.map(|limit_index| limit_index.0),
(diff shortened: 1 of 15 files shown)


zaino-state/src/fetch.rs
```diff
@@ -2379,8 +2461,18 @@ mod tests {
...
                 None,
                 None,
+                None,
+                None,
+                test_manager
+                    .local_net
+                    .data_dir()
+                    .path()
+                    .to_path_buf()
+                    .join("zaino"),
```
AloeareV, line 2471: I see these same lines in several places in this file. Helper fn?

AloeareV (approved):
I'm noticing a lot of boilerplate in the tests, I think that is where I will focus next!

PR #167, opened 2025-01-24
propose new names to clarify roles of components in the IndexerInterface

zancas:
Not sure if this is the structure that's being suggested in the doc-comment?

### DIFF
--- a/zaino-serve/src/server/director.rs
+++ b/zaino-serve/src/server/director.rs
@@ -17,7 +17,7 @@ use crate::server::{
 };
 use zaino_state::{
     fetch::FetchServiceSubscriber,
-    indexer::IndexerSubscriber,
+    indexer::ChainStateInterface,
     status::{AtomicStatus, StatusType},
 };
 
@@ -73,7 +73,7 @@ impl Server {
     /// Spawns a new Server.
     #[allow(clippy::too_many_arguments)]
     pub async fn spawn(
-        service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+        service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
         tcp_ingestor_listen_addr: SocketAddr,
         max_queue_size: u16,
         max_worker_pool_size: u16,
--- a/zaino-serve/src/server/worker.rs
+++ b/zaino-serve/src/server/worker.rs
@@ -17,7 +17,7 @@ use crate::{
 use zaino_proto::proto::service::compact_tx_streamer_server::CompactTxStreamerServer;
 use zaino_state::{
     fetch::FetchServiceSubscriber,
-    indexer::IndexerSubscriber,
+    indexer::ChainStateInterface,
     status::{AtomicStatus, StatusType},
 };
 
@@ -50,13 +50,13 @@ impl Worker {
         _worker_id: usize,
         queue: QueueReceiver<ZingoIndexerRequest>,
         requeue: QueueSender<ZingoIndexerRequest>,
-        service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+        service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
         service_status: AtomicStatus,
         atomic_status: AtomicStatus,
         online: Arc<AtomicBool>,
     ) -> Self {
         let grpc_client = GrpcClient {
-            service_subscriber: service_subscriber.clone(),
+            chainstate_interface: service_subscriber.clone(),
             online: online.clone(),
         };
         Worker {
@@ -189,7 +189,7 @@ impl WorkerPoolStatus {
 #[derive(Clone)]
 pub(crate) struct WorkerPool {
     /// Chain fetch service subscriber.
-    service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+    service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
     /// Service status.
     service_status: AtomicStatus,
     /// Maximum number of concurrent workers allowed.
@@ -208,7 +208,7 @@ impl WorkerPool {
     /// Creates a new worker pool containing [idle_workers] workers.
     #[allow(clippy::too_many_arguments)]
     pub(crate) async fn spawn(
-        service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+        service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
         service_status: AtomicStatus,
         max_size: u16,
         idle_size: u16,
--- a/zaino-state/src/fetch.rs
+++ b/zaino-state/src/fetch.rs
@@ -5,7 +5,7 @@ use std::time;
 use crate::{
     config::FetchServiceConfig,
     error::FetchServiceError,
-    indexer::{IndexerSubscriber, LightWalletIndexer, ZcashIndexer, ZcashService},
+    indexer::{ChainStateInterface, LightWalletIndexer, ZcashIndexer, ZcashService},
     mempool::{Mempool, MempoolSubscriber},
     status::{AtomicStatus, StatusType},
     stream::{
@@ -135,8 +135,8 @@ impl ZcashService for FetchService {
     }
 
     /// Returns a [`FetchServiceSubscriber`].
-    fn get_subscriber(&self) -> IndexerSubscriber<FetchServiceSubscriber> {
-        IndexerSubscriber::new(FetchServiceSubscriber {
+    fn get_chainstate(&self) -> ChainStateInterface<FetchServiceSubscriber> {
+        ChainStateInterface::new(FetchServiceSubscriber {
             fetcher: self.fetcher.clone(),
             mempool: self.mempool.subscriber(),
             data: self.data.clone(),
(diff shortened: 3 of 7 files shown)


idky137 (commented):
I think this is definitely better. I am wondering if it is `IndexerServiceSubscriber` that should be renamed to `IndexerInterface` though. 

The `IndexerService` is the struct that holds internal chain "state" data and is responsible for handling child processes but does not have access to the actual chain fetch and tx submission methods.

It is the `IndexerServiceSubscriber` that is designed to be the actual chain fetcher. This has been done to allow a single struct to be responsible for maintaining states / processes and *any number of lightweight "subscribers" to be spawned to fetch data concurrently.

We will be adding a `no_state` config option (it is currently in the config but unused) to allow wallets to spawn a very lightweight `IndexerServiceSubscriber` that doesnt have any of the extra processes of internal states.

idky137:
What about something like this?
IndexerService -> IndexerManager (or IndexerStateManager?)
IndexerServiceSubscriber -> IndexerInterface

zancas:
Based on what you've said here..  I am wondering why "Indexer" is part of the name of IndexerService (it's already addressed at `indexer::`), is it a:

`ChainStateProcessHandler`

?

idky137:
I generally like having a descriptive name within a mod, that way if someone imports `zaino_state::indexer;` it is still clear within code what is being used. I not fixed on this though. 

We could use `ChainStateProcessHandler` and `ChainStateInterface`?

zancas:
OK it sounds like we have settled on `ChainStateProcessHandler` and `ChainStateInterface`, I will make the changes.

zancas:
@idky137 new commit pushed.

zaino-state/src/indexer.rs
```diff
@@ -90,18 +90,18 @@ pub trait ZcashService: Sized {
 ///
 /// The future plan is to also add a TonicServiceSubscriber and DarksideServiceSubscriber to this to enable wallets to use a single unified chain fetch service.
 #[derive(Clone)]
-pub struct IndexerSubscriber<Subscriber: Clone + ZcashIndexer + LightWalletIndexer> {
+pub struct ChainStateInterface<Subscriber: Clone + ZcashIndexer + LightWalletIndexer> {
     /// Underlying Service Subscriber.
     subscriber: Subscriber,
```
idky137, line 95: This should change to `interface` to go with the new names

zaino-serve/src/server/director.rs
```diff
@@ -73,7 +73,7 @@ impl Server {
     /// Spawns a new Server.
     #[allow(clippy::too_many_arguments)]
     pub async fn spawn(
-        service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+        service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
```
idky137, line 76: should this change to `service_interface` or `chain_state_interface`?

zaino-serve/src/server/worker.rs
```diff
@@ -50,13 +50,13 @@ impl Worker {
         _worker_id: usize,
         queue: QueueReceiver<ZingoIndexerRequest>,
         requeue: QueueSender<ZingoIndexerRequest>,
-        service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+        service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
```
idky137, line 53: similarly to above, should this change to `service_interface` or `chain_state_interface`

zaino-serve/src/server/worker.rs
```diff
@@ -208,7 +208,7 @@ impl WorkerPool {
     /// Creates a new worker pool containing [idle_workers] workers.
     #[allow(clippy::too_many_arguments)]
     pub(crate) async fn spawn(
-        service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+        service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
```
idky137, line 211: same as above

zaino-serve/src/server/worker.rs
```diff
@@ -189,7 +189,7 @@ impl WorkerPoolStatus {
 #[derive(Clone)]
 pub(crate) struct WorkerPool {
     /// Chain fetch service subscriber.
-    service_subscriber: IndexerSubscriber<FetchServiceSubscriber>,
+    service_subscriber: ChainStateInterface<FetchServiceSubscriber>,
```
idky137, line 192: same as above

zaino-state/src/fetch.rs
```diff
@@ -135,8 +135,8 @@ impl ZcashService for FetchService {
     }
 
     /// Returns a [`FetchServiceSubscriber`].
-    fn get_subscriber(&self) -> IndexerSubscriber<FetchServiceSubscriber> {
-        IndexerSubscriber::new(FetchServiceSubscriber {
+    fn get_chainstate(&self) -> ChainStateInterface<FetchServiceSubscriber> {
```
idky137, line 138: i think this should be `get_interface()`, as we are fetching an `*Interface` from a `*ProcessHandler`

zaino-state/src/indexer.rs
```diff
@@ -77,7 +77,7 @@ pub trait ZcashService: Sized {
     async fn spawn(config: Self::Config, status: AtomicStatus) -> Result<Self, Self::Error>;
 
     /// Returns a [`ServiceSubscriber`].
-    fn get_subscriber(&self) -> IndexerSubscriber<Self::Subscriber>;
+    fn get_chainstate(&self) -> ChainStateInterface<Self::Subscriber>;
```
idky137, line 80: i think this should be get_interface(), as we are fetching an *`Interface` from a *`ProcessHandler`

zaino-state/src/indexer.rs
```diff
@@ -29,23 +29,23 @@ use crate::{
 ///
 /// The future plan is to also add a TonicService and DarksideService to this to enable wallets to use a single unified chain fetch service.
 ///
-/// NOTE: Work to implement a unified endpoint for IndexerService will be completed in Milestone 3 of the Zaino Dev Grant.
+/// NOTE: Work to implement a unified endpoint for ChainStateProcessHandler will be completed in Milestone 3 of the Zaino Dev Grant.
 #[derive(Clone)]
-pub struct IndexerService<Service: ZcashService> {
+pub struct ChainStateProcessHandler<Service: ZcashService> {
```
idky137, line 34: Should the inner here change to `ProcessHandler` to work with new names?

zaino-state/src/indexer.rs
```diff
@@ -90,18 +90,18 @@ pub trait ZcashService: Sized {
 ///
 /// The future plan is to also add a TonicServiceSubscriber and DarksideServiceSubscriber to this to enable wallets to use a single unified chain fetch service.
 #[derive(Clone)]
-pub struct IndexerSubscriber<Subscriber: Clone + ZcashIndexer + LightWalletIndexer> {
+pub struct ChainStateInterface<Subscriber: Clone + ZcashIndexer + LightWalletIndexer> {
```
idky137, line 93: Should the inner `Subscriber` here change to `Interface` to go with new names?

zaino-state/src/indexer.rs
```diff
@@ -90,18 +90,18 @@ pub trait ZcashService: Sized {
...
 }
 
-impl<Subscriber> IndexerSubscriber<Subscriber>
+impl<Subscriber> ChainStateInterface<Subscriber>
 where
     Subscriber: Clone + ZcashIndexer + LightWalletIndexer,
 {
-    /// Creates a new [`IndexerSubscriber`].
+    /// Creates a new [`ChainStateInterface`].
     pub fn new(subscriber: Subscriber) -> Self {
```
idky137, line 103: similarly to above

zainod/src/indexer.rs
```diff
@@ -57,7 +57,7 @@ pub struct Indexer {
     /// GRPC server.
     server: Option<Server>,
     /// Internal block cache.
-    _service: IndexerService<FetchService>,
+    _service: ChainStateProcessHandler<FetchService>,
```
idky137, line 60: should this name change?

zainod/src/indexer.rs
```diff
@@ -99,7 +99,7 @@ impl Indexer {
             zebrad_uri
         );
         status.indexer_status.store(StatusType::Spawning.into());
-        let service = IndexerService::<FetchService>::spawn(
+        let service = ChainStateProcessHandler::<FetchService>::spawn(
```
idky137, line 102: similarly to above

zaino-serve/src/server/director.rs
zancas: I am in favor of fields of type "ChainStateInterface" being `chainstate`, since fields are generally assumed to be interfaces to something.

zaino-state/src/fetch.rs
zancas: Oh.. I didn't understand that it could be an interface to something other than a ChainState!

zancas: Where are the other *Interfaces?

idky137:
The only possible changes, should the inner of ChainStateInterface be interface instead of subscriber? Also the same with the ChainStateInterface::new() method input?

idky137 (approved):

zancas:
We will revisit naming later.

PR #168, opened 2025-01-25
Update docs

zancas:
I'm unclear on whether validator can now be used in place of miner, but I noticed that validator is used earlier in the doc, and then "miner" shows up later.  Maybe we can just switch to "validator" in all cases?


idky137:
This is actually an important distinction, maybe it could be rewritten better though. 

Zaino is taking over responsibility for serving all "non-miner" clients from zcashd, with this change, zebra is now only responsible for serving zaino or serving "miner clients" direct. So we should keep the distinction in the docs.

idky137 (changes requested):
distinction between miner and non-miner clients should be kept as is important use case detail.

zancas:
I understand that miner and non-miner are important distinctions..  my question was about changing all instances of the the word "miner" to "validator".

"miner" makes more sense in the context of `zcashd` but the word "validator" is used e.g. in the type system.

To put it another way, eventually there won't be **any** miners, but zaino will still be serving validators (once CrossLink lands).

idky137:
Ahh sorry I understand now. Yes  I am with this change.

idky137 (approved):
Looks good to me, Im not sure if this will conflict with the name change PR though?

zancas (commented):
I am not sure about the term validator and am concerned that it's ambiguous.

PR #169, opened 2025-01-26
Port client rpcs tests to zaino

zancas:
This depends on:

https://github.com/zingolabs/zcash-local-net/pull/75


zancas:
Won't land...

Issue #126, opened 2024-11-20
Task 2.2.1: Add error handling for specific errors to the `TcpIngestor`, `Worker` and `Director` in `Zaino-Serve`


idky137:

idky137:
This work was completed in #180 and #181

PR #170, opened 2025-01-27
Compact block cache pt3

idky137:
Final integration of the Compact Block Cache.

- Adds BlockCache to FetchService and uses BlockCache block fetch methods.
- Updates DB to use big-endian bytes for height instead of `Height(u32)` (zebra-chain). This is required to search LMDB height keys by order.
- Moves startup sync from FetchService::Spawn to BlockCache::Spawn.

### DIFF
--- a/docs/use_cases.md
+++ b/docs/use_cases.md
@@ -12,7 +12,7 @@
 3) Run `$ zebrad --config #PATH_TO_CONF/zebrad.toml start`
 4) Run `$ zainod --config #PATH_TO_CONF/zindexer.toml`
 
-NOTE: Unless the `no_db` option is set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits durin this process the database is saved in its current state, enabling the chain to be synced in several stages.
+NOTE: Unless the `no_db` option is set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits during this process the database is saved in its current state, enabling the chain to be synced in several stages.
 
 - To launch Zingo-Cli running through Zaino [from #PATH_TO/zingolib]:
 5) Run `$ cargo run --release --package zingo-cli -- --chain "CHAIN_TYPE" --server "ZAINO_LISTEN_ADDR" --data-dir #PATH_TO_WALLET_DATA_DIR`
--- a/docs/use_cases.md
+++ b/docs/use_cases.md
@@ -12,7 +12,7 @@
 3) Run `$ zebrad --config #PATH_TO_CONF/zebrad.toml start`
 4) Run `$ zainod --config #PATH_TO_CONF/zindexer.toml`
 
-NOTE: Unless the `no_db` option if set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits durin this process the database is saved in its current state, enabling the chain to be synced in several stages.
+NOTE: Unless the `no_db` option is set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits durin this process the database is saved in its current state, enabling the chain to be synced in several stages.
 
 - To launch Zingo-Cli running through Zaino [from #PATH_TO/zingolib]:
 5) Run `$ cargo run --release --package zingo-cli -- --chain "CHAIN_TYPE" --server "ZAINO_LISTEN_ADDR" --data-dir #PATH_TO_WALLET_DATA_DIR`
--- a/docs/use_cases.md
+++ b/docs/use_cases.md
@@ -12,6 +12,8 @@
 3) Run `$ zebrad --config #PATH_TO_CONF/zebrad.toml start`
 4) Run `$ zainod --config #PATH_TO_CONF/zindexer.toml`
 
+NOTE: Unless the `no_db` option if set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits durin this process the database is saved in its current state, enabling the chain to be synced in several stages.
+
 - To launch Zingo-Cli running through Zaino [from #PATH_TO/zingolib]:
 5) Run `$ cargo run --release --package zingo-cli -- --chain "CHAIN_TYPE" --server "ZAINO_LISTEN_ADDR" --data-dir #PATH_TO_WALLET_DATA_DIR`
 
--- a/zaino-state/src/config.rs
+++ b/zaino-state/src/config.rs
@@ -162,3 +162,17 @@ impl BlockCacheConfig {
         }
     }
 }
+
+impl From<FetchServiceConfig> for BlockCacheConfig {
+    fn from(value: FetchServiceConfig) -> Self {
+        Self {
+            map_capacity: value.map_capacity,
+            map_shard_amount: value.map_shard_amount,
+            db_path: value.db_path,
+            db_size: value.db_size,
+            network: value.network,
+            no_sync: value.no_sync,
+            no_db: value.no_db,
+        }
+    }
+}
--- a/zaino-state/src/fetch.rs
+++ b/zaino-state/src/fetch.rs
@@ -1,6 +1,31 @@
 //! Zcash chain fetch and tx submission service backed by zcashds JsonRPC service.
 
+use futures::StreamExt;
+use hex::FromHex;
 use std::time;
+use tokio::{sync::mpsc, time::timeout};
+use tonic::async_trait;
+
+use zebra_chain::subtree::NoteCommitmentSubtreeIndex;
+use zebra_rpc::methods::{
+    trees::{GetSubtrees, GetTreestate},
+    AddressBalance, AddressStrings, GetAddressTxIdsRequest, GetAddressUtxos, GetBlock,
+    GetBlockChainInfo, GetInfo, GetRawTransaction, SentTransactionHash,
+};
+
+use zaino_fetch::{
+    chain::{transaction::FullTransaction, utils::ParseFromSlice},
+    jsonrpc::connector::{test_node_and_return_uri, JsonRpcConnector, RpcError},
+};
+use zaino_proto::proto::{
+    compact_formats::CompactBlock,
+    service::{
+        AddressList, Balance, BlockId, BlockRange, Duration, Exclude, GetAddressUtxosArg,
+        GetAddressUtxosReply, GetAddressUtxosReplyList, GetSubtreeRootsArg, LightdInfo,
+        PingResponse, RawTransaction, SendResponse, ShieldedProtocol, SubtreeRoot,
+        TransparentAddressBlockFilter, TreeState, TxFilter,
+    },
+};
 
 use crate::{
     config::FetchServiceConfig,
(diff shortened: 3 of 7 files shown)


zaino-state/src/local_cache/non_finalised_state.rs
```diff
@@ -54,8 +54,40 @@ impl NonFinalisedState {
             config,
         };
 
+        // If no_db is active wait for server to sync with p2p network.
+        if non_finalised_state.config.no_db
```
pacu: can this be factored out to a function of non_finalized_state that has meaningful name? It seems a pretty specific condition that is being described here.

zaino-state/src/local_cache/non_finalised_state.rs
```diff
@@ -54,8 +54,40 @@ impl NonFinalisedState {
...
+            non_finalised_state.status.store(StatusType::Syncing.into());
+            loop {
+                let blockchain_info = fetcher.get_blockchain_info().await?;
+                if (blockchain_info.blocks.0 as i64 - blockchain_info.estimated_height.0 as i64)
+                    .abs()
+                    <= 10
+                {
+                    break;
+                } else {
+                    println!(" - Validator syncing with network. Validator chain height: {}, Estimated Network chain height: {}",
```
pacu: should this be a log entry instead?

zaino-state/src/local_cache/non_finalised_state.rs
```diff
@@ -286,8 +274,11 @@ impl NonFinalisedState {
             };
         }
 
-        // Refill from reorg_height[+1].
-        for block_height in (reorg_height.0 + 1)..self.fetcher.get_blockchain_info().await?.blocks.0
+        // Refill from max(reorg_height[+1], sapling_activation_height).
```
pacu, line 255: shouldn't you refill the cache from reorg_height - 1?

zaino-state/src/local_cache/non_finalised_state.rs
```diff
@@ -286,8 +274,11 @@ impl NonFinalisedState {
...
+        let validator_height = self.fetcher.get_blockchain_info().await?.blocks.0;
+        for block_height in ((reorg_height.0 + 1)
+            .max(self.config.network.sapling_activation_height().0))
+            ..=validator_height
         {
             // Either pop the reorged block or pop the oldest block in non-finalised state.
```
pacu, line 263: Document Reorg Logic outside code as well.

zaino-state/src/local_cache/non_finalised_state.rs
```diff
@@ -309,22 +300,29 @@ impl NonFinalisedState {
...
+                            if let Some(block) = self.hashes_to_blocks.get(&hash) {
+                                if self
+                                    .block_sender
+                                    .send((pop_height, *hash, block.as_ref().clone()))
+                                    .await
+                                    .is_err()
+                                {
+                                    self.status.store(StatusType::CriticalError.into());
+                                    return Err(NonFinalisedStateError::Critical(
+                                        "Critical error in database. Closing NonFinalisedState"
```
pacu, line 297: Can the error be more self explanatory given that the conditions of the failure are well-known?

pacu (changes requested):
Left some remarks and requests! Pretty exciting step forward!

zaino-state/src/local_cache/non_finalised_state.rs
idky137: this is being added in #177.

zaino-state/src/local_cache/non_finalised_state.rs
idky137: No the reorg_height is the last block that is valid, so we fill from the next block (the first block we do not have). I will update the comment for this.

zaino-state/src/local_cache/non_finalised_state.rs
idky137: sorry i dont understand where you want extra comments.

zaino-state/src/local_cache/non_finalised_state.rs
idky137: we do not actually no the true error here, just that the finalised state has fully closed. these errors will be handled more completely in https://github.com/zingolabs/zaino/issues/126 and https://github.com/zingolabs/zaino/issues/128

zaino-state/src/local_cache/non_finalised_state.rs
idky137: added in [d65ae0f](https://github.com/zingolabs/zaino/pull/170/commits/d65ae0f7b0f3ab2dac95d60bb4fb9a5596051774)

zaino-state/src/local_cache/non_finalised_state.rs
idky137: [d65ae0f](https://github.com/zingolabs/zaino/pull/170/commits/d65ae0f7b0f3ab2dac95d60bb4fb9a5596051774)

zaino-state/src/fetch.rs
```diff
@@ -523,7 +505,7 @@ impl LightWalletIndexer for FetchServiceSubscriber {
                 ));
             }
         };
-        match self.get_compact_block(&height).await {
+        match self.block_cache.get_compact_block(height.to_string()).await {
```
AloeareV, line 509: Why does block_cache take a string argument?

docs/use_cases.md
```diff
@@ -12,6 +12,8 @@
 3) Run `$ zebrad --config #PATH_TO_CONF/zebrad.toml start`
 4) Run `$ zainod --config #PATH_TO_CONF/zindexer.toml`
 
+NOTE: Unless the `no_db` option if set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits durin this process the database is saved in its current state, enabling the chain to be synced in several stages.
```
zancas: "if" --> "is"

zaino-state/src/config.rs
```diff
@@ -163,3 +163,17 @@ impl BlockCacheConfig {
         }
     }
 }
+
```
zancas, line 166: Maybe the `FetchServiceConfig` should simply refer to a `BlockCacheConfig` that bundles these 8 values?

zaino-state/src/local_cache.rs
```diff
@@ -101,22 +112,64 @@ impl BlockCacheSubscriber {
...
-                true => {
+                None => {
                     let (_, block) = fetch_block_from_node(&self.fetcher, hash_or_height).await?;
                     Ok(block)
                 }
             }
         }
     }
 
+    /// Returns a compact block holding only action nullifiers.
```
AloeareV, line 130: Why does this only supply nullifiers for orchard actions and not sapling outputs? Should the name reflect this?

zaino-state/src/local_cache.rs
```diff
@@ -101,22 +112,64 @@ impl BlockCacheSubscriber {
...
+    pub async fn get_compact_block_nullifiers(
```
zancas, line 133: Should this name mention orchard specifically?

E.G.

`get_compact_block_orchard_nullifiers`

AloeareV (dismissed):
Great! Approved, once the conversations are resolved!

zaino-state/src/fetch.rs
idky137: The block cache takes a string representing a Zebra `HashOrHeight`, so internally blocks are fetchable by either option. 

Currently the first line of `BlockCache::get_compact_block` is `let hash_or_height: HashOrHeight = hash_or_height.parse()?;` 

This could be changed so that `get_compact_block` and `get_compact_block_nullifiers` take a HashOrHeight directly but I think using a string is a nicer interface as a `String(hash)` can be passed directly and a `Height(u32)` can be very easily passed using `height.0.to_string()`.

Currently LightWalletD only supports fetching blocks by height so the `get_block` gRPC services replicate this behaviour but it will be very easy for us to add fetch by hash with this internal functionality.

zaino-state/src/local_cache.rs
idky137: This is replicating LightWalledD behaviour, we could update this but currently wallets expect only orchard nullifiers: https://github.com/zcash/lightwalletd/blob/339b6d37e839d27bbd167ed02627ab7ab4d7051f/frontend/service.go#L159

zaino-state/src/local_cache.rs
idky137: This comes directly from librustzcash proto [https://github.com/zcash/librustzcash/blob/612bc03d1bce415bdba902277c4c68c396b22913/zcash_client_backend/proto/service.proto#L165] so we should not change the name yet but In will update the comment.

There is a similar situation with `get_taddress_txids`, I believe this RPC should also be renamed [https://github.com/zingolabs/zaino/blob/38e167aed850236c33223965f3db01679d07cde9/zaino-serve/src/rpc/service.rs#L130].

My plan is to put forward a ZIP that unifies all of Zcash'd `Indexer RPCs` as several need to be renamed and there are several that overlap, but I think this should be done once we are ready for zcashd deprecation.

zaino-state/src/config.rs
idky137: I was thinking of doing this and I am happy to make the change. 

The thing that makes this approach not quite as simple is that `map_capacity` and `map_shard_amount` are also used as config for the `DashMaps` used in the `Mempool`. 

We could split this and have config options for both but until we are doing large scale testing there is not really any point in testing these options at all..

zaino-state/src/local_cache.rs
idky137: regarding the name I have been following the RPC names directly to show where the functionality for each is but I happy ton update it.

zaino-state/src/local_cache/non_finalised_state.rs
idky137: Ive updated the doc comment for the method in [[dfc2f75](https://github.com/zingolabs/zaino/pull/170/commits/dfc2f7504e65fb6fe20e8219e0d7143167162b67)], is this what you were thinking?

docs/use_cases.md
idky137: updated in [dfc2f75](https://github.com/zingolabs/zaino/pull/170/commits/dfc2f7504e65fb6fe20e8219e0d7143167162b67)

zaino-state/src/local_cache.rs
idky137: comment updated in [dfc2f75](https://github.com/zingolabs/zaino/pull/170/commits/dfc2f7504e65fb6fe20e8219e0d7143167162b67).

zaino-state/src/fetch.rs
AloeareV: The thoughts going through my head were that all the callsites I saw to block_cache.get_compact_block were from a place where there was a known height/hash. Having it take a string means that we stringify the argument, pass it in, parse the string back into a height/hash, and then we have error handling for a situation that could have been known at the compiler level to already be the right type.

docs/use_cases.md
```diff
@@ -12,6 +12,8 @@
 3) Run `$ zebrad --config #PATH_TO_CONF/zebrad.toml start`
 4) Run `$ zainod --config #PATH_TO_CONF/zindexer.toml`
 
+NOTE: Unless the `no_db` option is set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits durin this process the database is saved in its current state, enabling the chain to be synced in several stages.
```
pacu:
```suggestion
NOTE: Unless the `no_db` option is set to true in the config file zaino will sync its internal `CompactBlock` cache with the validator it is connected to on launch. This can be a very slow process the first time Zaino's DB is synced with a new chain and zaino will not be operable until the database is fully synced. If Zaino exits during this process the database is saved in its current state, enabling the chain to be synced in several stages.
```

zaino-state/src/local_cache/finalised_state.rs
```diff
@@ -21,6 +25,22 @@ use crate::{
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
 struct DbHeight(pub Height);
 
+impl DbHeight {
+    /// Converts `[DbHeight]` to 4-byte **big-endian** bytes.
+    /// Used when storing as an LMDB key.
+    fn to_be_bytes(self) -> [u8; 4] {
+        self.0 .0.to_be_bytes()
```
pacu, line 32:
```suggestion
        self.0.0.to_be_bytes()
```

zaino-state/src/local_cache/non_finalised_state.rs
pacu: I'm referring to explaining how Zaino handles chain re-orgs outside the code itself as something that's of interest of Zaino clients.

pacu: This is [non-blocking] but let's open a github issue document reorg handling on its on documentation section.

pacu (dismissed):
looks good to me provided those small typos are addressed :)

zaino-state/src/local_cache/finalised_state.rs
AloeareV: This whitespace is added by rust-format, as 0.0 can be interpreted by some versions of rust as a floating point number instead of a field access.

AloeareV (approved):

Issue #113, opened 2024-11-19
Task 2.1.4: Implement Indexer for `StateService`.


idky137:
- Implement Indexer for `StateService` (use function impls).

- Blocked on #111, #112

Issue #111, opened 2024-11-19
 Task 2.1.2: Implement Zcash RPC functionality required by lightwallet services.


idky137:
Implement Zcash RPC functionality required by lightwallet services. 
This should be implemented as methods on `StateService` with identical functionality to the following Zcash RPCs:
- [x] [[getinfo](https://zcash.github.io/rpc/getinfo.html)]
- [x] [[getblockchaininfo](https://zcash.github.io/rpc/getblockchaininfo.html)]
- [x] [[getaddressbalance](https://zcash.github.io/rpc/getaddressbalance.html)]
- [x] [[send_raw_transaction](https://zcash.github.io/rpc/sendrawtransaction.html)]
- [x] [[getblock](https://zcash.github.io/rpc/getblock.html)]
- [x] [[getrawmempool](https://zcash.github.io/rpc/getrawmempool.html)]
- [x] [[z_gettreestate](https://zcash.github.io/rpc/z_gettreestate.html)]
- [x] [[z_getsubtreesbyindex](https://zcash.github.io/rpc/z_getsubtreesbyindex.html)]
- [x] [[get_raw_transaction](https://zcash.github.io/rpc/getrawtransaction.html)]
- [x] [[getaddresstxids](https://zcash.github.io/rpc/getaddresstxids.html)]
- [x] [[getaddressutxos](https://zcash.github.io/rpc/getaddressutxos.html)]

The following RPCs will be completed in #59:
- [getaddressdeltas](https://zcash.github.io/rpc/getaddressdeltas.html)
- [getaddressutxos](https://zcash.github.io/rpc/getaddressutxos.html)
- [getblockcount](https://zcash.github.io/rpc/getblockcount.html)
- [getblockdeltas](https://zcash.github.io/rpc/getblockdeltas.html)
- [getblockhash](https://zcash.github.io/rpc/getblockhash.html)
- [getblockhashes](https://zcash.github.io/rpc/getblockhashes.html)
- [getblockheader](https://zcash.github.io/rpc/getblockheader.html)
- [getchaintips](https://zcash.github.io/rpc/getchaintips.html)
- [getdifficulty](https://zcash.github.io/rpc/getdifficulty.html)
- [getmempoolinfo](https://zcash.github.io/rpc/getmempoolinfo.html)
- [getspentinfo](https://zcash.github.io/rpc/getspentinfo.html)
- [gettxout](https://zcash.github.io/rpc/gettxout.html)
[...]

Issue #135, opened 2024-11-20
Task 3.2.1: Implement lightwallet gRPC services using `Indexer` trait defined and implemented in Milestone 2.


idky137:
- Blocked on #54, #55, #56

PR #178, opened 2025-01-31
impl ZcashIndexer for StateService

AloeareV:
This is in need of testing, which I plan to start on today.  
Completes #111 and #113


idky137 (commented):
Looks really good!

The only things i can see without more testing is that the `impl StateService` on line 770 is now mostly defunct. I think it is only the `get_block_header` method that is now used. 

Maybe this should be grouped in an `impl StateService` with the `get_compact_block` and `to_tx` methods from the impl block starting on line 1222.

The `get_block_range` could be left in its own impl that will be converted to the LightWalletIndexer trait in Milestone 3?

The other thing is that mempool data is also available by adding a `zaino-state::Mempool` to `StateService`. This would require the `StateService` to also implement a `StateServiceSubscriber` in the same way `FetchService` does so it may be best to wait until we implement the `LightWalletIndexer` trait in Milestone 3 for this work.

idky137 (dismissed):

idky137 (approved):

zaino-state/src/fetch.rs
```diff
@@ -1825,7 +1825,7 @@ mod tests {
             .unwrap();
 
         dbg!(recipient_balance.clone());
-        dbg!(fetch_service_balance.clone());
+        dbg!(fetch_service_balance);
```
pacu, line 1828: shouldn't these be log calls?

zaino-state/src/state.rs
```diff
@@ -949,34 +1056,171 @@ impl StateService {
...
+            chain_metadata,
+        };
+
+        Ok(compact_block)
+    }
+}
+
+/// This impl will hold the Lightwallet RPC method implementations for StateService.
+///
+/// TODO: Update this to be `impl LightWalletIndexer for StateService` once rpc methods are implemented and tested (or implement separately).
```
pacu, line 1092: track with github issue so it's actionable by others

zaino-state/src/state.rs
```diff
@@ -1056,12 +1300,53 @@ fn tx_to_compact(
 /// !!! NOTE / TODO: This code should be retested before continued development, once zebra regtest is fully operational.
```
pacu, line 1300: Track on Github. Also this seems kind of a blocker to me so we should track it

pacu (changes requested):

zaino-state/src/fetch.rs
AloeareV: If this was production logging instead of test code than absolutely, but accessing logs during test runs can be less convenient than on-screen logging.

zaino-state/src/state.rs
AloeareV: I believe this is covered by https://github.com/zingolabs/zaino/issues/135

zaino-state/src/state.rs
AloeareV: Tracked!

pacu (approved):

PR #185, opened 2025-02-17
Allow Zaino operators to accept Zcash donations

emersonian:
In order to incentivize more Zaino operators and more decentralization, allow operators to specify a Zcash donation address. Zec.rocks is using this on Lightwalletd to reimburse the hosting expenses of participants in our node workshop, we would love to see Zaino support it as well.

Related PRs:
- [Librustzcash #1707](https://github.com/zcash/librustzcash/pull/1707)
- [Zingolib #1660](https://github.com/zingolabs/zingolib/pull/1660)
- [Lightwalletd #506](https://github.com/zcash/lightwalletd/pull/506)


idky137:
This looks good!

I think it would be good to get the change merged in Librustzcash first to make sure we are staying in line with what wallets expect but Im definitely for this update!

zaino-state/src/config.rs
```diff
@@ -88,6 +88,8 @@ pub struct FetchServiceConfig {
     /// Disables FinalisedState.
     /// Used for testing.
     pub no_db: bool,
+    /// Zcash donation UA address
+    pub donation_address: String,
```
pacu: This should be Optiona<String>

zainod/zindexer.toml
```diff
@@ -101,3 +101,7 @@ no_db = false
 #
  no_state = false
 
+# Specify a Zcash donation UA address
+#
+# This is publicly shared with users to allow them to donate Zcash to support your Zaino instance.
+donation_address = ""
```
pacu: Given this is optional, it not being there shouldn't break anything. Let's change the code so that this value can be omitted and won't break anything

zainod/src/config.rs
```diff
@@ -92,6 +92,8 @@ pub struct IndexerConfig {
     ///
     /// NOTE: Currently unimplemented as will require either a Tonic backend or a JsonRPC server.
     pub no_state: bool,
+    /// Zcash donation UA address
+    pub donation_address: String,
```
pacu: also Optional<String>. Also, this should be validated so that invalid addresses can't be set.

zaino-state/src/fetch.rs
```diff
@@ -1765,6 +1766,7 @@ mod tests {
             Network::new_regtest(Some(1), Some(1)),
             true,
             true,
+            "".to_string(),
```
pacu: this should be `None`

zaino-testutils/src/lib.rs
```diff
@@ -382,6 +382,7 @@ impl TestManager {
                 no_sync: zaino_no_sync,
                 no_db: zaino_no_db,
                 no_state: false,
+                donation_address: "".to_string(),
```
pacu: This should be `None`

pacu (changes requested):

zainod/src/config.rs
emersonian: In my opinion validation is a good v2 feature for this. The Lightwalletd implentation does not yet support validation, and ultimately clients should also validate/sanitize before displaying the string to users.

When displaying donation addresses on Hosh, for participants of the node workshop, I am validating first before displaying.

If you think validation really needs to be in v1 I can try to work on it, but this is already live in lightwalletd on our servers so I'm just trying to keep Zaino up to par.

zaino-state/src/config.rs
```diff
@@ -109,6 +111,7 @@ impl FetchServiceConfig {
         network: zebra_chain::parameters::Network,
         no_sync: bool,
         no_db: bool,
+        donation_address: Option<String>,
```
pacu, line 114: @idky137 shall we include zcash vaildation and make Zaino failed if set up with an invalid address?

pacu (approved):
I approve this PR provided that @idky137 finds that not validating the provided address is not a problem

zancas:
Hey!  What's the state of this feature?  Can this merge?  It looks like there're conflicts?

zaino-state/src/config.rs
emersonian: This should be straightforward with str4d's new helpers:

https://github.com/zcash/librustzcash/pull/1760/files#diff-cf407df070451c5c95e7f21b6bfb79ba5331a45f694a60da52b94d6e8c7be447R334

emersonian:
It merged into librustzcash, so once Zaino is no longer depending on a fork of librustzcash it should be straightforward. Or when Zaino's fork is updated, I can update the PR here.

Address validation should be straightforward now, but it would be nice to have an Error state rather than just "None" - silently failing when someone provides an invalid address feels like bad CLI UX.

https://github.com/zcash/librustzcash/pull/1760/files#diff-cf407df070451c5c95e7f21b6bfb79ba5331a45f694a60da52b94d6e8c7be447R334

idky137:
Ye i think we are close to switching over to LRZ main now, if I find out we are further than I think I will update this branch and merge these changes into Zaino-Proto while we are waiting..

zancas:
@emersonian there's been a lot of work cleaning up tech-debt and updating dependencies over the last week.

I think this may be unblocked?

idky137:
I will add this in a new PR, I think so much has changed that it will be easier than dealing with conflicts. 

We now use up to date LRZ so it should just be a case of updating the proto and the config and linking the 2.

emersonian:
Did this make it in? Not seeing donations in config.rs, we're about to start to donating to node operators on a monthly basis.

PR #227, opened 2025-02-28
fix breaking state service tests

nachog00:
This PR aims to fix some problems that make the `state_service` fail.

A helper function (`create_test_manager_and_state_service`) isnt flexible enough and is being used for both `TESTNET` and `REGTEST` mode tests, withuot adjusting for each case.

Right now, the changes make the test `state_service_regtest_get_block_object` pass, but the helper function needs to be refactored to support other tests accordingly.


nachog00:
This grew stale after #228

PR #230, opened 2025-03-02
update zingolabs/infrastructure rev

ala-mode:
updated infrastructure dependency to most recent commit on `dev`

I don't have tests running smoothly here so I will assert anything about them. `cargo check` finishes without complaint.


AloeareV (approved):
Tests pass locally!

PR #234, opened 2025-03-04
update state tests to testnet

idky137:
updates state tests to run on either regtest or testnet


AloeareV (approved):
Tests are flaky (in part due to https://github.com/zingolabs/infrastructure/issues/43) but all will pass when ran individually when run with a fully-synced testnet zebrad cache!

PR #244, opened 2025-03-14
Lightwalletindexer and zcashindexer for stateservice

### LINKED ISSUE #68: 3.2  Implement lightwallet service using ReadStateService instead of the JSONRPC interface
Re-implement the lightwallet gRPC service using Zaino-State (backed by Zebra's ReadStateService). Legacy functionality relying on Zaino-Fetch (backed by Zebra's JsonRPC service) should be kept as a configurable option.

- Blocked on
    - #54 
    - #55 
    - #56

### Tasks
 - [x] [Task 3.2.1: Implement lightwallet gRPC services using `Indexer` trait defined and implemented in Milestone 2.](https://github.com/zingolabs/zaino/issues/135)
 - [x] [Task 3.2.2: Make fetching service configurable (StateService or FetchService) on system launch.](https://github.com/zingolabs/zaino/issues/136)

AloeareV:
Partial implementation of #68.


zaino-state/src/state.rs
```diff
@@ -1190,6 +1221,19 @@ impl ZcashIndexer for StateServiceSubscriber {
             })
             .collect())
     }
+
+    // Helper function, to get the chain height in rpc implementations
+    async fn chain_height(&self) -> Result<Height, Self::Error> {
```
idky137, line 1229: Should this request height from the internal NonFinalisedState rather than directly from the node?

zaino-state/src/state.rs
```diff
@@ -1198,50 +1242,40 @@ impl LightWalletIndexer for StateServiceSubscriber {
 
     /// Return the height of the tip of the best chain
     async fn get_latest_block(&self) -> Result<BlockId, Self::Error> {
```
idky137, line 1247: Similar to above, should this use `NonFinalisedState::chain_tip()` rather than fetching from the node?

zaino-state/src/state.rs
```diff
@@ -1259,227 +1293,57 @@ impl LightWalletIndexer for StateServiceSubscriber {
...
-        Ok(CompactBlockStream::new(channel_rx))
+        self.get_block_range_inner(request, true).await
     }
 
     /// Return the requested full (not compact) transaction (as from zcashd)
-    async fn get_transaction(&self, _request: TxFilter) -> Result<RawTransaction, Self::Error> {
-        Err(crate::error::StateServiceError::TonicStatusError(tonic::Status::unimplemented(
-            "Not yet implemented. If you require this RPC please open an issue or PR at the Zaino github (https://github.com/zingolabs/zaino.git)."
-        )))
+    async fn get_transaction(&self, request: TxFilter) -> Result<RawTransaction, Self::Error> {
```
idky137, line 1319: It looks like this is currently not checking the mempool for the transaction? Could we delegate this method to `StateService::get_raw_transaction` as that method is already checking both chain and mempool.

idky137 (approved):

PR #250, opened 2025-03-24
Docker builds and optimised CI (signed)

nachog00:
This PR is a review and copy of #184 with squashed and signed commits.

PR #251, opened 2025-03-24
new zaino ci docker

nachog00:
This PR is to be merged after #250 

It adds a new docker image that is used for Zaino CI runs.

Known issues:
- Cache isn't being fully taken advantage of yet
- Chain cache relying tests are failing since cached state isn't being initialized (we should cache it too)
- Tests matrix is stopping all tests execution on first fail


nachog00:
This is running most tests now. A subsequent PR will disable the ones we agreed on @idky137 @AloeareV

nachog00:
The "Build docker image, push if tag ..." job fails only cause there's no dockerhub token configured on this repo yet. It's working fine on my fork.

docker-ci/Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+  
+# Stage 5: Final Image
+FROM rust:latest
+
+LABEL maintainer="nachog00" \
+    version="1.0" \
+    description="This image provides several testing zcash-related executables in $PATH. Use this image in CI workflows for testing, or as a base for running local integration tests." \
+    usage="For CI, run with the container option defined in your workflow YAML and use the checkout action. For local testing, you can run: docker run --rm -v /path/to/zaino/repo:/app zaino-ci [nextest-options]" \
+    zcashd.version="v6.1.0" \
+    lightwalletd.version="v0.4.16"
```
zancas, line 59: What happened to more recent versions of lightwalletd, especially ones that include a donation address?

zancas (approved):

docker-ci/Dockerfile
nachog00: I'll probably update this image a bit with templated versions to ease up updating the deps. The current versions are just the ones I used while developing this.

PR #257, opened 2025-04-04
Use Rustls instead of OpenSSL for `reqwest`

str4d:
This doesn't alter the repo's `Cargo.lock` due to the internal dependency on zingolabs/infrastructure, but for downstream Zaino library users this removes multiple dependencies and avoids OpenSSL. This also matches what the Zebra dependencies were doing internally (which was already pulling Rustls into Zaino's dependency tree).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ tracing-futures = "0.2"
 # Network / RPC
 http = "1.1"
 url = "2.5"
-reqwest = { version = "0.12", features = ["cookies"] }
+reqwest = { version = "0.12", default-features = false, features = ["cookies", "rustls-tls"] }
 tower = { version = "0.4", features = ["buffer", "util"] }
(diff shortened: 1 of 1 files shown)


idky137 (approved):
Thanks for this! Looks good!

idky137:
I have added this change in https://github.com/zingolabs/zaino/pull/259 as unverified commits are not allowed in Zingolabs repos.

str4d:
Not sure why a new PR was necessary, instead of signing the commit yourself (which should have been possible as I always set "allow edits by maintainers"). Could similarly cherry-pick the commit to sign locally in a new branch if preferred.

idky137:
Sorry will do this next time.

I know some people don't like their branches being messed with so I try not to unless I have explicit consent.

I wanted to get these changes in as quick as possible and as just a single line change I thought it would be quickest to just add this in a new PR.

PR #258, opened 2025-04-04
reverse get_latest_block block hash

### LINKED ISSUE #256: `FetchServiceSubscriber::get_latest_block` incorrectly reverses the hash bytes
The block hash is being byte-reversed before being returned:

https://github.com/zingolabs/zaino/blob/2b2ebb4725d771724c724587e2dfe44af7b46377/zaino-state/src/fetch.rs#L490-L504

AFAICT this is replicating a bug in `lightwalletd` that was introduced 2.5 years ago without anyone noticing: https://github.com/zcash/lightwalletd/issues/512

For annoying Bitcoin-inherited reasons we are forced to represent the hex string forms of block hashes or txids in the "big-endian" order for compatibility. However, it is **always** incorrect to represent block hashes or txids as raw bytes in the "big-endian" order, because we have no way of detecting it.

The current output of `get_latest_block` is also incompatible with the block hash support added to `get_block` in #255 because that (correctly) takes the raw bytes in the "little-endian" order.

idky137:
Reverses the byte order of the block hash returned by get_latest_block.

- Closes #256

### DIFF
--- a/integration-tests/tests/fetch_service.rs
+++ b/integration-tests/tests/fetch_service.rs
@@ -608,10 +608,7 @@ async fn fetch_service_get_latest_block(validator: &str) {
 
     let json_service_get_latest_block = dbg!(BlockId {
         height: json_service_blockchain_info.blocks.0 as u64,
-        hash: json_service_blockchain_info
-            .best_block_hash
-            .bytes_in_display_order()
-            .to_vec(),
+        hash: json_service_blockchain_info.best_block_hash.0.to_vec(),
     });
 
     assert_eq!(fetch_service_get_latest_block.height, 2);
--- a/integration-tests/tests/wallet_to_validator.rs
+++ b/integration-tests/tests/wallet_to_validator.rs
@@ -615,6 +615,8 @@ mod wallet_basic {
                 .await
         );
 
+        recipient_client.do_sync(true).await.unwrap();
+
         assert_eq!(
             recipient_client
                 .do_balance()
--- a/zaino-state/src/fetch.rs
+++ b/zaino-state/src/fetch.rs
@@ -490,12 +490,11 @@ impl LightWalletIndexer for FetchServiceSubscriber {
     /// Return the height of the tip of the best chain
     async fn get_latest_block(&self) -> Result<BlockId, Self::Error> {
         let latest_height = self.block_cache.get_chain_height().await?;
-        let mut latest_hash = self
+        let latest_hash = self
             .block_cache
             .get_compact_block(latest_height.0.to_string())
             .await?
             .hash;
-        latest_hash.reverse();
 
         Ok(BlockId {
             height: latest_height.0 as u64,


zaino-state/src/fetch.rs
```diff
@@ -490,12 +490,11 @@ impl LightWalletIndexer for FetchServiceSubscriber {
     /// Return the height of the tip of the best chain
     async fn get_latest_block(&self) -> Result<BlockId, Self::Error> {
         let latest_height = self.block_cache.get_chain_height().await?;
-        let mut latest_hash = self
+        let latest_hash = self
             .block_cache
             .get_compact_block(latest_height.0.to_string())
             .await?
             .hash;
-        latest_hash.reverse();
```
ala-mode, line 498: removing core::slice reverse()
which reverses elements of the slice, in place.

integration-tests/tests/fetch_service.rs
```diff
@@ -608,10 +608,7 @@ async fn fetch_service_get_latest_block(validator: &str) {
 
     let json_service_get_latest_block = dbg!(BlockId {
         height: json_service_blockchain_info.blocks.0 as u64,
-        hash: json_service_blockchain_info
-            .best_block_hash
-            .bytes_in_display_order()
-            .to_vec(),
+        hash: json_service_blockchain_info.best_block_hash.0.to_vec(),
```
ala-mode, line 611: when I revert `zaino-state/src/fetch.rs` to prior state, this modified test fails, with the contents of the Vec being reversed.

integration-tests/tests/wallet_to_validator.rs
```diff
@@ -615,6 +615,8 @@ mod wallet_basic {
                 .await
         );
 
+        recipient_client.do_sync(true).await.unwrap();
```
ala-mode, line 618: I'm not sure why this was added here, my run of `cargo nextest run fetch_service_get_latest_block` passes both with and without this line.

Also respectfully, I'd prefer an `.expect("...")` here.

I don't notice this changing any logic, just adding a sync for the lightclient... so these won't be blocking.

integration-tests/tests/fetch_service.rs
```diff
@@ -608,10 +608,7 @@ async fn fetch_service_get_latest_block(validator: &str) {
 
     let json_service_get_latest_block = dbg!(BlockId {
         height: json_service_blockchain_info.blocks.0 as u64,
-        hash: json_service_blockchain_info
-            .best_block_hash
-            .bytes_in_display_order()
```
ala-mode, line 613: removing zebra_chain::block::hash::Hash::bytes_in_display_order()
(returning in big-endian byte-order)... and replacing with `.0` (first in `Hash`). 

Reverting to this state in isolation causes this test to fail, with reversed bytes - `bytes_in_display_order()` reverses the bytes and this file has been correctly updated to check for the code change in [zaino-state/src/fetch.rs](https://github.com/zingolabs/zaino/pull/258/files#diff-42db1ef5e528df3647aec681decb6fe3c768eeb7df64feaad1ab11fa6e53e805)

ala-mode (commented):
Looks good! A few questions on the change in [wallet_to_validator.rs](https://github.com/zingolabs/zaino/pull/258/files#diff-112e92995bed042cce0f96d61528eea0dfa3b6db87eff301560aaaa05b21caa0) but nothing blocking. I am going to run full tests locally and merge if they pass.

ala-mode:
Noting that now `integration-tests::client_rpcs client_rpcs::get_latest_block` fails.

Even though these `client_rpc` tests might not be needed anymore, I'll open a PR in `infrastructure` to change this test. IIUC this shows a divergence from `lightwalletd` behavior, with `zaino` being correct, and `lightwalletd` being wrong.

ala-mode (approved):

PR #260, opened 2025-04-04
improved ci

nachog00:
### Improvements:
- Tests are built and ran by larger Github Hosted runners (faster)
- Github's proprietary cache is replaced by Builjet's
- Non CI ready tests (cache dependant) are ignored for now (upcomming PR's will work on supporting them)
- Cargo cache is managed completely by swatinem's action (possible room for improvement, but much simpler than custom cacheing)

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -4,7 +4,7 @@ fail-fast = false
 retries = 1
 
 [profile.ci]
-default-filter = "not test(testnet)"
+default-filter = "not test(testnet) & not test(regtest) & not test(client_rpcs)"
 
 [profile.ci.junit]  # this can be some other profile, too
 path = "junit.xml"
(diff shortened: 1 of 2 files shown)


.config/nextest.toml
```diff
@@ -4,7 +4,7 @@ fail-fast = false
 retries = 1
 
 [profile.ci]
-default-filter = "not test(testnet)"
+default-filter = "not test(testnet) & not test(regtest) & not test(client_rpcs)"
```
zancas, line 7: I am fairly confident that this should be "&&" not "&".

zancas (changes requested):

.config/nextest.toml
nachog00: It's indeed just "&"

https://nexte.st/docs/filtersets/reference/#operators

zancas (approved):

PR #262, opened 2025-04-07
fix try from for GetTreeStateResponse

idky137:
Fixes conversion from Height(i32) to Height(u32).


zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -780,12 +780,16 @@ impl TryFrom<GetTreestateResponse> for zebra_rpc::methods::trees::GetTreestate {
 
     fn try_from(value: GetTreestateResponse) -> Result<Self, Self::Error> {
         let parsed_hash = zebra_chain::block::Hash::from_hex(&value.hash)?;
+        let height_u32 = u32::try_from(value.height).map_err(|_| {
```
ala-mode, line 783: due to the binding to `u32::...`, a non-error height_u32 will always be a `u32`. 

`u32::try_from` from `height` (a i32) does indeed return an error if it is a negative value.

ala-mode (approved):
all tests pass except `client_rpcs::get_latest_block` which should be patched with #266

I will look into creating a test for this particular case, if one does not exists already.

PR #264, opened 2025-04-07
State service subscriber

idky137:
Converts `StateService` to Service-Subscriber model and makes changes required to integrate the RemoteStateService:
- Removes `checked_call`, now StateService uses Tower traits directly.
- Adds `Mempool` and `BlockCache` to `StateService`.
- Implements `ZcashService for StateService` and moves RPC trait implementations to `StateServiceSubscriber`.
- Adds `ReadStateService` based block fetch to `FinalisedState`.

NOTE: Leaves the majority of the LightwalletIndexer methods unimplemented. These will be implemented in a separate PR.


zaino-state/src/local_cache.rs
```diff
@@ -51,6 +58,7 @@ impl BlockCache {
 
         Ok(BlockCache {
             fetcher: fetcher.clone(),
+            state: state.map(|state| state.clone()),
```
AloeareV: Non-blocking, `Option::cloned` is a helper for this. 
```suggestion
            state: state.cloned(),
```

zaino-state/src/local_cache/finalised_state.rs
```diff
@@ -168,6 +171,7 @@ impl FinalisedState {
 
         let mut finalised_state = FinalisedState {
             fetcher: fetcher.clone(),
+            state: state.map(|state| state.clone()),
```
AloeareV: See earlier cloned comment

zaino-state/src/state.rs
```diff
@@ -93,8 +111,45 @@ pub struct StateService {
 }
 
 impl StateService {
+    /// Uses poll_ready to update the status of the `ReadStateService`.
+    ///
+    /// TODO: Remove use of `tokio::task::block_in_place`.
+    fn fetch_status_from_validator(&self) -> StatusType {
+        tokio::task::block_in_place(|| {
```
AloeareV: I see the todo to remove the use of this, and I don't understand why it's here in the first place. Why block in place?

zaino-state/src/state.rs
```diff
@@ -1255,9 +1396,612 @@ impl StateService {
                 }
             }
         });
+        Ok(CompactBlockStream::new(channel_rx))
+    }
 
+    /// Same as GetBlockRange except actions contain only nullifiers
+    ///
+    /// NOTE: Currently this only returns Orchard nullifiers to follow Lightwalletd functionality but Sapling could be added if required by wallets.
```
AloeareV: It turns out sapling nullifiers are part of the spends field, and not the outputs field..so they are retained, already.

AloeareV (changes requested):

zaino-state/src/local_cache.rs
idky137: Oh nice! Added in [53f1839](https://github.com/zingolabs/zaino/pull/264/commits/53f1839e30d23173a41167039f69b967830763f1)

zaino-state/src/local_cache/finalised_state.rs
idky137: Added in [53f1839](https://github.com/zingolabs/zaino/pull/264/commits/53f1839e30d23173a41167039f69b967830763f1)

zaino-state/src/state.rs
idky137: Sweet removed comment in [53f1839](https://github.com/zingolabs/zaino/pull/264/commits/53f1839e30d23173a41167039f69b967830763f1)

zaino-state/src/state.rs
idky137: This is because `ZcashService::status()` is not async, so i was using `block_in_place` to run async code in the non-async method.

We could make `ZcashService::status()` async but im not sure how this will effect the rest of the code. Or we could try to run this code sync?

If you have any opinions on how we should implement this im not sure what is best yet..

idky137: Ive updated `ZcashIndexer::status()` to be async in [b7ed5ba](https://github.com/zingolabs/zaino/pull/264/commits/b7ed5baf8bdb01c6e059f547bd828f7834bc7f69). I think this should all now be good to go.

AloeareV (approved):

PR #272, opened 2025-04-10
ziserv backed ci

nachog00:
This PR makes ziserv-2 (self hosted) handle all ci jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,7 +119,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - run: rustup toolchain install stable --profile minimal
+      - uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          components: clippy
      
       - name: Cargo cache
         uses: Swatinem/rust-cache@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   deps-cache:
     name: Cache Dependencies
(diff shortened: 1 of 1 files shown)


idky137 (dismissed):
Hey sorry I just merged "use smaller runners which has caused conflicts with this. 

Ill let you resolve them so I dont break anything..

emersonian:
Looking good. We should move to the S3-compatible caching (on localhost) soon because those cache restore times are extremely slow. Let's get Minio running on Ziserv-2 and use that locally for much faster cache restores/saves.

Some options:
https://github.com/whywaita/actions-cache-s3

https://runs-on.com/caching/s3-cache-for-github-actions/

nachog00:
> Looking good. We should move to the S3-compatible caching (on localhost) soon because those cache restore times are extremely slow. Let's get Minio running on Ziserv-2 and use that locally for much faster cache restores/saves.
> 
> Some options: https://github.com/whywaita/actions-cache-s3
> 
> https://runs-on.com/caching/s3-cache-for-github-actions/

I tried using the runs-on. Hooking up non AWS credentials wasn't all that trivial. Maybe we can pair on it? I remember you had an S3 instance for trying it out, not sure if it's still up. 

A local Minio in `ziserv-2` would definitely be the best. Would you recommend deploying it on the cluster?

nachog00:
> Hey sorry I just merged "use smaller runners which has caused conflicts with this.
> 
> Ill let you resolve them so I dont break anything..

It's fine. That other one uses small Github Hosted runners, I opened before being certain whether the ziserv approach would work reliably. Should have closed that one before opening this one tho, my bad.

idky137 (approved):

PR #275, opened 2025-04-11
Docker: Build zaino with container support

emersonian:
Use the new feature flag to build Zaino with support for containers, only in Docker builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,7 @@ RUN cargo fetch && \
 COPY . .
 
 RUN find . -name "*.rs" -exec touch {} + && \
-    cargo build --release
+    cargo build --release --features test_only_very_insecure
 
 FROM debian:bookworm-slim
 


HayashidaMakoto:
IMHO, it would be better to leave the choice to the user, in case they don't want to build the docker image to be used with a reverse proxy.

For history, this patch has been used first in https://github.com/stakeholdrs/zcash-infra/commit/30bb9454e4d52d1d00a5bc316b31d88dab4ae818 to be able to run it with the reverse proxy Caddy.

emersonian:
Perhaps two tags could be made for every build? 

v0.0.1 (no feature flag)
v0.0.1-no-tls (with feature flag)

My goal here is to no longer need to maintain a fork in order to have working Docker builds.

idky137:
Ye if it's possible to add this it would be good! That way people can explicitly see what they're using.

nachog00:
Closing this as it's implemented in #291

PR #276, opened 2025-04-14
apply clippy fix

zancas:
Please run this command with every commit:

`cargo check --all-features && cargo check --tests --all-features && cargo fmt && cargo clippy && ./utils/trailing-whitespace.sh reject`


zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -781,7 +781,7 @@ impl TryFrom<GetTreestateResponse> for zebra_rpc::methods::trees::GetTreestate {
     fn try_from(value: GetTreestateResponse) -> Result<Self, Self::Error> {
         let parsed_hash = zebra_chain::block::Hash::from_hex(&value.hash)?;
         let height_u32 = u32::try_from(value.height).map_err(|_| {
-            zebra_chain::serialization::SerializationError::Parse("negative block height".into())
+            zebra_chain::serialization::SerializationError::Parse("negative block height")
```
ala-mode, line 784: Clippy reports this `.into()` as a useless conversion to the same type: `&str`
`Parse` expects a `&'static str`
Therefore I think this is a good adjustment.

ala-mode (approved):
Good adjustment, removed unneeded conversion.

PR #277, opened 2025-04-14
add pre-commit util scripts

ala-mode:
pre-commit script as suggested by @zancas 

trailing whitespace script sourced from zingolabs `infrastructure`
https://github.com/zingolabs/infrastructure/blob/dev/utils/trailing-whitespace.sh


zancas (changes requested):
Why does this PR affect Cargo.lock?

ala-mode:
> Why does this PR affect Cargo.lock?

Almost certainly due to the LSP issue we were discussing yesterday, and then I probably added the file thinking it was a needed bump, without thinking much about it.

ala-mode:
Something went wrong in my merge attempt. closing.

PR #282, opened 2025-04-15
increment threads in nextest config

ala-mode:
In a way this is a work in progress.

The idea is to speed development velocity by simply increasing our number of parallel threads running tests, in a way that works for our core devs.

I was able to increment the thread count to 5 without any issues. At 6 I have a single test being flaky, `integration-tests::state_service state_service_get_address_balance_regtest_zebrad`

Confirming that even incrementing 5 works for people will increase our testrun speed by 25%!

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,5 +1,5 @@
 [profile.default]
-test-threads = 4
+test-threads = 5
 fail-fast = false
 retries = 1
 
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,5 +1,5 @@
 [profile.default]
-test-threads = 5
+test-threads = 6
 fail-fast = false
 retries = 1
 


zancas:
Sanity testing..

ala-mode:
I was able to increment again to 6 without a flaky test showing up.

I have a one week old version of `zebra` and an older `lwd` running so may move to change those. Getting results across machines is very useful though, I'll connect with you to make sure we're running binaries from the same commits.

zancas (approved):

zancas:
This test uniquely fails:

`integration-tests::wallet_to_validator wallet_basic::zebrad_monitor_unverified_mempool`

idky137:
Hey Ive been testing this a bit locally. 

With 5 threads I have been getting 1-2 failures and with 6 threads 3-4.

If this works for everyone else Im happy for it to merge as I can just update what I run locally but it might be a good idea to get everyone to run tests with this change as we want the tests to be as robust as possible for anyone to run..

EDIT: For note I still sometimes get 1 failure with 4 threads that looks like the same error..

idky137:
One thing that may help with this is adding a small delay after we call `test_manager.local_net.generate_blocks(1).await.unwrap();`, this is a bit of an annoying change to make though and I'm not sure it will actually fix the problem.

The reason I think this might help is that most failures I get ("flaky" failures not "real" failures) are due to the test being ahead of the zebra instance. (eg. the test thinks there should be 100 blocks in the chain but zebra is behind by 20)

The best way to do this would probably be to add a "generate_with_delay" fn and replace `test_manager.local_net.generate_blocks(1).await.unwrap();` in all tests with this.

zancas:
Are these block generation races due to a constraint on spending transparent transactions, that would be non-problematic if we could mine to orchard in regtest on zebra?

zancas:
If we are actively debugging temporarily ignored tests while everyone is running the full suite with an extra two threads... I bet that's a cost-effective tradeoff.

idky137:
> Are these block generation races due to a constraint on spending transparent transactions, that would be non-problematic if we could mine to orchard in regtest on zebra?

Not directly. this would probably help the issue though as instead of having to generate a few hundred blocks per test we would only have to generate a few (single blocks) so there is much less room for the test to get ahead. .

ala-mode:
> One thing that may help with this is adding a small delay after we call `test_manager.local_net.generate_blocks(1).await.unwrap();`, this is a bit of an annoying change to make though and I'm not sure it will actually fix the problem.
> 
> The reason I think this might help is that most failures I get ("flaky" failures not "real" failures) are due to the test being ahead of the zebra instance. (eg. the test thinks there should be 100 blocks in the chain but zebra is behind by 20)
> 
> The best way to do this would probably be to add a "generate_with_delay" fn and replace `test_manager.local_net.generate_blocks(1).await.unwrap();` in all tests with this.

This is a good note. I've charged ahead a bit with #294 (now merged - I may open another one to refine), where we can get visibility from multiples developers into what's hanging things up. When we or I pivot to trying to "unflake" or "unfail" tests or set configs to help streamline things, I'll revisit this comment.

PR #288, opened 2025-04-17
CI add checks

### LINKED ISSUE #279: Missing checks on CI
We need to add the following checks in CI:
- [ ] cargo check --all-features
- [ ] cargo check --tests --all-features
- [ ] ./utils/trailing-whitespace.sh reject

ala-mode:
fixes #279

WIP, currently tinkering.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,10 +62,10 @@ jobs:
 
       - name: Cargo cache
         uses: Swatinem/rust-cache@v2
-        with: 
+        with:
           cache-provider: buildjet
 
-      - name: Copy bins into test_binries
+      - name: Copy bins into test_binaries
         run: |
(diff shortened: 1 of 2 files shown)


nachog00:
Nice! I think we probably want those checks to be their own separate job tho, don't we? like clippy and fmt.

As it currently stands, I think they would check the whole codebase on each test partition, wouldn't it?

ala-mode:
> Nice! I think we probably want those checks to be their own separate job tho, don't we? like clippy and fmt.


Okay, I even prepared a version like that, so I'll just push that up.

> As it currently stands, I think they would check the whole codebase on each test partition, wouldn't it?

Not entirely sure, the idea of the test partition is new to me but I can look into it!

zancas (approved):

PR #296, opened 2025-04-22
refine nextest benchmarking

ala-mode:
so far:
-remove retries, improve readability

with this change there are no retries: there are no longer 'flaky' tests; a single failure is a FAIL when running the script.

tried also to improve readability.


ala-mode:
When I ran with this config, I had a fail at 7 threads from
`integration-tests::state_service zebrad::get::block_object_regtest`

I need to update local clients, including `zcashd` to `6.2.0`

zancas:
What's the state of this PR?   Is it's still viable, then should we "update branch"?

PR #297, opened 2025-04-22
Json rpc server

### LINKED ISSUE #190: Task 3.3.2: Add JsonRPC server to Zaino.
Zaino needs a JsonRPC server to serve the Zcash RPCs it is taking over from zcashd.

idky137:
- Adds jsonRPC server to zaino-serve.
- Implements RPCs: `getinfo`, `getblockchaininfo`, `getaddressbalance`, `sendrawtransaction`, `getblock`, `getrawmempool`, `z_gettreestate`, `z_getsubtreesbyindex`, `getrawtransaction`, `getaddresstxids`, `getaddressutxos`.
- updates zainod conf for jsonRPC server.
- uses cookie auth from zebra.

- Closes #190


zaino-testutils/src/lib.rs
```diff
@@ -334,11 +339,15 @@ impl TestManager {
     /// If chain_cache is given a path the chain will be loaded.
     ///
     /// If clients is set to active zingolib lightclients will be created for test use.
+    ///
+    /// TODO: Add TestManagerConfig struct and constructor methods of common test setups.
```
AloeareV, line 343: Agreed!

zaino-serve/src/rpc/jsonrpc/service.rs
```diff
@@ -0,0 +1,441 @@
...
+use jsonrpsee::types::ErrorObjectOwned;
+use jsonrpsee::{proc_macros::rpc, types::ErrorCode};
+
+use crate::rpc::JsonRpcClient;
+
+/// Zcash RPC method signatures.
+///
+/// Doc comments taken from Zebra for consistency.
+#[rpc(server)]
+pub trait ZcashIndexerRpc {
```
AloeareV, line 20: Non-blocking, I think this can be refactored to get rid of a lot of the boilerplate. This is the same set of methods that ZcashIndexer has, and we can probably avoid the need to define all these things multiple times.

AloeareV (approved):

zaino-serve/src/rpc/jsonrpc/service.rs
zancas: @AloeareV did you do the refactor?   Unifying logic into a single source of truth is more than reducing boilerplate, it's an important upgrade.

Issue #293, opened 2025-04-21
`JsonRpSeeConnector::get_block` errors when getting testnet blocks from Zebra


str4d:
Using libraries as of 3f1a50782ba43356b877766c3e9d2a604d6f0cb1

Request is `height = 3358850, verbosity = 1`

Response from Zebra is:
```json
[...]

PR #301, opened 2025-04-22
Fix getblock

idky137:
Adds derive default to fields in GetBlockTrees to allow for missing fields.


str4d:
Tested this locally, and AFAICT in its current state it does not fix #293 for me. (EDIT and I've now confirmed this PR is unrelated to the necessary fix.)

PR #320, opened 2025-05-12
Update Zcash Deps to latest

idky137:
Updates zebra to v2.3.0, updates Librustzcash to latest, updates zingolib to 2.0, fixes errors.

- Built atop https://github.com/ZcashFoundation/zebra/pull/9522

PR #326, opened 2025-05-13
new attempt at aligning dependencies

AloeareV:


idky137:
#320 now passes all tests and updates all zcash deps. I think we should go with that update unless there are any reasons not to?

AloeareV:
> #320 now passes all tests and updates all zcash deps. I think we should go with that update unless there are any reasons not to?

Fantastic! I wasn't expecting this to merge, this was a fact-finding PR!

PR #327, opened 2025-05-16
start moving cryptpad draft code into rust

AloeareV:


zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -1 +1,214 @@
...
+struct NonFinalzedState {
+    /// We need access to the validator's best block hash, as well
+    /// as a source of blocks
+    source: BlockchainSource,
+    staged: RwLock<mpsc::UnboundedReceiver<ChainBlock>>,
+    staging_sender: mpsc::UnboundedSender<ChainBlock>,
+    /// This lock should not be exposed to consumers. Rather,
+    /// clone the Arc and offer that. This means we can overwrite the arc
+    /// without interfering with readers, who will hold a stale copy
+    current: RwLock<Arc<NonfinalizedBlockCacheSnapshot>>,
```
idky137: Should this be an `ArcSwap<N>` instead of a `RWLock<Arc<N>>` ?

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -1 +1,214 @@
...
+use zebra_state::{HashOrHeight, ReadStateService};
+
+use crate::ChainBlock;
+
+/// Holds the block cache
+struct NonFinalzedState {
+    /// We need access to the validator's best block hash, as well
+    /// as a source of blocks
+    source: BlockchainSource,
+    staged: RwLock<mpsc::UnboundedReceiver<ChainBlock>>,
```
idky137, line 18: Does this need to be a `RWLock<N>`, could it just be a `Mutex<N>` (Do we need the extra functionality of a RWLock here?)

Also do we want this to be a BoundedChannel? Is there a risk of flooding?

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -1 +1,214 @@
...
+    staging_sender: mpsc::UnboundedSender<ChainBlock>,
```
idky137, line 19: Similar to above, do we want an UnboundedChannel here?

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -1 +1,214 @@
...
+    /// without interfering with readers, who will hold a stale copy
+    current: RwLock<Arc<NonfinalizedBlockCacheSnapshot>>,
+}
+
+pub(crate) struct NonfinalizedBlockCacheSnapshot {
+    /// the set of all known blocks < 100 blocks old
+    /// this includes all blocks on-chain, as well as
+    /// all blocks known to have been on-chain before being
+    /// removed by a reorg. Blocks reorged away have no height.
+    blocks: HashMap<Hash, ChainBlock>,
```
idky137, line 31: Should this be:
`blocks: HashMap<Hash, Arc<ChainBlock>>,`

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -1 +1,214 @@
...
+                .await
+            {
+                Ok(block) => block,
+                Err(_) => todo!(),
+            };
+
+            match next_block {
+                Some(block) => {
+                    // If this block is next in the chain, we sync it as normal
+                    if Hash::from(block.header.previous_block_hash) == initial_state.best_tip.1 {
```
idky137, line 69: Should this be:
`if Hash::from(block.header.previous_block_hash) == best_tip.1 {`

So the best tip that is tracked updates in this loop..

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -1 +1,214 @@
...
+                        best_tip = (best_tip.0 + 1, block.hash().into());
+                        new_blocks.push(block);
+                    } else {
+                        // If not, there's been a reorg, and we need to adjust our best-tip
```
idky137, line 73: Do we want to look back more than one block here if required?

Possible code:
```
while !initial_state.blocks.contains_key(&Hash::from(block.header.previous_block_hash)) {
    if best_tip.0 == 0 { break }
    best_tip.0 -= 1;

    best_tip.1 = *initial_state
        .blocks
        .values()
        .find(|b| b.height() == Some(best_tip.0))
        .expect("cache hole during re-org")
        .hash();
}
```

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: I have never heard of an `ArcSwap`. If that does what it sounds like it does, yes, it should be!

Edit: Yes, an ArcSwap is an improvement here.

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: Good point, I don't think we ever need a readlock. 

Actually, I am not sure we need this staging area at all.

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: I expect this channel to be small enough that I'm starting that we might not even need it

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: Yes...I have this change locally already, it's not pushed as it's part of some currently non-compiling changes!

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: Possibly? For the sake of reliably serving sidechain data it might be simpler to go back one block at a time until we find the fork point, and doing something with the intermediate blocks.

AloeareV:
Closing in favour of #373 as in-repo PR.

PR #331, opened 2025-05-20
refactor(docker): Multistage Dockerfile with dynamic conf generation

gustavovalverde:
## Context & Scope

This PR includes a set of changes to demonstrate a functional Zaino within the Z3 ecosystem. While the scope is broad for a single PR, its purpose is to present a working solution validated against the Z3 Docker Compose setup.

The intention is to follow up with smaller, more focused Pull Requests for easier review and merging. This PR can serve as a testable reference.

## Purpose

This Pull Request introduces changes to `zaino` to enable its integration within the Z3 stack (Zebra, Zaino, Zallet). The goal is to allow these components to work together in a dockerized environment, focusing on dynamic configuration and operational practices suitable for production.

These changes support the Z3 Docker Compose setup available at the https://github.com/ZcashFoundation/z3 repo.

## Key Changes & Design Principles

The modifications in this PR update Zaino by:

*   **Updating Dockerization:**
    *   Introduced a new multi-stage `docker/Dockerfile` for improved build caching and separation of build/runtime dependencies, with explicit user/group management.
    *   Updated `.dockerignore` to control the build context.
    *   The previous root `Dockerfile` has been moved (but this is optional).
*   **New Entrypoint Script (`docker/entrypoint.sh`):**
    *   Added an entrypoint script that handles dynamic generation of `zindexer.toml` from environment variables, allowing configuration without mounted config files.
    *   Includes logic for directory creation, ownership management, and uses `gosu` for privilege dropping.
    *   Formats optional path configurations in `zindexer.toml` (e.g., using the literal string `"None"`) to align with Zaino's Rust configuration parsing. This is documented in `docs/docker/adr-001-config-none-string-handling.md`.
*   **Integrated Healthchecks:** The new `docker/Dockerfile` includes a `HEALTHCHECK` instruction using `curl`.
*   **Dynamic Configuration via Environment Variables:** Runtime configuration of `zainod` within Docker is managed via environment variables processed by the `entrypoint.sh`.
*   **Modified Zebra Cookie Handling:** The RPC cookie parsing logic in `zaino-fetch/src/jsonrpsee/connector.rs` has been updated to handle Zebra's `__cookie__:<password>` format.
*   **Reduced Connectivity Assumptions:** Reliance on environment variables for service endpoints and configurations minimizes hardcoded assumptions about service connectivity.

**Testing:**
[...]

PR #332, opened 2025-05-21
fix(rpc): correctly parse cookie file for authentication with Zebra

gustavovalverde:
When authenticating to Zebra using cookie auth, Zebra writes the cookie file with the format `__cookie__:<token>`. However, Zebra's RPC server expects the client to use `__cookie__` as the username and `<token>` (the part after the prefix) as the password in the Basic Authentication header.

Previously, `zaino-fetch` was using the entire content of the cookie file as the password if the file contained the `__cookie__:` prefix.

This updates the cookie parsing logic in `new_with_cookie_auth` and `test_node_and_return_url` to strip the `__cookie__:` prefix from the cookie file content if it exists. The remaining part (the actual token) is then used as the password when constructing the `Authorization` header, aligning with Zebra's expectations.

This is a subset of:
- #331

### DIFF
--- a/zaino-fetch/src/jsonrpsee/connector.rs
+++ b/zaino-fetch/src/jsonrpsee/connector.rs
@@ -96,6 +96,24 @@ impl fmt::Display for RpcError {
 
 impl std::error::Error for RpcError {}
 
+// Helper function to read and parse the cookie file content.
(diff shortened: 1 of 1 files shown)


idky137 (approved):

idky137:
Thanks for catching this!

PR #333, opened 2025-05-21
feat(docker): refactor Dockerfile for better caching and flexibility

gustavovalverde:
This PR updates the `Dockerfile` and build mechanisms. The main goals are to improve build efficiency, provide more flexibility for image generation, use an FHS-compliant base directory, and add runtime health monitoring.

### Key Changes:

1.  **Updated Docker Syntax & Multi-Stage Build:**
    *   The Dockerfile now uses `syntax=docker/dockerfile:1`, which allows features like improved cache mount capabilities.
    *   The build process has been changed from two to **three distinct stages**:
        *   `deps`: Prepares build dependencies and captures ARGs. This stage is more likely to remain cached if only application code changes.
        *   `builder`: Compiles the `zainod` binary, using cache mounts for Cargo's target and home directories.
        *   `runtime`: A slim image that includes only the necessary runtime dependencies and the compiled binary.
    *   This multi-stage setup improves layer caching, which can lead to faster rebuilds.

2.  **Improved Caching with Cache Mounts:**
    *   The `builder` stage now uses `RUN --mount=type=cache` for `CARGO_HOME` (`~/.cargo`) and `CARGO_TARGET_DIR` (`target/`). This speeds up rebuilds when dependencies (`Cargo.lock`) have not changed, as pre-compiled dependencies can be reused.

3.  **More Rebuild Flexibility with ARGs:**
    *   Several build arguments (`ARG`) have been introduced or standardized, allowing users to customize the image build without modifying the Dockerfile directly:
        *   `UID`, `GID`, `USER`, `HOME`: Users can specify the user ID, group ID, username, and home directory for the `zaino` user within the container. This is useful for environments with specific security policies or UID/GID mapping requirements.
        *   `RUST_VERSION`: Allows for building Zaino with different Rust toolchain versions for testing and compatibility.
        *   `NO_TLS`: (Existing) Controls whether to build with TLS support.
        *   `ZAINO_GRPC_PORT`, `ZAINO_JSON_RPC_PORT`: Define the default gRPC and JSON-RPC ports, used in `EXPOSE` and the `HEALTHCHECK` instruction.

4.  **Runtime Health Monitoring:**
    *   `curl` has been added to the `runtime` stage.
    *   A `HEALTHCHECK` instruction is now included. It uses `curl` to periodically check the availability of the Zaino gRPC service on `127.0.0.1:${ZAINO_GRPC_PORT}`. This helps Docker determine if the container is healthy.

5.  **Standard Binary Location:**
    *   The `zainod` binary is now copied to `/usr/local/bin/` in the runtime image, making it accessible from any location within the container's `PATH`.

6.  **Updated Build Context:**
    *   The `.dockerignore` file has been updated to use an "exclude all, then include specific" pattern (`*` followed by `!pattern`). This sends a more precise and minimal build context to the Docker daemon, which can speed up the initial phase of the build.
    
This is a subset of:
- #331

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,23 @@
-target/
-.git/
-.github/
-.gitignore
-.dockerignore
-Dockerfile
-*.md
-.env*
+# Before the docker CLI sends the context to the docker daemon, it looks for a file
+# named .dockerignore in the root directory of the context. If this file exists, the
+# CLI modifies the context to exclude files and directories that match patterns in it.
+#
+# You may want to specify which files to include in the context, rather than which
+# to exclude. To achieve this, specify * as the first pattern, followed by one or
+# more ! exception patterns.
+#
+# https://docs.docker.com/engine/reference/builder/#dockerignore-file
+
+# Exclude everything:
+#
+*
+
+# Now un-exclude required files and folders:
+#
+!.cargo
+!*.toml
+!*.lock
+!integration-tests
+!utils-*
+!zaino-*
+!zainod
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,26 @@
-# Specify the base image for building
-FROM rust:1.86.0-bookworm AS builder
+# syntax=docker/dockerfile:1
+
+# Set the build arguments used across the stages.
+# Each stage must define the build arguments (ARGs) it uses.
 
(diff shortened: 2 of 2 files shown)


nachog00 (approved):
This looks great! Thanks for the input

PR #334, opened 2025-05-21
refactor(config)!: Standardize TOML parsing using direct Serde deserialization

gustavovalverde:
## Description

This PR refactors the configuration parsing logic in `zaino/zainod/src/config.rs` to standardize it based on idiomatic Serde practices.

The key changes include:

-   **Elimination of Special "None" String Handling**: The previous logic that treated the string literal `"None"` as a way to unset optional path fields has been removed.
-   **Direct Serde Deserialization**:
    -   The `IndexerConfig` struct now directly uses `#[derive(serde::Deserialize)]`.
    -   The custom `load_config` function has been significantly simplified to use `toml::from_str` for parsing the entire configuration into `IndexerConfig`.
    -   The `parse_field_or_warn_and_default!` macro has been removed.
-   **Custom Deserializers**: For fields requiring specific string-to-type transformations (e.g., `SocketAddr` from a hostname, `BackendType` from "fetch" or "state"), `#[serde(deserialize_with = "...")]` is now used with dedicated deserialization functions.
-   **Post-Deserialization Logic**: Conditional default logic, such as determining `cookie_dir` based on `enable_cookie_auth`, is now handled in a new `finalize_config_logic` method within `IndexerConfig`, called after the initial Serde parsing.
-   **Error Handling**: If the TOML file cannot be read or parsed, `load_config` now logs a warning and falls back to `IndexerConfig::default()`, which is then also finalized and checked.
-   **Unit Tests**: A comprehensive suite of unit tests has been added to `config.rs` to validate:
    -   Parsing of various valid and invalid TOML inputs.
    -   Correct defaulting of optional fields.
    -   The new behavior of how literal `"None"` strings are handled for different `Option` types.
    -   Inclusion of the project's actual `zaino/zainod/zindexer.toml` (via `include_str!`) to confirm its parsing behavior (it will now error on fields like `map_capacity = "None"`).

This approach aligns Zaino's configuration handling more closely with practices seen in Zebra and Zallet, and standard Rust/Serde idioms, improving clarity and maintainability.

## Breaking Change

This change modifies how optional fields in `zindexer.toml` are interpreted:

1.  **For `Option<String>` or `Option<PathBuf>` fields**: If a field like `cookie_dir` or `tls_cert_path` is set to the string `"None"` in the TOML (e.g., `cookie_dir = "None"`), it will now be parsed as `Some(PathBuf::from("None"))` or `Some("None".to_string())` respectively. To have these fields be `std::option::Option::None`, the key should be entirely **omitted** from the TOML file.
    *   *Exception*: `cookie_dir` will still become `None` if `enable_cookie_auth` is `false`, due to the `finalize_config_logic` method.
2.  **For other `Option<T>` fields (e.g., `Option<usize>`, `Option<bool>`)**: Setting such a field to the string `"None"` (e.g., `map_capacity = "None"`) will now correctly result in a **TOML parsing error**. To have these fields be `std::option::Option::None`, the key must be **omitted** from the TOML file.

Users who previously relied on setting fields to the string `"None"` to indicate an unset optional value will need to update their `zindexer.toml` files by removing these lines instead.

Relates to:
- #331 
- ADR: https://github.com/zingolabs/zaino/pull/331/files#diff-e44d81a13db5cbf1c5e3ffffe53f0ff1206522bc72e3e8d4595d3d89897d107d

### DIFF
--- a/zainod/src/config.rs
+++ b/zainod/src/config.rs
@@ -3,41 +3,45 @@
 use std::{
     net::{IpAddr, SocketAddr, ToSocketAddrs},
     path::PathBuf,
-    str::FromStr,
 };
 
-use toml::Value;
-use tracing::warn;
+// Added for Serde deserialization helpers
+use serde::{
+    de::{self, Deserializer},
+    Deserialize,
+};
+use toml; // Ensure toml crate is available for from_str
+use tracing::{info, warn};
 use zaino_state::{BackendConfig, BackendType, FetchServiceConfig, StateServiceConfig};
 
 use crate::error::IndexerError;
 
-/// Arguments:
-/// config: the config to parse
-/// field: the name of the variable to define, which matches the name of
-///     the config field
-/// kind: the toml::Value variant to parse the config field as
-/// default: the default config to fall back on
-/// handle: a function which returns an Option. Usually, this is
-///     simply [Some], to wrap the parsed value in Some when found
-macro_rules! parse_field_or_warn_and_default {
-    ($config:ident, $field:ident, $kind:ident, $default:ident, $handle:expr) => {
-        let $field = $config
-            .get(stringify!($field))
-            .map(|value| match value {
-                Value::String(string) if string == "None" => None,
-                Value::$kind(val_inner) => $handle(val_inner.clone()),
-                _ => {
-                    warn!("Invalid `{}`, using default.", stringify!($field));
-                    None
-                }
-            })
-            .unwrap_or_else(|| {
-                warn!("Missing `{}`, using default.", stringify!($field));
-                None
-            })
-            .unwrap_or_else(|| $default.$field);
-    };
+/// Custom deserialization function for `SocketAddr` from a String.
+/// Used by Serde's `deserialize_with`.
+fn deserialize_socketaddr_from_string<'de, D>(deserializer: D) -> Result<SocketAddr, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    fetch_socket_addr_from_hostname(&s)
+        .map_err(|e| de::Error::custom(format!("Invalid socket address string '{}': {}", s, e)))
+}
+
+/// Custom deserialization function for `BackendType` from a String.
+/// Used by Serde's `deserialize_with`.
+fn deserialize_backendtype_from_string<'de, D>(deserializer: D) -> Result<BackendType, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let s = String::deserialize(deserializer)?;
+    match s.to_lowercase().as_str() {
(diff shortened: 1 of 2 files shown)


zainod/src/config.rs
```diff
@@ -611,3 +467,499 @@ impl TryFrom<IndexerConfig> for BackendConfig {
...
+    }
+
+    #[test]
+    fn test_parse_existing_zindexer_toml_content() {
+        // Include the actual zindexer.toml file content at compile time
+        let zindexer_toml_content = include_str!("../zindexer.toml");
+
+        // Assert that parsing the original zindexer.toml content fails
+        // due to string "None" for Option<usize> fields (and similar issues).
+        assert!(
```
idky137: Should this assert be here? 

It looks like it was used to test that the old "zindexer.toml" would fail, however since this has been updated to the new format this test is now failing for me.

idky137 (changes requested):
Thanks for this! Looks much nicer!

The only issue I can see is `test_parse_existing_zindexer_toml_content()` looking at the new format instead of the old.

zainod/src/config.rs
gustavovalverde: Sorry about that, this was an oversight while I was testing with the previous file. Then I updated the `zindexer.toml` as there was no use in having it with the previous format, and I forgot to update the tests to reflect this.

I updated the test and ran a `cargo test -p zainod` to confirm, this is now fixed here
[test(config): Update zindexer.toml parsing test for new format](https://github.com/zingolabs/zaino/pull/334/commits/eb9cbcbad0071d4ded99b436ad982479ebc2a6ac)

gustavovalverde:
@idky137 I'm closing this PR in favor of #335

That implementation is cleaner, the test are separated (to avoid cluttering the module logic) and it avoid implementing a custom docker entrypoint with all the variables consideration as we have in Zebra https://github.com/ZcashFoundation/zebra/blob/main/docker/entrypoint.sh#L41-L130 and as I was trying to do here: https://github.com/zingolabs/zaino/pull/331/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7ca

gustavovalverde:
@idky137 I'm closing this PR in favor of #335

That implementation is cleaner, the test are separated (to avoid cluttering the module logic) and it avoid implementing a custom docker entrypoint with all the variables as we have in Zebra https://github.com/ZcashFoundation/zebra/blob/main/docker/entrypoint.sh#L41-L130 and as I was trying to do here: https://github.com/zingolabs/zaino/pull/331/files#diff-4f5cabe26761257a4d685a6edc7a43e0fe0f78762f50eeb48530f2bd3b3ee7ca

gustavovalverde:
Hey @idky137, let me know if you have questions for #335

PR #338, opened 2025-05-22
better zebra error propagation

### LINKED ISSUE #337: RpcResponse needs its response field to be optional, as it's not present in error case
> ... zebra, for example, gives an Ok status with a json block (that is currently not being parsed correctly, it should be extracting the error from `{\"jsonrpc\":\"2.0\",\"id\":0,\"error\":{\"code\":-5,\"message\":\"No such mempool or main chain transaction\"}}` but is instead returning a SerdeJsonError parsing error due to the `result` field of an RpcResponse not being marked optional). 

 _Originally posted by @AloeareV in [#336](https://github.com/zingolabs/zaino/issues/336#issuecomment-2902663510)_

AloeareV:
fixes #337

### DIFF
--- a/zaino-fetch/src/jsonrpsee/connector.rs
+++ b/zaino-fetch/src/jsonrpsee/connector.rs
@@ -285,7 +285,7 @@ impl JsonRpSeeConnector {
                 .await
                 .map_err(JsonRpSeeConnectorError::ReqwestError)?;
 
-            let body_str = dbg!(String::from_utf8_lossy(&body_bytes));
+            let body_str = String::from_utf8_lossy(&body_bytes);
 
             if body_str.contains("Work queue depth exceeded") {
                 if attempts >= max_attempts {
@@ -313,9 +313,9 @@ impl JsonRpSeeConnector {
                     error.code, error.message
                 ))),
                 (None, Some(result)) => Ok(result),
-                (None, None) => Err(JsonRpSeeConnectorError::new(format!(
-                    "error: no response body"
-                ))),
+                (None, None) => Err(JsonRpSeeConnectorError::new(
+                    "error: no response body".to_string(),
+                )),
             };
         }
     }
--- a/zaino-fetch/src/jsonrpsee/connector.rs
+++ b/zaino-fetch/src/jsonrpsee/connector.rs
@@ -41,7 +41,7 @@ struct RpcRequest<T> {
 struct RpcResponse<T> {
     id: i64,
     jsonrpc: Option<String>,
-    result: T,
+    result: Option<T>,
     error: Option<RpcError>,
 }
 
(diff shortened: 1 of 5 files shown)


zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -285,7 +285,7 @@ impl JsonRpSeeConnector {
                 .await
                 .map_err(JsonRpSeeConnectorError::ReqwestError)?;
 
-            let body_str = String::from_utf8_lossy(&body_bytes);
+            let body_str = dbg!(String::from_utf8_lossy(&body_bytes));
```
nuttycom: dbg! left in code

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -307,12 +307,15 @@ impl JsonRpSeeConnector {
...
+            return match (response.error, response.result) {
+                (Some(error), _) => Err(JsonRpSeeConnectorError::new(format!(
                     "Error: Error from node's rpc server: {} - {}",
                     error.code, error.message
                 ))),
-                None => Ok(response.result),
+                (None, Some(result)) => Ok(result),
+                (None, None) => Err(JsonRpSeeConnectorError::new(format!(
+                    "error: no response body"
+                ))),
```
nuttycom: Are you sure that this isn't a valid response?

zaino-fetch/src/jsonrpsee/connector.rs
zancas: @nachog00 can we please add automation to our CI that fails on `dbg!` detection?

Or do we intend to leave some in some cases?

zaino-fetch/src/jsonrpsee/connector.rs
AloeareV: I don't think it's a meaningful response, and I am unsure of what we could do in this case other than give an error response.

zaino-fetch/src/jsonrpsee/connector.rs
AloeareV: Good catch!

AloeareV: > @nachog00 can we please add automation to our CI that fails on `dbg!` detection?
> 
> Or do we intend to leave some in some cases?

I think we intend to leave it in test-related code. I don't think we want dbg!s in anything not test-only, that should be updated to be tracing logs, etc.

zaino-fetch/src/jsonrpsee/connector.rs
AloeareV: https://docs.rs/jsonrpsee-types/0.25.1/jsonrpsee_types/response/enum.ResponsePayload.html

This field is intended to be parsed as an result, `(None, None)` is an invalid response.

idky137 (approved):

PR #329, opened 2025-05-17
Implement getbestblockhash

ala-mode:
Fixes #198

PR #340, opened 2025-05-27
Implement `getblockcount`

### LINKED ISSUE #199: 1.3: Add Zcash RPC `getblockcount`
Spec: https://zcash.github.io/rpc/getblockcount.html

dorianvp:
Implements `getblockcount`. Closes #199.


zaino-state/src/backends/fetch.rs
```diff
@@ -318,6 +318,15 @@ impl ZcashIndexer for FetchServiceSubscriber {
             .try_into()?)
     }
 
+    /// Returns the current block count in the best valid block chain.
+    ///
+    /// zcashd reference: [`getblockcount`](https://zcash.github.io/rpc/getblockcount.html)
+    /// method: post
+    /// tags: blockchain
+    async fn get_block_count(&self) -> Result<Height, Self::Error> {
```
ala-mode, line 326: I see multiple `get_block_count` functions...

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -408,6 +408,17 @@ impl JsonRpSeeConnector {
         }
     }
 
+    /// Returns the height of the most recent block in the best valid block chain
+    /// (equivalently, the number of blocks in this chain excluding the genesis block).
+    ///
+    /// zcashd reference: [`getblockcount`](https://zcash.github.io/rpc/getblockcount.html)
+    /// method: post
+    /// tags: blockchain
+    pub async fn get_block_count(&self) -> Result<GetBlockCountResponse, JsonRpSeeConnectorError> {
```
ala-mode, line 420: I see multiple `get_block_count` functions...

zaino-state/src/backends/state.rs
```diff
@@ -1018,6 +1018,15 @@ impl ZcashIndexer for StateServiceSubscriber {
         ))
     }
 
+    /// Returns the current block count in the best valid block chain.
+    ///
+    /// zcashd reference: [`getblockcount`](https://zcash.github.io/rpc/getblockcount.html)
+    /// method: post
+    /// tags: blockchain
+    async fn get_block_count(&self) -> Result<Height, Self::Error> {
```
ala-mode, line 1026: I think this is the correct file in which to put this definition. Therefore I suspect the others are not the correct places.

zaino-state/src/indexer.rs
```diff
@@ -240,6 +240,13 @@ pub trait ZcashIndexer: Send + Sync + 'static {
         verbosity: Option<u8>,
     ) -> Result<GetBlock, Self::Error>;
 
+    /// Returns the current block count in the best valid block chain.
+    ///
+    /// zcashd reference: [`getblockcount`](https://zcash.github.io/rpc/getblockcount.html)
+    /// method: post
+    /// tags: blockchain
+    async fn get_block_count(&self) -> Result<Height, Self::Error>;
```
ala-mode, line 248: I do think this is the correct file to define the function within the Trait, but I don't think this is the correct place to do it. Please see the review comments for more.

ala-mode (commented):
I don't entirely know the way this repo is working, but after taking some time, and consulting with @AloeareV - I made a WIP [you can look at here](https://github.com/zingolabs/zaino/pull/329). I'm not sure if this is the right architecture, but I think there's a decent chance it is.

Because looking at the [RPC docs for this call](https://zcash.github.io/rpc/getblockcount.html) I see this in a group of RPCs under the simple heading "Blockchain" along with the one I am working toward, [`getbestblockhash`](https://zcash.github.io/rpc/getbestblockhash.html) I don't think any of those RPCs are yet present in `zaino-state/src/backends/state.rs` so a new Trait was implemented (`BlockExplorerIndexer`) rather than trying to fit in with others.

Let's talk more, and find out more about the `StateServiceSubscriber` as well as consulting with the devs who might be able to shine some light on this. In the mean time, I recommend trimming this PR back, only adding tests once there's set functionality we expect to be working (I am assuming this is a WIP right now).

ala-mode:
After discussion, I will model my RPC PR (#329 ) on this PR, so one design can be considered for more review, instead of two.

idky137 (changes requested):
This looks really good!

I think we should add a test for the StateService::GetBlock count and a test for the JsonRPC server before this lands.

Im happy to open a Pr into this branch with these tests.

dorianvp:
> This looks really good!
> 
> I think we should add a test for the StateService::GetBlock count and a test for the JsonRPC server before this lands.
> 
> Im happy to open a Pr into this branch with these tests.

Tests added! And those are now passing!

ala-mode:
Observed some flakey tests documented in #251

idky137 (dismissed):

zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -502,6 +502,18 @@ pub enum GetBlockResponse {
     Object(Box<BlockObject>),
 }
 
+/// Contains the height of the most recent block in the best valid block chain
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+pub struct GetBlockCountResponse(Height);
+
+impl TryFrom<GetBlockCountResponse> for Height {
```
AloeareV: Why is this `TryFrom` and not `From`? There is no possibility for error.

AloeareV (changes requested):

AloeareV (approved):

zaino-fetch/src/jsonrpsee/response.rs
zancas: If this is resolved, then @AloeareV should click "Resolve Conversation".

zancas:
@ala-mode what remains to be done to merge this?

ala-mode:
Got caught trying to review with full testing which was failing for me against `zebra` on `main` (bleeding edge) and put some notes down on #355 . I'll also follow up with related works on a `contributing.md` type document for present description and eventual proscription.

PR #330, opened 2025-05-19
ChainBlock

idky137:
Adds new chain_index module, adds ChainBlock struct.

Implements getters, setters, conversions and dcbor encoding.

NOTE: This will be used by the new internal finalised and non-finalised state.

PR #342, opened 2025-05-27
Finalised state 2.0

idky137:
This PR is the initial work in implementing the updated internal finalised-state (ZainoDB).

- Adds db wrapper structs to types.rs.
- Adds Stored entry db structs to finalised_state.rs.
- Adds ZainoDB:
  - Spawn, close, write_block, delete_block methods.
  - Background DB validation system.
  - Transparent address history indexing.
  - Transparent spent outpoint indexing.
- Adds DbReader:
  - Immutable view over a running ZainoDB instance.
  - Implements fetcher methods.
- Testing:
  - Builds test chain using regtest and stores data for use in chain_index unit tests.
  - Initial unit tests added for ZainoDB.

Remaining Work:
- Add shielded spent outpoint indexing.
- Improve errors returned by DB.
- Add explicit version handling path.


idky137:
build on top of #330

idky137:
This is now close to complete. I am now fixing the last couple bugs in the transparent address index.

Tests currently fail.

EDIT: this was a bug in test data not in db code.

integration-tests/tests/test_vectors.rs
```diff
@@ -0,0 +1,627 @@
...
+            buf1, buf2,
+            "CompactBlock protobuf mismatch at height {h_orig}"
+        );
+    }
+
+    assert_eq!(faucet_data, re_faucet, "faucet tuple mismatch");
+    assert_eq!(recipient_data, re_recipient, "recipient tuple mismatch");
+}
+
+// Test-only helper: takes big-endian hex‐encoded txids (`Vec<Vec<u8>>`)
```
dorianvp:
```suggestion
/// Test-only helper: takes big-endian hex‐encoded txids (`Vec<Vec<u8>>`)
```

integration-tests/tests/test_vectors.rs
```diff
@@ -0,0 +1,627 @@
...
+
+    let state_subscriber = state_service.get_subscriber().inner();
+
+    tokio::time::sleep(std::time::Duration::from_millis(500)).await;
+
+    (test_manager, state_service, state_subscriber)
+}
+
+#[tokio::test]
+#[ignore = "Not a test! Used to build test vector data for zaino_state::chain_index unit tests."]
```
dorianvp, line 126: I am seeing some assertions being made in this function. Does it make sense to enable this test to check whether something in the serialization step changed? Or what about creating a separate test (in a different PR) where we just test that serialization still generates the same data?

dorianvp (changes requested):

integration-tests/tests/test_vectors.rs
dorianvp: This, if done, should be done in `encoding.rs`. Closing...

dorianvp (dismissed):

zaino-state/src/chain_index/encoding.rs
```diff
@@ -233,113 +260,138 @@ impl CompactSize {
 
 /* ───────────────────────────── integer helpers ───────────────────────────── */
 
+/// Reads a u16 in LE format.
```
Oscar-Pepper, line 263: are you familiar with the `byteorder` crate?

in zingolib we use `use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};`

which extends Read and Write to include byte sized RW for LE or BE

Oscar-Pepper: not a request for changes BTW

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+        }
+        Ok(())
+    }
+
+    /// Returns the status of ZainoDB.
+    pub(crate) async fn status(&self) -> StatusType {
+        self.status.load().into()
+    }
+
+    /// Awaits untile the DB returns a Ready status.
```
Oscar-Pepper:
```suggestion
    /// Awaits until the DB returns a Ready status.
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+            metadata: self.metadata,
+            validated_tip: Arc::clone(&self.validated_tip),
+            validated_set: self.validated_set.clone(),
+            db_handler: None,
+            status: self.status.clone(),
+            config: self.config.clone(),
+        };
+
+        let handle = tokio::spawn({
+            let zaino_db = zaino_db;
```
Oscar-Pepper: [minor] doesnt the `move` keyword achieve this already? i tihnk this can be deleted

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+                }
+
+                info!(
+                    "initial validation complete – tip={}",
+                    zaino_db.validated_tip.load(Ordering::Relaxed)
+                );
+                zaino_db.status.store(StatusType::Ready.into());
+
+                // *** steady-state loop ***
+                let mut maintenance = interval(Duration::from_secs(60));
```
Oscar-Pepper, line 354: its worth noting that tick() has not been called after the interval is created so it will perform maintenence immediately when `zaino_db_handler_sleep` is first called. not sure if thats intended behaviour or not

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+        if !db_path.exists() {
+            fs::create_dir_all(&db_path)?;
+        }
+
+        // Check system rescources to set max db reeaders, clamped between 256 and 1024.
+        let cpu_cnt = std::thread::available_parallelism()
+            .map(|n| n.get())
+            .unwrap_or(4);
+        let max_readers = u32::try_from((cpu_cnt * 4).clamp(256, 1024))
+            .expect("max_readers was clamped to fit in u32");
```
Oscar-Pepper, line 186: if we say a "general" case is the computer has 16 cores, then 16 x 4 = 64 which is well below the min clamped value of 256. is the scalar set correctly?

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+    /// Used for hash based fetch of the best chain (and random access).
+    heights: Database,
+    /// Spent outpoints: Outpoint -> StoredEntry<Vec<TxIndex>>
+    ///
+    /// Used to check spent status of given outpoints, retuning spending tx.
+    spent: Database,
+    /// Transparent address history: AddrScript -> StoredEntry<AddrEventBytes>
+    ///
+    /// Used to search all transparent address indexes (txids, utxos, balances, deltas)
+    addrhist: Database,
```
Oscar-Pepper: [minor] could we add an underscore for readability or is this naming mimicing something historic for consistency? `addr_hist` or even `address_history`?

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+            Self::open_or_create_db(&env, "transparent", DatabaseFlags::empty()).await?;
+        let sapling = Self::open_or_create_db(&env, "sapling", DatabaseFlags::empty()).await?;
+        let orchard = Self::open_or_create_db(&env, "orchard", DatabaseFlags::empty()).await?;
+        let commitment_tree_data =
+            Self::open_or_create_db(&env, "commitment_tree_data", DatabaseFlags::empty()).await?;
+        let hashes = Self::open_or_create_db(&env, "hashes", DatabaseFlags::empty()).await?;
+        let spent = Self::open_or_create_db(&env, "spent", DatabaseFlags::empty()).await?;
+        let addrhist = Self::open_or_create_db(
+            &env,
+            "addrhist",
```
Oscar-Pepper: see underscore comment above

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+
+        let handle = tokio::spawn({
+            let zaino_db = zaino_db;
+            async move {
+                // *** initial validation ***
+                zaino_db.status.store(StatusType::Syncing.into());
+                let (r1, r2, r3) = tokio::join!(
+                    async { zaino_db.initial_spent_scan().await },
+                    async { zaino_db.initial_addrhist_scan().await },
+                    async { zaino_db.initial_block_scan().await },
```
Oscar-Pepper:
```suggestion
                    zaino_db.initial_spent_scan(),
                    zaino_db.initial_addrhist_scan(),
                    zaino_db.initial_block_scan(),
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+        let _ = self.clean_trailing().await;
+        if let Err(e) = self.env.sync(true) {
+            warn!("LMDB fsync before close failed: {e}");
+        }
+        Ok(())
+    }
+
+    /// Returns the status of ZainoDB.
+    pub(crate) async fn status(&self) -> StatusType {
+        self.status.load().into()
```
Oscar-Pepper: [minor] load() is not needed as SyncStatus impls From for AtomicStatus directly

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+    }
+
+    /// Awaits untile the DB returns a Ready status.
+    pub(crate) async fn wait_until_ready(&self) {
+        loop {
+            if self.status.load() == StatusType::Ready as usize {
+                break;
+            }
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
Oscar-Pepper: i prefer intervals with MissedTickBehvaiour set to Delay or Skip instead of sleeps in these loop cases.

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+            txids: self.txids,
+            transparent: self.transparent,
+            sapling: self.sapling,
+            orchard: self.orchard,
+            commitment_tree_data: self.commitment_tree_data,
+            heights: self.heights,
+            spent: self.spent,
+            addrhist: self.addrhist,
+            metadata: self.metadata,
+            validated_tip: Arc::clone(&self.validated_tip),
```
Oscar-Pepper, line 487: no issue here but interesting to note that Arc also implements Clone which is the same operation so could use `.clone()` also

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+                    }
+                }
+                // no block in db, this must be genesis block.
+                Err(lmdb::Error::NotFound) => {
+                    if block_height.0 != 1 {
+                        return Err(FinalisedStateError::Custom(format!(
+                            "first block must be height 1, got {:?}",
+                            block_height
+                        )));
+                    }
```
Oscar-Pepper:
```suggestion
                    // Height must be exactly +1 over the current tip
                    if block_height.0 != last_height.0 + 1 {
                        return Err(FinalisedStateError::Custom(format!(
                            "cannot write block at height {block_height:?}; \
                     current tip is {last_height:?}"
                        )));
                    }
                }
                // no block in db, this must be genesis block.
                Err(lmdb::Error::NotFound) => {
                    if block_height.0 != 1 {
                        return Err(FinalisedStateError::Custom(format!(
                            "first block must be height 1, got {block_height:?}"
                        )));
                    }
```

clippy warning suggests -^

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+
+        let mut spent_map: HashMap<Outpoint, TxIndex> = HashMap::new();
+        #[allow(clippy::type_complexity)]
+        let mut addrhist_inputs_map: HashMap<
+            AddrScript,
+            Vec<(AddrHistRecord, (AddrScript, AddrHistRecord))>,
+        > = HashMap::new();
+        let mut addrhist_outputs_map: HashMap<AddrScript, Vec<AddrHistRecord>> = HashMap::new();
+
+        for (tx_pos, tx) in block.transactions().iter().enumerate() {
```
Oscar-Pepper, line 616: i see some slight inconsistencies regarding the the use of "index"
the tx index here is called `tx_pos` and then a TxIndex struct is created from it which includes a block_index... but this block_index seems to be a block height and not the BlockIndex struct which is more like metadata than just an index. maybe i am unclear on what index actually means so i am not requesting changes per se i am just a little confused

zaino-state/src/chain_index/types.rs
```diff
@@ -2133,57 +2434,78 @@ impl ZainoVersionedSerialise for AddrHistRecord {
...
 
 impl AddrEventBytes {
     const LEN: usize = 17;
 
-    /// Create an [`AddrEventBytes`] from an [`AddrHistRecord`].
-    #[allow(dead_code)]
-    pub(crate) fn from_record(rec: &AddrHistRecord) -> Self {
-        use byteorder::{BigEndian, ByteOrder, LittleEndian};
+    /// Flag mask for is_mined.
+    pub const FLAG_MINED: u8 = 0x01;
```
Oscar-Pepper, line 2468: i think it is much clearer to show these values in binary for flags set by each bit in a byte i.e. `0b00000001`

zaino-state/src/chain_index/types.rs
```diff
@@ -2133,57 +2434,78 @@ impl ZainoVersionedSerialise for AddrHistRecord {
...
-        let mut buf = [0u8; Self::LEN];
-        BigEndian::write_u32(&mut buf[0..4], rec.tx_index.block_index);
-        BigEndian::write_u16(&mut buf[4..6], rec.tx_index.tx_index);
-        BigEndian::write_u16(&mut buf[6..8], rec.out_index);
-        buf[8] = rec.flags;
-        LittleEndian::write_u64(&mut buf[9..17], rec.value);
-        Self(buf)
-    }
+    /// Flag mask for is_spent.
+    pub const FLAG_SPENT: u8 = 0x02;
```
Oscar-Pepper, line 2471:
```suggestion
    pub const FLAG_SPENT: u8 = 0b00000010;
```

zaino-state/src/chain_index/types.rs
```diff
@@ -2133,57 +2434,78 @@ impl ZainoVersionedSerialise for AddrHistRecord {
...
+
+    /// Flag mask for is_input.
+    pub const FLAG_IS_INPUT: u8 = 0x04;
```
Oscar-Pepper, line 2474:
```suggestion
    pub const FLAG_IS_INPUT: u8 = 0b00000100;
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+                tx_index,
+                output_idx as u16,
+                output.value(),
+                AddrHistRecord::FLAG_MINED,
+            );
+            match map.entry(addr_script) {
+                Entry::Occupied(mut e) => e.get_mut().push(output_record),
+                Entry::Vacant(e) => {
+                    e.insert(vec![output_record]);
+                }
```
Oscar-Pepper: see doc comments example on `entry` method for using `and_modify` and `or_insert` chains in these cases

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+            Ok((_k, mut val)) => loop {
+                // Parse AddrEventBytes:
+                // - [0] StoredEntry tag
+                // - [1] record tag
+                // - [2..5] height
+                // - [6..7] tx_index
+                // - [8..9] vout
+                // - [10] flags
+                // - [11..18] value
+                // - [19..50] checksum
```
Oscar-Pepper: its a little confusing to deviate from end exclusive here as its end exclusive in the code.

could we do one of the following?
- add equals to show end inclusive i.e. `[2..=5]`
- list them end exclusive i.e. `[2..6]`
- add a note above saying `end inclusive`

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+    /// Delete all `addrhist` duplicates for `addr_bytes` that
+    ///   * belong to `block_height`, **and**
+    ///   * match the requested record type(s).
+    ///
+    /// * `delete_inputs`  – remove records whose flag-byte contains FLAG_IS_INPUT
+    /// * `delete_outputs` – remove records whose flag-byte contains FLAG_MINED
+    ///
+    /// `expected` is the number of records to delete;
+    ///
+    /// WARNINNG: This is a blocking function and **MUST** be called within a blocking thread / task.
```
Oscar-Pepper:
```suggestion
    /// WARNING: This is a blocking function and **MUST** be called within a blocking thread / task.
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+        txn.commit()?;
+        Ok(())
+    }
+
+    /// Mark a specific AddrHistRecord as spent in the addrhist DB.
+    /// Looks up a record by script and tx_index, sets FLAG_SPENT, and updates it in place.
+    ///
+    /// Returns Ok(true) if a record was updated, Ok(false) if not found, or Err on DB error.
+    ///
+    /// WARNINNG: This is a blocking function and **MUST** be called within a blocking thread / task.
```
Oscar-Pepper:
```suggestion
    /// WARNING: This is a blocking function and **MUST** be called within a blocking thread / task.
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+
+                // Parse the tx_index out of arr:
+                // - [0] StoredEntry tag
+                // - [1] record tag
+                // - [2..5] height
+                // - [6..7] tx_index
+                // - [8..9] vout
+                // - [10] flags
+                // - [11..18] value
+                // - [19..50] checksum
```
Oscar-Pepper: see similar comment above

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+        txn.commit()?;
+        Ok(false)
+    }
+
+    /// Mark a specific AddrHistRecord as unspent in the addrhist DB.
+    /// Looks up a record by script and tx_index, sets FLAG_SPENT, and updates it in place.
+    ///
+    /// Returns Ok(true) if a record was updated, Ok(false) if not found, or Err on DB error.
+    ///
+    /// WARNINNG: This is a blocking function and **MUST** be called within a blocking thread / task.
```
Oscar-Pepper:
```suggestion
    /// WARNING: This is a blocking function and **MUST** be called within a blocking thread / task.
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+
+                // Parse the tx_index out of arr:
+                // - [0] StoredEntry tag
+                // - [1] record tag
+                // - [2..5] height
+                // - [6..7] tx_index
+                // - [8..9] vout
+                // - [10] flags
+                // - [11..18] value
+                // - [19..50] checksum
```
Oscar-Pepper: see similar comment above

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+        txn.commit()?;
+        Ok(false)
+    }
+
+    /// Fetches the previous transparent output for the given outpoint.
+    /// Returns `TxOutCompact` or an explicit error if not found or invalid.
+    ///
+    /// Used to build addrhist records.
+    ///
+    /// WARNINNG: This is a blocking function and **MUST** be called within a blocking thread / task.
```
Oscar-Pepper:
```suggestion
    /// WARNING: This is a blocking function and **MUST** be called within a blocking thread / task.
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+    ///
+    /// # Arguments
+    /// - `stored`: the raw LMDB byte buffer
+    /// - `target_tx_idx`: index in the tx list
+    /// - `target_output_idx`: index in the outputs of that tx
+    ///
+    /// # Returns
+    /// - `Some(TxOutCompact)` if found and present, otherwise `None`
+    #[inline]
+    fn find_txout_in_stored_transparenttxlist(
```
Oscar-Pepper: [minor] could we add underscores to make it `transparent_tx_list` for readability?

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+    ///
+    /// # Arguments
+    /// - `target_txid`: A `[u8; 32]` representing the transaction ID to match.
+    /// - `stored`: Raw LMDB byte slice from a `StoredEntryVar<TxidList>`.
+    ///
+    /// # Returns
+    /// - `Some(index)` if a matching txid is found
+    /// - `None` if the format is invalid or no match
+    #[inline]
+    fn find_txid_position_in_stored_txidlist(
```
Oscar-Pepper: [minor] could we add an underscore to make it `txid_list` for readability?

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+
+        Self::find_txout_in_stored_transparenttxlist(stored_bytes, tx_pos, out_pos).ok_or_else(
+            || FinalisedStateError::Custom("Previous output not found at given index".into()),
+        )
+    }
+
+    /// Finds a TxIndex [block_height, tx_index] from a given txid.
+    /// Used for Txid based lookup in transaction DBs.
+    ///
+    /// WARNINNG: This is a blocking function and **MUST** be called within a blocking thread / task.
```
Oscar-Pepper:
```suggestion
    /// WARNING: This is a blocking function and **MUST** be called within a blocking thread / task.
```

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1 +1,4383 @@
...
+        let mut body = vec![0u8; T::VERSIONED_LEN];
+        r.read_exact(&mut body)?;
+        let item = T::deserialize(&body[..])?;
+
+        let checksum = read_fixed_le::<32, _>(r)?;
+        Ok(Self { item, checksum })
+    }
+
+    fn decode_v1<R: Read>(r: &mut R) -> io::Result<Self> {
+        Self::decode_latest(r)
```
Oscar-Pepper, line 4229: is this resistent to future versions?

i am probably missing something but I would have thought we want the code in `decode_latest` to be in `decode_v1` and then `decode_latest` calls `decode_v1`, so its essentially swapped... this protects from decode latest being updated to call later versions and then decode_v1 calling decode_latest

Oscar-Pepper (dismissed):
There is a lot of code here so i consider this a "first pass" review. Most of this is minor or not requesting changes but i thought it was worth commenting

zaino-state/src/chain_index/finalised_state.rs
AloeareV: > no issue here but interesting to note that Arc also implements Clone which is the same operation so could use `.clone()` also

With method call syntax it can be hard to disambiguate between a clone of the arc, and a clone of the validated tip, Arc::clone is probably clearer here.

zaino-state/src/chain_index/encoding.rs
idky137: I believe we were using this before, we have had quite a few discussions regarding having a disk serialisation schema that is as resistant as possible to changes coming from dependency updates. For this reason I decided to implement this scheme just using rust core (and core2). 

I do wonder if this is a case where it would be worth having the extra dep for the simplicity though. Maybe something we should discuss in a meeting?

zaino-state/src/chain_index/finalised_state.rs
idky137: Removed in review changes commit

zaino-state/src/chain_index/finalised_state.rs
idky137: I think this is actually ok, on startup there can be a lot of database read / write activity so calling this method as soon as the startup process is finished and then starting the maintenance tick is probably a good thing.

zaino-state/src/chain_index/finalised_state.rs
idky137: I have updated the comment for this, max reader does not need to be directly tied to cpu_cores but it is a useful metric to use to scale this input. 

In the future we will probably add more fined grained control over this in the config file for low concurrency use cases..

zaino-state/src/chain_index/finalised_state.rs
idky137: updated to address_history

zaino-state/src/chain_index/finalised_state.rs
idky137: updated to address_history

zaino-state/src/chain_index/finalised_state.rs
idky137: change added in review changes commit

zaino-state/src/chain_index/finalised_state.rs
idky137: Im not sure I follow this, could you expand?

zaino-state/src/chain_index/finalised_state.rs
idky137: updated in review changes pr, currently set to delay.

zaino-state/src/chain_index/finalised_state.rs
idky137: Ive left as arc::clone() here but can update if we think it should be..

zaino-state/src/chain_index/finalised_state.rs
idky137: this is weird I fixed all the warnings I get and i cant see any actual changes in this code..

zaino-state/src/chain_index/finalised_state.rs
idky137: In the database transactions are indexed by:
`TxIndex{ block_height, transaction_position }`

This is much more efficient that hash based indexing (we then have separate methods for converting txid -> txindex).

So for each transaction index we create we use the block height and the transactions position within the block.

Im not sure if this answers your question / concern?

zaino-state/src/chain_index/types.rs
idky137: updated to binary form

zaino-state/src/chain_index/types.rs
idky137: updated to binary form

zaino-state/src/chain_index/types.rs
idky137: updated to binary form

zaino-state/src/chain_index/finalised_state.rs
idky137: updated to:
```
map.entry(addr_script)
                .and_modify(|v| v.push(output_record))
                .or_insert_with(|| vec![output_record]);
```

zaino-state/src/chain_index/finalised_state.rs
idky137: updated to show end inclusive

zaino-state/src/chain_index/finalised_state.rs
idky137: updated to show end inclusive

zaino-state/src/chain_index/finalised_state.rs
idky137: updated to show end inclusive

zaino-state/src/chain_index/finalised_state.rs
idky137: fixed in review changes commit

zaino-state/src/chain_index/finalised_state.rs
idky137: fixed in review changes commit

zaino-state/src/chain_index/finalised_state.rs
idky137: fixed in review changes commit

zaino-state/src/chain_index/finalised_state.rs
idky137: added underscore

zaino-state/src/chain_index/finalised_state.rs
idky137: added underscore

zaino-state/src/chain_index/finalised_state.rs
Oscar-Pepper: i will make this change globally in a following PR

idky137: I think this is a good change to make but i think this should be made in a separate pr that updates for all trait implementers

zaino-state/src/chain_index/finalised_state.rs
Oscar-Pepper: after discussion in the call i will create a following PR for these changes

zaino-state/src/chain_index/finalised_state.rs
Oscar-Pepper: resolved in call

zancas (approved):
We now have more people working on code downstream of this branch than dev.  It has passed review.  I think it's better to merge it, so that developers are working off of dev!

PR #343, opened 2025-05-28
Implement `getdifficulty`

### LINKED ISSUE #205: 1.9: Add Zcash RPC `getdifficulty`
Spec: https://zcash.github.io/rpc/getdifficulty.html

dorianvp:
Implements `getdifficulty`. Closes #205.


nachog00:
I recommend you merge dev into this to get the whitespace check passing

The fmt check on the other hand, i think is inherent to this PR's code

integration-tests/tests/fetch_service.rs
```diff
@@ -497,6 +497,33 @@ async fn fetch_service_get_latest_block(validator: &ValidatorKind) {
     test_manager.close().await;
 }
 
+async fn fetch_service_get_difficulty(validator: &ValidatorKind) {
+    let (test_manager, _fetch_service, fetch_service_subscriber) =
+        create_test_manager_and_fetch_service(validator, None, true, true, true, true).await;
+
+    let fetch_service_get_difficulty =
+        dbg!(fetch_service_subscriber.get_difficulty().await.unwrap());
```
zancas: Let's have a policy from `dbg!` I think we should NEVER allow a `dbg!` into the codebase.

integration-tests/tests/fetch_service.rs
```diff
@@ -497,6 +497,33 @@ async fn fetch_service_get_latest_block(validator: &ValidatorKind) {
     test_manager.close().await;
 }
 
+async fn fetch_service_get_difficulty(validator: &ValidatorKind) {
```
zancas: Should this test helper fn be called:

`assert_fetch_vs_json_difficulty_equality`

?

What's the *purpose* of this fn?  I think the purpose is to assert the equality of the difficulty factor reported by each of two different sources.

Something that immediately comes up for me, when I slow down, and try to clarify *that* name, is that `JsonRpSeeConnector` is a very low-information name, in this context.

It's possible that it's more useful in other contexts, but if we name the thing after its **purpose** rather than what it is, then..    as long as it has a single purpose (which is consistent with Unix philosophy (do one thing well)) then it's name will be informative in all contexts.

Is it a:

(a) wood-stick with metal end
(b) nail-pounder
(c) hammer
(d) all of the above

integration-tests/tests/json_server.rs
```diff
@@ -247,6 +247,26 @@ async fn get_block_count_inner() {
     test_manager.close().await;
 }
 
+async fn get_difficulty_inner() {
+    let (mut test_manager, _zcashd_service, zcashd_subscriber, _zaino_service, zaino_subscriber) =
+        create_test_manager_and_fetch_services(false, false).await;
+
+    let zcashd_difficulty = dbg!(zcashd_subscriber.get_difficulty().await.unwrap());
```
zancas: I'm seeing lots of `dbg!`..  what's the policy?

integration-tests/tests/json_server.rs
```diff
@@ -611,6 +631,11 @@ mod zcashd {
             get_block_count_inner().await;
         }
 
+        #[tokio::test]
+        async fn get_difficulty() {
```
zancas, line 623: How about:

`assert_equality_of_zebrad_and_zcashd_difficulty`

For the name of this test?

zancas: Can we add a doc-comment, describing the test, and saying why we expect the two things to be equal?

integration-tests/tests/state_service.rs
```diff
@@ -1143,7 +1143,7 @@ mod zebrad {
         #[tokio::test]
         async fn block_count() {
```
zancas, line 1146: I recognize that this test was not named in this PR, but when I read the test name I don't learn what's being tested...

integration-tests/tests/state_service.rs
```diff
@@ -1164,6 +1164,49 @@ mod zebrad {
             let state_service_block_count =
                 dbg!(state_service_subscriber.get_block_count().await.unwrap());
             assert_eq!(fetch_service_block_count, state_service_block_count);
+
+            test_manager.close().await;
+        }
+
+        #[tokio::test]
+        async fn difficulty() {
```
zancas, line 1174: Remove `dbg!`, rename test so that someone who sees the name of the test on their terminal after a test run, has some intuition for what the test shows.

zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -186,6 +186,10 @@ pub struct GetBlockchainInfoResponse {
     commitments: u64,
 }
 
+/// Response to a `getdifficulty` RPC request.
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
```
zancas, line 190: Do we have any workspace/project-or-crate level policies on deser?

@AloeareV --^ ?

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -342,6 +344,16 @@ impl JsonRpSeeConnector {
             .await
     }
 
+    /// Returns the proof-of-work difficulty as a multiple of the minimum difficulty.
```
zancas, line 347: Let's enforce SSOT with an iron fist!!

Where's the proper place for this?  I think this level of doc-comment ought to go on the declaration in the trait?

Then the doc-comment on a specific implementation should about about those details?

zaino-serve/src/rpc/jsonrpc/service.rs
```diff
@@ -48,6 +48,14 @@ pub trait ZcashIndexerRpc {
     #[method(name = "getblockchaininfo")]
     async fn get_blockchain_info(&self) -> Result<GetBlockChainInfo, ErrorObjectOwned>;
 
+    /// Returns the proof-of-work difficulty as a multiple of the minimum difficulty.
+    ///
+    /// zcashd reference: [`getdifficulty`](https://zcash.github.io/rpc/getdifficulty.html)
+    /// method: post
+    /// tags: blockchain
+    #[method(name = "getdifficulty")]
```
zancas, line 56: Why the map from `get_difficulty` to `getdifficulty`?

zancas (dismissed):

zainod/tests/config.rs
```diff
@@ -377,10 +377,10 @@ fn test_figment_env_override_toml_and_defaults() {
         let config = load_config(&temp_toml_path).expect("load_config should succeed");
 
         assert_eq!(config.network, "Mainnet");
-        assert_eq!(config.enable_json_server, true);
+        assert!(config.enable_json_server);
         assert_eq!(config.map_capacity, Some(12345));
         assert_eq!(config.cookie_dir, Some(PathBuf::from("/env/cookie/path")));
-        assert_eq!(config.grpc_tls, false);
+        assert!(!config.grpc_tls);
```
AloeareV: It's allegedly best practice to use `assert` rather than `assert_eq`...but I find when comparing against a `false`, it's easy to miss the second `!` in `assert!(!`

AloeareV: If clippy complains, we'll need to add an `allow` somewhere

zaino-state/src/backends/state.rs
```diff
@@ -799,6 +802,17 @@ impl ZcashIndexer for StateServiceSubscriber {
             .map_err(Self::Error::from)
     }
 
+    async fn get_difficulty(&self) -> Result<f64, Self::Error> {
+        match self
+            .rpc_client
```
AloeareV: The rpc_client is a fallback option for when `read_state_service` and `block_cache` don't have the information we need. I don't think getting difficultly should need to fallback to the rpc_client, but I am not sure of that.

zaino-fetch/src/jsonrpsee/connector.rs
dorianvp: I agree, we should move all doc comments to the trait, instead of having them in this impl. I'll do it in a separate PR.

zaino-serve/src/rpc/jsonrpc/service.rs
dorianvp: This exposes the JSON-RPC under that name instead of using `get_difficulty`.

dorianvp:
Addressed comments, except the clippy issue

integration-tests/tests/json_server.rs
```diff
@@ -615,6 +615,34 @@ mod zcashd {
             get_block_count_inner().await;
         }
 
+        /// Checks that the difficulty is the same between zcashd and zaino.
+        ///
+        /// This tests generates blocks and checks that the difficulty is the same between zcashd and zaino
```
zancas, line 620: Should be difficulty between zcashd and zebrad.. not "zaino".

zancas (dismissed):

zancas:
Hi!   As Far As I Can Tell this PR is blocked on a *lint*?!

@AloeareV @dorianvp we need velocity.  Let's fix the lint, and whatever else it takes to complete this PR, and merge it.

zaino-fetch/src/jsonrpsee/response.rs
idky137: Do you mean regarding the use of Serde? We have recently started moving away from Serde and towards using dCBOR. 

The first place this is changing is in the disc format serialisation (started in https://github.com/zingolabs/zaino/pull/330).

However updating all our wire format serialisation is a much larger task that we probably do not have scope for in our current grants.

I think a long term Ideal would be to use dCBOR everywhere possible (Im not sure how easy this is with gRPC) and only derive serde under  `#[cfg_attr(test, derive(serde::Serialize, serde::Deserialize))]` (useful for testing).

zaino-state/src/backends/state.rs
idky137: Currently we do not have this information in our local cache so this is probably the best way to fetch the data but when the ChainIndex lands we will be able to check this locally: https://github.com/zingolabs/zaino/blob/de14dddadbf82c476e7e3ee7c1ba03eb1ede88ff/zaino-state/src/chain_index/types.rs#L877

EDIT: Sorry I just realised this is referencing the StateService not the FetchService. Difficulty is not served by the ReadStateService currently but it may be added in the future. We could fetch the Header Via the ReadStateService and calculate using nBits but due to the fact we will be adding this in the ChainIndex update I dont see a problem with using the fetcher until then.

integration-tests/tests/state_service.rs
idky137: These tests used to have longer names but were updated during the test structure refactor a couple months ago (when they were split into multiple mods), I think if we are going to update all the test names again we should have a meeting about what they are changing to so we dont have to continuously spend time updating this.

zaino-serve/src/rpc/jsonrpc/service.rs
idky137: This comes from https://zcash.github.io/rpc/getdifficulty.html

zancas (approved):

zancas:
@ala-mode if there are missing merge requirements for this PR then please:

(1) list them in a "changes requested" comment on a review
(2) do what you can to fix them

zancas:
Because the merge requirements are not yet specified I am merging without them.

PR #348, opened 2025-06-02
Block explorer indexer trait

ala-mode:
Simple stub to demonstrate possible architecture for a `BlockExplorerIndexer` trait.

This structure was suggested by @AloeareV but I factored it out of my #329 PR to be considered as a separate concern.


ala-mode:
I don't think this is a wanted pattern, closing.

PR #358, opened 2025-06-09
Fix trailing whitespace in nextest.toml 

nachog00:
Removed trailing whitespace from excluded tests in `.config/nextest.toml`


nachog00:
After seeing the checks, the trailing whitespace job isn't failing anymore, which is the whole aim of this PR. I call it mergeable after review

AloeareV (approved):

PR #362, opened 2025-06-11
Update contributing.md

ala-mode:
updating CONTRIBUTING.md with requirements for testing, platform expectations, details about submissions, and cleanup edits.


ala-mode:
Following on feedback from @AloeareV I want to note the following:

1. passing all CI is proper and should be required
2. a `cargo update` run might be considered as a requirement
3. certain situations may arise where experienced developers might bypass the merge-contraints, on a case-by-case basis.

1 and 2 are not contemporary with this repo.

idky137:
CI has still never reliably worked. 

While I agree this is something we should try push for, I don't think we can yet require CI to pass for PRs to be merged. 

Until we have a reliably, working CI this has too big a potential to disrupt development.

ala-mode:
> CI has still never reliably worked.
> 
> While I agree this is something we should try push for, I don't think we can yet require CI to pass for PRs to be merged.
> 
> Until we have a reliably, working CI this has too big a potential to disrupt development.

I agree.

These comments could be followed up as issues, if this branch merges.

zancas (approved):
Doc only change.   This is an improvement.

PR #363, opened 2025-06-11
fix cargo update and update

AloeareV:
Cargo update was failing on dev. I have added a workaround, and ran `cargo update`.


ala-mode:
This built for me using `cargo clean && cargo update && cargo build --release`.

idky137 (approved):

PR #364, opened 2025-06-11
Chaintip update subscriber

AloeareV:
This adds a chaintip upgrade subscriber to the StateService. Ideally, this should be a ZcashIndexer trait method, but the StateService implementation is easier than the FetchService implementation, and I didn't want to block one on the other.

### DIFF
--- a/integration-tests/tests/state_service.rs
+++ b/integration-tests/tests/state_service.rs
@@ -1050,7 +1050,7 @@ mod zebrad {
         #[tokio::test]
         async fn state_service_chaintip_update_subscriber() {
             let (
-                mut test_manager,
+                test_manager,
                 _fetch_service,
                 _fetch_service_subscriber,
                 _state_service,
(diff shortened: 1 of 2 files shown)


zancas (dismissed):
This fails CI.  Please ensure that CI passes, before merging this branch, as that MUST be a requirement in our CONTRIBUTING.md "merge requirements" section.

@ala-mode will be providing that "merge requirements" section today or tomorrow.

idky137:
Ci currently fails on all branches so I don't think we can have this as a requirement until we are sure it is working and robust.

idky137 (approved):
It would be good to add this functionality to the fetch service at some point but I have a feeling it will be very easy after we update the ChainIndex so I am happy to add this here for now so Zallet has access to it.

idky137:
@zancas I am dismissing the requested changes as currently CI is failing on all branches and we cannot stop development completely because CI is still not working. I have tested this locally and all tests pass.

ala-mode:
Taking CI discussion to chat.

PR #366, opened 2025-06-12
Zebrad version check

### LINKED ISSUE #361: we need better error messages for zebrad version mismatches
Currently when the zebrad version we run against doesn't match the version of the zebra libraries we are depending on, we get error messages that aren't understandable to someone who hasn't run into this exact issue before. (Serialization/Deserialization errors, io read errors, etc). We need a short-circuit that detects this mismatch and fails early with message that explains what version of zebrad we need to depend on.

AloeareV:
This fixes #361 by adding a zebrad version check the case of commit-hash zebra dependencies. If we depend on tagged/version-based zebra, the check won't be ran, but ideally when we're stable enough to depend on tagged/version-based zebra, this check shouldn't be necessary.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -592,6 +592,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "cargo-lock"
+version = "10.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c06acb4f71407ba205a07cb453211e0e6a67b21904e47f6ba1f9589e38f2e454"
+dependencies = [
+ "semver",
+ "serde",
+ "toml 0.8.23",
+ "url",
+]
+
+[[package]]
 name = "cbc"
 version = "0.1.2"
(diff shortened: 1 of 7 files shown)


nachog00:
If all this does is check the condition and fail gracefully, merging it asap would help in debugging the solution to the issue itself. Looks good at a glance, but I do notice there are some clippy annotations.

nachog00:
Does this consider comparisons between branch like refs and commit like?
Right on this PR i notice the checks do throw a new and descriptive error:
```
thread 'zebrad::lightwallet_indexer::get_block' panicked at integration-tests/tests/state_service.rs:132:6:
called Result::unwrap() on an Err value: ZebradVersionMismatch { expected_zebrad_version: "8b9fdd073d822a02768927634c635e84edb4eee4", connected_zebrad_version: "v2.3.0" }
```

zancas (approved):
Per @nachog00 's suggestion I am merging this.

PR #371, opened 2025-06-13
Impl getbestblockhash

### LINKED ISSUE #198: 1.2: Add Zcash RPC `getbestblockhash`
Spec: https://zcash.github.io/rpc/getbestblockhash.html

ala-mode:
Local to `Zingolabs/zaino`, this currently contains the same code as #367 

Fixes #198


integration-tests/tests/json_server.rs
```diff
@@ -235,6 +235,18 @@ async fn launch_json_server_check_info(enable_cookie_auth: bool) {
     test_manager.close().await;
 }
 
+async fn get_best_blockhash_inner() {
+    let (mut test_manager, _zcashd_service, zcashd_subscriber, _zaino_service, zaino_subscriber) =
+        create_test_manager_and_fetch_services(false, false).await;
+
+    let zcashd_bbh = dbg!(zcashd_subscriber.get_best_blockhash().await.unwrap());
```
zancas, line 246: Let's yoink these dbgs.

zaino-state/src/backends/state.rs
```diff
@@ -1073,6 +1073,22 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+        // return should be valid hex encoded.
+        // Hash from zebra says:
+        // Return the hash bytes in big-endian byte-order suitable for printing out byte by byte.
+        //
+        // Zebra displays transaction and block hashes in big-endian byte-order,
+        // following the u256 convention set by Bitcoin and zcashd.
+        Ok(self
+            .read_state_service
+            .best_tip()
+            .expect("best_tip() to unwrap")
```
AloeareV: This should be some `Self::Error` variant, and not an unwrap/expect. I don't expect this to fail, but I think it can if this call is made during the wrong place in sync.

AloeareV: Also, `block_cache` access over `read_state_service` should be preferred. 

(In cases where the information we need isn't in the cache, we should ideally use the `block_cache`'s information to determine what to look up in the `read_state_service`, although I don't think that applies here.)

zancas: I think disallowing `.unwrap` and `.expect` from mergeable PRs is a reasonable goal..

Something to put into "Merge Requirements"?

ala-mode: > I think disallowing `.unwrap` and `.expect` from mergeable PRs is a reasonable goal..
> 
> Something to put into "Merge Requirements"?

Because this is framed as a question, and most closely resembles something that would fit into a PR about `CONTRIBUTING.md`, I am not going to block closing this conversation as resolved on this comment.

One such PR was just merged in, but I will also add this topic to the [discussion of a follow up issue](https://github.com/zingolabs/zaino/issues/377) for which I have already gotten some guidance.

ala-mode: > Also, `block_cache` access over `read_state_service` should be preferred.
> 
> (In cases where the information we need isn't in the cache, we should ideally use the `block_cache`'s information to determine what to look up in the `read_state_service`, although I don't think that applies here.)

Digging more into this, the `block_cache` is a `BlockCacheSubscriber` - which does not offer the same methods as `read_state_service`, a `StateServiceSubscriber`.

In fact, `block_cache` looks like it reads from `CompactBlock`s, and in the case of this RPC, I think I can only access the hash via `BlockCacheSubscriber` by providing the height (to `get_compact_block()`), obtaining and then inspecting a `CompactBlock`. Therefore it looks like additionally this RPC would then lean on another function (another RPC call) if the height was not already known. Am I correct? 

These non-direct techniques give me pause, is this what we really prefer? I know the API calls are intended to be atomic but what if I need to chain 2 together?

I will certainly implement whatever is right, but in this case an RPC + RPC to read an indirect record seems questionable. `StateServiceSubscriber` provides `.best_tip()` and it seems much more direct.

I'm still learning this system so I could be off, but wanted to open a discussion on this point.

ala-mode: > This should be some `Self::Error` variant, and not an unwrap/expect. I don't expect this to fail, but I think it can if this call is made during the wrong place in sync.

I worked on this note a while, and I've pushed up what might be a solution, but this is new to me so please take a look.

The error `code:` is specified in `zcash/src/rpc/protocol.h` as `RPC_MISC_ERROR`. I have some recollection of a certain code almost always being used, but I don't properly remember and I'm not sure if it pertains to this kind of situation.

zaino-state/src/backends/state.rs
```diff
@@ -1073,6 +1073,27 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+    async fn get_best_blockhash(&self) -> Result<Hash, Self::Error> {
+        // return should be valid hex encoded.
+        // Hash from zebra says:
+        // Return the hash bytes in big-endian byte-order suitable for printing out byte by byte.
+        //
+        // Zebra displays transaction and block hashes in big-endian byte-order,
+        // following the u256 convention set by Bitcoin and zcashd.
+        match self.read_state_service.best_tip() {
+            Some(x) => return Ok(x.1),
+            None => {
```
idky137, line 1107: We could try calling `self.rpc_client.get_best_blockhash().await?.into()` if `best_tip()` returns `None` (trying the JsonRPC server if the ReadStateService fails..). But we dont need this so not blocking this landing.

idky137 (approved):

idky137 (dismissed):
I am currently getting test failures running the new tests:

------------
     Summary [  42.345s] 2 tests run: 0 passed, 2 failed, 167 skipped
        FAIL [  18.581s] integration-tests::json_server zcashd::zcash_indexer::get_best_blockhash
        FAIL [  23.763s] integration-tests::state_service zebrad::get::best_blockhash
error: test run failed

trying to debug now.

I have also realise there is no new test in integration-tests::fetch_service, is there a reason for this?

These failures were due to 2 errors:

1. The GetBestBlockHashResponse did not use hex for serialisation / deserialisation, this can be fixed by using the GetBlockHash struct already defined.

2. TheJsonRPC service method was returning a raw Hash rather than a GetBlockHash as its response type, this should be fixed by updating both the JsonRPC server method as well as the underlying ZcashIndexer trait method.

zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -502,6 +502,15 @@ pub enum GetBlockResponse {
     Object(Box<BlockObject>),
 }
 
+/// Contains the hex-encoded hash of the best (tip) block of the longest chain.
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+pub struct GetBestBlockHashResponse(Hash);
```
idky137: We actually already have a structure for this response, `GetBlockHash` on line 323. I think we should use that one stead of introducing multiple. 

Also a note here, the serde "with_hex" attribute must be used for Hashes (see line 323).

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -411,6 +411,21 @@ impl JsonRpSeeConnector {
...
+    /// zcashd reference: [`z_gettreestate`](https://zcash.github.io/rpc/getbestblockhash.html)
+    /// method: post
+    /// tags: blockchain
+    ///
+    /// # Notes
+    ///
+    /// The zcashd doc reference above says there are no parameters and the result is a "hex" (string) of the block hash hex encoded.
+    pub async fn get_best_blockhash(
+        &self,
+    ) -> Result<GetBestBlockHashResponse, JsonRpSeeConnectorError> {
```
idky137: This should use the current GetBlockHash struct https://github.com/zingolabs/zaino/blob/de14dddadbf82c476e7e3ee7c1ba03eb1ede88ff/zaino-fetch/src/jsonrpsee/response.rs#L323

zaino-state/src/indexer.rs
```diff
@@ -247,6 +249,13 @@ pub trait ZcashIndexer: Send + Sync + 'static {
     /// tags: blockchain
     async fn get_block_count(&self) -> Result<Height, Self::Error>;
 
+    /// Returns the hash of the best block (tip) of the longest chain.
+    /// zcashd reference: [`z_gettreestate`](https://zcash.github.io/rpc/getbestblockhash.html)
+    /// The zcashd doc reference above says there are no parameters and the result is a "hex" (string) of the block hash hex encoded.
+    /// method: post
+    /// tags: blockchain
+    async fn get_best_blockhash(&self) -> Result<Hash, Self::Error>;
```
idky137: This should probably use GetBlockHash instead of Hash, matching the rest of the methods in these RPC traits (the output type of the method matches exactly the output type of the RPC).

zaino-serve/src/rpc/jsonrpc/service.rs
```diff
@@ -48,6 +53,17 @@ pub trait ZcashIndexerRpc {
...
+    /// Returns the hash of the best block (tip) of the longest chain.
+    /// zcashd reference: [`z_gettreestate`](https://zcash.github.io/rpc/getbestblockhash.html)
+    /// method: post
+    /// tags: blockchain
+    ///
+    /// # Notes
+    ///
+    /// The zcashd doc reference above says there are no parameters and the result is a "hex" (string) of the block hash hex encoded.
+    #[method(name = "getbestblockhash")]
+    async fn get_best_blockhash(&self) -> Result<Hash, ErrorObjectOwned>;
```
idky137: This should use GetBlockHash rather than Hash, Matching the type that we expect from zcashd

integration-tests/tests/fetch_service.rs
```diff
@@ -553,6 +553,54 @@ async fn fetch_service_get_block(validator: &ValidatorKind) {
...
+            difficulty: _,
+            trees: _,
+            previous_block_hash: _,
+            next_block_hash: _,
+        } => Some(hash),
+        _ => None,
+    };
+
+    let fetch_service_get_best_blockhash: GetBlockHash =
+        dbg!(fetch_service_subscriber.get_best_blockhash().await.unwrap());
```
dorianvp, line 594:
```suggestion
        fetch_service_subscriber.get_best_blockhash().await.unwrap();
```

integration-tests/tests/state_service.rs
```diff
@@ -1142,6 +1142,32 @@ mod zebrad {
...
+                false,
+                false,
+                Some(services::network::Network::Regtest),
+            )
+            .await;
+            test_manager.local_net.generate_blocks(2).await.unwrap();
+            tokio::time::sleep(std::time::Duration::from_millis(999)).await;
+
+            let fetch_service_bbh =
+                dbg!(fetch_service_subscriber.get_best_blockhash().await.unwrap());
```
dorianvp, line 1165:
```suggestion
                fetch_service_subscriber.get_best_blockhash().await.unwrap();
```

integration-tests/tests/state_service.rs
```diff
@@ -1142,6 +1142,32 @@ mod zebrad {
...
+            let state_service_bbh =
+                dbg!(state_service_subscriber.get_best_blockhash().await.unwrap());
```
dorianvp, line 1167:
```suggestion
                state_service_subscriber.get_best_blockhash().await.unwrap();
```

dorianvp (changes requested):

dorianvp (approved):
Removing those dbg! calls in a separate PR.

zaino-state/src/backends/state.rs
AloeareV: One thing that is noteworthy is that in the new cache architecture, best-tip is initialized before the block cache construction function finishes, so it will shortly no-longer be an Option.

PR #390, opened 2025-06-19
docs(OpenRpc): document JSON-RPC endpoints

dorianvp:


dorianvp:
The OpenRPC file should be automatically generated when building. We should use `schemars` in a buildscript.

zancas:
@ala-mode do you agree?

ala-mode:
Zaino PR #410  was my 'stub' trying to start mocking up an analog of what `zallet` is doing with openRPC started in `zaino`. It didn't really auto-generate documentation, but there are hooks for it. 

After working with it a bit this week, I would be surprised if this file + a short addition to our build script would auto-generate documentation. If we want to continue, I would suggest trying to add the build scripts and firing it up, or I can spend some time trying to advance this idea.

zancas:
Over 1 year old..   closing as stale.

PR #391, opened 2025-06-20
refactor test_binaries/bins pattern

### LINKED ISSUE #393: copying binaries from image is a bad pattern
Currently, zaino expects testing binaries to be stashed in `testing_binaries/bins`.
This is cumbersome with the `cargo-make` setup, since the whoel `zaino` clone directory  is mounted into the testing container, then those binaries have to be copied from the container into the host's  `testing_binaries/bins`, deriving in de-syncs.

nachog00:
This PR moves away from the current pattern of counting on the testing binaries to be stashed in `test_binaries/bins`.
It should prevent issues with de-synced bins copied over from the testing images, making a yet stronger case for `cargo make test` being a plug-and-test command.

New pattern:
---
The new pattern expects an env var `TEST_BINARIES_DIR` to be set while running tests.
This var must contain the path to a directory with all the needed bins.

This also updates the `cargo-make` tools to make sure tests run with `cargo make test` always have this variable set correctly.

Testers who won't use the docker approach can opt for manually setting the env var before running tests or keeping it unset if they add all the binaries to their `$PATH`

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -5,6 +5,7 @@ default_to_workspace = false
 
 [env]
 IMAGE_NAME = "zingodevops/zaino-ci"
+TEST_BINARIES_DIR = "/usr/local/bin"
 
 [tasks.help]
 description = "List available commands and usage notes"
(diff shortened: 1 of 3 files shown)


nachog00:
I meant to update `docs/testing.md` with these changes, to realise that it's heavily outdated and due for a big rewrite.
Opened #392 and think this PR shouldn't lock on that as it's crucial for people's local dev experience.

dorianvp (dismissed):

zancas:
Have we considered symlinks?

nachog00:
> Have we considered symlinks?

I'm aware of them, but I don't see how they would play a role here.

nachog00:
Merging `dev` into this branch has triggered a run of the job that builds and pushes the ci docker image.
This is unexpected (that image has already been built and pushed before), but makes sense since the condition for that job is that a diff contains a change to the `.env.testing-artifacts`. From this branch's perspective, that has changed...
I suppose we want to have that job first check for the image on Dockerhub, and skip the whole build if it is there already.

zancas (approved):

PR #395, opened 2025-06-24
begin cargo make docs

ala-mode:
WIP, cargo-make documentation


docs/cargo_make.md
```diff
@@ -0,0 +1,44 @@
...
+Requirements:
+1. `cargo-make`
+`cargo install --force cargo-make` or prebuilt [binary](https://github.com/sagiegurari/cargo-make/releases)
+2. `docker` *
+https://docs.docker.com/
+https://docs.docker.com/engine/install/
+
+You can run `cargo make help` for a list of commands available.
+
+`cargo make compute-image tag`
```
nachog00, line 18: typo: missing "-" between "image" and "tag". Not a big deal, but if people copy and run int, it wont work

docs/cargo_make.md
```diff
@@ -0,0 +1,44 @@
...
+using `./utils/get-ci-image-tag.sh`, formats what's found in the checked-in `.env.testing-artifacts` file (based on commits of `zcahsd`, `zebrad`, and Rust version) into a docker image tag, prints them to stdout and makes the CARGO_MAKE_IMAGE_TAG environment variable.
+
+`cargo make test`
+Runs integration tests in a docker container, defined from `Dockerfile.ci`, and mounts the local Zaino directory on the  host system.
```
nachog00, line 22: Wording is a bit off here (i think). I would say the volumes are mounted on the container.
Also, maybe listing all the volumes would be good? it mounts the zaino source code (the repo clone), the target directory and a `.cargo` that isn't the global one, but a dedicated location to avoid messing with your global cargo stuff.

nachog00: I would also explain the flow followed by this command to get your image:
1. it gets the image tag described in `cargo make compute-image-tag`
2. it checks if the local docker daemon has that image available
3. if not, it tries to pull the image from Dockerhub
4. if not available there either, it builds it (using the `build-image`) task
5. after all that (or just some of that, if any step finds the image already), it runs:
```bash
cargo nextest run --profile ci
```
User can pass in extra arguments too, for example
```bash
cargo make test zebrad::get::address
```
will result in the container running
```bash
cargo nextest run --profile ci zebrad::get::address
```

docs/cargo_make.md
```diff
@@ -0,0 +1,44 @@
...
+
+`cargo ensure-image-exists`
+Check if the required image exists locally, build if not
+
+`cargo copy-bins`
+Extract binaries from the image to ./test_binaries/bins
+
+`base-script`
+sources `helpers.sh`
+runs tests with appropriate image,
```
nachog00, line 38: i would'nt say that base script runs tests. It just sources the helpers and computes the TAG

nachog00:
This current setup is a bit limiting since it only makes it easy to do `cargo nextest run` commands with the container, but we probably want to extend it so people can start the container, keep it running and pass whatever nextest command they want, and maybe even regular cargo commands too.

zancas:
Because the `makers` tooling is so new, and so many changes are incoming, I have marked this low-priority..  we should circle back to documentation when we onboard someone new AND more than 4 weeks have passed.

zancas:
I think this PR has likely been superceded.   Is it still providing relevant documentation @ala-mode ?

zancas:
Makefile.toml has substantially changed.

PR #373, opened 2025-06-13
New chain cache nonfinalized state

AloeareV:

PR #396, opened 2025-06-24
Update chain index types.

idky137:
Updates the chain index types.

### DIFF
--- a/zaino-state/src/chain_index/types.rs
+++ b/zaino-state/src/chain_index/types.rs
@@ -182,8 +182,6 @@ impl TryFrom<zcash_protocol::consensus::BlockHeight> for Height {
 pub struct Index(pub u32);
 
 /// 20-byte hash (RIPEMD-160 or Blake2b-160) of a transparent output script.
-///
-/// DCBOR encoding: raw 20-byte byte-string wrapped with tag 74244.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 #[cfg_attr(test, derive(serde::Serialize, serde::Deserialize))]
 pub struct AddrScript([u8; 20]);
@@ -242,9 +240,6 @@ impl From<AddrScript> for [u8; 20] {
 }
 
 /// Reference to a spent transparent UTXO.
-///
-/// DCBOR encoding (tag 74243):  
-/// `[ <byte-string 32>, <uint32> ]`
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 #[cfg_attr(test, derive(serde::Serialize, serde::Deserialize))]
 pub struct Outpoint {
@@ -565,9 +560,6 @@ impl<'a> TryFrom<&'a [u8]> for EquihashSolution {
 
 /// Pair of commitment-tree roots and their corresponding leaf counts
 /// after the current block has been applied.
-///
-/// *tag 74248*  
-/// `[ CommitmentTreeRoots, CommitmentTreeSizes ]`
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 #[cfg_attr(test, derive(serde::Serialize, serde::Deserialize))]
 pub struct CommitmentTreeData {
@@ -1376,9 +1368,6 @@ impl CompactSaplingOutput {
 }
 
 /// Compact summary of all shielded activity in a transaction.
-///
-/// *Encoding*  
-/// `[ sapling_value | null, orchard_value | null, SaplingCompactTx, Vec<CompactOrchardAction> ]`
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[cfg_attr(test, derive(serde::Serialize, serde::Deserialize))]
 pub struct OrchardCompactTx {
--- a/zaino-state/src/chain_index/types.rs
+++ b/zaino-state/src/chain_index/types.rs
@@ -517,7 +517,8 @@ pub enum EquihashSolution {
     #[cfg_attr(test, serde(with = "serde_arrays::fixed_1344"))]
     Standard([u8; 1344]),
     /// 48-5 solution (regtest).
-    Regtest([u8; 32]),
+    #[cfg_attr(test, serde(with = "serde_arrays::fixed_36"))]
+    Regtest([u8; 36]),
 }
 
 impl EquihashSolution {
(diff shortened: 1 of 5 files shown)


zaino-state/src/chain_index/types.rs
```diff
@@ -790,15 +399,16 @@ pub struct BlockData {
...
+    /// - < V4: [`hashFinalSaplingRoot`] - Sapling note commitment tree root.
+    /// - => V4: [`hashBlockCommitments`] - digest over hashLightClientRoot and hashAuthDataRoot.``
+    block_commitments: [u8; 32],
     /// Compact difficulty target used for proof-of-work and difficulty calculation.
     bits: u32,
+    /// Equihash nonse.
+    nonse: [u8; 32],
+    /// Equihash solution
+    #[cfg_attr(test, serde(with = "serde_arrays::fixed_1344"))]
+    solution: [u8; 1344],
```
AloeareV: This is only 1344 bytes on Mainnet and Testnet! On (default) regtest, it is 36.

AloeareV: I have enumed it in #373

idky137: Hey I couldnt see the enum you added in #373. I have added  an `EquihashSolution` enum in [c2fc80b](https://github.com/zingolabs/zaino/pull/396/commits/c2fc80be3f6ee4771210ab48c2f5b852e7c0b7f0) to fix this issue, if you want this implemented differently I am happy to update this code!

AloeareV: This one looks great!

AloeareV: Well, except that it isn't `32` bytes, but `36`

idky137: Ah thanks for catching this!

idky137: fixed in [3bcf278](https://github.com/zingolabs/zaino/pull/396/commits/3bcf27885d348d9d51889ab6a6ee2054fd7e7f7d)

AloeareV (approved):

zaino-state/src/chain_index/types.rs
```diff
@@ -790,15 +394,15 @@ pub struct BlockData {
...
-    ///
-    /// [`hashAuthDataRoot`]
-    auth_data_root: [u8; 32],
-    /// Roots of the Sapling and Orchard note-commitment trees (for shielded tx verification).
-    commitment_tree_roots: CommitmentTreeRoots,
-    /// Number of leaves in Sapling/Orchard note commitment trees at this block.
-    commitment_tree_size: CommitmentTreeSizes,
+    /// - < V4: [`hashFinalSaplingRoot`] - Sapling note commitment tree root.
+    /// - => V4: [`hashBlockCommitments`] - digest over hashLightClientRoot and hashAuthDataRoot.``
+    block_commitments: [u8; 32],
```
AloeareV, line 399: I find this name confusing. Not sure what else to call it, though, so non-blocking.

AloeareV (approved):

zaino-state/src/chain_index/types.rs
```diff
@@ -645,58 +173,157 @@ impl TryFrom<zcash_protocol::consensus::BlockHeight> for Height {
...
-        Self::from_tagged_cbor(cbor)
+impl FromHex for AddrScript {
+    type Error = <[u8; 20] as FromHex>::Error;
+    fn from_hex<T: AsRef<[u8]>>(hex: T) -> Result<Self, Self::Error> {
+        Ok(Self(<[u8; 20]>::from_hex(hex)?))
     }
 }
 
-/// Numerical index of subtree / shard roots.
+impl From<[u8; 20]> for AddrScript {
```
nuttycom, line 231: The existence of this impl makes the global assertion that every 20 byte value is a valid `AddrScript`; I would prefer a constructor method like `AddrScript::from_bytes` that documents validity constraints.

zaino-state/src/chain_index/types.rs
nuttycom: The documentation on this field is incorrect; version 4 is the _only_ supported version for Zcash blocks. The semantics of this field change based upon the network upgrade, and those semantics should be documented here.

zaino-state/src/chain_index/types.rs
```diff
@@ -808,19 +412,19 @@ impl BlockData {
         version: u32,
         time: i64,
         merkle_root: [u8; 32],
-        auth_data_root: [u8; 32],
-        commitment_tree_roots: CommitmentTreeRoots,
-        commitment_tree_size: CommitmentTreeSizes,
+        block_commitments: [u8; 32],
         bits: u32,
+        nonse: [u8; 32],
```
nuttycom, line 417: nit: This is usually spelled `nonce`

zaino-state/src/chain_index/types.rs
```diff
@@ -901,64 +495,93 @@ impl BlockData {
         }
         result
     }
+
+    /// Returns Equihash Nonse.
+    pub fn nonse(&self) -> [u8; 32] {
+        self.nonse
+    }
```
nuttycom, line 502: `nonce`

zaino-state/src/chain_index/types.rs
```diff
@@ -989,135 +612,312 @@ impl CommitmentTreeRoots {
...
+/// Provides efficient indexing for blockchain state queries and updates.
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(test, derive(serde::Serialize, serde::Deserialize))]
+pub struct ChainBlock {
+    /// Metadata and indexing information for this block.
+    index: BlockIndex,
+    /// Essential header and metadata information for the block.
+    data: BlockData,
+    /// Compact representations of transactions in this block.
+    tx: Vec<CompactTxData>,
```
nuttycom, line 652: Does it make sense to fix this to compact data only? I would expect that the compact block cache would separate; is it intended that `to_compact_block` is only invoked once for each `ChainBlock` value? The conversion to protobuf encoding is relatively costly, so I assume that it's not going to be done on the fly.

zaino-state/src/chain_index/types.rs
```diff
@@ -989,135 +612,312 @@ impl CommitmentTreeRoots {
...
+/// - Final sapling root:
+///  - Must be fetched from separate RPC.
+/// - Final orchard root:
+///  - Must be fetched from separate RPC.
+/// - parent_block_sapling_tree_size:
+///   - Used to calculate sapling tree size.
+/// - parent_block_orchard_tree_size:
+///   - Used to calculate sapling tree size.
+impl
+    TryFrom<(
```
nuttycom, line 767: This is a very weird `TryFrom`, using a giant tuple like this; I think it would be much better to have a `ChainBlock::from_block` method.

zaino-state/src/chain_index/types.rs
nuttycom, line 1: Overall, these types still seem excessively tied to the compact data representation, although changing that may be a subject for a different PR.

PR #401, opened 2025-06-26
Zaino-state serialiser

idky137:
Defines Zaino's serialisation scheme and implements for chain-index types.

- adds trait `ZainoVersionedSerialise`
- adds primitive read / write functuions
- implements `ZainoVersionedSerialise` for chain-index types

- Built atop #396

### DIFF
--- a/zaino-state/src/chain_index/encoding.rs
+++ b/zaino-state/src/chain_index/encoding.rs
@@ -1,0 +1,433 @@
+//! Holds traits and primitive functions for Zaino's Serialisation schema.
+#![allow(dead_code)]
+
+use core::iter::FromIterator;
+use core2::io::{self, Read, Write};
+
+/// Wire-format version tags.
+/// `pub(crate)` → visible everywhere inside *this* crate,
+/// not re-exported to downstream users.
+pub mod version {
+    /// Tag byte for data encoded with *v1* layout.
+    pub const V1: u8 = 1;
+
+    /// Tag byte for data encoded with *v2* layout.
+    pub const V2: u8 = 2;
+
+    // Add new versions as required.
+    // pub const V3: u8 = 3;
+}
+
+/* ────────────────────────── Zaino Serialiser Trait ─────────────────────────── */
+/// # Zaino wire-format: one-byte version tag
+///
+/// ## Quick summary
+///
+/// ┌─ byte 0 ─┬──────────── body depends on that tag ────────────┐
+/// │ version  │              (little-endian by default)          │
+/// └──────────┴──────────────────────────────────────────────────┘
+///
+/// * `Self::VERSION` = the tag **this build *writes***.
+/// * On **read**, we peek at the tag:
+///   * if it equals `Self::VERSION` call `decode_latest`;
+///   * otherwise fall back to the relevant `decode_vN` helper  
+///     (defaults to “unsupported” unless overwritten).
+///
+/// ## Update guide.
+///
+/// ### Initial release (`VERSION = 1`)
+/// 1. `pub struct TxV1 { … }`   // layout frozen forever  
+/// 2. `impl ZainoVersionedSerialise for TxV1`  
+///    * `const VERSION = 1`  
+///    * `encode_body` – **v1** layout  
+///    * `decode_latest` – parses **v1** bytes  
+///    * (optional) `decode_v1 = Self::decode_latest`
+///
+/// ### Bump to v2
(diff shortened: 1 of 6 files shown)


idky137:
Serialisation scheme summary:

```
/* ────────────────────────── Zaino Serialiser Trait ─────────────────────────── */
/// # Zaino wire-format: one-byte version tag
///
/// ## Quick summary
///
/// ┌─ byte 0 ─┬──────────── body depends on that tag ────────────┐
/// │ version  │              (little-endian by default)          │
/// └──────────┴──────────────────────────────────────────────────┘
///
/// * `Self::VERSION` = the tag **this build *writes***.
/// * On **read**, we peek at the tag:
///   * if it equals `Self::VERSION` call `decode_latest`;
///   * otherwise fall back to the relevant `decode_vN` helper
///     (defaults to “unsupported” unless overwritten).
///
/// ## Update guide.
///
/// ### Initial release (`VERSION = 1`)
/// 1. `pub struct TxV1 { … }`   // layout frozen forever
/// 2. `impl ZainoVersionedSerialise for TxV1`
///    * `const VERSION = 1`
///    * `encode_body` – **v1** layout
///    * `decode_latest` – parses **v1** bytes
///    * (optional) `decode_v1 = Self::decode_latest`
///
/// ### Bump to v2
/// 1. `pub struct TxV2 { … }`   // new “current” layout
/// 2. `impl From<TxV1> for TxV2` // loss-less upgrade path
/// 3. `impl ZainoVersionedSerialise for TxV2`
///    * `const VERSION = 2`
///    * `encode_body` – **v2** layout
///    * `decode_latest` – parses **v2** bytes
///    * `decode_v1` – `TxV1::decode_latest(r).map(Self::from)`
///    * (optional) `decode_v2 = Self::decode_latest`
///
/// ### Next bumps (v3, v4, …)
/// * Set `const VERSION = N`.
/// * Implement `decode_latest` for N’s layout.
/// * Add `decode_v(N-1)` (and older).
/// * Extend the `match` table inside **this trait** when a brand-new tag first appears.
///
/// ## Mandatory items per implementation
/// * `const VERSION`
/// * `encode_body`
/// * `decode_latest` — **must** parse `Self::VERSION` bytes.
///
/// Historical helpers (`decode_v1`, `decode_v2`, …) must be implemented
/// for compatibility with historical versions

**Comment by zancas** [comment]
Should we update from `dev`?

File: zaino-state/src/chain_index/encoding.rs [hunk 1]
```diff
@@ -0,0 +1,423 @@
...
+    fn decode_body<R: Read>(r: &mut R, version_tag: u8) -> io::Result<Self> {
+        if version_tag == Self::VERSION {
+            Self::decode_latest(r)
+        } else {
+            match version_tag {
+                version::V1 => Self::decode_v1(r),
+                version::V2 => Self::decode_v2(r),
+                _ => Err(io::Error::new(
+                    io::ErrorKind::InvalidData,
+                    format!("unsupported Zaino version tag {version_tag}"),
```
nachog00, line 113: I think this would print
```
unsupported Zaino version tag 4
```
which might look a bit confusing ?
what about:
```
unsupported Zaino encoding version: v4
```

zaino-state/src/chain_index/encoding.rs
```diff
@@ -0,0 +1,423 @@
...
+        match result {
+            s if s > <u64>::from(MAX_COMPACT_SIZE) => Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "CompactSize too large",
+            )),
+            s => Ok(s),
+        }
+    }
+
+    /// Reads an integer encoded in contact form and performs checked conversion
```
nachog00: typo: contact -> compact

nachog00 (dismissed):
This looks great
Left a few comments but don't see them as blockers really

zaino-state/src/chain_index/encoding.rs
dorianvp:
```suggestion
    /// Reads an integer encoded in compact form and performs checked conversion
```

dorianvp (changes requested):

dorianvp (approved):

zaino-state/src/chain_index/encoding.rs
dorianvp: Sounds like this is a good place to do some logging!

PR #403, opened 2025-06-27
Update Rust version in .env.testing-artifacts

nachog00:
- RUST_VERSION updated from 1.87 to 1.88

To ensure the environment uses the latest compatible Rust version.


zancas (changes requested):
OK..   but why not "stable"?

zancas (approved):

nachog00:
currently, RUST_VERSION is a required build argument to our ci image. moving into another pattern isn't trivial, so I wanted to get ci going as is

PR #404, opened 2025-06-27
remove redundant toolchain setups

### LINKED ISSUE #375: Unecessary steps in CI
I noticed some steps in CI checks still using Github's cache. This slows jobs down.
Luckily, those steps are toolchain setups steps we should not need really, since the ci image `(zaino-ci)` already has those baked in...

What to do: get rid of those steps while checking that jobs keep running the same

nachog00:
Eliminated unnecessary steps since Rust toolchain setup is handled elsewhere (baked into ci image), increasing CI efficiency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: CI check
+name: CI checks
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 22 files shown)


nachog00:
Ok, this passes on checks, let's see how tests behave...

removing `draft` status to trigger a full test run

zancas:
Are we not running our static checks locally?   How can lints errors like this be getting pushed to GH?

zancas (dismissed):
I ran `cargo clippy --fix` and applied the changes.

zancas (approved):

PR #381, opened 2025-06-16
Add semantic error handling

dorianvp:

PR #409, opened 2025-07-02
refactor(jsonrpsee): method error variants

dorianvp:
On top of #381.

### DIFF
--- a/zaino-fetch/src/jsonrpsee/response.rs
+++ b/zaino-fetch/src/jsonrpsee/response.rs
@@ -352,21 +352,16 @@ pub struct GetBalanceResponse {
     pub balance: u64,
 }
 
-/// Error type for the `getbalance` RPC request.
-/// TODO: check for variants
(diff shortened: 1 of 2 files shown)


zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -16,6 +16,12 @@ use crate::jsonrpsee::connector::ResponseToError;
 
 use super::connector::RpcError;
 
+impl From<RpcError> for Infallible {
+    fn from(_: RpcError) -> Self {
+        unreachable!()
```
AloeareV: This is reachable, and will panic if reached. This could implement `try_from` and always return an error?

AloeareV (changes requested):

zaino-fetch/src/jsonrpsee/response.rs
dorianvp: Sounds good!

AloeareV (approved):

PR #410, opened 2025-07-04
openRPC from zallet, modified, compiling

ala-mode:
Taken from from [`zallet` `main` @ 3a36f3d](https://github.com/zcash/wallet/commit/3a36f3db134c827efe3ce92170a02ae1566e124c) and adapted. 

Currently only in `zaino-state`. 

This is compiling.


zancas:
Current direction tracked in #1026

PR #415, opened 2025-07-11
fix clippy warnings and add deny warnings flag to ci

AloeareV:


nachog00 (approved):
There's a few unrelated (AFAIK) comments commited into this PR.
Maybe they are related to the clippy warnings? in any case, them just being comments, I don't see them as blockers at all. this looks great!

AloeareV:
> There's a few unrelated (AFAIK) comments commited into this PR. Maybe they are related to the clippy warnings? in any case, them just being comments, I don't see them as blockers at all. this looks great!

Yes, these were all undocumented which caused clippy warnings, and this upgraded those warnings to errors!

PR #418, opened 2025-07-15
feat(rpc): add `getmempoolinfo` endpoint

### LINKED ISSUE #206: 1.10: Add Zcash RPC `getmempoolinfo`
Spec: https://zcash.github.io/rpc/getmempoolinfo.html

dorianvp:
Fixes: #206 
This PR adds the `getmempoolinfo` RPC method. It also:

- [x] Moves `Mempool` and `MempoolInfo` from the `local_cache` mod to `chain_index`: #421


zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -402,6 +402,18 @@ impl JsonRpSeeConnector {
             .await
     }
 
+    /// Returns details on the active state of the TX memory pool.
+    ///
+    /// zcashd reference: [`getmempoolinfo`](https://zcash.github.io/rpc/getmempoolinfo.html)
```
zancas: This is not a link to C++ source code.   It should be a link to the lines of code in Zcashd that implement the get_mempool_info functionality.

dorianvp: Reference added!

dorianvp:
The remaining thing to do here is to also return the memory usage of the mempool.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,22 @@ impl ZcashIndexer for StateServiceSubscriber {
         ))
     }
 
+    async fn get_mempool_info(&self) -> Result<GetMempoolInfoResponse, Self::Error> {
+        let mempool_state = self.mempool.get_mempool().await;
```
ala-mode, line 985: `self.mempool.subscriber` would be equivalent without making a function call to re-create this data, but `subscriber` is set as a private field.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,22 @@ impl ZcashIndexer for StateServiceSubscriber {
         ))
     }
```
ala-mode, line 970: We need to add some doc comments, including links to the zcash source code, to this and other get_mempool_info functions. See line 1091 in this file.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,22 @@ impl ZcashIndexer for StateServiceSubscriber {
...
 
+    async fn get_mempool_info(&self) -> Result<GetMempoolInfoResponse, Self::Error> {
+        let mempool_state = self.mempool.get_mempool().await;
+
+        Ok(GetMempoolInfoResponse {
+            size: mempool_state.len() as u64,
+            bytes: mempool_state
+                .iter()
+                .map(|tx| match &tx.1 .0 {
```
ala-mode: I am finding this `.map()` difficult to parse. Specifically I'm not confident this `match` is matching on both fields - I think it might only be matching on the first field.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,22 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+                    GetRawTransaction::Object(tx) => tx.hex.as_ref().len() as u64,
```
ala-mode: You're ultimately looking for byte lengths, which is what we want, but I am also not sure about the use of `GetRawTransaction`.

I'll try to find another way to get these sizes...

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,22 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+                    GetRawTransaction::Raw(tx) => tx.as_ref().len() as u64,
+                })
+                .sum(),
+            usage: std::mem::size_of_val(&self.mempool) as u64,
```
ala-mode: I couldn't find a general method that reflects the size of an allocation without calculation, so I suggested `size_of_val`, which on the `mempool` here is the size of the `MempoolSubscriber` struct in bytes... doing some testing I convinced myself that a struct containing a field that represents another complex field is all totaled with the full capacity of the nested types... however I need to do more testing to see about types with a field of `Vec<(_,_)>` or hashmaps or similar, like our `MempoolSubscriber` struct contains.

ala-mode (commented):
Here are a few notes. I think we definitely need to add doc comments (I'll see about getting a commit in on this) - and I think our `bytes` field needs refactoring, and I'm not sure about `usage`.

I'm going to turn this PR status back to `Draft` to work on it more, and we can re-open soon.

zaino-state/src/backends/state.rs
dorianvp: The creation of this method is specifically required for the ChainIndex upgrade.

zaino-state/src/backends/state.rs
dorianvp: Done! Thx

zaino-state/src/backends/state.rs
dorianvp: `GetRawTransaction` contains the raw serialized transaction data, so I'm not sure what else we could put there.

`zcashd` defines that field as the `sum of all mempool tx' byte sizes`.

zaino-state/src/backends/state.rs
dorianvp: I refactored it to make it more readable.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,39 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+        let mempool_state = self.mempool.get_mempool().await;
+
+        Ok(GetMempoolInfoResponse {
+            size: mempool_state.len() as u64,
+            bytes: mempool_state
+                .iter()
+                .map(|tx| {
+                    // First item = txid, second item = GetRawTransaction
+                    let raw_txn = &tx.1 .0;
+                    match raw_txn {
```
ala-mode: This is a bit more readable, but I think this is probably still matching only the first found `match`. I'll try to make a minimal recreation and test it.

integration-tests/tests/fetch_service.rs
```diff
@@ -242,6 +242,102 @@ async fn fetch_service_get_raw_mempool(validator: &ValidatorKind) {
     test_manager.close().await;
 }
 
+// Zebra does not support the `getmempoolinfo` RPC
```
zancas: What is my takeaway from this comment?

*Should* Zebra support the getmempoolinfo RPC?   Did it in the past?  Was it expected to?   I need a bit more for this to be useful without more context, and it probably ought to be handy without more context.

integration-tests/tests/fetch_service.rs
```diff
@@ -242,6 +242,102 @@ async fn fetch_service_get_raw_mempool(validator: &ValidatorKind) {
...
+async fn fetch_service_get_mempool_info(validator: &ValidatorKind) {
```
zancas: I am having a lot of difficulty understanding this name.

In this context is "fetch" a verb?  Like this thing is "fetching a service", or is it a "fetch_service" which is a service that helps fetchers?   Or is it a "fetch_service" which is a service that "fetches"?   I haven't event parsed the first two words!

zancas: Maybe it's a:

fetch of a "service", and then a get of a "mempool_info".

integration-tests/tests/state_service.rs
```diff
@@ -1268,6 +1268,54 @@ mod zebrad {
...
+            clients.faucet.quick_shield().await.unwrap();
+            test_manager.local_net.generate_blocks(1).await.unwrap();
+            tokio::time::sleep(std::time::Duration::from_millis(500)).await;
+            clients.faucet.sync_and_await().await.unwrap();
+
+            let _recipient_ua = clients.get_recipient_address("unified").await;
+
+            tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+
+            let result = fetch_service_subscriber.get_mempool_info().await;
```
ala-mode: This test in tests/state_service.rs is only showing us something about the fetch_service_subscriber, but I think we should be looking at the state_service_subscriber too, right?

Should this test go to `fetch_service.rs` and some logic about the fields in state service should be tested?

zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -1437,3 +1437,18 @@ where
 {
     type RpcError = T::RpcError;
 }
+
+/// Response type for the `getmempoolinfo` RPC request
```
ala-mode, line 1441: Probably should add the doc comments here too.

ala-mode (commented):
Still working my way through things. I'll try to either write a test directly in the code here or make a minimal rust playground link to try to see the details of the `.map(|tx| ...)` function for the `bytes` field.

zaino-state/src/backends/state.rs
ala-mode: Here's a link, unfortunately, I believe demonstrating that we are _not_ totaling the memory of the inner struct's `Vec` :

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=300cf9406bb0e74d7c7d4dd1c4dfbb08

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,39 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+    /// `usage` is the total memory usage for the mempool, in bytes.
+    /// the [optional `fullyNotified` field](<https://github.com/zcash/zcash/blob/654a8be2274aa98144c80c1ac459400eaf0eacbe/src/rpc/blockchain.cpp#L1549>), is only utilized for zcashd regtests, is deprecated, and is not included.
+    async fn get_mempool_info(&self) -> Result<GetMempoolInfoResponse, Self::Error> {
+        let mempool_state = self.mempool.get_mempool().await;
+
+        Ok(GetMempoolInfoResponse {
+            size: mempool_state.len() as u64,
+            bytes: mempool_state
+                .iter()
+                .map(|tx| {
```
ala-mode: `tx` here is a `Vec<(MempoolKey,MempoolValue)>`

`MempoolKey` is a wrapper for `String`
`MempoolValue` holds the `enum` `GetRawTransaction` which is either `Raw` or `Object.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,39 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+                    // First item = txid, second item = GetRawTransaction
+                    let raw_txn = &tx.1 .0;
```
ala-mode: this line is matching the `MempoolValue` only.

zaino-state/src/backends/state.rs
ala-mode: ... It must be matching the `MempoolValue` only, because the match is only comparing against the two `enum` variants.

Therefore, only the raw transaction, or (boxed) transaction `Object` will be matched...

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,39 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+                    match raw_txn {
+                        GetRawTransaction::Object(tx) => tx.hex.as_ref().len() as u64,
+                        GetRawTransaction::Raw(tx) => tx.as_ref().len() as u64,
+                    }
+                })
+                .sum(),
```
ala-mode, line 1002: ...to be summed.

ala-mode: I think based on this line https://github.com/zcash/zcash/blob/fa9d32b3e0ad7920d11ddcfec4fbfd9de4c99f7e/src/txmempool.h#L349

The comment says that yours is the _correct_ approach, just summing the transactions alone without their `MempoolKey`s.

I'm not sure when we read the state, if we will have both enum variants present, and so this code might be boiled down or changed, but I now do think it's functioning in the match.

It also looks like in `zcash` the variable is set to 0, and I don't see the place this is altered!

ala-mode (commented):
Looking deeper into the `.map(...)` for `bytes` - I think you've got it.

zaino-state/src/backends/state.rs
dorianvp: Updated!

integration-tests/tests/fetch_service.rs
dorianvp: Added clearer comments.

integration-tests/tests/fetch_service.rs
dorianvp: Renamed to `test_get_mempool_info`

integration-tests/tests/state_service.rs
dorianvp: Thank you! I added checks for the state service return value, and now we expect to have one transaction in mempool.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,42 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+        let mempool_state = self.mempool.get_mempool().await;
+
+        let aprox_mem_usage = std::mem::size_of::<Vec<(MempoolKey, MempoolValue)>>()
+            + mempool_state.capacity() * std::mem::size_of::<(MempoolKey, MempoolValue)>();
+
+        Ok(GetMempoolInfoResponse {
+            size: mempool_state.len() as u64,
+            bytes: mempool_state
+                .iter()
+                .map(|tx| {
```
ala-mode: I am going to name-change the `tx` here, because the `tx` within the match below is of a different type.

zaino-state/src/backends/state.rs
```diff
@@ -968,6 +968,42 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+    /// `bytes` is the sum memory size in bytes of all transactions in the mempool: the sum of all transaction byte sizes.
+    /// the `usage` field is called by [this line of code](<https://github.com/zcash/zcash/blob/654a8be2274aa98144c80c1ac459400eaf0eacbe/src/rpc/blockchain.cpp#L1546>), and returns an int64 derived from the return of this function(<https://github.com/zcash/zcash/blob/654a8be2274aa98144c80c1ac459400eaf0eacbe/src/txmempool.h#L1199>), which includes a number of elements.
+    /// `usage` is the total memory usage for the mempool, in bytes.
+    /// the [optional `fullyNotified` field](<https://github.com/zcash/zcash/blob/654a8be2274aa98144c80c1ac459400eaf0eacbe/src/rpc/blockchain.cpp#L1549>), is only utilized for zcashd regtests, is deprecated, and is not included.
+    async fn get_mempool_info(&self) -> Result<GetMempoolInfoResponse, Self::Error> {
+        let mempool_state = self.mempool.get_mempool().await;
+
+        let aprox_mem_usage = std::mem::size_of::<Vec<(MempoolKey, MempoolValue)>>()
```
ala-mode: This looks good at a glance!

integration-tests/tests/fetch_service.rs
```diff
@@ -242,6 +242,102 @@ async fn fetch_service_get_raw_mempool(validator: &ValidatorKind) {
...
+        }
+        ValidatorKind::Zcashd => {
+            // Zcashd implements the RPC, so the call should succeed and match JSON-RPC output.
+            let fetch_info = fetch_service_subscriber.get_mempool_info().await.unwrap();
+
+            let json_info = json_service.get_mempool_info().await.unwrap();
+
+            // We only assert that the size (number of transactions) and bytes match.
+            assert_eq!(json_info.size, fetch_info.size);
+            assert_eq!(json_info.bytes, fetch_info.bytes);
```
ala-mode, line 326: This looks good, it compares the function of `JsonRpSeeConnector` to the `FetchServiceSubscriber` function.

integration-tests/tests/fetch_service.rs
```diff
@@ -242,6 +242,102 @@ async fn fetch_service_get_raw_mempool(validator: &ValidatorKind) {
...
+
+            // The memory usage is tied to how it was implemented,
+            // and cannot be guaranteed to be the same set of transactions.
+            // We only test that the memory usage is non-zero.
+            assert!(fetch_info.usage > 0);
+            assert!(json_info.usage > 0);
```
ala-mode: I found these are equal in the current code, so I added an assert_eq! for that, and just left one of the `> 0` asserts.

integration-tests/tests/json_server.rs
```diff
@@ -394,6 +394,39 @@ async fn get_raw_mempool_inner() {
...
+    from_inputs::quick_send(&mut clients.faucet, vec![(recipient_ua, 250_000, None)])
+        .await
+        .unwrap();
+
+    tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+
+    let zcashd_mempool_info = zcashd_subscriber.get_mempool_info().await.unwrap();
+    let zaino_mempool_info = zaino_subscriber.get_mempool_info().await.unwrap();
+
+    assert_eq!(zcashd_mempool_info, zaino_mempool_info);
```
ala-mode: These two call the same function, testing the equality across the subscribers.

Looking at these, I wasn't completely sure of what was diverging. There is a mention of a zebrad_rpc_listen_address in association
see line 40
and they seem to have the same zaino `db_path` in their config. 
see line 54 and line 100

integration-tests/tests/state_service.rs
```diff
@@ -1268,6 +1268,66 @@ mod zebrad {
...
+        /// Direct fetch of `getmempoolinfo` RPC not supported by Zebra
+        #[tokio::test]
+        async fn get_mempool_info() {
+            let (
+                mut test_manager,
+                _fetch_service,
+                fetch_service_subscriber,
+                _state_service,
+                state_service_subscriber,
+            ) = create_test_manager_and_services(&ValidatorKind::Zebrad, None, true, true, None)
```
ala-mode, line 1280: in this state_service test we only are looking at the `zebrad` validator - is that correct? Would we want to compare against `zcashd`?

integration-tests/tests/state_service.rs
```diff
@@ -1268,6 +1268,66 @@ mod zebrad {
...
+            // Zebra does not support this RPC.
+            // When they implement it, this will start failing and will need to be updated.
+            assert!(fetch_service_result.is_err());
+            let fetch_service_error = fetch_service_result.unwrap_err();
+            assert!(matches!(
+                fetch_service_error,
+                FetchServiceError::Critical(_)
+            ));
+
+            // The state service does not fail.
```
ala-mode, line 1320: I'm going to add some more asserts because this test aginst `zebrad` seems locally to be deterministic.

We could go further and unpack the `Object` we're getting back is my only other thought. For example:
`let tx = &state_service_subscriber.mempool.get_mempool().await[0].1 .0;`

ala-mode (dismissed):

integration-tests/tests/state_service.rs
zancas: Ideally, but what's the cost-benefit?

integration-tests/tests/state_service.rs
zancas: Go for it.

integration-tests/tests/state_service.rs
dorianvp: We cannot test the state service against `zcashd`, that I am aware of.

AloeareV: > in this state_service test we only are looking at the `zebrad` validator - is that correct? Would we want to compare against `zcashd`?

A state service uses direct file-system access to a zebrad database, it's not compatible with zcashd!

AloeareV (approved):
Fantastic code archeology!

PR #421, opened 2025-07-20
refactor: move `Mempool` & `MempoolSubscriber` to chain_index mod

dorianvp:
Merges into #418.


dorianvp:
The tests that are failing do because zebra does not yet implement `getmempoolinfo`.

ala-mode (approved):
This is merging into a feature, not `dev` - so the review is simply that the test for an error when fetching using this method is passing, and we understand the movement of code to be the intended direction of the project.

PR #423, opened 2025-07-21
Index trait definition, separate from in-progress implementation work

AloeareV:


zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+}
+
+/// The interface to the chain index
+pub trait Index: Sized {
+    /// Creates a new chainindex from a connection to a validator
+    /// Currently this is a ReadStateService or JsonRpSeeConnector
+    async fn new<T: Into<BlockchainSource>>(
+        source: T,
+        network: ZebraNetwork,
+    ) -> Result<Self, non_finalised_state::InitError>;
```
nuttycom: This doesn't need to be a member of this trait; a function will just take an `Index` (maybe rename to `ChainIndex`, and have something like `ZebraChainIndex` for the concrete type)  and will use that as its capability. Tying this type to `BlockchainSource` ties it to much to the underlying implementation.

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+    /// Takes a snapshot of the non_finalized state. All NFS-interfacing query
+    /// methods take a snapshot. The query will check the index
+    /// it existed at the moment the snapshot was taken.
+    fn snapshot_nonfinalized_state(&self) -> Arc<NonfinalizedBlockCacheSnapshot>;
```
nuttycom: `NonfinalizedBlockCacheSnapshot` is probably a concrete type and should not be; instead this should be an associated type (see suggestion above.)

zaino-state/src/chain_index.rs
nuttycom:
```suggestion
    /// The type of state snapshot handles provided by this index.
    type Snapshot;
    /// The type of errors returned by this index.
    type Error;
```

or:
```suggestion
    /// The type of state snapshot handles provided by this index.
    type Snapshot: Clone;
    /// The type of errors returned by this index.
    type Error;
```

zaino-state/src/chain_index.rs
nuttycom:
```suggestion
    fn snapshot_nonfinalized_state(&self) -> Arc<Self::Snapshot>;
```
Is this truly infallible? If not:
```suggestion
    fn snapshot_nonfinalized_state(&self) -> Result<Arc<Self::Snapshot>, Self::Error>;
```

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+    /// Given inclusive start and end indexes, stream all blocks
+    /// between the given indexes. Can be specified
+    /// by hash or height.
+    fn get_block_range<'snapshot: 'future, 'self_lt: 'future, 'future>(
+        &'self_lt self,
+        nonfinalized_snapshot: impl AsRef<NonfinalizedBlockCacheSnapshot> + Clone + 'snapshot,
```
nuttycom:
```suggestion
        nonfinalized_snapshot: impl AsRef<Self::Snapshot> + Clone + 'snapshot,
```

nuttycom: It's not clear to me why this has the complexity of the `AsRef` and the `Clone`.

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+        start: Option<HashOrHeight>,
+        end: Option<HashOrHeight>,
+    ) -> Result<impl Stream<Item = Result<Vec<u8>, ()>> + 'future, ()>;
```
nuttycom: The unit error type here doesn't tell us much about what can go wrong, maybe this is preferable?
```suggestion
    ) -> Result<impl Stream<Item = Result<Vec<u8>, Self::Error>> + 'future, ()>;
```
If it's a distinct set of errors that could be produced here, use a distinct associated type from `Self::Error`

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+    /// Finds the newest ancestor of the given block on the main
+    /// chain, or the block itself if it is on the main chain.
+    fn find_fork_point(
+        &self,
+        snapshot: impl AsRef<NonfinalizedBlockCacheSnapshot>,
+        block_hash: &types::Hash,
+    ) -> Result<Option<(types::Hash, types::Height)>, ()>;
```
nuttycom:
```suggestion
    ) -> Result<Option<(types::Hash, types::Height)>, Self::Error>;
```

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+        &'self_lt self,
+        nonfinalized_snapshot: impl AsRef<NonfinalizedBlockCacheSnapshot> + Clone + 'snapshot,
+        start: Option<HashOrHeight>,
+        end: Option<HashOrHeight>,
+    ) -> Result<impl Stream<Item = Result<Vec<u8>, ()>> + 'future, ()>;
+    /// Finds the newest ancestor of the given block on the main
+    /// chain, or the block itself if it is on the main chain.
+    fn find_fork_point(
+        &self,
+        snapshot: impl AsRef<NonfinalizedBlockCacheSnapshot>,
```
nuttycom:
```suggestion
        snapshot: impl AsRef<Self::Snapshot>,
```

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+        block_hash: &types::Hash,
+    ) -> Result<Option<(types::Hash, types::Height)>, ()>;
+    /// given a transaction id, returns the transaction
+    async fn get_raw_transaction(
+        &self,
+        snapshot: impl AsRef<NonfinalizedBlockCacheSnapshot>,
```
nuttycom:
```suggestion
        snapshot: impl AsRef<Self::Snapshot>,
```

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+        txid: types::Hash,
+    ) -> Result<Option<Vec<u8>>, ()>;
```
nuttycom:
```suggestion
    ) -> Result<Option<Vec<u8>>, Self::Error>;
```

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+    /// Given a transaction ID, returns all known
+    async fn get_transaction_status(
+        &self,
+        snapshot: impl AsRef<NonfinalizedBlockCacheSnapshot>,
```
nuttycom:
```suggestion
        snapshot: impl AsRef<Self::Snapshot>,
```

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+        txid: types::Hash,
+    ) -> HashMap<types::Hash, Option<types::Height>>;
```
nuttycom: Is this actually infallible in the general case? It should probably also return in `Result`.

nuttycom (changes requested):

zaino-state/src/chain_index.rs
nuttycom: I'm not sure whether the concrete `Arc` should actually be exposed here; if instead, it returned just `Self::Snapshot` then in the implementation for the concrete type, you might choose `Self::Snapshot = Arc<NonfinalizedBlockCacheSnapshot>`. Then, you wouldn't need all of the `AsRef + Clone`.

zaino-state/src/chain_index.rs
AloeareV: Yes, I moved all the pub functions over but new should not be a trait method!

zaino-state/src/chain_index.rs
AloeareV: As the cache is exposed as an arc, I thought it would simplify access to be able to pass in an `Arc<snapshot>`, `&snapshot`, or `&Arc<snapshot>` interchangeably. If we make the snapshot into an associated type, this might all go under the hood anyway.

zaino-state/src/chain_index.rs
AloeareV: > The unit error type here doesn't tell us much about what can go wrong, maybe this is preferable?
> 
> If it's a distinct set of errors that could be produced here, use a distinct associated type from `Self::Error`

I think there will be a distinct set of errors that could be produced here, errors are all unit-typed as these error types will all be implementation defined...an associated error type for each method should be the way to go here.

zaino-state/src/chain_index.rs
```diff
@@ -20,3 +20,65 @@ pub mod mempool;
...
+///
+/// TODO: Currently only contains the non-finalized state.
+pub struct ChainIndex {
+    // TODO: finalized state
+    // TODO: mempool
+    non_finalized_state: non_finalised_state::NonFinalizedState,
+}
+
+/// The interface to the chain index
+pub trait Index: Sized {
```
zancas: I am curious about this naming convention.  If there's a wider pattern of using traits to abstract implementation from interface, and that's applied in cases like this where there's a single implementor that "inspired" the trait..  would it make sense to use a name like:

ChainIndexInterface

?

"Index" seems not very helpful to me.  Apologies if there's context that makes this obvious.

AloeareV: Index is a placeholder name, I agree it's unhelpful!

zaino-state/src/chain_index.rs
nuttycom: I would avoid having an associated type for each method; that ends up being a nightmare from a type ascription and constraint perspective. Forcing the trait implementer to create a single error type is not terrible.

zaino-state/src/chain_index.rs
nuttycom: I would suggest that the name be `ChainIndex` - I dislike the `Interface` suffix as interface-based programming is the norm, not the exception; concrete types have to justify their existence and document their specificity, whereas interfaces do not.

zancas: You suggest that the type and the trait have the same name?

nuttycom: > You suggest that the type and the trait have the same name?

No; I think that the trait should be named `ChainIndex` and the concrete type should have a different name; maybe something like `DiskChainIndex` or `MemoryChainIndex` or `LmdbChainIndex` or whatever, to indicate the concrete type of the backing store.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,90 @@
...
+    fmt::{Debug, Display},
+    future::Future,
+};
+
+use super::types::{self, ChainBlock};
+use futures::Stream;
+use zebra_state::HashOrHeight;
+
+/// The interface to the chain index
+pub trait ChainIndex: Sized {
```
nuttycom: Why is the Sized bound necessary here? Can that bound instead be added at the call site where an argument bounded by `ChainIndex` is required?

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// A snapshot of the nonfinalized state, needed for atomic access
+    type Snapshot: NonFinalizedSnapshot;
+
+    /// How it can fail
+    type Error: std::error::Error;
```
nuttycom: This is also probably not the right place for this bound.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// Takes a snapshot of the non_finalized state. All NFS-interfacing query
+    /// methods take a snapshot. The query will check the index
+    /// it existed at the moment the snapshot was taken.
+    fn snapshot_nonfinalized_state(&self) -> &Self::Snapshot;
+
+    /// Given inclusive start and end indexes, stream all blocks
+    /// between the given indexes. Can be specified
+    /// by hash or height.
+    fn get_block_range<'snapshot, 'self_lt, 'future>(
```
nuttycom: The 'future lifetime here is a little puzzling to me. Can you explain why it is necessary, given that it is not constrained by the arguments to this function?

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,90 @@
...
+    /// chain, or the block itself if it is on the main chain.
+    fn find_fork_point(
+        &self,
+        snapshot: impl AsRef<Self::Snapshot>,
+        block_hash: &types::Hash,
+    ) -> Result<Option<(types::Hash, types::Height)>, Self::Error>;
+    /// given a transaction id, returns the transaction
+    fn get_raw_transaction(
+        &self,
+        snapshot: &Self::Snapshot,
```
nuttycom, line 43: I'm wondering whether the snapshot is really necessary here; even if a transaction doesn't exist in a particular fork, asking for the raw transaction is still valid.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,90 @@
...
+};
+
+use super::types::{self, ChainBlock};
+use futures::Stream;
+use zebra_state::HashOrHeight;
+
+/// The interface to the chain index
+pub trait ChainIndex: Sized {
+    /// A snapshot of the nonfinalized state, needed for atomic access
+    type Snapshot: NonFinalizedSnapshot;
```
nuttycom: This bound is also a candidate for the call site; a function that takes a `ChainIndex` argument doesn't need to make use of the operations on `NonFinalizedSnapshot` and so it doesn't need to know that they exist. This bound would be useful as a constraint in the place where `ChainIndex` is impl'ed.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,90 @@
...
+        snapshot: impl AsRef<Self::Snapshot>,
+        block_hash: &types::Hash,
+    ) -> Result<Option<(types::Hash, types::Height)>, Self::Error>;
+    /// given a transaction id, returns the transaction
+    fn get_raw_transaction(
+        &self,
+        snapshot: &Self::Snapshot,
+        txid: [u8; 32],
+    ) -> impl Future<Output = Result<Option<Vec<u8>>, Self::Error>>;
+    /// Given a transaction ID, returns all known
```
nuttycom, line 46: Incomplete documentation.

zaino-state/src/chain_index/interface.rs
AloeareV: Well....the sized bound doesn't seem to be necessary at all here! I think it was when I added it..

zaino-state/src/chain_index/interface.rs
AloeareV: Again I am confused. When I added these lifetime bounds, it was due to needing to constrain the returned future to not outlive either the snapshot reference or self reference passed as arguments. I don't understand why this requirement seems to have gone away, but removing these lifetime parameters (and constraints) seems to still work fine. 

Maybe they were needed when this was defined as an `async` function and are now inferable....

zaino-state/src/chain_index/interface.rs
AloeareV: Hmm. I also don't know that it's necessary here, but I've been unable to convince myself it isn't. If it doesn't take a snapshot, it can no longer be used to check whether a txid is contained in the given snapshot.

zaino-state/src/chain_index/interface.rs
```diff
@@ -9,29 +9,27 @@ use futures::Stream;
...
 
     /// How it can fail
-    type Error: std::error::Error;
+    type Error;
+
     /// Takes a snapshot of the non_finalized state. All NFS-interfacing query
     /// methods take a snapshot. The query will check the index
     /// it existed at the moment the snapshot was taken.
-    fn snapshot_nonfinalized_state(&self) -> &Self::Snapshot;
+    fn snapshot_nonfinalized_state(&self) -> Self::Snapshot;
```
nuttycom, line 22: This is a good change; `Self::Snapshot` can always be a reference type if necessary.

AloeareV:
Closing, as #446 will contain these changes in addition to the implementation.

PR #428, opened 2025-07-24
Upgrade makefile toml to symlink to binaries

zancas:
@nachog00 this is draft because in order to debug CI I created a stub job, and turned off most of the jobs.

I believe you have much better tools for this.


nachog00:
I know it's unconventional, but CI passing on this PR is mostly irrelevant, since this changes are for local tooling and for the image that runs CI, but before that is built and pushed, this runs on the old image. Pushing it straight away and having this run would work, but it would contaminate all other branches with new untested CI image.

For us to say that this is tested, I would say we need this dockerfile built, pushed with some tag like "trying-out" and another branch (or even this one, but it would have to be reverted before merging) to run the test suite with that image. That is, with "container: zingodevops/zaino-ci:trying out"

nachog00:
> I believe you have much better tools for this.

I usually do end up making dummy PR's adn branches... I tried ACT some time back, but it started deviating quite a bit from the setup i was running stuff on. It's probably worth it to revisit that tool, as it's really good when you want to  'unit test' a workflow or step... not so much when you are trying to test out github related behaviors like triggers and so on.

zancas:
Superseded.

PR #429, opened 2025-07-24
Zaino db version handling

idky137:
Add Version handling layer to ZainoDB, moved code into submodules.

- built on top of #342

PR #431, opened 2025-07-25
Deserialization for pre-v4 blocks

dorianvp:
On top of #429.

- Update `BlockHeaderData::parse_from_slice` to conditionally skip fields not present in pre-v4 block headers.
- Implement `FullTransaction::parse_vX` for older transaction formats.


zaino-fetch/src/chain/transaction.rs
```diff
@@ -146,7 +146,46 @@ fn parse_transparent(data: &[u8]) -> Result<(&[u8], Vec<TxIn>, Vec<TxOut>), Pars
...
+/// The fields that this function parses are:
+///
+/// - nJoinSplit: compactSize
+/// - vJoinSplit: JSDescriptionBCTV14\[nJoinSplit\] (if version 1 < tx_version < 4)
+/// - vJoinSplit: JSDescriptionGroth16\[nJoinSplit\] (if tx_version >= 4)
+/// - joinSplitPubKey: byte\[32\]
+/// - joinSplitSig: byte\[64\]
+///
+/// TODO: Enable skipped fields (all of them currently).
+fn parse_joinsplits(data: &[u8], version: u32) -> Result<(&[u8], Vec<JoinSplit>), ParseError> {
```
AloeareV: iiuc, we're reading joinsplits...and then after a JoinSplit is a pubkey and sig, which are separate and we ignore them?

AloeareV (commented):
Everything here that I understand looks great, I am confused by what's happening in the joinsplit code!

zaino-fetch/src/chain/transaction.rs
dorianvp: Updated!

dorianvp:
Adding tests. Will mark as ready for review once those land.

dorianvp:
For some reason, `zcashd` hangs when launching it with activation heights other than 1.

zaino-fetch/src/chain/transaction.rs
```diff
@@ -422,6 +422,10 @@ struct TransactionData {
     ///
     /// Size\[bytes\]: Vec<8+CompactSize>
     transparent_outputs: Vec<TxOut>,
+    // /// Unix-epoch UTC time or block height, encoded as in Bitcoin (TODO: what does this mean?).
+    // ///
+    // /// Size\[bytes\]: 4
+    // lock_time: Option<u32>,
```
idky137: Why has an extra lock_time been added above the current NLockTime? Unless I am mistaken there is only one lock_time filed here so trying to skip or read extra bytes would break deserialisation.

idky137 (dismissed):
Looks really close, I left one comment about the add lock_time field and I am a bit confused about the tests which have (or will be?) added. Are there anybtests that have been added to test deserialising v1, v2 and v3 blocks?

zaino-fetch/src/chain/transaction.rs
dorianvp: You're right, my mistake. Removed!

dorianvp:
I'm trying to do a `quick_shield`, but zingolib seems to be generating a transaction with the wrong transaction version:

```bash
thread 'varying_block_and_transaction_deserialization' panicked at integration-tests/tests/test_vectors.rs:709:41:
called `Result::unwrap()` on an `Err` value: SendError(TransmissionError(TransmissionFailed("Send Error: status: Internal, message: \"unhandled rpc-specific zaino_fetch::jsonrpsee::response::SendTransactionError error: RPC Error (code: -25): failed to validate tx: WtxId(\\\"private\\\"), error: transaction did not pass consensus validation: transaction version 5 not supported by the network upgrade Canopy\", details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Tue, 05 Aug 2025 20:42:50 GMT\", \"content-length\": \"0\"} }")))
```

AloeareV:
> I'm trying to do a `quick_shield`, but zingolib seems to be generating a transaction with the wrong transaction version:
> 
> ```shell
> thread 'varying_block_and_transaction_deserialization' panicked at integration-tests/tests/test_vectors.rs:709:41:
> called `Result::unwrap()` on an `Err` value: SendError(TransmissionError(TransmissionFailed("Send Error: status: Internal, message: \"unhandled rpc-specific zaino_fetch::jsonrpsee::response::SendTransactionError error: RPC Error (code: -25): failed to validate tx: WtxId(\\\"private\\\"), error: transaction did not pass consensus validation: transaction version 5 not supported by the network upgrade Canopy\", details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Tue, 05 Aug 2025 20:42:50 GMT\", \"content-length\": \"0\"} }")))
> ```

Zingolib is capable of syncing pre-orchard blocks, but it looks like it's assuming that any sends it does will be post-orchard...this is fine for any non-regtest scenario, but it makes testing older network conditions awkward...I think to generate these kinds of test vectors in the past we've used the zcash-cli wallet

nachog00:
The Bug

  The `JoinSplit::parse_from_slice()` method was using incorrect proof sizes across all transaction versions. It was hardcoded to skip 192 bytes for zk-SNARK proofs, but this is
  only correct for v4+ transactions that use Groth16 proofs. v2/v3 transactions use the older BCTV14 proof system which produces 296-byte proofs.

  This meant that for every `JoinSplit` in v2/v3 transactions, the parser was missing 104 bytes (296 - 192). Since many test vectors contain 2 JoinSplits, this resulted in
  exactly 208 bytes of unparsed data per transaction.

  The Fix

  Made `JoinSplit` proof parsing version-dependent:
  - v2/v3 transactions now correctly parse 296-byte BCTV14 proofs
  - v4+ transactions continue using 192-byte Groth16 proofs
  - Updated all callers to pass the transaction version to `JoinSplit::parse_from_slice()`

  The fix ensures complete data consumption for all transaction versions without breaking existing v4+ parsing.

  New Tests

  Added comprehensive unit tests for transaction parsing across all versions (v1, v2, v3, v4) using the existing test vector data. These tests validate that:
  - All transaction data is consumed during parsing (no remaining bytes)
  - Version-specific parsing logic works correctly
  - Transaction properties match expected test vector metadata

AloeareV (approved):
Fantastic!

PR #432, opened 2025-07-25
fix clippy in CI

### LINKED ISSUE #434: Clippy skew

nachog00:
To enforce errors to be addressed before proceeding.
I'm not sure that this is the fix, as in the past I noticed clippy being a bit quirky in the way it's output is interpreted by the job runner, hence why this config was in place IIUC.

Let's see what CI make's of it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,6 +205,20 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # This is key to decide on te clippy job to run.
+      # The one from actions-rs prints nice annotations accross the pr, when browsing the changed files, right where the actual warning/error was found.
+      # It requires access to the GITHUB_TOKEN tho, a secret github automatically injects into every workflow run, but only for internally owned branches, not for forks.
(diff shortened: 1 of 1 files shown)


nachog00:
**Update**

I realise now that we were previously running two rounds of clippy, and ignoring both of them on the job status.

Now, there are two actions:
- `actions-rs/clippy-check@v1`: This one takes the clippy output and annotates the PR so we can see the clippy snippets  right where they matter on the file diffs. It needs access to GITHUB_TOKEN, a secret github injects on every workflow run, but isn't available on runs for external contributors.
- console output clippy: this is just a manual `clippy` run, it doesn't annotate the PR, but simply outputs the result and makes the step and job fail. It's good enough for an external contributor.

I made it so that the clippy job first checks if the `GITHUB_TOKEN` secret is available, and run either of the clippy steps accordingly.

nachog00:
Looks like it's working! I say we merge this as is and apply the modifications to the source code on a separate PR.
Not only does that keep this PR's scope concise, but also, I plan on making a dummy PR from a fork of `dev` once this merges, to see how this updated job handles the very same code from an external contributor.

zancas (approved):

PR #441, opened 2025-07-28
Bake protoc in ci image

nachog00:
- Bake protoc in ci image
- remove rocksdb bake as it was causing plenty of errors


zancas:
Hmm...   I think the rocksdb bake is necessary for this image file to support building the repo when mounted into a container running on an Arch kernel.

zancas:
Ahh..   in the newer version rocksdb is built during the image build to ensure compatibility with both Debian and Arch user environments.

nachog00:
not longer relevant after #449

PR #445, opened 2025-07-29
Diminish container user privs

zancas:


.github/workflows/ci.yml
```diff
@@ -169,7 +172,7 @@ jobs:
         uses: test-summary/action@v2
         with:
           paths: "target/nextest/ci/junit.xml"
-        if: always()
```
nachog00, line 172: why disabling this? it's what reports the test results in github's UI

Cargo.toml
```diff
@@ -9,6 +9,7 @@ members = [
     "zaino-fetch",
     "zaino-proto",
 ]
+exclude = [ "externals" ]
```
nachog00: Maybe this doesn't belong on this PR ? also not an issue

nachog00:
We agreed to have the `DOCKER_DIR` hash not written into `.env.testing-artifacts` but loaded at runtime

dorianvp (approved):

PR #416, opened 2025-07-14
chain_cache conglomerate object api

AloeareV:

PR #446, opened 2025-07-29
Chain index interface

AloeareV:
Supercedes #416


idky137 (dismissed):
Should the new error types in interface.rs live in the error mod rather than in the interface mod? Not blocking but maybe we should move all errors into error mods to tidy up the crates layout?

dorianvp (dismissed):

AloeareV:
CI is failing, tests are passing locally for me. If anyone else runs these tests locally, could you confirm if they pass or fail for you?

zancas:
I am CLTing da8a916 ...

zancas:
clt_335c5

zancas:
I used **makers container-test**

nachog00 (approved):

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+use zebra_state::HashOrHeight;
+
+/// The combined index. Contains a view of the mempool, and the full
+/// chain state, both finalized and non-finalized, to allow queries over
+/// the entire chain at once. Backed by a source of blocks, either
+/// a zebra ReadStateService (direct read access to a running
+/// zebrad's database) or a jsonRPC connection to a validator.
+///
+/// TODO: Currently only contains the non-finalized state.
+pub struct NodeBackedChainIndex {
```
nuttycom, line 22: nit: I'm surprised to look at the `interface` module, and have the first thing I see be the concrete implementation. Why is the `interface` module necessary at all?

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// Given inclusive start and end indexes, stream all blocks
+    /// between the given indexes. Can be specified
+    /// by hash or height.
+    #[allow(clippy::type_complexity)]
+    fn get_block_range(
+        &self,
+        nonfinalized_snapshot: &Self::Snapshot,
+        start: Option<HashOrHeight>,
+        end: Option<HashOrHeight>,
+    ) -> Result<Option<impl Stream<Item = Result<Vec<u8>, Self::Error>>>, Self::Error>;
```
nuttycom, line 130: This is a weird interface; you could have a starting hash and an ending height, or vice versa, or optional mixtures of these. My sense is that this range should be strictly in terms of block height; it's always possible for a caller to look up a block height by block hash if they need to. Since this API works in terms of a specific snapshot, the block height is always well-defined.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn find_fork_point(
+        &self,
+        snapshot: &Self::Snapshot,
+        block_hash: &types::Hash,
+    ) -> Result<Option<(types::Hash, types::Height)>, Self::Error>;
+    /// given a transaction id, returns the transaction
+    fn get_raw_transaction(
+        &self,
+        snapshot: &Self::Snapshot,
+        txid: [u8; 32],
```
nuttycom, line 142: nonblocking suggestion: Since `zaino` already relies on `zcash_protocol`, you could use the `zcash_protocol::txid::TxId` newtype here.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+    ) -> impl Future<Output = Result<Option<Vec<u8>>, Self::Error>>;
+    /// Given a transaction ID, returns all known hashes and heights of blocks
+    /// containing that transaction. Height is None for blocks not on the best chain.
+    fn get_transaction_status(
+        &self,
+        snapshot: &Self::Snapshot,
+        txid: [u8; 32],
+    ) -> Result<HashMap<types::Hash, Option<types::Height>>, Self::Error>;
```
nuttycom, line 150: Although, looking at this API, it would be weird to use a `zcash_protocol` txid but not a `zcash_protocol` `BlockHash` and/or `BlockHeight`.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+    fn get_chainblock_by_hashorheight<'snapshot, 'self_lt, 'output>(
+        &'self_lt self,
+        non_finalized_snapshot: &'snapshot NonfinalizedBlockCacheSnapshot,
+        height: &HashOrHeight,
+    ) -> Option<&'output ChainBlock>
+    where
+        'snapshot: 'output,
+        'self_lt: 'output,
+    {
+        //TODO: finalized state
```
nuttycom, line 82: My suspicion is that access to the finalized state will have potential error cases as it may involve disk/database access, so I'm guessing that the API of this will have to change to wrap the output in a `Result`?

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+        end: Option<HashOrHeight>,
+    ) -> Result<Option<impl Stream<Item = Result<Vec<u8>, Self::Error>>>, Self::Error> {
+        // with no start supplied, start from genesis
+        let Some(start_block) = self.get_chainblock_by_hashorheight(
+            nonfinalized_snapshot,
+            &start.unwrap_or(HashOrHeight::Height(zebra_chain::block::Height(1))),
+        ) else {
+            return if start.is_some() {
+                Ok(None)
+            } else {
```
nuttycom, line 181: I don't understand why this returns `Ok(None)` if the caller specifies a start height, but an error otherwise. If `None` is a valid argument, and the default is height `1`, then I guess that this means that a genesis block exists but no chain exists? That's a bit weird, and the weirdness exists because of the default; if a start height isn't specified, why does this not start at the genesis block?

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+                })
+            };
+        };
+        let Some(end_block) = self.get_chainblock_by_hashorheight(
+            nonfinalized_snapshot,
+            &end.unwrap_or(HashOrHeight::Hash(nonfinalized_snapshot.best_tip.1.into())),
+        ) else {
+            return if start.is_some() {
+                Ok(None)
+            } else {
```
nuttycom, line 195: This is even weirder; here, the `Ok(None)` makes no sense to me at all. If the user doesn't specify an end height, that just means that they want the whole chain, right? But if we can't get the chain tip block, how does the user having specified a start height make any difference for what is returned.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+                    message: "chaintip missing from database".to_string(),
+                    source: None,
+                })
+            };
+        };
+
+        let mut nonfinalized_block = nonfinalized_snapshot.get_chainblock_by_hash(end_block.hash());
+        let first_nonfinalized_hash = nonfinalized_snapshot
+            .get_chainblock_by_hash(start_block.hash())
+            .map(|block| block.index().hash());
```
nuttycom, line 207: Now I'm even more puzzled. If `get_chainblock_by_hashorheight` returned `None` for either start or end, we've already returned now. What happens if the requested height exists only within the nonfinalized snapshot? It looks to me like this is expecting the nonfinalized range to start within the part of the chain provided by the node. Am I totally misunderstanding what's happening here?

nuttycom (commented):
Flushing a few would-have-been-blocking review comments.

zaino-state/src/chain_index/interface.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// At most one of these blocks will be on the best chain
+    ///
+    fn get_transaction_status(
+        &self,
+        snapshot: &Self::Snapshot,
+        txid: [u8; 32],
+    ) -> Result<
+        HashMap<super::types::Hash, std::option::Option<super::types::Height>>,
+        ChainIndexError,
+    > {
```
nuttycom, line 285: I don't understand why this function takes a snapshot; given a snapshot, it can't possibly occur in more than one block, and therefor the return type being a HashMap is weird. It should either not take a snapshot, and instead should look at all *available* snapshots, or if it takes a snapshot it should just return a `Result<Option<(Hash, Option<Height>)>, ...>`

zaino-state/src/chain_index/interface.rs
AloeareV: Well caller is allowed to specify a block hash or height that isn't in the chain, and a None will be returned. The error return is the special case, as the only block that we can be absolutely certain is in zaino's index is the genesis block, and its nonpresence should be an error imo. 

Edit: I see the confusion, "if a start height isn't specified, why does this not start at the genesis block?" this else case is only hit when zaino cannot find the genesis block, which is an error. If the caller supplied a block which zaino can't find, we return None.

zaino-state/src/chain_index/interface.rs
AloeareV: Again, this case is only hit where the boundary block isn't found in zaino's database. If the boundary block was user-provided, then it's not an error as the user could have specified a block that doesn't exist, but if the boundary block wasn't supplied, we start from a block that is known to be in zaino's database, and then if we immediately can't find it, something is wrong.

zaino-state/src/chain_index/interface.rs
AloeareV: I'm not sure about "given a snapshot, it can't possibly occur in more than one block". I think it makes sense to keep known non-best-chain blocks in a snapshot.

nuttycom: Ah, okay, this is a different meaning of "snapshot" than I was understanding; I had thought that a snapshot referred to a particular branch of the non-finalized state, not multiple branches of the non-finalized state.

zaino-state/src/chain_index/interface.rs
AloeareV: Yes. This is a private helper anyway, mostly to handle conversion between public interface types, and internal methods that work with zebra HashOrHeights.

zaino-state/src/chain_index/interface.rs
nuttycom: But why is the start height/hash relevant, given that this is determining the end block?

zaino-state/src/chain_index/interface.rs
nuttycom: The genesis block has height 0, not height 1, IIRC? But the default is height 1.

zaino-state/src/chain_index/interface.rs
AloeareV: Oh, huh..here are zebra types in the public interface still. That's an oversight... The reasoning behind making it accept hashes as arguments is to allow for getting a range containing blocks that are on a non-best chain in the provided snapshot. If this method only needs to support the best chain as of the provided snapshot, that can simplify things.

zaino-state/src/chain_index/interface.rs
AloeareV: This is collecting blocks from the nonfinalized state top-down, starting from the highest block in the nonfinalized snapshot. If the range starts within the nonfinalized snapshot, we'll hit the equality check and stop. If the range doesn't start in the nonfinalized snapshot, we'll stop when we search up a block that's not in the nonfinalized snapshot. 

This range will then need to be reversed, and appended to the end of the (not yet implemented) finalized section of the range.

zaino-state/src/chain_index/interface.rs
AloeareV: Yeah, the snapshot is the entirety of the non-finalized state as it existed at the time the snapshot was taken.

zaino-state/src/chain_index/interface.rs
AloeareV: Oh, whoops, that's supposed to be end... The dangers of duplicate code! This is already (inadvertently) fixed in https://github.com/zingolabs/zaino/pull/448 where I helperised start and end determination.

zaino-state/src/chain_index/interface.rs
AloeareV: I thought block height started at 1. If it was, height zero should be invalid...so it must be 0? I think this is an off-by-one error

zaino-state/src/chain_index/interface.rs
AloeareV: the `chain_index` mod was getting too big...maybe the public interface only should go back into the parent `chain_index` mod, and the internals stay deeper in..

zaino-state/src/chain_index/interface.rs
nuttycom: Genesis is block 0: https://3xpl.com/zcash/block/0

zaino-state/src/chain_index/interface.rs
AloeareV: Hm. I'll have to take a look to see if anything from zcash_protocol is in the public API already...if it is, might as well use those types in here. However, we already have a block hash newtype in zaino, a zaino txid newtype would be more consistent with the rest of the interface as you mention below...

nuttycom: Yeah, if you're not using `zcash_protocol` in the public API then it's better to use your own newtypes.

zaino-state/src/chain_index/interface.rs
nuttycom: Yeah, I can't think of a scenario where you want a block range that's not on the best chain. Certainly not one that we'd want for now; I think it's better to implement this as just taking a range of heights.

I would also suggest making the start height non-optional; if someone wants to start from zero, they can explicitly start from zero - only the end height may be inefficient to separately determine.

PR #448, opened 2025-07-30
incorporate finalised_state into chain_index 

AloeareV:


zaino-fetch/src/chain/utils.rs
```diff
@@ -14,6 +14,7 @@ pub trait ParseFromSlice {
     /// tx_version is used for deserializing sapling spends and outputs.
     fn parse_from_slice(
         data: &[u8],
+        // TODO: Why is txid a vec of vecs?
```
idky137, line 17: This was used due to the output type of `JsonRpcConnector::get_block()` ( https://github.com/zingolabs/zaino/blob/636cf5343b26ce21d241fa5f2833cf225f800fb5/zaino-fetch/src/jsonrpsee/response.rs#L797 ).

This could be updated to use the new `TransactionHash` types in the future if we wanted.

zaino-state/src/chain_index.rs
```diff
@@ -27,3 +38,319 @@ pub mod types;
...
+impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
+    pub(super) fn start_sync_loop(
+        &self,
+    ) -> tokio::task::JoinHandle<Result<std::convert::Infallible, SyncError>> {
+        let nfs = self.non_finalized_state.clone();
+        let fs = self.finalized_db.clone();
+        tokio::task::spawn(async move {
+            loop {
+                nfs.sync(fs.clone()).await?;
+                //TODO: configure
```
idky137: Is the plan for finalised blocks ( > 100 blocks deep ) to be pushed from the non-finalised state to the finalised state in this method?

I think this can be added in a separate PR but maybe we should expand the TODO?

zaino-state/src/chain_index.rs
```diff
@@ -27,3 +38,319 @@ pub mod types;
...
+    async fn blocks_containing_transaction<'snapshot, 'self_lt, 'iter>(
+        &'self_lt self,
+        snapshot: &'snapshot NonfinalizedBlockCacheSnapshot,
+        txid: [u8; 32],
+    ) -> Result<impl Iterator<Item = ChainBlock> + use<'iter, Source>, FinalisedStateError>
+    where
+        'snapshot: 'iter,
+        'self_lt: 'iter,
+    {
+        Ok(snapshot
```
idky137, line 174: Should we also check mempool for unconfirmed? (This can be done in a separate PR)

zaino-state/src/chain_index.rs
```diff
@@ -27,3 +38,319 @@ pub mod types;
...
+        }
+    }
+
+    /// given a transaction id, returns the transaction
+    async fn get_raw_transaction(
+        &self,
+        snapshot: &Self::Snapshot,
+        txid: [u8; 32],
+    ) -> Result<Option<Vec<u8>>, Self::Error> {
+        let Some(block) = self
```
idky137, line 297: As above, once mempool is integrated we may want to check there also.

idky137: Maybe we should add a TODO?

zaino-state/src/chain_index.rs
```diff
@@ -27,3 +38,319 @@ pub mod types;
...
+
+    /// Given a transaction ID, returns all known blocks containing this transaction
+    /// At most one of these blocks will be on the best chain
+    ///
+    async fn get_transaction_status(
+        &self,
+        snapshot: &Self::Snapshot,
+        txid: [u8; 32],
+    ) -> Result<HashMap<types::Hash, std::option::Option<types::Height>>, ChainIndexError> {
+        Ok(self
```
idky137, line 341: Similar to above, will we want to check for "in_mempool" status? (also can be added in separate PR but we may want a todo.

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -238,7 +245,7 @@ impl NonFinalizedState {
             transactions.push(txdata);
         }
 
-        let height = Some(Height(1));
+        let height = Some(Height(0));
```
idky137: Should this use the Genesis height const?

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -103,26 +112,24 @@ pub enum InitError {
...
-        source: ValidatorConnector,
-        network: Network,
-    ) -> Result<Self, InitError> {
+    pub async fn initialize(source: Source, network: Network) -> Result<Self, InitError> {
         // TODO: Consider arbitrary buffer length
         let (staging_sender, staging_receiver) = mpsc::channel(100);
         let staged = Mutex::new(staging_receiver);
-        // The comment on the Height type is somewhat unclear if genesis block is height 0 or height 1
-        // TODO: Confirm
         let genesis_block = source
```
idky137: Do we always want to initialise the non-finalised state with Genesis block? Does this mean we would have to look through the whole chain to find the current tip? Maybe I am mistake as to how this works but if this is the case then would it be better to load the finalised state tip (zaino_db_tip) and initialise from there? Maybe this could be passed into initialise?

zaino-state/src/chain_index/non_finalised_state.rs
idky137: Or updated as above.

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -656,10 +589,181 @@ impl NonFinalizedState {
         }
     }
 
-    /// Get a copy of the block cache as it existed at the last [`Self::update`] call
+    /// Get a snapshot of the block cache
     pub fn get_snapshot(&self) -> Arc<NonfinalizedBlockCacheSnapshot> {
         self.current.load_full()
     }
+
+    async fn block_to_chainblock(
```
idky137, line 616: There is a try_from impl for this transform available in types.rs, it may be cleaner to remove this method and rely on a single try_from used in all places. ( https://github.com/zingolabs/zaino/blob/636cf5343b26ce21d241fa5f2833cf225f800fb5/zaino-state/src/chain_index/types.rs#L1347 )

idky137 (dismissed):
Ive added a few comments but I think we can merge this with a couple todo's and make the final changes in subsequent PRs. 

Im also happy fo these changes to be made here with the exception of the mempool changes as they will require changes in #472

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: This comment made me realize that I hadn't added functionality to the chain_index at all for save/load. I've added an optional chainblock argument to initialize, but I haven't added in a load wrapper that calls it yet.

zaino-state/src/chain_index.rs
AloeareV: Hmm. When I originally wrote the sync code for the non-finalized state, I had envisioned it checking the height of the finalized state to know when to truncate...at the time, I didn't know what sync interface the finalized state would support. Finalized state syncing may want to go into a follow-up PR, it shouldn't be too big an addition, but this PR is already huge.

zaino-state/src/chain_index.rs
AloeareV: Yeah, mempool is still in flight

idky137 (dismissed):

idky137 (dismissed):

zancas (approved):

PR #458, opened 2025-07-31
cargo update

zancas:
I *think* that this should *just work* and likely we should run it quite often.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,9 +162,9 @@ checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "async-compression"
-version = "0.4.25"
+version = "0.4.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "40f6024f3f856663b45fd0c9b6f2024034a702f453549449e0d84a305900dad4"
+checksum = "ddb939d66e4ae03cee6091612804ba446b12878410cfa17f785f4dd67d4014e8"
 dependencies = [
  "futures-core",
  "memchr",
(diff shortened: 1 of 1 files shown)


nachog00:
What about something like this?

```yaml
name: Cargo Update and Pull Request
                                                                                                                                                                                                                                          
on:
  schedule:
    - cron: '0 0 * * 1'  # Run every Monday at 00:00 UTC
  workflow_dispatch:     # Allows manual triggering
                                                                                                                                                                                                                                          
jobs:
  update-dependencies:
    runs-on: ubuntu-latest

    steps:
    - name: Checkout Repository
      uses: actions/checkout@v3

    - name: Set up Rust
      uses: actions-rs/toolchain@v1
      with:
        toolchain: stable
        override: true

    - name: Update Cargo Dependencies
      run: cargo update

    - name: Commit Changes
      run: |
        git config --local user.name "github-actions[bot]"
        git config --local user.email "github-actions[bot]@users.noreply.github.com"
        git checkout -b update-dependencies
        git add -A
        git commit -m "Update Cargo dependencies" || echo "No changes to commit"

    - name: Push Changes
      env:
        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      run: |
        if git diff-index --quiet HEAD; then
          echo "No changes, nothing to push."
        else
          git push origin update-dependencies
        fi

    - name: Create Pull Request
      env:
        GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
      run: |
        if git diff-index --quiet HEAD; then
          echo "No changes, no PR needed."
        else
          gh pr create --title "Update Cargo dependencies" --body "This PR updates Cargo dependencies to their latest versions." --head update-dependencies --base main
        fi

```

zancas:
Can we leverage dependabot so that there's an event driven update each time a dependency updates.   What we really want is to detect any issue as early as possible.

nachog00:
We can probably have dependabot act as a trigger for that action i shared. Not sure tho,haven't done much with dependabot.

nachog00 (approved):

PR #465, opened 2025-08-01
sync test targets

nachog00:
## Motivation

CI test targets were out of sync and empty target set wouldn't fail CI.

## Solution

- ran `makers update-test-targets` to sync targets in CI
- set `--no-tests fail` in CI job to fail empty sets


nachog00:
There's some bug with the task `update-test-targets` since it didn't get rid of int-tests::test-vectors even tho it has zero testcases

dorianvp (approved):

zancas (approved):

zancas (approved):

PR #471, opened 2025-08-12
Replace light wallet .proto files with canonical versions obtained from https://github.com/zcash/lightwallet-protocol

nuttycom:
## Motivation

The Zcash ecosystem needs a canonical definition of the lightwalletd protocol; https://github.com/zcash/lightwallet-protocol is proposed as the location for that canonical set of definitions.

## Solution

This PR replaces the copies of the `service.proto` and `compact_formats.proto` files that have previously been used for local builds based on these definitions with symbolic links to the canonical versions, which will be maintained in the future using `git subtree`. The command used to bring in those files was:

```
git subtree add --prefix=zaino-proto/lightwallet-protocol --squash git@github.com:zcash/lightwallet-protocol.git main
```

### Tests

This PR relies on existing build and integration testing; as it does not actually modify any of the content of the .proto files (the canonical versions that were brought in in this PR are identical to the existing .proto files) there is no semantic change, nor any delta to any of the generated files produced by the build using this files.


zancas (approved):
Thanks for doing this @nuttycom !  Single Source Of Truth!!

PR #472, opened 2025-08-13
Chain index mempool integration update

idky137:
## Motivation

Required updates for the current mempool code to work withing the new ChainIndex.

### PR Checklist

- [ ] The solution is tested.


idky137:
There are currently 2 failing tests on the PR due to a bug in the testvectors used in unti tests, fixing this is currently blocked on #431.

Apart from this issue, this PR is now complete.

zaino-state/src/backends/state.rs
```diff
@@ -1335,10 +1324,32 @@ impl ZcashIndexer for StateServiceSubscriber {
                     .await
                 {
                     Some(tx) => {
-                        if let GetRawTransaction::Object(transaction_object) = &tx.0 {
-                            Ok(GetRawTransaction::Object(transaction_object.clone()))
-                        } else {
-                            unreachable!("unmatched response to a `Transaction` read request")
+                        let serialized = tx.as_ref().0.clone();
```
Oscar-Pepper, line 1312: [non-blocking] its not clear to me why the TxId and SerializedTransaction's are wrapped in MempoolKey and MempoolValue structs? 
And also why the `get` methods return an Arc<Tx> where Tx has been cloned into the arc anyway. We could just return the cloned Tx and then wrap in an arc where needed after `get` is called.... or it could already be wrapped in an arc so the `get` method just clones the Arc instead of creating one which is more efficient as it only has to clone the ref to the Vec<u8> instead of cloning the Vec<u8> itself . This may need structural changes though, i havent looked into it deeply

zaino-state/src/chain_index/source.rs
```diff
@@ -50,11 +65,26 @@ pub struct InvalidData(String);
 
 type BlockchainSourceResult<T> = Result<T, BlockchainSourceError>;
 
+/// Currently the Mempool cannot utilise the mempool change endpoint in the ReadStateService,
+/// for this reason the lagacy jsonrpc inteface is used until the Mempool updates required can be implemented.
```
Oscar-Pepper, line 80:
```suggestion
/// for this reason the legacy jsonrpc interface is used until the Mempool updates required can be implemented.
```

zaino-state/src/chain_index/source.rs
```diff
@@ -50,11 +65,26 @@ pub struct InvalidData(String);
...
+///
+/// Due to the difference if the mempool inteface provided by the ReadStateService and the Json RPC service
+/// two seperate Mempool implementation will likely be required.
```
Oscar-Pepper, line 83:
```suggestion
/// Due to the difference between the mempool interface provided by the ReadStateService and the Json RPC service,
/// two seperate Mempool implementations will likely be required.
```

zaino-state/src/chain_index/source.rs
```diff
@@ -215,6 +243,178 @@ impl ValidatorConnector {
...
+
+    pub(super) async fn get_mempool_txids(
+        &self,
+    ) -> BlockchainSourceResult<Option<Vec<zebra_chain::transaction::Hash>>> {
+        let mempool_fetcher = match self {
+            ValidatorConnector::State(state) => &state.mempool_fetcher,
+            ValidatorConnector::Fetch(json_rp_see_connector) => json_rp_see_connector,
+        };
+
+        let txid_strings = mempool_fetcher
```
Oscar-Pepper, line 266: [non-blocking] as mentioned in a doc comment you added, these should be stored as bytes

zaino-state/src/chain_index/source.rs
```diff
@@ -215,6 +243,178 @@ impl ValidatorConnector {
...
+
+                if let zebra_state::ReadResponse::Transaction(opt) = resp {
+                    if let Some(found) = opt {
+                        return Ok(Some((found).tx.clone()));
+                    }
+                } else {
+                    unreachable!("unmatched response to a `Transaction` read request");
+                }
+
+                // Else heck mempool for transaction.
```
Oscar-Pepper, line 320:
```suggestion
                // Else check mempool for transaction.
```

zaino-state/src/chain_index/source.rs
```diff
@@ -215,6 +243,178 @@ impl ValidatorConnector {
...
+        match self {
+            ValidatorConnector::State(State {
+                read_state_service,
+                mempool_fetcher,
+            }) => {
+                // Check state for transaction
+                let mut read_state_service = read_state_service.clone();
+                let mempool_fetcher = mempool_fetcher.clone();
+
+                let txid_tr: zebra_chain::transaction::Hash =
```
Oscar-Pepper: tr?

zaino-state/src/chain_index/source.rs
```diff
@@ -235,13 +435,35 @@ impl BlockchainSource for ValidatorConnector {
...
+        txid: Hash,
+    ) -> BlockchainSourceResult<Option<Arc<zebra_chain::transaction::Transaction>>> {
+        self.get_transaction(txid).await
+    }
+
+    async fn get_best_block_hash(
+        &self,
+    ) -> BlockchainSourceResult<Option<zebra_chain::block::Hash>> {
+        self.get_best_block_hash().await
+    }
```
Oscar-Pepper: shouldnt we just move the implemtenations here instead of re-calling them? i dont see any benefit for having both but in this module (and parent module) there are now two methods with the same name

zaino-state/src/chain_index/source.rs
```diff
@@ -268,27 +491,77 @@ pub(crate) mod test {
...
+                roots,
+                hashes,
+                active_chain_height: Arc::new(AtomicU32::new(active_chain_height)),
             }
+        }
 
-            // Block height indexing starts at 1, vec indexing starts at 0..
-            let index = (height - 1) as usize;
+        pub(crate) fn mine_blocks(&self, blocks: u32) {
+            let max_tip = self.blocks.len().saturating_sub(1) as u32;
```
Oscar-Pepper: call max_chain_height to DRY as I believe i found a bug with this so best to call throughout. see below

zaino-state/src/chain_index/source.rs
```diff
@@ -268,27 +491,77 @@ pub(crate) mod test {
...
+        }
+
+        fn height_to_index(&self, height: u32) -> Result<usize, BlockchainSourceError> {
+            let index = height as usize;
 
             if index >= self.blocks.len() {
                 return Err(BlockchainSourceError::Unrecoverable(format!(
                     "Height {height} is out of range (max height = {})",
-                    self.blocks.len()
+                    self.blocks.len().saturating_sub(1)
```
Oscar-Pepper: call max_chain_height to DRY as I believe i found a bug with this so best to call throughout. see above

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+            };
+
+            Ok(Some(txids))
+        }
+
+        async fn get_transaction(
+            &self,
+            txid: Hash,
+        ) -> BlockchainSourceResult<Option<Arc<zebra_chain::transaction::Transaction>>> {
+            let txid_tr: zebra_chain::transaction::Hash =
```
Oscar-Pepper: tr?

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+                zebra_chain::transaction::Hash::from(txid.0);
+
+            let active_height = self.active_height() as usize;
+            let mempool_index = active_height + 1;
```
Oscar-Pepper:
```suggestion
            let mempool_height = active_height + 1;
```

Oscar-Pepper: i think mempool index can be confusing, its better to use mempool height througout and then -1 from it for the index where needed

zaino-state/src/chain_index/source.rs
```diff
@@ -268,27 +491,77 @@ pub(crate) mod test {
...
+                        None
+                    } else {
+                        Some(target)
+                    }
+                },
+            );
+        }
+
+        pub(crate) fn max_chain_height(&self) -> u32 {
+            self.blocks.len().saturating_sub(1) as u32
```
Oscar-Pepper, line 555: shouldnt this be +1 instead of -1? my understanding is the first block is height 1 and index 0

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+            &self,
+            txid: Hash,
+        ) -> BlockchainSourceResult<Option<Arc<zebra_chain::transaction::Transaction>>> {
+            let txid_tr: zebra_chain::transaction::Hash =
+                zebra_chain::transaction::Hash::from(txid.0);
+
+            let active_height = self.active_height() as usize;
+            let mempool_index = active_height + 1;
+
+            for index in 0..=active_height {
```
Oscar-Pepper: the list index for the active height block is active_height - 1 right? so this should be `end exclusive` instead of inclusive... unless we are using index as a synonym for height and therefore we should be consistent with our naming if possible

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+                if index >= self.blocks.len() {
+                    break;
+                }
```
Oscar-Pepper, line 659: relating to the previous comment, this if statement tells me we are talking about a list index which is infact height - 1. if the equals is removed from the for loop it should be removed from here too. although this is correct logic, its a bug surface to mix heights and indexes IMO

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+                if let Some(found) = self.blocks[index]
```
Oscar-Pepper: its much safer in rust (and also helps avoid the bugs i mentioned previously with index off by 1 errors) to use the iterators instead of looping and hard calling the indexes on a list. for exmaple instead of the above for loop, can we just iterate over a slice of `self.blocks` in the range 0..active_height?

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+                    .transactions
+                    .iter()
+                    .find(|transaction| transaction.hash() == txid_tr)
+                {
+                    return Ok(Some(Arc::clone(found)));
+                }
+            }
+
+            if mempool_index < self.blocks.len() {
+                if let Some(found) = self.blocks[mempool_index]
```
Oscar-Pepper: see above comment

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+                }
+            }
+
+            Ok(None)
+        }
+
+        async fn get_best_block_hash(
+            &self,
+        ) -> BlockchainSourceResult<Option<zebra_chain::block::Hash>> {
+            let active_tip = self.active_height() as usize;
```
Oscar-Pepper:
```suggestion
            let active_height = self.active_height() as usize;
```
i like constisency :)

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+            &self,
+            txid: Hash,
+        ) -> BlockchainSourceResult<Option<Arc<zebra_chain::transaction::Transaction>>> {
+            self.get_transaction(txid).await
+        }
+
+        async fn get_best_block_hash(
+            &self,
+        ) -> BlockchainSourceResult<Option<zebra_chain::block::Hash>> {
+            self.get_best_block_hash().await
```
Oscar-Pepper: as previously stated, can we move the impls here instead of duplicating methods?

Oscar-Pepper (changes requested):

zaino-state/src/chain_index/source.rs
```diff
@@ -268,27 +491,77 @@ pub(crate) mod test {
...
+        pub(crate) fn max_chain_height(&self) -> u32 {
+            self.blocks.len().saturating_sub(1) as u32
+        }
+
+        pub(crate) fn active_height(&self) -> u32 {
+            self.active_chain_height.load(Ordering::SeqCst)
+        }
+
+        fn height_to_index(&self, height: u32) -> Result<usize, BlockchainSourceError> {
+            let index = height as usize;
```
Oscar-Pepper: index should be height - 1 and logic below should be `if index > self. blocks.len()` as length is related to a list index

zaino-state/src/chain_index/source.rs
```diff
@@ -306,19 +579,26 @@ pub(crate) mod test {
                     BlockchainSourceError::Unrecoverable("Block hash not found".to_string())
                 })
         }
-    }
 
-    #[async_trait]
-    impl BlockchainSource for MockchainSource {
         async fn get_block(&self, id: HashOrHeight) -> BlockchainSourceResult<Option<Arc<Block>>> {
+            let active_tip = self.active_height() as usize;
```
Oscar-Pepper:
```suggestion
            let active_height = self.active_height() as usize;
```

zaino-state/src/chain_index/source.rs
```diff
@@ -306,19 +579,26 @@ pub(crate) mod test {
...
+
             match id {
-                HashOrHeight::Height(h) => {
-                    let i = self.height_to_index(h.0)?;
-                    Ok(Some(Arc::clone(&self.blocks[i])))
+                HashOrHeight::Height(height) => {
+                    let index = self.height_to_index(height.0)?;
+                    if index <= active_tip {
```
Oscar-Pepper:
```suggestion
                    if index <= active_height - 1 {
```

zaino-state/src/chain_index/source.rs
```diff
@@ -306,19 +579,26 @@ pub(crate) mod test {
...
+                        Ok(Some(Arc::clone(&self.blocks[index])))
+                    } else {
+                        Ok(None)
+                    }
                 }
                 HashOrHeight::Hash(hash) => {
-                    let i = self.hash_to_index(&hash)?;
-                    Ok(Some(Arc::clone(&self.blocks[i])))
+                    let index = self.hash_to_index(&hash)?;
+                    if index <= active_tip {
```
Oscar-Pepper:
```suggestion
                    if index <= active_height - 1 {
```

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
             id: Hash,
         ) -> BlockchainSourceResult<(Option<(sapling::Root, u64)>, Option<(orchard::Root, u64)>)>
         {
-            let index = self.hashes.iter().position(|h| h == &id);
-            Ok(index.map(|i| self.roots[i]).unwrap_or((None, None)))
+            let active_tip = self.active_height() as usize; // serve up to active tip
```
Oscar-Pepper:
```suggestion
            let active_height = self.active_height() as usize; // serve up to active tip
```

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+
+            if let Some(index) = self.hashes.iter().position(|h| h == &id) {
+                if index <= active_tip {
```
Oscar-Pepper:
```suggestion
                if index <= active_height - 1 {
```

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+                }
+            } else {
+                Ok((None, None))
+            }
+        }
+
+        async fn get_mempool_txids(
+            &self,
+        ) -> BlockchainSourceResult<Option<Vec<zebra_chain::transaction::Hash>>> {
+            let mempool_index = self.active_height() as usize + 1;
```
Oscar-Pepper:
```suggestion
            let mempool_height = self.active_height() as usize + 1;
```

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+
+            let txids = if mempool_index < self.blocks.len() {
```
Oscar-Pepper:
```suggestion
            let txids = if mempool_height - 1 < self.blocks.len() {
```

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+
+            Ok(None)
+        }
+
+        async fn get_best_block_hash(
+            &self,
+        ) -> BlockchainSourceResult<Option<zebra_chain::block::Hash>> {
+            let active_tip = self.active_height() as usize;
+
+            if self.blocks.is_empty() || active_tip >= self.blocks.len() {
```
Oscar-Pepper:
```suggestion
            if self.blocks.is_empty() || active_height - 1 > self.blocks.len() {
```

zaino-state/src/chain_index/source.rs
```diff
@@ -328,8 +608,122 @@ pub(crate) mod test {
...
+                return Ok(None);
+            }
+
+            Ok(Some(self.blocks[active_tip].hash()))
```
Oscar-Pepper: its best to use `get` and return errors for safety, otherwise it may panic unless its guaranteed to always find the value. also its off by one:
```suggestion
            Ok(Some(self.blocks[active_height - 1].hash()))
```

Oscar-Pepper (dismissed):

zaino-state/src/backends/state.rs
idky137: This is historic, it was useful to be able to easily update what the mempool held but is less useful in the new ChainIndex.

We will be updating the mempool for zebra's new interface soon and should make these updates then as I think this change could spiral.

zaino-state/src/chain_index/source.rs
idky137: ah no block height is 0, a mistake we also made.

zaino-state/src/chain_index/source.rs
idky137: I think we are ok here as this is purely unit test code.

zaino-state/src/chain_index/source.rs
idky137: as height is also 0 indexed i think this is correct

zaino-state/src/chain_index/source.rs
idky137: updated in review changes pr

zaino-state/src/chain_index/source.rs
idky137: updated in review changes pr

zaino-state/src/chain_index/source.rs
idky137: all index mentions have now been removed and height indexing is standard across code (zero indexed)

zaino-state/src/chain_index/source.rs
idky137: this is something we will need to update in a bigger mempool update.

zaino-state/src/chain_index/source.rs
idky137: as this is only test code i think this is ok.

zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -232,7 +226,9 @@ impl ZainoDB {
             }
         };
 
-        for height_int in (db_height.0 + 1)..=height.0 {
+        let start_height = if db_height.0 == 0 { 0 } else { db_height.0 + 1 };
```
AloeareV: I don't understand this line. If db height is 0, we start at 0...but if db height is 1, we start at 2? Why do we skip 1?

zaino-state/src/chain_index/finalised_state/migrations.rs
```diff
@@ -145,15 +146,19 @@ impl<T: BlockchainSource> Migration<T> for Migration0_0_0To1_0_0 {
...
                         parent_chain_work = *shadow
                             .get_block_header(shadow_db_height)
                             .await?
                             .index()
                             .chainwork();
                     }
 
-                    for height in (shadow_db_height.0 + 1)..=primary_db_height.0 {
+                    if shadow_db_height.0 != 0 {
+                        shadow_db_height.0 += 1;
```
AloeareV: As above, I don't understand skipping 1. Ohhh...hang on, is this to correct for version 0 having genesis height wrong? As far as I'm aware version 0 was built starting at sapling activation height, so an incorrect genesis height wouldn't have effected our heights

AloeareV (commented):
Looks great, aside from what's currently outstanding. Reviewing to keep track of what changes I've looked at.

zaino-state/src/chain_index/finalised_state.rs
idky137: This is has been fixed for the Genesis height fix in #448 - https://github.com/zingolabs/zaino/blob/d16aa2527e478839e7235ca0ba304fd9b873a2da/zaino-state/src/chain_index/finalised_state.rs#L212.

We will need to either land #448 and #431 before merging this or rebase this on top of #448.

zaino-state/src/chain_index/finalised_state/migrations.rs
idky137: similar to above.

zaino-state/src/chain_index/finalised_state.rs
idky137: Updates to Block deserialisation and test vectors made in #448 are also required for testing this PR.

idky137: Also for reference, updated code runs as follows:
- try to fetch database_tip (returns none if no data found - new database)
- If None, build from Height(0), else build from Height(database_tip + 1).

AloeareV: Oh...If height is found, it's the height that's already in there. If not, then we start from zero. I see!

zaino-state/src/chain_index/source.rs
AloeareV: I also don't understand what `tr` means here

idky137:
Built on top of #448

idky137:
Yes there were several which we went through on a call and worked out that the original code was correct and I have pushed fixes for the others. This will be much easier to see once #448 lands.

zaino-state/src/chain_index/source.rs
idky137: I have changed this to zebra_txid in 
3b189ea

zaino-state/src/chain_index/source.rs
idky137: These have now ben moved into the trait implementation

zaino-state/src/chain_index/source.rs
idky137: updated to call max_chain_height in 
3b189ea

zaino-state/src/chain_index/source.rs
idky137: this has been updated to use max_chain_height in 
3b189ea

zaino-state/src/chain_index/source.rs
idky137: updated to zebra_txid in 
3b189ea

Oscar-Pepper (approved):

PR #478, opened 2025-08-16
Add transaction hash type

Oscar-Pepper:
## Motivation

Hash is used for both blocks and tx's

## Solution

Seperated into two distinct types

### PR Checklist

- [ ] The solution is tested.


zaino-state/src/chain_index/types.rs
```diff
@@ -77,44 +77,44 @@ impl FromHex for Hash {
...
     }
 }
 
-impl From<Hash> for zcash_primitives::block::BlockHash {
-    fn from(h: Hash) -> Self {
+impl From<BlockHash> for zcash_primitives::block::BlockHash {
+    fn from(hash: BlockHash) -> Self {
         // Convert to display order (big-endian)
-        zcash_primitives::block::BlockHash(h.bytes_in_display_order())
+        zcash_primitives::block::BlockHash(hash.bytes_in_display_order())
```
idky137: Looking over the docs in zcash_primitives, I think this should actually be the non-flipped bytes. I think this is a mistake that slipped through review on the original PR as currently zcash_primitives are unused in zaino. Looking at the docs, zcash_primitives expects server byte order.

zaino-state/src/chain_index/types.rs
```diff
@@ -77,44 +77,44 @@ impl FromHex for Hash {
...
     }
 }
 
-impl From<zcash_primitives::block::BlockHash> for Hash {
-    fn from(h: zcash_primitives::block::BlockHash) -> Self {
-        Hash::from_bytes_in_display_order(&h.0)
+impl From<zcash_primitives::block::BlockHash> for BlockHash {
+    fn from(hash: zcash_primitives::block::BlockHash) -> Self {
+        BlockHash::from_bytes_in_display_order(&hash.0)
```
idky137: Similar to above.

zaino-state/src/chain_index/types.rs
```diff
@@ -127,12 +127,130 @@ impl ZainoVersionedSerialise for Hash {
...
+impl From<zebra_chain::transaction::Hash> for TransactionHash {
+    fn from(hash: zebra_chain::transaction::Hash) -> Self {
+        TransactionHash(hash.0)
+    }
+}
+
+impl From<TransactionHash> for zcash_primitives::transaction::TxId {
+    fn from(hash: TransactionHash) -> Self {
+        // Convert to display order (big-endian)
+        zcash_primitives::transaction::TxId::from_bytes(hash.bytes_in_display_order())
```
idky137: See block hash comment

zaino-state/src/chain_index/types.rs
```diff
@@ -127,12 +127,130 @@ impl ZainoVersionedSerialise for Hash {
...
+    }
+}
+
+impl From<zcash_primitives::transaction::TxId> for TransactionHash {
+    fn from(hash: zcash_primitives::transaction::TxId) -> Self {
+        TransactionHash::from_bytes_in_display_order(&hash.into())
```
idky137: Similar to above.

idky137 (changes requested):
One small change which I think you already raised in an internal channel, but other than that looks good!

zaino-state/src/chain_index/types.rs
nuttycom: The in-memory type should not be byte-flipped; transaction and block hashes are *only* byte-flipped for display.

Oscar-Pepper:
in line 2888 of finalised_state/db/v1.rs the bytes loaded into the CompactTxData is flipped compared with the the TxId byte order in the TxidList. @idky137  could you verify this is intended?

idky137:
This matches the current format where ChainBlock holds BE txids, I have updated this in #479 as it should be done at the same time as updating all other logic that uses txids in chainBlock.

idky137:
theses changes wee made in #481

PR #476, opened 2025-08-14
Update to Zcashd 6.3.0 since 6.2.0 EOL'd

pacu:
Please disregard if unnecessary.

Motivation: Zcashd 6.2.0 EOL'd last week

PR #484, opened 2025-08-22
update zcashd version

zancas:
This PR replicates @pacu 's version bump, but I have also pushed a new image to dockerhub.


zancas:
Duplicates #476

PR #487, opened 2025-08-25
Chain index mempool

idky137:
Adds Mempool to chain index

## Motivation

Currently the chain index does not use mempool data, this PR fixes that.

## Solution

- Add mempool to Chain Index.
- Makes mempool broadcast current chain tip for mempoolsubscriber to use.
- Update ChainIndex::get_transaction and transaction status methods to use mempool.
- Add get_mempool_transactions and get_mempool_stream to ChainIndex

### PR Checklist

- [ ] The solution is tested.


zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -400,7 +404,7 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
                     match initial_state
                         .blocks
                         .values()
-                        .find(|block| block.height() == Some(best_tip.0 - 1))
+                        .find(|block| block.height() == Some(next_height_down))
```
AloeareV, line 407: Oh dear. Good catch!

zaino-state/src/chain_index.rs
```diff
@@ -63,41 +64,70 @@ pub trait ChainIndex {
...
+    fn get_mempool_transactions(
+        &self,
+        exclude_list: Vec<String>,
+    ) -> impl std::future::Future<Output = Result<Vec<Vec<u8>>, Self::Error>>;
+
+    /// Returns a stream of mempool transactions, ending the stream when the chain tip block hash
+    /// changes (a new block is mined or a reorg occurs).
+    ///
+    /// If the chain tip has changed from the given spanshot an IncorrectChainTip error is returned.
+    /// NOTE: Is this correct here?
```
AloeareV: I think this should be a none, rather than an error.

zaino-state/src/chain_index.rs
```diff
@@ -63,41 +64,70 @@ pub trait ChainIndex {
...
 /// The combined index. Contains a view of the mempool, and the full
 /// chain state, both finalized and non-finalized, to allow queries over
 /// the entire chain at once. Backed by a source of blocks, either
 /// a zebra ReadStateService (direct read access to a running
 /// zebrad's database) or a jsonRPC connection to a validator.
-///
-/// Currently does not support mempool operations
 pub struct NodeBackedChainIndex<Source: BlockchainSource = ValidatorConnector> {
-    // TODO: mempool
+    _mempool_state: std::sync::Arc<mempool::Mempool<Source>>,
```
AloeareV: Why is this underscored?

zaino-state/src/chain_index.rs
```diff
@@ -333,6 +381,10 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndex<Source> {
             return Ok(None);
         };
 
+        // NOTE: Could we safely use zebra's get transaction method here without invalidating the snapshot?
```
AloeareV, line 384: I think that if we get_transaction, we could get the transaction having been reorged into a different block...it's possible the transaction will be identical in that case. This will need some careful thought to ensure it is safe, whereas getting the block by hash and then the transaction from the block is guaranteed...it would be more efficient.

AloeareV (changes requested):

zaino-state/src/chain_index.rs
idky137: behaviour updated and new unit test added in [80aa6a1](https://github.com/zingolabs/zaino/pull/487/commits/80aa6a17ede3280848c3d2412a02b80312c8778f)

zaino-state/src/chain_index.rs
idky137: The actual mempool _state_ is currently unused once launched so this was added to stop clippy warnings. This will change once we implement subscriber, status and shutdown as these methods will require the actual mempool state rather than a subscriber. I have updated the name and added a temporary clippy allow in 
8a45fa6.

zaino-state/src/chain_index.rs
idky137: Ye I think this can be an optimisation we look at implementing at a later date, I have updated the comment to specify the danger in 
8a45fa6

idky137: Although if we can use the snapshot (and db) to verify that the txid is in the chain then maybe this method doesn't need to know where in the chain it is? (get_transaction_status could be the API for that). 

Either way its not something we need to implement now, but something to mull over..

AloeareV (approved):

PR #493, opened 2025-08-27
Load chain index, shutdown method, and status display

AloeareV:
### Follow-up Work

The status is only ever set to Spawning, Syncing, Ready, and Closing.
Error statuses are not yet reported

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -103,8 +103,7 @@ pub struct NodeBackedChainIndex<Source: BlockchainSource = ValidatorConnector> {
     pub(crate) finalized_db: std::sync::Arc<finalised_state::ZainoDB>,
     finalized_state: finalised_state::reader::DbReader,
     sync_loop_handle: Option<tokio::task::JoinHandle<Result<(), SyncError>>>,
-    status: Arc<AtomicStatus>,
-    shutdown_signal: std::sync::Mutex<futures::channel::oneshot::Sender<()>>,
+    status: AtomicStatus,
 }
 
 impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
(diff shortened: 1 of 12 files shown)


zaino-state/src/chain_index.rs
```diff
@@ -102,6 +102,9 @@ pub struct NodeBackedChainIndex<Source: BlockchainSource = ValidatorConnector> {
     // pub crate required for unit tests, this can be removed once we implement finalised state sync.
     pub(crate) finalized_db: std::sync::Arc<finalised_state::ZainoDB>,
     finalized_state: finalised_state::reader::DbReader,
+    sync_loop_handle: Option<tokio::task::JoinHandle<Result<(), SyncError>>>,
+    status: Arc<AtomicStatus>,
```
idky137: I think this can be a bare `AtomicStatus`, as `AtomicStatus` is just a wrapper for an `Arc<AtomicUsize>`

zaino-state/src/chain_index.rs
```diff
@@ -102,6 +102,9 @@ pub struct NodeBackedChainIndex<Source: BlockchainSource = ValidatorConnector> {
...
+    shutdown_signal: std::sync::Mutex<futures::channel::oneshot::Sender<()>>,
```
idky137: Im not sure we need this extra shutdown signal, other child processes use the AtomicStatus as their shutdown signal which i think is probably also adequate here. 

Below is how the Mempool shutdown signal is sent and handled (the Finalised state also works like this):
Sending shutdown signal: https://github.com/zingolabs/zaino/blob/5130a3e21fc0d3df165d913341547dc74cb68e51/zaino-state/src/chain_index/mempool.rs#L278

Reacting to shutdown signal: https://github.com/zingolabs/zaino/blob/5130a3e21fc0d3df165d913341547dc74cb68e51/zaino-state/src/chain_index/mempool.rs#L180

Unless we have a reason to deviate from this format I think it makes sense to keep interface consistency here.

zaino-state/src/chain_index.rs
```diff
@@ -114,30 +117,74 @@ impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
...
+                        self.status.store(StatusType::Closing as usize);
+                        Ok(())
+                    }
+                    Err(_) => Err(()),
+                }
+            }
+            Err(_) => Err(()),
+        }
+    }
+
```
idky137, line 186: We should also call `mempool.close() and finalised_state.shutdown() in this method to ensure we are closing all long lived child processes and closing the database properly.

We may also want to also implement drop for safety (I think the mempool and finalised state already do so if the non-finalised state does then we may not need to here).

zaino-state/src/chain_index.rs
```diff
@@ -114,30 +117,74 @@ impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
...
+    /// Displays the status of the chain_index
+    pub fn get_status(&self) -> StatusType {
+        StatusType::from(self.status.load())
```
idky137: I think here we may want to call `.status()` on all sub processes (mempool, finalised_state, non_finaliaed_state) and then use combine https://github.com/zingolabs/zaino/blob/5130a3e21fc0d3df165d913341547dc74cb68e51/zaino-state/src/status.rs#L139 , then update self.status with `.store()` and then return self.status().

idky137 (changes requested):
Looks good, I think there are a few places we might want to change before merging though

zaino-state/src/chain_index.rs
AloeareV: Ahhh, I wrote the shutdown signal before the status, and was originally detecting the signal and using it to set the status...this makes much more sense

idky137 (dismissed):

idky137 (dismissed):

idky137 (approved):

PR #494, opened 2025-08-28
Chain index subscriber

idky137:
Move ChainIndex to subscriber model

## Motivation

Required to work in zainod.

## Solution

Reduce publicity of dangerous methods, create clone-safe, ready-only index, move trait implementations to reader, update tests.

### Tests

Uses current tests.

Updated ChainIndex to follow interface set out in zaino previously.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -124,17 +124,14 @@ pub trait ChainIndex {
 /// the entire chain at once. Backed by a source of blocks, either
 /// a zebra ReadStateService (direct read access to a running
 /// zebrad's database) or a jsonRPC connection to a validator.
-pub struct NodeBackedChainIndex<Source: BlockchainSource = ValidatorConnector> {
+pub struct NodeBackedChainIndexer<Source: BlockchainSource = ValidatorConnector> {
     #[allow(dead_code)]
-    mempool_state: std::sync::Arc<mempool::Mempool<Source>>,
-    mempool: mempool::MempoolSubscriber,
+    mempool: std::sync::Arc<mempool::Mempool<Source>>,
     non_finalized_state: std::sync::Arc<crate::NonFinalizedState<Source>>,
-    // pub crate required for unit tests, this can be removed once we implement finalised state sync.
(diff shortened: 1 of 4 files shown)


idky137:
Build atop #487 as requires mempool changes made there.

zaino-state/src/chain_index.rs
```diff
@@ -153,19 +150,25 @@ where {
...
         };
         chain_index.start_sync_loop();
         Ok(chain_index)
     }
-}
 
-impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
-    pub(super) fn start_sync_loop(
+    /// Creates a [`NodeBackedChainIndex`] from self,
+    /// a clone-safe, ready only view onto the running indexer.
```
AloeareV:
```suggestion
    /// a clone-safe, read-only view onto the running indexer.
```

AloeareV (changes requested):
Could we make the names a bit more distinct than ChainIndex and ChainIndexer? ChainIndexSubscriber is bit more of a mouthful, but glancing at Index vs Indexer it's not clear to me from the name which one is which

zaino-state/src/chain_index.rs
```diff
@@ -161,18 +169,38 @@ where {
     }
 
     /// Creates a [`NodeBackedChainIndex`] from self,
-    /// a clone-safe, ready only view onto the running indexer.
-    pub async fn to_index(&self) -> NodeBackedChainIndex<Source> {
-        NodeBackedChainIndex {
+    /// a clone-safe, read-only view onto the running indexer.
+    pub async fn to_index(&self) -> NodeBackedChainIndexSubscriber<Source> {
```
AloeareV:
```suggestion
    pub async fn get_subscriber(&self) -> NodeBackedChainIndexSubscriber<Source> {
```

AloeareV (changes requested):
to_index as a name no longer makes sense. get_subscriber maybe?

AloeareV (approved):

PR #497, opened 2025-08-29
use derived deser for automatic hex encode/decode, remove hex transfromation from tryfrom

### LINKED ISSUE #495: `GetTreestateResponse` returns invalid tree state encodings
https://github.com/zingolabs/zaino/commit/a23e1beb23a68a06e9d67127ffec1f6833cbf53b (#360) migrated across a Zebra crate change, and as part of that `GetTreestateResponse`'s internal uses of `zebra_rpc::client::Treestate` needed to be updated for its representation changing from `String` (storing the hex-encoded data as-is) to `Vec<u8>` (relying on serde to decode the hex to store the raw bytes). Zaino doesn't use the Zebra serde logic and implements `serde::Deserialize` itself, and its migration was incorrect, and after the change, the hex-encoded data from the JSON-RPC response is being stored in byte form, instead of hex-decoded.

At a minimum `GetTreestateResponse` needs to be updated to parse the hex-encoded data to store the raw bytes as required by the current Zebra API. The partial duplication of RPC client types should also be examined (in particular to enumerate the reasons why `zebra-rpc` types are insufficient for `zaino-fetch` to use), as there are likely other instances of this class of bug elsewhere.

AloeareV:
## Motivation
Fixes #495 

## Solution

We now rely on derived deserialization, which ensures that `zebra_rpc::client::Treestate`s are serialized and deserialized as intended. We also remove the hex-decoding when transforming between GetTreestateResponse types, as that is handled automatically at serialization time.

### DIFF
--- a/zaino-fetch/src/jsonrpsee/response.rs
+++ b/zaino-fetch/src/jsonrpsee/response.rs
@@ -957,7 +957,7 @@ impl<'de> serde::Deserialize<'de> for TxidsResponse {
 /// Encoded using v0 frontier encoding.
 ///
 /// This is used for the output parameter of [`crate::jsonrpsee::connector::JsonRpSeeConnector::get_treestate`].
-#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize)]
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct GetTreestateResponse {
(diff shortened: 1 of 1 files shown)


zancas (commented):
LGTM!

nachog00 (approved):
My only caveat is with the doc lines in the beggining, but i don't consider it blocking. we can udpate it later on after clarifying the reasons for the duplicated response typings.

PR #510, opened 2025-09-02
refactor Dockerfile

y4ssi:
## Motivation

- The previous Dockerfile setup caused build errors due to missing workspace members (e.g., `zaino-testvectors`).
- It used inefficient caching and overly complex mount setups, which slowed down builds.
- The runtime image included unnecessary dependencies, making it larger than required.

## Solution

- **Workspace & build flow**
  - Copy the **entire workspace** (`COPY . .`) to avoid missing members.
  - Install minimal build deps + **`protobuf-compiler`** (provides `protoc`) in the **builder** stage.
  - Build via `cargo install --path zainod --bin zainod --root /out` to get a **stable artifact path** (`/out/bin/zainod`) instead of relying on `target/release`.
  - Keep the optional **`NO_TLS`** toggle (`--features disable_tls_unencrypted_traffic_mode`).
- **Caching**
  - Use BuildKit caches for `cargo` **registry**, **git**, and **target**:
    - `--mount=type=cache,target=/usr/local/cargo/registry`
    - `--mount=type=cache,target=/usr/local/cargo/git`
    - `--mount=type=cache,target=/app/target`
  - Recommend a tight `.dockerignore` to keep context small and preserve cache hits.
- **Runtime**
  - Base on `debian:bookworm-slim`; install only `ca-certificates`, `libssl3`, and `libgcc-s1`.
  - Run as **non-root** (`container_user`), set `WORKDIR` to the user’s home.
  - Copy only the installed binary from `/out/bin/zainod`.
  - Generic, safe `HEALTHCHECK` using `zainod --version`.
  - Expose default gRPC/JSON-RPC ports (configurable via build args).
- **Simplicity & production**
  - No bind mounts, no toolchains in the final image, minimal surface area.
  - Clear place to add `libclang-dev` in builder if `bindgen` is required later.

### Tests

- **Builder sanity**
  - `docker build -t zainod:prod .` completes successfully (TLS enabled).
  - `docker build --build-arg NO_TLS=true -t zainod:notls .` completes successfully (TLS disabled).
  - Verified `protoc` available in builder: `protoc --version` during build logs.
- **Runtime sanity**
  - `docker run --rm zainod:prod zainod --version` exits 0.
  - Container runs as non-root: `id -u` inside container is not `0`.
  - `HEALTHCHECK` passes (`zainod --version` returns success).
- **Behavioral**
  - Basic invocation `docker run --rm zainod:prod` starts the service without missing-library errors.
- **Caching**
  - Rebuild after code-only changes confirms BuildKit cache hits on cargo registry/git and target.

- Docker BuildKit caching: https://docs.docker.com/build/cache/
- `prost-build` documentation on sourcing `protoc` (requires `protobuf-compiler`).
- [Official Rust Docker images](https://hub.docker.com/_/rust) and Debian slim best practices
- `cargo install --path` for producing stable install artifacts.

### Follow-up Work

- Supply a **distroless** or **musl-static** variant if the project and dependencies allow it.
- Add CI pipeline to build and push Docker images automatically.
- Add SBOM and image scanning (e.g., Syft/Grype) to the pipeline.
- Consider a health endpoint check if the service exposes one (replacing the `--version` healthcheck).

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,18 @@
+# Build artifacts
+target/
+**/target/
+**/*.rs.bk
+
+# VCS / CI
+.git/
+.gitignore
+.github/
+
+# IDE / OS
+.vscode/
+.idea/
+*.swp
+.DS_Store
+
+# Misc
+**/*.log
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,131 +1,81 @@
 # syntax=docker/dockerfile:1
 
-# Set the build arguments used across the stages.
-# Each stage must define the build arguments (ARGs) it uses.
-
-# Use UID/GID 1000 to match host user
-ARG UID=1000
-ARG GID=${UID}
-ARG USER="container_user"
-ARG HOME="/home/container_user"
-ARG CARGO_HOME="${HOME}/.cargo"
-ARG CARGO_TARGET_DIR="${HOME}/target"
-
+############################
+# Global build args
+############################
 ARG RUST_VERSION=1.86.0
-
-  # This stage prepares Zaino's build deps and captures build args as env vars.
-FROM rust:${RUST_VERSION}-bookworm AS deps
-SHELL ["/bin/bash", "-xo", "pipefail", "-c"]
-
-# Accept an argument to control no-tls builds
-ARG NO_TLS=false
-ENV NO_TLS=${NO_TLS:-false}
-
-# Install build deps (if any beyond Rust).
-RUN apt-get update && apt-get install -y --no-install-recommends \
-    musl-dev \
-    gcc \
-    clang \
-    llvm-dev \
-    libclang-dev \
-    cmake \
-    make \
-    # Install OpenSSL only if not building with no-tls
-    && if [ "$NO_TLS" = "false" ]; then apt-get install -y libssl-dev; fi \
-    && rm -rf /var/lib/apt/lists/*  /tmp/*
-
-# Build arguments and variables
-ARG CARGO_INCREMENTAL
-ENV CARGO_INCREMENTAL=${CARGO_INCREMENTAL:-0}
-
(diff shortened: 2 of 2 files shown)


zancas:
Hi!  Thanks for this.   Can you confirm that this image still works with the local workflows specified in the Makefile.toml..  or modify that code to work with this image (if it doesn't)?

y4ssi:
> Hi! Thanks for this. Can you confirm that this image still works with the local workflows specified in the Makefile.toml.. or modify that code to work with this image (if it doesn't)?

From what I’m seeing, it should still maintain compatibility. That said, the `Makefile.toml` looks outdated because, for example, under `[tasks.build-image]` I see:

```
docker build -f Dockerfile \
  --target "$TARGET" \
  --build-arg ZCASH_VERSION=$ZCASH_VERSION \
  --build-arg ZEBRA_VERSION=$ZEBRA_VERSION \
  --build-arg RUST_VERSION=$RUST_VERSION \
  --build-arg UID=$(id -u) \
  --build-arg GID=$(id -g) \
  -t ${IMAGE_NAME}:$TAG \
  ${@} \
```

But the build-args `ZCASH_VERSION` and `ZEBRA_VERSION` don’t appear in the current Dockerfile, and they aren’t used in the Dockerfile in my PR either.

zancas (approved):

zancas:
> 
> But the build-args `ZCASH_VERSION` and `ZEBRA_VERSION` don’t appear in the current Dockerfile, and they aren’t used in the Dockerfile in my PR either.

The Dockerfile referenced here is test_environment/Dockerfile, hence the confusion.

zancas:
This can merge once CI passes, and I verify the builds.

PR #467, opened 2025-08-04
Configuration System Refactoring

nachog00:
# Configuration System Refactoring: From Flat to Structured

This document outlines the comprehensive refactoring of Zaino's configuration system, transforming it from a flat structure with individual parameters to a nested, structured configuration architecture that improves maintainability and developer experience.

## Overview

The configuration refactoring addressed several challenges in code maintainability, API design, and configuration management. By introducing structured configuration types, eliminating parameter drilling, and creating a clean public API, we've improved the codebase's long-term sustainability while maintaining compatibility with the existing Figment-based configuration system.

### 1. Parameter Drilling and Function Complexity

The original system required numerous individual parameters to be passed through multiple function layers:

```rust
// Before: Multiple individual parameters
pub async fn create_test_manager(
    validator: &ValidatorKind,
    backend_type: &BackendType,
    network: Option<services::network::Network>,
    chain_cache: Option<std::path::PathBuf>,
    enable_zaino: bool,
    enable_json_server: bool,
    enable_cookie_auth: bool,
    zaino_no_sync: bool,
    zaino_no_db: bool,
    enable_clients: bool,
    // ... additional parameters
) -> Result<TestManager, Box<dyn std::error::Error>>
```

The new approach uses structured configuration objects:

```rust
// After: Single configuration struct
pub async fn create_test_manager(
    validator: &ValidatorKind,
    config: &IndexerConfig,
) -> Result<TestManager, Box<dyn std::error::Error>>
```

### 2. Configuration Structure and Readability

The original flat configuration structure made it difficult to understand relationships between related settings and required manual management of parameter dependencies.

**Before (Flat Structure)**:
```rust
pub struct IndexerConfig {
    pub backend: zaino_state::BackendType,
    pub enable_json_server: bool,
    pub json_rpc_listen_address: SocketAddr,
    pub enable_cookie_auth: bool,
    pub cookie_dir: Option<PathBuf>,
    pub grpc_listen_address: SocketAddr,
    pub grpc_tls: bool,
    pub tls_cert_path: Option<String>,
    pub tls_key_path: Option<String>,
    pub validator_listen_address: SocketAddr,
    pub validator_grpc_listen_address: SocketAddr,
    pub validator_cookie_auth: bool,
    pub validator_cookie_path: Option<String>,
    pub validator_user: Option<String>,
    pub validator_password: Option<String>,
    pub map_capacity: Option<usize>,
    pub map_shard_amount: Option<usize>,
    pub zaino_db_path: PathBuf,
[...]

Issue #515, opened 2025-09-03
Inappropriate Intimacy - Public API tightly coupled to Zebra's internal types


nachog00:
# Inappropriate Intimacy - Public API tightly coupled to Zebra's internal types

**Labels:** `configs`, `tech-debt`, `architecture`, `coupling`

## Problem

Zaino's public configuration API suffers from **Inappropriate Intimacy** with Zebra's internal implementation details. The `StateServiceConfig` directly exposes `zebra_state::Config` as a public field, creating tight coupling that violates abstraction boundaries.

### Current Problematic Code

```rust
// zaino-state/src/config.rs:28
pub struct StateServiceConfig {
    /// Zebra [`zebra_state::ReadStateService`] config data
    pub validator_config: zebra_state::Config,  // Direct Zebra type exposure!
    pub validator_rpc_address: std::net::SocketAddr,
    // ... other fields
}

// zainod/src/config.rs:414 - Public API constructs Zebra types
impl TryFrom<IndexerConfig> for BackendConfig {
    fn try_from(cfg: IndexerConfig) -> Result<Self, Self::Error> {
        match cfg.backend {
            zaino_state::BackendType::State => Ok(BackendConfig::State(StateServiceConfig {
                validator_config: zebra_state::Config {  // Users must understand Zebra internals!
                    cache_dir: cfg.zebra_db_path.clone(),
                    ephemeral: false,
                    delete_old_database: true,
                    debug_stop_at_height: None,
                    debug_validity_check_interval: None,
                },
                // ...
            })),
        }
    }
}
```

### Impact

- **Breaking Changes**: Zebra updates force breaking changes in Zaino's public API
- **Abstraction Violation**: Users must understand Zebra's internal config structure
- **Version Lock-in**: Zaino's API evolution tied to Zebra's internal changes
- **Cognitive Load**: Users need knowledge of both Zaino and Zebra configurations  
- **Serialization Coupling**: Config persistence depends on Zebra's serialization format
- **Testing Complexity**: Tests must import and understand Zebra's config types

### Locations

- `zaino-state/src/config.rs:28` - Direct `zebra_state::Config` field
- `zainod/src/config.rs:414` - Public API construction of Zebra types
- `zaino-state/src/config.rs:70` - Constructor parameter exposure
- Integration tests throughout codebase construct `zebra_state::Config`

## Solution

Define Zaino-specific configuration types that map internally to Zebra types:

```rust
// Zaino's own state configuration
[...]

Issue #508, opened 2025-09-02
Power-of-Two Constraint - Cache shard count causes runtime panics


nachog00:
# Power-of-Two Constraint - Cache shard count causes runtime panics

**Labels:** `configs`, `tech-debt`, `runtime-safety`

## Problem

Cache configuration has a **hidden power-of-two constraint** that causes runtime panics when violated, with no type-level enforcement or clear error messages.

### Current Problematic Code

```rust
pub struct BlockCacheConfig {
    // Documentation says "should be power of two" but type doesn't enforce
    pub map_shard_amount: Option<usize>,
    pub map_capacity: Option<usize>,
}

// User creates seemingly valid config
let config = BlockCacheConfig {
    map_capacity: Some(1000),
    map_shard_amount: Some(10),  // Not power of 2!
};

// Later, deep in the code...
let cache = DashMap::with_capacity_and_shard_amount(
    capacity, 
    shard_amount  // PANIC: "shard_amount must be a power of two"
);

// Thread panics with unhelpful message:
// thread 'main' panicked at 'assertion failed: shard_amount.is_power_of_two()'
```

### Impact

- **Runtime Panics**: Invalid configurations crash the application
- **Poor Error Messages**: Panic message doesn't indicate config issue
- **Late Discovery**: Problem only found when cache is created
- **User Confusion**: Why must shards be power of two?
- **Documentation Dependency**: Constraint only mentioned in comments

## Solution

Encode the constraint in the type system:

```rust
// Use shard_power instead of shard_amount
pub struct CacheConfig {
    pub capacity: usize,
    pub shard_power: u8,  // 2^n shards - can't be invalid!
}

impl CacheConfig {
    pub fn shard_count(&self) -> usize {
        1 << self.shard_power  // Always power of 2
    }
    
    pub fn build_cache<K, V>(&self) -> DashMap<K, V> 
    where 
        K: Eq + Hash,
        V: Clone,
    {
        DashMap::with_capacity_and_shard_amount(
            self.capacity,
            self.shard_count()  // Never panics
        )
    }
}

// User-friendly constructors
impl CacheConfig {
    pub fn new(capacity: usize, shard_power: u8) -> Self {
        Self { capacity, shard_power }
    }
    
    pub fn with_shards(capacity: usize, desired_shards: usize) -> Self {
        // Convert to nearest power of 2
        let shard_power = desired_shards.next_power_of_two().trailing_zeros() as u8;
        Self { capacity, shard_power }
[...]

Issue #507, opened 2025-09-02
Stringly-Typed Configuration - Using strings where enums would be safer


nachog00:
# Stringly-Typed Configuration - Using strings where enums would be safer

**Labels:** `configs`, `tech-debt`, `type-safety`

## Problem

Configuration uses **Stringly-Typed** patterns - representing domain concepts as strings instead of proper types, requiring string comparison and making typos runtime errors.

### Current Problematic Code

```rust
pub struct IndexerConfig {
    // Network as string - typos become runtime errors
    pub network: String,  
    
    // Backend type as string
    pub backend: String,  // "zcashd", "zebra", etc.
}

// String comparison everywhere
impl IndexerConfig {
    pub fn check_config(&self) -> Result<(), IndexerError> {
        // String literals scattered through code
        if (self.network != "Regtest") && 
           (self.network != "Testnet") && 
           (self.network != "Mainnet") {
            return Err(IndexerError::ConfigError(
                "Incorrect network name given...".to_string(),
            ));
        }
        
        // Case sensitivity issues
        if self.backend.to_lowercase() == "zcashd" {
            // ...
        } else if self.backend.to_lowercase() == "zebra" {
            // ...
        }
    }
}

// Typos compile but fail at runtime
let config = IndexerConfig {
    network: "MainNet",  // Wrong case - runtime error!
    backend: "zcash",    // Typo - runtime error!
};

// String manipulation for configuration
if config.network.starts_with("Test") {
    // Is this Testnet? Something else?
}
```

### Impact

- **Runtime Errors**: Typos and case mismatches caught late
- **Poor IDE Support**: No autocomplete for valid values
- **String Comparison Overhead**: Performance cost of string operations
- **Magic Strings**: Valid values scattered as literals through code
- **Refactoring Difficulty**: Renaming requires finding all string occurrences
- **Documentation Burden**: Must document valid string values

## Solution

Replace strings with proper enums and types:

```rust
// Network as enum - compile-time safety
#[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
#[serde(rename_all = "lowercase")]
pub enum Network {
    Mainnet,
    Testnet,
    Regtest,
}

// Backend types explicit
#[derive(Debug, Clone, serde::Deserialize, serde::Serialize)]
#[serde(rename_all = "lowercase")]
pub enum BackendType {
    Zcashd,
    Zebra,
    Custom(String),  // Extensibility if needed
}

// No string comparison needed
[...]

PR #524, opened 2025-09-04
zaino-common-foundation

nachog00:
# Configuration Foundation: Shared Types and Parameter Cleanup

## Motivation

This PR addresses configuration architecture technical debt and establishes a foundation for shared configuration types across Zaino crates. The previous approach had scattered configuration parameters (`map_capacity`, `map_shard_amount`) that were performance tuning parameters exposed in consumer-facing config, making them difficult to maintain and error-prone (shard amounts had to be power-of-2).

## Solution

- **New `zaino-common` crate**: Centralized shared configuration types (`Network`, `ServiceConfig`, `StorageConfig`, `CacheConfig`, `DatabaseConfig`)
- **Modernized cache parameters**: Replaced raw `map_capacity`/`map_shard_amount` with structured `CacheConfig` containing `capacity` and `shard_power` (power-of-2 automatically enforced)
- **Network abstraction**: Unified network handling with conversion utilities between Zaino and Zebra network types
- **Type safety improvements**: `shard_power` as `u8` ensures valid power-of-2 values via `1 << shard_power`
- **Database defaults**: Set sensible 128GB default database size, removed unlimited option

### Tests

- All existing tests pass with new configuration structure
- Maintains backward compatibility for essential functionality 
- Database operations continue to work with new `StorageConfig`
- Network conversion utilities tested across Mainnet/Testnet/Regtest
- Config parsing tests updated with new structured fields (pending)

- DashMap performance tuning via capacity and shard configuration
- Zebra network types integration for seamless chain parameter access

### Follow-up Work

- Continue replacement of config fields into structs towards structure outlined in #467

### PR Checklist

- [X] The PR name is suitable for the release notes.
- [ ] The solution is tested.
- [X] The documentation is up to date.


nachog00:
## Configuration Architecture Improvements - Relationship to Config Issues

This PR begins addressing several configuration antipatterns identified in our Config-labeled issues. While it doesn't fully resolve these systemic issues, it establishes a foundation and demonstrates the path forward.

### Issues Directly Addressed

** #508 - Power-of-Two Constraint - Cache shard count causes runtime panics**
- **Resolved**: Replaced raw `map_shard_amount: Option<usize>` with `shard_power: u8` in `CacheConfig`
- **Solution**: `1 << shard_power` automatically ensures power-of-2 values, eliminating runtime panics
- **Impact**: Type-level safety prevents invalid shard configurations

**=� #500 - Primitive Obsession - Configuration uses primitive types requiring runtime validation**
- **Partially Addressed**: Introduced structured types (`StorageConfig`, `CacheConfig`, `ServiceConfig`)
- **Progress**: Replaced primitive cache parameters with proper structs containing semantic meaning
- **Remaining**: `IndexerConfig` still uses primitive types (string network, raw booleans, etc.)

**=� #499 - Configuration Shotgun Surgery - Duplicated fields across config structs**
- **Partially Addressed**: Created shared `zaino-common` crate with reusable config types
- **Progress**: `StorageConfig` and `ServiceConfig` now shared between `StateServiceConfig` and `FetchServiceConfig`
- **Remaining**: `IndexerConfig` still duplicates patterns, hasn't been updated to use shared types

### Issues Partially Addressed

**=� #515 - Inappropriate Intimacy - Public API tightly coupled to Zebra's internal types**
- **Progress**: Created `zaino_common::Network` with conversion utilities to Zebra types
- **Progress**: New configs use Zaino's `Network` type instead of Zebra's directly
- **Remaining**: Some lower-level library configs still expose `zebra_chain::parameters::Network` publicly
- **Remaining**: `IndexerConfig` conversion still directly creates Zebra network types

**=� #507 - Stringly-Typed Configuration - Using strings where enums would be safer**
- **Progress**: Created proper `zaino_common::Network` enum to replace network strings
- **Critical Gap**: `IndexerConfig` still uses `pub network: String` - not updated to use new enum!
- **Impact**: Main consumer-facing config still suffers from string typing issues

**=� #502 - God Object - IndexerConfig mixes concerns across multiple domains**
- **Progress**: Extracted storage and service concerns into structured subtypes for internal configs
- **Critical Gap**: `IndexerConfig` itself unchanged - still mixes TLS, auth, network, cache, database concerns
- **Remaining**: Consumer-facing config structure needs the same treatment

**=� #505 - Implicit Invariants - Cache configuration relies on undocumented constraints** 
- **Progress**: Made power-of-2 constraint explicit through `shard_power` type design in internal configs
- **Remaining**: `IndexerConfig` still has `map_capacity`/`map_shard_amount` with implicit constraints

**=� #501 - Telescoping Constructor - Config constructors with 15+ parameters**
- **Progress**: New structured config types reduce parameter count in internal construction
- **Remaining**: `IndexerConfig` constructors and conversion logic still telescoping

### Issues Not Yet Addressed (Future Work)

**L #506 - Long Method - Config validation contains complex branching logic**
- **Status**: `IndexerConfig::check_config()` validation logic remains complex and monolithic
- **Future**: Domain-specific validation methods per concern area

**L #504 - Feature Envy - Config structs manipulate external state instead of owning behavior**
- **Status**: Configs still primarily data containers
- **Future**: Move behavior into config types (validation, conversion, etc.)

**L #503 - Boolean Blindness - Authentication uses bool flags allowing invalid states**
- **Status**: Still using `validator_cookie_auth: bool`, `grpc_tls: bool` etc. in `IndexerConfig`
- **Future**: Enum-based authentication and TLS configuration

### Critical Missing Piece: IndexerConfig Updates

**The main consumer-facing `IndexerConfig` hasn't been updated to use any of the new structured types!** This is a significant oversight that should be addressed:

1. **Network**: Should use `zaino_common::Network` instead of `String`
2. **Cache/Storage**: Should use structured `StorageConfig` instead of raw primitives
3. **Services**: Should use `ServiceConfig` for timeout/channel settings
4. **Structure**: Should be decomposed into domain-specific sub-configs

### Foundation for Future Improvements

This PR establishes `zaino-common` as the foundation, but **we need to actually update `IndexerConfig` to use these new types** to realize the benefits. The internal config improvements show the path forward, but the main consumer API needs the same treatment.

**Immediate Next Steps**:
1. **Update `IndexerConfig`** to use `zaino_common::Network` instead of `String` 
2. **Integrate `StorageConfig`** into `IndexerConfig` to replace primitive cache fields
3. **Continue domain separation** per #467 for TLS, auth, service configs

zaino-common/src/network.rs
```diff
@@ -0,0 +1,112 @@
...
+        self.into()
+    }
+
+    /// Convert to Zebra's network type for internal use (alias for to_zebra_default).
+    pub fn to_zebra_network(&self) -> zebra_chain::parameters::Network {
+        self.to_zebra_default()
+    }
+
+    /// Get the standard regtest activation heights used by Zaino.
+    pub fn zaino_regtest_heights() -> zebra_chain::parameters::testnet::ConfiguredActivationHeights
```
idky137, line 164: Would it be possible to add a separate method for setting custom regtest heights? This is sometimes useful for testing so if possible to add here could  be good.

zaino-common/src/storage.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl Default for DatabaseSize {
+    fn default() -> Self {
+        DatabaseSize::Gb(128)  // Default to 128 GB
+    }
+}
+
+impl DatabaseSize {
+    /// Convert to bytes
+    pub fn as_bytes(&self) -> usize {
```
idky137: Could we update this name to  `as_byte_count` or something similar? 

`as_bytes` methods are generally used to convert to a sequence of raw u8 bytes so I think this could be confusing.

zaino-common/src/network.rs
```diff
@@ -0,0 +1,112 @@
...
+            nu6_1: None,
+            nu7: None,
+        }
+    }
+
+    /// Determines if sync should be skipped for testing.
+    ///
+    /// - Mainnet/Testnet: Skip sync (false) because we don't want to sync real chains in tests
+    /// - Regtest: Enable sync (true) because regtest is local and fast to sync
+    pub fn should_sync_for_testing(&self) -> bool {
```
idky137: Is this to replace the `no_sync` setting used for regtest connection?

If so then I think there is a slight misunderstanding of what the original bool did (probably because the docs were not clear).

What that bool actually did was told the `BlockCache` not to wait for the server to sync.

When running on mainnet or testnet the old `BlockCache` code used the `Blocks` (the amount of blocks the local validator has synced) and the `estimated height` (the estimated height of the zcash network, calculated by the local validator) fields from `get_blockchain_info` to determine when the validator was synced, and therefore when to start service.

With regtest this did not work as the estimated height was always incorrect, and due to the fact there is no network to sync with we used this bool to completely remove this check.

Currently the new `ChainIndex` code does not use this at all, it currently uses a different set of notifiers that should not be effected in the same way.

However we are still in the process of integrating this code so may still need to use something similar so could be good to get in. 

Given what this is doing (if i have read it correct) I think a more suitable name may be "wait_on_server_sync" or something similar.

zaino-common/src/network.rs
AloeareV: > Would it be possible to add a separate method for setting custom regtest heights? (not ) This is sometimes useful for testing so if possible to add here could be good.

Not what? I think you started to type a parenthetical here and didn't finish

nachog00: we thought of adding the heights as fields on the Testnet variant, so they can be passed from the config. It will surely not be part of this PR to keep it simpler. But it wouldn't be too hard to add later on. Good think is all that logic we can keep tight and together on the Network enum and aleviate the cosnumer components from dealing with it.

idky137: I think having them on Regtest is more useful than Testnet. Currently the only custom Testnet we use is Regtest, with Testnet internally meaning the public "Testnet".

This could be useful in the future but having a configurable Regtest would be more useful in the short term.

idky137: Ah the (not) was meant to say: Not necessarily exposed in the actual config file, but accessible internally for test usage. eg TestManager..

nachog00: > I think having them on Regtest is more useful than Testnet. Currently the only custom Testnet we use is Regtest, with Testnet internally meaning the public "Testnet".
> 
> This could be useful in the future but having a configurable Regtest would be more useful in the short term.

my bad, I meant regtest variant

zaino-common/src/storage.rs
```diff
@@ -0,0 +1,99 @@
...
+            size: DatabaseSize::default(),
+        }
+    }
+}
+
+/// Storage configuration combining cache and database settings.
+///
+/// This is used by services that need both in-memory caching and persistent storage.
+#[derive(Debug, Clone, serde::Deserialize, serde::Serialize, Default)]
+pub struct StorageConfig {
```
idky137, line 93: Non blocking, just thought I'd point out that `CacheConfig` here is only related to the in-memory cache rather than on disc storage so `StorageConfig` may not be completely accurate.

Im happy for this to land and for us to look at some of these names at a later date if best but, afaict it will mean an external, client facing config change, so we should confirm this name before release.

zaino-state/src/chain_index/tests/finalised_state/v0.rs
```diff
@@ -22,26 +24,16 @@ pub(crate) async fn spawn_v0_zaino_db(
...
-                nu7: None,
+        storage: StorageConfig {
+            database: DatabaseConfig {
+                path: db_path,
+                ..Default::default()
             },
-        ),
+            ..Default::default()
+        },
+        db_version: 1,
```
idky137:
```suggestion
        db_version: 0,
```

zaino-state/src/chain_index/tests/finalised_state/v0.rs
```diff
@@ -212,26 +204,16 @@ async fn load_db_from_file() {
...
```
idky137:
```suggestion
        db_version: 0,
```

idky137 (changes requested):
Looks good! 

Just the 2 small changes in the v0 ZainoDB tests. One was already present before this PR (slipped through original impl and review) but would be good to fix both of these and is only a one line change.

Also it looks like there are a couple failing config tests now, it may be that the actual tests need updating.

zaino-common/src/storage.rs
nachog00: agreed. We also had some chats about possibly moving some of these behind a 'dev' feature

idky137 (approved):

PR #525, opened 2025-09-04
prune build dependencies

### LINKED ISSUE #480: Prune build dependencies to limit impact of ZingoLib internal dependency
reported by @nuttycom
> Hey, I just realized that my current `Zallet` testing is stymied because NU6.1 activated on testnet but the Zaino version we're depending on doesn't yet support it. Zebra's recent releases have been updated, so I decided to give a go at updating `Zaino` but I found myself somewhat stymied here, because it looks like to really do things properly I also need to update `zingolib`. We've got a bit of a knot of dependencies here, and I'm wondering what the best approach is to untying it; for example, it looks like `zingolib` is only needed by `zaino-testutils` - is there a way that we can prune down the workspace so that we don't run into this kind of issue?

ala-mode:
Fixes #480 

DRAFT / WIP
[I have created a "dummy" tag in `infra`](https://github.com/zingolabs/infrastructure/releases/tag/pin_number_one), with the idea that `zingo-infra-testutils` (and maybe `zingo-infra-services`) will be moved to a [dev-dependency]. This gesture also means I should be able to remove `zingolib` (and, I think, `testvectors`) from `zaino`'s `Cargo.toml` altogether, in favor of them being a dependency in `infra`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2178,9 +2178,9 @@ dependencies = [
  "zaino-state",
  "zaino-testutils",
  "zcash_primitives 0.24.0",
- "zebra-chain 2.0.0",
- "zebra-rpc 2.0.0",
- "zebra-state 2.0.0",
+ "zebra-chain",
+ "zebra-rpc",
+ "zebra-state",
  "zingolib",
 ]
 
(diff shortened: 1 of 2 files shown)


zancas:
Obsolete.

PR #529, opened 2025-09-09
update get_transaction_status to new interface

### LINKED ISSUE #518: update ChainIndex `get_transaction_status` output to `(Option<(BlockHash, Height)>, HashSet<BlockHash>)`
This has been requested by the zallet team.

Currently get_transaction_status returns a hashmap<block hash, block>, it is not clear which, if any blocks are in the best chain, this method response should be updated to clearly show this.

The suggested response is `(Option<(BlockHash, Height)>, HashSet<BlockHash>)`, where the `Option` shows any "best chain" block that contains the transaction (if exists), and the `Hashset` shows any orphan blocks that contain the transaction in question. An improvement to this would be to also indicate whether the transaction is currently in the mempool.

AloeareV:
## Motivation

Fix #518

## Solution

Update get_transaction_status to return the new type

### Tests

Existing tests adapted to the new interface

### PR Checklist

- [*] The PR name is suitable for the release notes.
- [*] The solution is tested.
- [*] The documentation is up to date.

### DIFF
--- a/zaino-state/src/chain_index/types.rs
+++ b/zaino-state/src/chain_index/types.rs
@@ -41,6 +41,29 @@ impl BlockHash {
     }
 }
 
+/// The location of a transaction in the best chain
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum BestChainLocation {
+    /// the block containing the transaction
+    Block(BlockHash, Height),
+    /// If the transaction is in the mempool,
+    /// which matches the snapshot's chaintip
+    Mempool,
+}
+
+/// The location of a transaction not in the best chain
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum NonBestChainLocation {
+    /// the block containing the transaction
+    Block(BlockHash),
+    /// if the transaction is in the mempool
+    /// but the mempool does not match the
+    /// snapshot's chaintip
+    /// This likely means that the provided
+    /// snapshot is out-of-date
+    Mempool,
+}
+
 impl fmt::Display for BlockHash {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         f.write_str(&self.encode_hex::<String>())
--- a/zaino-state/src/chain_index/types.rs
+++ b/zaino-state/src/chain_index/types.rs
@@ -46,9 +46,11 @@ impl BlockHash {
 pub enum BestChainLocation {
     /// the block containing the transaction
     Block(BlockHash, Height),
-    /// If the transaction is in the mempool,
-    /// which matches the snapshot's chaintip
-    Mempool,
+    /// If the transaction is in the mempool and the mempool
+    /// matches the snapshot's chaintip
+    /// Return the target height, which is known to be a block above
+    /// the provided snapshot's chaintip and is returned for convenience
+    Mempool(Height),
 }
 
 /// The location of a transaction not in the best chain
@@ -56,12 +58,15 @@ pub enum BestChainLocation {
 pub enum NonBestChainLocation {
     /// the block containing the transaction
     Block(BlockHash),
+    // TODO: in this case, returning a consensus branch
+    // ID would be useful
     /// if the transaction is in the mempool
     /// but the mempool does not match the
-    /// snapshot's chaintip
+    /// snapshot's chaintip, return the target height if known
+    ///
     /// This likely means that the provided
     /// snapshot is out-of-date
-    Mempool,
+    Mempool(Option<Height>),
 }
 
 impl TryFrom<&ChainBlock> for NonBestChainLocation {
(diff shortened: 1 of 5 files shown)


idky137 (dismissed):
Overall looks good! 

On thing, I think we want to return the mempool chaintip hash in the NonBestChainLocation mempool variant.

Currently this is the only option that does not have a clear path for zallet to find all the correct data it needs. 

If we returned the mempool chain tip block hash in this variant then there is a clear route for zallet to follow to sync to the new best chain tip:
- end = get_block_height( non best chain mempool chain tip block hash )
- start = get_fork_point( internal chain tip block hash )
- get_block_range( start, end )

I may have misunderstood this situation in which case this can be dismissed.

I have also detailed this in #518.

zaino-state/src/chain_index/types.rs
```diff
@@ -41,6 +41,50 @@ impl BlockHash {
...
 }
 
+/// The location of a transaction in the best chain
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum BestChainLocation {
+    /// the block containing the transaction
+    Block(BlockHash, Height),
+    /// If the transaction is in the mempool,
+    /// which matches the snapshot's chaintip
+    Mempool,
```
str4d: This should include the target height (the height at which the txs in the mempool would be mined by the next block), i.e. chain_tip_height + 1 (for the chain that the mempool is on top of).

zaino-state/src/chain_index/types.rs
```diff
@@ -41,6 +41,50 @@ impl BlockHash {
...
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum NonBestChainLocation {
+    /// the block containing the transaction
+    Block(BlockHash),
+    /// if the transaction is in the mempool
+    /// but the mempool does not match the
+    /// snapshot's chaintip
+    /// This likely means that the provided
+    /// snapshot is out-of-date
+    Mempool,
```
str4d: Ditto (as even here this is descending from the backend node's view of the best chain, which in this case is not descended from the snapshot).

zaino-state/src/chain_index/types.rs
```diff
@@ -41,6 +41,50 @@ impl BlockHash {
...
+    /// If the transaction is in the mempool,
+    /// which matches the snapshot's chaintip
+    Mempool,
+}
+
+/// The location of a transaction not in the best chain
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum NonBestChainLocation {
+    /// the block containing the transaction
+    Block(BlockHash),
```
str4d, line 62: Ideally this would include the consensus branch ID that the block was validated against, if known. (It cannot include a height, as heights are by definition best-chain heights, and the best chain is defined by the snapshot here).

zaino-state/src/chain_index/types.rs
```diff
@@ -41,6 +41,50 @@ impl BlockHash {
     }
 }
 
+/// The location of a transaction in the best chain
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum BestChainLocation {
+    /// the block containing the transaction
+    Block(BlockHash, Height),
```
str4d, line 48: By comparision, here height is fine, because it is an ambient requirement that the caller and Zaino agree on the consensus rules, so the caller can always determine the consensus branch ID from the height.

str4d (dismissed):
Comments from the zcashd deprecation meeting.

zaino-state/src/chain_index/types.rs
nuttycom: @str4d how is that data going to be used in `zallet`? This seems like a nice-to-have that could be added at a later point.

zaino-state/src/chain_index/types.rs
nuttycom: Same question here; do we actually need this now?

zaino-state/src/chain_index/types.rs
AloeareV: This can be determined by the caller as it will always be the tip of the snapshot's best chain + 1. That said, this means it's really easy to return it here as a convenience to the caller.

AloeareV:
> Overall looks good!
> 
> On thing, I think we want to return the mempool chaintip hash in the NonBestChainLocation mempool variant.
> 
> Currently this is the only option that does not have a clear path for zallet to find all the correct data it needs.
> 
> If we returned the mempool chain tip block hash in this variant then there is a clear route for zallet to follow to sync to the new best chain tip:
> 
>     * end = get_block_height( non best chain mempool chain tip block hash )
> 
>     * start = get_fork_point( internal chain tip block hash )
> 
>     * get_block_range( start, end )
> 
> 
> I may have misunderstood this situation in which case this can be dismissed.
> 
> I have also detailed this in #518.

I've added an Optional height to the return which I think does the job. The height returned will be None only in the case where the mempool's hash is not on the validator's best chain. This implies that there was a reorg between when the call was made and when we responded, and in this case I think the correct reaction is to rerun the call.

zaino-state/src/chain_index/types.rs
str4d: I was describing the ideal API here; I don't think Zallet will be using this aspect (at least not initially), so it could be deferred for now.

Oscar-Pepper (approved):

zancas (approved):

PR #530, opened 2025-09-09
feature flag disable_tls_unencrypted_traffic_mode -> no_tls_use_unenc…

### LINKED ISSUE #511: `disable_tls_unencrypted_traffic_mode` is a confusing name
It is meant as "disable TLS and use unencrypted traffic" but can easily be misread as "disable (TLS unencrypted traffic mode)" which technically could be a thing but isn't.

(TLS up to and including version 1.2 technically had a [TLS_NULL_WITH_NULL_NULL](https://datatracker.ietf.org/doc/html/rfc5246#section-6.2.3.1) ciphersuite, although it was prohibited to negotiate it after the handshake. The [SSL 3.0 spec](https://www.ietf.org/rfc/rfc6101.txt) had quite a few null-encryption ciphersuites and did not prohibit negotiating them, if I'm reading correctly. The SSH spec also specifies a null cipher although it is [usually compiled out](https://serverfault.com/questions/116875/how-can-i-disable-encryption-on-openssh/606367).)

Please rename it to something like `disable_tls_and_use_unencrypted_traffic` to avoid the ambiguity.

ala-mode:
fixes #511 

Simple name change. Slightly shorter, seems to me less ambiguous.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,7 +34,7 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     --mount=type=cache,target=/usr/local/cargo/git \
     --mount=type=cache,target=/app/target \
     if [ "${NO_TLS}" = "true" ]; then \
-      cargo install --locked --path zainod --bin zainod --root /out --features disable_tls_unencrypted_traffic_mode; \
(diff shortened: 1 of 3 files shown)


idky137 (commented):
Looks good to me.

idky137 (approved):

daira (commented):
Post-hoc ACK

PR #534, opened 2025-09-10
Update get transaction

### LINKED ISSUE #520: Fix output type of ChainIndex get_transaction
It has been requested that we update the response type of get_transaction.

The exact format to be returned has not been decided, this decision should be made with the Zallet devs.

Currently `get_transaction` returns the inner bytes of a `zebra_chain::transaction::SerializedTransaction`, we could update this to return an actual `zebra_chain::transaction::Transaction` but this would move away from the current standard of returning raw bytes.

AloeareV:
## Motivation

Fixes #520 

## Solution

Adds an optional consensus branch id u32 as a return type to get_raw_transaction. 

### Tests

Updated get_raw_transaction tests to assert we return the latest consensus branch ID, except the genesis block (which has no ID).

### PR Checklist

- [*] The PR name is suitable for the release notes.
- [*] The solution is tested.
- [*] The documentation is up to date.

### DIFF
--- a/integration-tests/tests/chain_cache.rs
+++ b/integration-tests/tests/chain_cache.rs
@@ -267,12 +267,13 @@ mod chain_query_interface {
         test_manager.generate_blocks_with_delay(5).await;
         let snapshot = indexer.snapshot_nonfinalized_state();
         assert_eq!(snapshot.as_ref().blocks.len(), 7);
-        for txid in snapshot
-            .blocks
-            .values()
-            .flat_map(|block| block.transactions().iter().map(|txdata| txdata.txid().0))
-        {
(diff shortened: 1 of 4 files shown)


idky137:
Looks good! As soon as we can confirm the response type of `get_transaction_status and land #529 I will approve this.

zancas (approved):
I resolved conflicts and assume that the hunk that used "snapshot" was preferable and that the additional import was the correct choice.

zancas:
@AloeareV I reset this to the pre-merge state, and opened a revert PR that sets dev back to the pre-merge state:

https://github.com/zingolabs/zaino/pull/558

Issue #516, opened 2025-09-04
Direct Field Access Pattern - Newtype structs expose internal fields via .0 access


nachog00:
# Direct Field Access Pattern - Newtype structs expose internal fields via .0 access

**Labels:** `tech-debt`, `type-safety`, `encapsulation`

## Problem

The codebase extensively uses direct `.0` field access on newtype structs (tuple structs), which undermines type safety and encapsulation. This pattern is found in 41+ files and affects core types like `Height`, `BlockHash`, and various wrapper types.

### Current Problematic Code

Throughout the codebase, we see patterns like:
```rust
// From zaino-state/src/local_cache/finalised_state.rs
height.0.saturating_sub(100)
blockchain_info.blocks.0 as i64
reorg_height.0 + 1

// From various files
hash.0  // [u8; 32] access
block.hash().0  // Direct byte access
shard_index.0  // u32 access

// Semantic confusion example from finalised_state.rs:520
if (blockchain_info.blocks.0 as i64 - blockchain_info.estimated_height.0 as i64).abs() <= 10
```

### Issues with Current Approach

**1. Loss of Type Safety**
```rust
// These are semantically different but all u32 internally
pub struct Height(pub u32);
pub struct ShardIndex(pub u32);
pub struct BlockNumber(pub u32);

// With .0 access, this compiles but may not make semantic sense:
let height = Height(100);
let shard = ShardIndex(5);
let result = height.0 + shard.0;  // Adding height to shard index
```

**2. Refactoring Challenges**
If we need to change the internal representation:
```rust
// Current
pub struct BlockHash(pub [u8; 32]);
let bytes = hash.0;  // Works

// If we later need to add functionality
pub struct BlockHash {
    bytes: [u8; 32],
    cached_string: Option<String>,  // Performance optimization
}
// Now every hash.0 in the codebase breaks
```

**3. Bypassing Validation**
Direct field access can bypass any invariants the type should maintain:
```rust
let mut height = Height(100);
height.0 = 0;        // What if 0 is invalid?
height.0 = u32::MAX; // Could cause overflow in calculations
```

**4. Semantic Clarity Issues**
The suspicious line in `zaino-state/src/local_cache/finalised_state.rs:520` demonstrates how `.0` access can obscure semantic meaning:
```rust
if (blockchain_info.blocks.0 as i64 - blockchain_info.estimated_height.0 as i64).abs() <= 10
```

**What could go wrong:**
[...]

PR #535, opened 2025-09-11
Field Access Patterns for ChainTipSubscriber, MempoolKey, MempoolValue, BestTip, StatusError, AtomicStatus

ala-mode:
Improves #516

### DIFF
--- a/zaino-state/src/status.rs
+++ b/zaino-state/src/status.rs
@@ -20,22 +20,27 @@ use std::{
 /// - [>=7: Critical-Error].
 ///   TODO: Refine error code spec.
 #[derive(Debug, Clone)]
-pub struct AtomicStatus(Arc<AtomicUsize>);
+pub struct AtomicStatus {
+    counter: Arc<AtomicUsize>,
+}
 
 impl AtomicStatus {
     /// Creates a new AtomicStatus
     pub fn new(status: u16) -> Self {
-        Self(Arc::new(AtomicUsize::new(status as usize)))
+        Self {
+            counter: Arc::new(AtomicUsize::new(status as usize)),
+        }
     }
 
     /// Loads the value held in the AtomicStatus
     pub fn load(&self) -> usize {
-        self.0.load(Ordering::SeqCst)
+        self.counter.load(Ordering::SeqCst)
     }
 
     /// Sets the value held in the AtomicStatus
+    // TODO do we need to have status be usize here? why not strong types?
     pub fn store(&self, status: usize) {
-        self.0.store(status, Ordering::SeqCst);
+        self.counter.store(status, Ordering::SeqCst);
     }
 }
 
--- a/zaino-state/src/status.rs
+++ b/zaino-state/src/status.rs
@@ -38,7 +38,6 @@ impl AtomicStatus {
     }
 
     /// Sets the value held in the AtomicStatus
-    // TODO do we need to have status be usize here? why not strong types?
     pub fn store(&self, status: usize) {
         self.counter.store(status, Ordering::SeqCst);
     }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,3 @@
-# dummy change to trigger ci. will revert inmediately
 # Workspace Members are in dependency order, do not change this without prior consideration.
 [workspace]
 members = [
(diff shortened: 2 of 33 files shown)


nachog00:
> fixes #516 WIP

Most of these config issues are quite broad. Almost describing a type of issue that is widespread across the codebase. maybe it's best to open sub-issues of it that point to specific instances? unless you do plan in fixing all of them on this pr! i suggest keeping it atomic (pun intended)

zaino-state/src/status.rs
```diff
@@ -20,22 +20,27 @@ use std::{
 /// - [>=7: Critical-Error].
 ///   TODO: Refine error code spec.
 #[derive(Debug, Clone)]
-pub struct AtomicStatus(Arc<AtomicUsize>);
+pub struct AtomicStatus {
```
nachog00, line 23: having a named field is already an improvement in clarity (compared with obscured ".0"). But is think it would be even better to work with getters ( .counter() ). Cause that way you aren't tied to a specific internal representation.

zaino-state/src/status.rs
```diff
@@ -20,22 +20,27 @@ use std::{
...
-        Self(Arc::new(AtomicUsize::new(status as usize)))
+        Self {
+            counter: Arc::new(AtomicUsize::new(status as usize)),
+        }
     }
 
     /// Loads the value held in the AtomicStatus
     pub fn load(&self) -> usize {
-        self.0.load(Ordering::SeqCst)
+        self.counter.load(Ordering::SeqCst)
```
nachog00, line 37: see, if this was `counter()` then say we did want to move back to tuple internals or simply change it somehow, then while  `.counter` could stop being a valid accessor and one would have to update the usage everywhere (not a massive hit on this simpler case ofc, but the alternative is simple too), a `.counter()` call could be left untouched everywhere and simply update it's definition to get the right value.

zaino-state/src/status.rs
```diff
@@ -20,22 +20,27 @@ use std::{
...
     }
 
     /// Sets the value held in the AtomicStatus
+    // TODO do we need to have status be usize here? why not strong types?
```
nachog00: I wonder this too. I would pursue this change.

zaino-state/src/status.rs
ala-mode: I agree on this point.

To answer the above comment first, I _was_ planning on doing everything reasonable with one PR, and only break up the PRs if I needed to. I'm trying to make each commit its own pattern-alteration, or at least not have more than one pattern change in any given commit.

If we want to add more or change logic to address these kind of observations, I propose putting _those_ into more issues, and breaking them off one at a time.

I'd like to try to get these patterns changed in the next few days, so it shouldn't go too stale.

zaino-state/src/status.rs
ala-mode: Yes, I like this.

Like I said, maybe we could break these kinds of improvements up into additional PRs, and have this PR be in effect 'non-changing'?

ala-mode:
I'm opening this for review because I think I attacked the problem significantly enough to make an impact.

Due to @idky137 's remarks, I tried to avoid changing anything where we are using zebra-types directly (particularly `Height` and `BlockHash`), and anything in `local_cache`. I'm not 100% confident I succeeded, but that was the aim and I hope a review will confirm these changes are appropriate and will help. I don't think I changed any logic such that Zaino will run any differently.

The work could continue: above, @nachog00 suggests a few things. One is to more strongly type the `status` - which is a change I support, but I think would need a little logic overhaul. A related idea brought up is to use getter(/setter?) patterns and create a `.counter()` function.

There are probably many more things that could be addressed as well. Looking at the code, these changes are diverging (in a good way, I think) from patterns found in `zebra` - so one thought would be to see if ZF is interested in following our lead on this.

Open to any thoughts.

nachog00:
I agree that we shouldn't simply copy zebra's approach if it is a sub-optimal pattern. Unless there's explicit reasons to stay coherent with it.

zancas:
@nachog00 what more needs to be done to merge this into dev?

If there are improvements on the feature, we want them to benefit other developers sooner than later, right?

nachog00:
Code looks good, but i think it hasn't run the full CI tests yet? I think that happens when one un-drafts a PR and doesn't push to it ever as non-draft, since the push event is the one that triggers CI

nachog00 (approved):

PR #542, opened 2025-09-17
height is not currently a correct indicator of whether the BlockIndex's block is on the best chain

AloeareV:
## Motivation

Height of the IndexedBlock is not an indicator of presence on the best chain. There may be more to this...but this is an unused function that can be cleanly removed.

### Follow-up Work

It's possible height is being relied on somewhere where it shouldn't be.


idky137:
Ah this is interesting, does this mean that non-best chain blocks in the non-finalised state currently have a height?

Oscar-Pepper (approved):

Issue #538, opened 2025-09-12
Custom Deserialization Duplicates Consensus-Critical Logic


nachog00:
## Problem Statement

Zaino currently implements a custom block and transaction deserialization system that mixes parsing concerns with indexing concerns. This creates several serious issues:

### 1. Mixed Responsibilities in `ParseFromSlice`

The current `ParseFromSlice` trait conflates pure deserialization with external data correlation:

```rust
fn parse_from_slice(
    data: &[u8],
    txid: Option<Vec<Vec<u8>>>,  // ❌ External RPC data during parsing
    tx_version: Option<u32>,     // ❌ Context mixed with parsing
) -> Result<(&[u8], Self), ParseError>
```

This violates separation of concerns by:
- Requiring RPC-derived transaction IDs during raw byte parsing
- Mixing consensus-critical parsing with validation logic
- Making the parser dependent on external data sources

### 2. Duplication of Consensus-Critical Logic

Zaino reimplements complex consensus-critical parsing that already exists in Zebra:
- **Block header parsing** - custom `BlockHeaderData` vs Zebra's `Header`
- **Transaction parsing** - custom `FullTransaction` vs Zebra's `Transaction`  
- **Serialization logic** - custom `CompactSize` vs Zebra's implementation
- **Validation rules** - partial reimplementation vs Zebra's complete consensus validation

**Risk**: Parsing divergence from consensus rules, potential security vulnerabilities

### 3. Maintenance Burden

Every Zcash protocol upgrade requires:
- ✅ Updates to Zebra (maintained by Zcash core developers)
- ❌ **Additional** updates to Zaino's custom parser (maintenance burden on Zaino team)
- ❌ Risk of falling behind or introducing bugs during protocol changes

### 4. Impedance Mismatch

The current approach conflates:
- **Parsing** (mechanical conversion of bytes to structures)
- **Enrichment** (adding derived/external metadata for indexing)
- **Validation** (correlating with RPC data)

This creates artificial coupling between low-level parsing and high-level indexing logic.

### Stack Responsibility Separation

**Zebra's Role (Parsing Layer)**:
- ✅ Convert raw bytes to validated block/transaction structures
- ✅ Enforce consensus rules during deserialization
- ✅ Provide cryptographically verified data

**Zaino's Role (Indexing Layer)**:
- ✅ Derive frequently-accessed data (height, txids)
- ✅ Enrich with external metadata (tree sizes from RPC)
[...]

PR #543, opened 2025-09-17
refactor ChainIndex types module into smaller modules

nachog00:
## Motivation

The types.rs module on chain_index is too big (3333 lines). Splitting it into smaller, focused modules following the `module.rs + module/` pattern preferred at ZingoLabs.

Also added a new type for TreeSize that validates the size and safely handles edge cases, preventing dangerous `as u32` casts that could cause silent truncation when tree sizes reach their theoretical maximum.

### Module Organization
- **`commitment.rs`**: Extracted commitment tree types (`CommitmentTreeData`, `CommitmentTreeRoots`, `CommitmentTreeSizes`) - ~160 lines
- **`primitives/`**: Created new primitives module structure:
  - `primitives.rs`: Index file for primitive types
  - `primitives/tree_size.rs`: Validated tree size types with bounds checking

### New Validated Tree Size Types
Added type-safe, validated representations of commitment tree sizes:
- **`TreeSize` trait**: Common behavior with validation, conversion methods, and bounds checking
- **`SproutTreeSize`**: Max 2^29 = 536,870,912 notes
- **`SaplingTreeSize`**: Max 2^32 = 4,294,967,296 notes  
- **`OrchardTreeSize`**: Max 2^32 = 4,294,967,296 notes
- **`TreeSizeError`**: Comprehensive error handling for edge cases

The TreeSize types address a critical issue where tree sizes are stored as u64 in Zebra but often cast to u32 in Zaino, creating potential for silent truncation when trees reach their theoretical maximum capacity (exactly u32::MAX + 1 for Sapling/Orchard).

### Tests

Compilation verified across all packages. TreeSize types include comprehensive validation and conversion methods with proper error handling.

Part of #538 - Replace custom block/transaction parsing with Zebra's consensus-validated deserialization

### Follow-up Work

The primitives module is prepared for migration of additional primitive types from the main types.rs file:
- Block and transaction hashes
- Block heights and chain work  
- Basic indexing primitives

### PR Checklist

- [x] The solution is tested.


zaino-state/src/chain_index/types.rs
```diff
@@ -1,6 +1,35 @@
...
+//! - [`types::primitives`] - Foundational types (hashes, heights, tree sizes, etc.)
+//! - [`types::commitment`] - Commitment tree data structures and utilities
+//!
+//! ## Planned Module Organization
+//! The remaining types in this file will be migrated to:
+//! - `block.rs` - Block-related structures (BlockIndex, BlockData, IndexedBlock)
+//! - `transaction.rs` - Transaction types (CompactTxData, TransparentCompactTx, etc.)
+//! - `address.rs` - Address and UTXO types (AddrScript, Outpoint, etc.)
+//! - `shielded.rs` - Shielded pool types (SaplingCompactTx, OrchardCompactTx, etc.)
+
```
idky137, line 16: Can we add a clear doc comment that outlines rules for this module:
- structs in this module _must_ _never_ use external types as fields directly, instead fundamental data should be saved into the struct, and from / into (or appropriate getters / setters) should be implemented.
- structs in this module _must_ implement `ZainoVersionedSerialize` _and_ abide by the stringent version rules outlined in that trait.
- structs in this module _must_ _never_ be changed without implementing a new version _and_ implementing the necessary ZainoDB updates and migrations.

zaino-state/src/chain_index/types/primitives/tree_size.rs
```diff
@@ -0,0 +1,151 @@
+//! Validated commitment tree size types.
```
idky137: We currently have a CommitmentTreeSizes struct that holds the data that we require, changing this would require updates to ZainoDB (see comment above) i think the most straight forward way to gain the validation implemented here without wider cvhanges to the codebase would be to use the struct we already have and implement the new validation there.

idky137 (dismissed):
Separating out into sub-modules would be good but we need to be very careful with changing any of the structs in this file / module (see first comment).

zaino-state/src/chain_index/types/primitives/tree_size.rs
AloeareV: Currently these types aren't actually used anywhere. Any issues with how these types fit into the database won't come to light until we actually try to incorporate them, at which point they may prove to be incompatible with how our database is written/read. For that reason I'm hesitant to approve them. I'm a big fan of splitting `types.rs` into different modules for readability, and I don't want that to get blocked on discussion of new types.

zaino-state/src/chain_index/types.rs
AloeareV: Agreed, but as these comments don't currently exist I don't think the need to add them should block this PR. I've opened #574 as follow-up work.

AloeareV (approved):
Moves only, this makes this clearer to read!

Issue #544, opened 2025-09-18
Follow up work on direct accessors pattern


nachog00:
see, if this was `counter()` then say we did want to move back to tuple internals or simply change it somehow, then while  `.counter` could stop being a valid accessor and one would have to update the usage everywhere (not a massive hit on this simpler case ofc, but the alternative is simple too), a `.counter()` call could be left untouched everywhere and simply update it's definition to get the right value.

_Originally posted by @nachog00 in https://github.com/zingolabs/zaino/pull/535#discussion_r2341241855_

ala-mode:
See https://github.com/zingolabs/zaino/pull/546

ala-mode:
Closed as #546 was the active issue.

There is still #516 open because the fields pattern still exists in the codebase.

PR #545, opened 2025-09-18
getter fn for AtomicStatus

ala-mode:
following #544 

This sets the getter pattern as a impl on AtomicStatus

IIUC, another thing we were interested in looking at here was the elimination of using integers for StatusTypes, if possible. I could add that work on top of this PR or open another.


ala-mode:
closing in favor of (WIP) #546 - which advances this idea.

PR #551, opened 2025-09-19
Workspace dependencies do not include zingolib.

fluidvanadium:
Move zingolib out of workspace dependency so that it cannot accidentally became a production dependency.

## Motivation

Cleaning the dependency tree as requested by Str4d, pacu, and nuttyc0m. Including Issue 480

## Partial Solution

Zingolib is not a dependency in Zaino whenever it can be avoided. Todayz PR demonstrates that this concern is taken care of, and enforces that it cannot revert.

### Tests

See follow-up work

https://hackmd.io/wcqYQAHJRxiaYx60ExIDOQ

### Follow-up Work

Setting up nightly CI cargo hack

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -7,6 +7,7 @@ homepage = { workspace = true }
 edition = { workspace = true }
 license = { workspace = true }
 version = { workspace = true }
+publish = false
 
 [dependencies]
 # Test utility
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -18,7 +18,10 @@ zaino-state = { workspace = true, features = ["bench"] }
 zebra-chain.workspace = true
 zebra-state.workspace = true
 zebra-rpc.workspace = true
-zingolib = { workspace = true }
+zingolib = { git = "https://github.com/zingolabs/zingolib.git", tag = "2.0.0-beta2", features = [
+    "test-elevation",
+] }
+
 core2 = { workspace = true }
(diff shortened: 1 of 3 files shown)


zancas:
The way to think about this is that it's mostly beneficial for external consumers, and mostly costly for internal developers.

zancas (commented):
#480 is merged, a follow on issue #561 tracks future work, #561 discusses the tradeoffs implied by this approach.

integration-tests/Cargo.toml
```diff
@@ -18,7 +19,10 @@ zaino-state = { workspace = true, features = ["bench"] }
 zebra-chain.workspace = true
 zebra-state.workspace = true
 zebra-rpc.workspace = true
-zingolib = { workspace = true }
+zingolib = { git = "https://github.com/zingolabs/zingolib.git", tag = "2.0.0-beta2", features = [
```
zancas, line 24: What happens if the version in integrations-tests and zaino-testutils diverge?

You could prevent this by re-exporting from zaino-testutils, and consuming the re-export in integration-tests...

zancas (dismissed):

integration-tests/Cargo.toml
zancas: I am opening up a folow-up  PR that includes augmentations that I think make this work better.

zancas (approved):
Nice work @fluidvanadium!  I am opening a follow-up PR with some augmentations.

PR #553, opened 2025-09-19
run autoinherit and machete

zancas:
## Motivation

Having multiple versions of the same crate in a single workspace is a significant source if bugs.   Moreover by unifying dependency policy to a single source of truth we reduce the possibility of version skew.

In this case I discovered that there were 2 versions of zcash_protocol specified one in the workspace, and another in zaino-testutils!!

## Solution

Use the **cargo autoinherit** tool.

This unifies polciy into the workspace level crate.

### Tests

If the current set on dev pass, that's sufficient for this dependency related PR.

### Follow-up Work

Once dependencies are unified it will be easier to know that updates in a single place have the same effect across the workspace.

### PR Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -973,16 +973,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ctrlc"
-version = "3.4.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


zancas:
Superceded.

PR #557, opened 2025-09-19
chore(`integration-tests`): mark as private

dorianvp:
## Motivation

Explicitly mark crates that are not supposed to be released as private.

## Solution

A `publish = false` should do. This also hooks up with future CI checks when we do releases.


zancas (commented):
I like this!  I want Ala to review so she stays in the loop.

zancas (approved):

PR #560, opened 2025-09-20
Depend on zingolib 2 0 7 testvectors

zancas:
As can be seen here, the older version of zingolib brought older versions of zcash crates in as dependencies of zaino-testutils:

skews

The new version of zingolib 2.0.7 depends on some variant of zcash_protocol 0.6.1


zancas:
All lints and the CLT pass locally.

Oscar-Pepper (changes requested):
I'm not sure whats going on here but according to current dev the git diff is wrong. Dev does not have a `zingolib` dep anymore so this should show as removed by this PR. Anyways, the change i am requesting is that *all* crates in zingolib repo should be updated to this latest tag. it may also make the most sense to update our infrastructure dependency to the latest tag although infra doesnt depend on zingolib anymore so this can be done in a future PR now

Oscar-Pepper:
stale

PR #562, opened 2025-09-21
Enhance codependencies

zancas:
## Motivation

The remove codepencies feature had costs and benefits.

One cost is that a viewer of the Workspace Manifest would not realize that zingolib is used in test members, another is that it was possible to skew the versions of zingolib between the tightly couples non-published integration-tests and zaino-testutils packages.

## Solution

Couple zaino-testutil and integration-tests to a single version of zingolib with a re-export.
Add a comment to the workspace Manifest warning readers that zingolib is around.

### Tests

Yes.

This work builds on https://github.com/zingolabs/zaino/pull/551 which references relevant context.

### Follow-up Work

We're updating zingolib, and will likely make it easier to depend on it less, here.

### PR Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,3 +138,6 @@ zingo-netutils = { git = "https://github.com/zingolabs/zingolib.git", tag = "2.0
 [profile.test]
 opt-level = 3
 debug = true
+
+#[workspace.dev-dependencies]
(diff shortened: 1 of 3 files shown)


zancas:
I assumed that integration-tests was referring to zingolib..  it's not, so there's no need to re-export it.

PR #563, opened 2025-09-21
Rm comment

zancas:
The refurbish StatusType PR commented out this code.   Is this code that's live?   Why is it commented out?   If it's not live, why not remove it?

If it's necessary for correct function, then why didn't a test fail when it was commented out?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -138,3 +138,6 @@ zingo-netutils = { git = "https://github.com/zingolabs/zingolib.git", tag = "2.0
 [profile.test]
 opt-level = 3
 debug = true
+
+#[workspace.dev-dependencies]
+# zingolib is imported in zaino-testutils!!
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -20,9 +20,6 @@ zaino-state = { workspace = true, features = ["bench"] }
 zebra-chain.workspace = true
 zebra-state.workspace = true
 zebra-rpc.workspace = true
-zingolib = { git = "https://github.com/zingolabs/zingolib.git", tag = "2.0.0-beta2", features = [
-    "test-elevation",
-] }
 
 core2 = { workspace = true }
 prost = { workspace = true }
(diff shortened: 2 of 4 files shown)


zaino-state/src/status.rs
```diff
@@ -89,28 +89,6 @@ impl From<StatusType> for usize {
     }
 }
 
-/*
```
ala-mode, line 92: Yup! It was the goal to eliminate these, so removal is proper.

Cargo.toml
```diff
@@ -138,3 +138,6 @@ zingo-netutils = { git = "https://github.com/zingolabs/zingolib.git", tag = "2.0
 [profile.test]
 opt-level = 3
 debug = true
+
+#[workspace.dev-dependencies]
```
ala-mode, line 142: Though a triumph worthy of two exclamations,
we could probably get rid of these comments.

zaino-testutils/Cargo.toml
```diff
@@ -7,6 +7,7 @@ homepage = { workspace = true }
 edition = { workspace = true }
 license = { workspace = true }
 version = { workspace = true }
+publish = false
```
ala-mode, line 10: IIUC, This means the zaino-testutils package will be prevented from being published to any package registry.

ala-mode, line 10: IIUC, This means the zaino-testutils package will be prevented from being published to any package registry.

integration-tests/Cargo.toml
```diff
@@ -20,9 +20,6 @@ zaino-state = { workspace = true, features = ["bench"] }
 zebra-chain.workspace = true
 zebra-state.workspace = true
 zebra-rpc.workspace = true
-zingolib = { git = "https://github.com/zingolabs/zingolib.git", tag = "2.0.0-beta2", features = [
```
ala-mode, line 23: I'll run some tests and see that everything works..

ala-mode (commented):
Running tests...

ala-mode:
All tests pass locally, approving!

ala-mode (approved):
LGTM

PR #564, opened 2025-09-21
cargo nextest run -p zaino-state

zancas:
This doesn't work on dev, and will work after the PR lands.

### DIFF
--- a/zaino-state/Cargo.toml
+++ b/zaino-state/Cargo.toml
@@ -66,6 +66,7 @@ zingo-infra-services = { workspace = true }
 tempfile = { workspace = true }
 tracing-subscriber = { workspace = true }
 once_cell = { workspace = true }
+zaino-state = { path = ".", features = ["bench"] }
 
 [build-dependencies]
 whoami = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6023,6 +6023,7 @@ dependencies = [
  "zaino-common",
  "zaino-fetch",
  "zaino-proto",
+ "zaino-state",
  "zcash_address 0.9.0",
  "zcash_keys 0.10.1",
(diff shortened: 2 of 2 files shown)


zaino-state/Cargo.toml
```diff
@@ -66,6 +66,7 @@ zingo-infra-services = { workspace = true }
 tempfile = { workspace = true }
 tracing-subscriber = { workspace = true }
 once_cell = { workspace = true }
+zaino-state = { path = ".", features = ["bench"] }
```
ala-mode, line 69: I needed to run `cargo nextest run --all-features -p zaino-state` instead of `cargo nextest run -p zaino-state`. Should we document this?

ala-mode (approved):

zaino-state/Cargo.toml
zancas: Really?  What error did you get when you tried to run?  Can you please share that?

fluidvanadium: wfm

ala-mode: > Really? What error did you get when you tried to run? Can you please share that?

```
error[E0432]: unresolved import `crate::bench`
  --> zaino-state/src/chain_index/tests/finalised_state/migrations.rs:8:12
   |
8  | use crate::bench::BlockCacheConfig;
   |            ^^^^^ could not find `bench` in the crate root
   |
note: found an item that was configured out
  --> zaino-state/src/lib.rs:59:9
   |
59 | pub mod bench {
   |         ^^^^^
note: the item is gated behind the `bench` feature
  --> zaino-state/src/lib.rs:57:7
   |
57 | #[cfg(feature = "bench")]
   |       ^^^^^^^^^^^^^^^^^
```
...and more like this.

PR #565, opened 2025-09-22
Rewrite testutils without zaino-common-zingo-infra-services frankenterface

### LINKED ISSUE #541: `zaino-common` uncontrollably depends on `zingo-infra-services`
The only thing the dependency is used for in `zaino-common` is to implement a conversion between `Network` types:
https://github.com/zingolabs/zaino/blob/528ef7d1e043c9abe1897bd00c6ffa4f184c9f8e/zaino-common/src/network.rs#L192-L212

However, this causes the entirety of `zingo-infra-services` to be pulled into downstreams like Zallet, including all of its diverging `zcash_*` crate versions as well as (currently) a dependency on OpenSSL that we will not accept.

fluidvanadium:
## Motivation

Fixes #541

## Solution

Move all true shared interface to small shared crates in infrastructure

(https://hackmd.io/@fluidvana/SJgpu_gqoge)

also updates the zingolib dependency of zaino-testutils

### PR Checklist

- [x] The solution is tested.


fluidvanadium:
Integration tests are building and running!

zancas:
Why isn't this review ready?

zancas:
The sooner upgrades like this get into dev, the sooner everyone will be building on them!

fluidvanadium:
> Why isn't this review ready?

Good question!

1. https://github.com/zingolabs/infrastructure/pull/124 should probably land first

2. 49 integration tests have turned to fail by this PR

zaino-testutils/src/lib.rs
```diff
@@ -668,14 +487,13 @@ impl TestManager {
         let clients = if enable_clients {
             let lightclient_dir = tempfile::tempdir().unwrap();
             let (lightclient_faucet, lightclient_recipient) = build_lightclients(
-                lightclient_dir.path().to_path_buf(),
+                lightclient_dir,
```
Oscar-Pepper: ah so this pattern of creating the tempdir and holding it is necessary to keep t in scope, otherwise it is deleted OR you have to persist it and then you lose the benefits on tempdir altogether. can you revert the "build_lightclients" fn to take a pathbuf and the Clients struct to have a lightclient_dir field so we can keep it scope while the tests run

zaino-state/Cargo.toml
```diff
@@ -62,7 +62,7 @@ derive_more = { workspace = true, features = ["from"] }
 
 
 [dev-dependencies]
-zingo-infra-services = { workspace = true }
+zingo-infra-services = { git = "https://github.com/zingolabs/infrastructure.git", rev = "6c8ad003491371c2e387c5f9d3ba35c45286257b" }
```
Oscar-Pepper: there seems to me many different versions of the infra repo, even for the saem crate. can we just unify this update to depend on one version of infra?

zaino-testutils/src/lib.rs
```diff
@@ -111,10 +113,15 @@ pub enum LocalNet {
 
 impl zingo_infra_services::validator::Validator for LocalNet {
     const CONFIG_FILENAME: &str = "";
+    const PROCESS: zingo_infra_services::Process = zingo_infra_services::Process::Empty; // todo
```
Oscar-Pepper: i dont think this needs a "todo" this is correct

zaino-testutils/src/lib.rs
```diff
@@ -111,10 +113,15 @@ pub enum LocalNet {
...
 
     type Config = ValidatorConfig;
 
-    fn activation_heights(&self) -> zingo_infra_services::network::ActivationHeights {
+    fn default_test_config() -> Self::Config {
+        todo!()
```
Oscar-Pepper, line 121: work for a following PR?

zaino-testutils/src/lib.rs
```diff
@@ -262,10 +269,9 @@ impl zingo_infra_services::validator::Validator for LocalNet {
     }
 }
 
-/// Holds zingo lightclients along with their TempDir for wallet-2-validator tests.
+/// Holds zingo lightclients for wallet-2-validator tests.
+// used to hold a tempdir, but it was used elsewhere instead
 pub struct Clients {
-    /// Lightclient TempDir location.
-    pub lightclient_dir: TempDir,
```
Oscar-Pepper, line 268: as state below, please revert this to hold the TempDir and reveret the comment to match

zaino-testutils/src/lib.rs
```diff
@@ -318,16 +324,18 @@ fn make_uri(indexer_port: portpicker::Port) -> http::Uri {
 // NOTE: this should be migrated to zingolib when LocalNet replaces regtest manager in zingoilb::testutils
 /// Builds faucet (miner) and recipient lightclients for local network integration testing
 async fn build_lightclients(
-    lightclient_dir: PathBuf,
+    lightclient_dir: TempDir,
```
Oscar-Pepper: as stated below, please revert this to take a pathbuf

zaino-testutils/src/lib.rs
```diff
@@ -368,227 +376,38 @@ impl TestManager {
...
             ValidatorKind::Zcashd => {
-                let cfg = zingo_infra_services::validator::ZcashdConfig {
-                    zcashd_bin: ZCASHD_BIN.clone(),
-                    zcash_cli_bin: ZCASH_CLI_BIN.clone(),
-                    rpc_listen_port: Some(zebrad_rpc_listen_port),
-                    activation_heights: activation_heights.unwrap_or_default(),
-                    miner_address: Some(REG_O_ADDR_FROM_ABANDONART),
-                    chain_cache: chain_cache.clone(),
-                };
+                let mut cfg = ZcashdConfig::default_test();
```
Oscar-Pepper, line 403: the "launch_with_activation_heights" seems to be deleted and now there is only a "ZcashdConfig::default_test()" option with default activation heights... please can we add this fn back in so we can configure the validator with custom activation heights

Oscar-Pepper: maybe a better option would be to just have an option activation heights and set to default activation heights if none, removing the need for two fns.. but im happy to just put it back to how it was also

zaino-testutils/src/lib.rs
```diff
@@ -368,227 +376,38 @@ impl TestManager {
...
-                    activation_heights: activation_heights.unwrap_or_default(),
-                    miner_address: zingo_infra_services::validator::ZEBRAD_DEFAULT_MINER,
-                    chain_cache: chain_cache.clone(),
-                    network,
-                };
+                let mut cfg = ZebradConfig::default_test();
+                cfg.rpc_listen_port = Some(rpc_listen_port);
+                cfg.indexer_listen_port = Some(grpc_listen_port);
+                cfg.chain_cache = chain_cache.clone();
+                cfg.network = network_kind;
```
Oscar-Pepper, line 413: same here, activation heights needs to be configurable

Oscar-Pepper (changes requested):

zaino-testutils/src/lib.rs
fluidvanadium: Okay, this is a reason that the zingolib dependency is a problem. Because this change is in another crate!

zaino-state/Cargo.toml
fluidvanadium: updated

zaino-testutils/src/lib.rs
fluidvanadium: You were the one who changed it in zingolib!

fluidvanadium: Done!

zaino-testutils/src/lib.rs
fluidvanadium: I dont understand why. But its done.

zaino-common/Cargo.toml
```diff
@@ -13,7 +13,9 @@ version = { workspace = true }
 zebra-chain = { workspace = true }
 
 # Zingo Infrastructure
```
Oscar-Pepper: can you remove this line if there is nothing left in this category please

Oscar-Pepper (changes requested):

zancas (dismissed):

Oscar-Pepper:
is this still relevant after #589 landed?

fluidvanadium:
> is this still relevant after #589 landed?

No, superceded.

PR #566, opened 2025-09-22
remove spurious wrapper

zancas:
AFAICT, this wrapper is just kruft.


idky137 (changes requested):
Looks good!

However if we are removing this wrapper then we should change the name of `to_zebra_default()` to `to_zebra_network()` or something similar.

I think default is very confusing here as we are not defaulting at all, we re just calling from / into

PR #573, opened 2025-09-23
[#214] Add Zcash RPC `getnetworksolps` passthrough RPC method

### LINKED ISSUE #214: 1.18: Add Zcash RPC `getnetworksolps`
Spec: https://zcash.github.io/rpc/getnetworksolps.html

pacu:
closes #214

In order to learn more about Zaino I decided to implement something relatively straight forward
with the help of AI.

## Motivation
Contribute to Zaino completion

## Solution
Create a passthrough solution

### Tests

I tested it manually. 

Assumes that Zebra RPC actually follows Zcashd's RPC spec

### Follow-up Work
Not sure there's more to it 

### PR Checklist

- [ ] The solution is tested.


zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -693,6 +693,31 @@ impl JsonRpSeeConnector {
         let params = vec![serde_json::json!({ "addresses": addresses })];
         self.send_request("getaddressutxos", params).await
     }
+
+    /// Returns the estimated network solutions per second based on the last n blocks.
+    ///
+    /// zcashd reference: [`getnetworksolps`](https://zcash.github.io/rpc/getnetworksolps.html)
```
ala-mode, line 728: For this particular RPC, I expect the implementation to be simple and probably match the online documentation.

However, in addition to the online RPC docs, we've been (recently, at least) documenting the zcash codebase itself with doc comments, as the online docs are not entirely accurate, and we've been warned to not implement features based solely off them.

Again, in this case I would not think there would be much to adjust, but we want to double-check and also minimally leave a trail of documentation to the canonical source: `zcash/zcash`. If anything I think looking closely at `blocks` and `height` parameters might reveal something. And, if something is skewed with a zebra response, we'll already be on top of the distinction.

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -33,7 +33,7 @@ use crate::jsonrpsee::{
     },
 };
 
-use super::response::GetDifficultyResponse;
+use super::response::{GetDifficultyResponse, GetNetworkSolPsResponse};
```
ala-mode, line 37: This seems to me to be a good response to be 'next to'.

But to me it seems like `GetDifficultyResponse` might be listed here in a somewhat unusual way, as the Response types listed just above link to the same file.

Maybe we should change `GetDifficultyResponse` to fit in with those? The only difference I see is that `GetDifficultyResponse` is defined as a relative path... it could make things more clear during refactoring and later reading.

zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -222,6 +222,14 @@ impl ResponseToError for GetBlockchainInfoResponse {
 #[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
 pub struct GetDifficultyResponse(pub f64);
 
+/// Response to a `getnetworksolps` RPC request.
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+pub struct GetNetworkSolPsResponse(pub f64);
```
ala-mode: I have reason to believe an `i32` would be sufficient here.

Also, this is an awkward name to begin with, but I might try
`GetNetworkSolpsResponse`. (no capital `P`).

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -693,6 +693,31 @@ impl JsonRpSeeConnector {
...
+    /// tags: blockchain
+    ///
+    /// # Parameters
+    ///
+    /// - `blocks`: (number, optional, default=120) Number of blocks, or -1 for blocks over difficulty averaging window.
+    /// - `height`: (number, optional, default=-1) To estimate network speed at the time of a specific block height.
+    pub async fn get_network_sol_ps(
+        &self,
+        blocks: Option<i32>,
+        height: Option<i32>,
```
ala-mode, line 742: height is .. not contingent on blocks being present...?

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -693,6 +693,31 @@ impl JsonRpSeeConnector {
...
+    ) -> Result<GetNetworkSolPsResponse, RpcRequestError<Infallible>> {
+        let mut params = Vec::new();
+        if let Some(b) = blocks {
+            params.push(serde_json::json!(b));
+            if let Some(h) = height {
```
ala-mode: The way that `height` appears here, nested inside the `if` for blocks... means that `height` will only be added to `params` if `blocks` is `Some()`..

(still without looking at the `zcash/zcash` code) I think this may be a mistake. I'm going to make an alteration, and open a PR to your branch, @pacu  ...

ala-mode (commented):
Hey @pacu ! Thanks for making this PR!

I had a few ideas, comments included. On one point I am just going to make a PR - if you agree with the change, you could merge it into your branch.

https://github.com/pacu/zaino/pull/1

ala-mode:
If I'm reading the graph right, the fork of this feature came a ways back from `dev` so it would have to be updated as well.

zaino-fetch/src/jsonrpsee/connector.rs
pacu: I don't understand the comment, could you elaborate more precisely what the suggestion is?

zaino-fetch/src/jsonrpsee/connector.rs
pacu: IIUC the change request here is that I add a link to the Zcashd code? 

In a second thought the link to Zcashd doesn't make much sense. probably I should link the Zebra RPC that's being served..

Do yo agree?

zaino-fetch/src/jsonrpsee/connector.rs
pacu: Don't. This is a draft PR and therefore unfinished

pacu:
@ala-mode @nachog00 I think there's something else I'm missing. Because I'm getting "method not found" for zaino. The zebra response is ok. Probably it's something really simple that you folks already know as obvious.

Zebra:
```
pacu@zerver:~$ curl  -H 'content-type: text/plain'   --data-binary '{"jsonrpc":"1.0","id":"zebra","method":"getnetworksolps","params":[120,200000]}'   http://127.0.0.1:18234
pacu@zerver:~$ ","id":"zebra","result":389,"error":null}
```

Zaino:
```
curl  -H 'content-type: text/plain'   --data-binary '{"jsonrpc":"1.0","id":"zaino","method":"getnetworksolps","params":[120,200000]}'   http://127.0.0.1:8237
{"jsonrpc":"1.0","id":"zaino","result":null,"error":{"code":-32601,"message":"Method not found"}}
```

pacu:
Manual Testing Results:

image

zaino-fetch/src/jsonrpsee/connector.rs
AloeareV: The request here is a link to the Zcashd code! Linking to the zebra code would also be useful, but if there's a discrepancy between zcashd and zebrad, then the right move is probably going to be to raise an issue in zebra that they're not doing what zcashd does. The goal of both zaino and zebra is generally to match zcashd (at least, in terms of this API)

zaino-fetch/src/jsonrpsee/response.rs
pacu: resolved to u64 because that's the inner type upstream

zaino-fetch/src/jsonrpsee/connector.rs
ala-mode: For an example of this in this same file, added with https://github.com/zingolabs/zaino/pull/371/files ,
see https://github.com/zingolabs/zaino/blob/c6e317c44798e5f452305088584da91970fff011/zaino-fetch/src/jsonrpsee/connector.rs#L520-L535

zaino-fetch/src/jsonrpsee/connector.rs
ala-mode: My comment was based off a misunderstanding. The way the arguments are used is...
from the (non-authoritative) readthedocs:
```
Returns the estimated network solutions per second based on the last n blocks.
Pass in [blocks] to override # of blocks, -1 specifies over difficulty averaging window.
Pass in [height] to estimate the network speed at the time when a certain block was found.
```

According to this document, in this case, without specifying `height` it defaults to -1, which is the top block, IIUC.
The default for `blocks` is 120, but I don't understand the comment about `difficulty averaging window` exactly, I think that might be being pulled from another place (IIRC there's a lot of logic around `difficulty`).

I think the readthedocs are saying without either argument, it's going to give you the solutions per second, as an average based on the last 120 blocks from the current tip. But if `blocks` is set to `-1`, I'm not sure how it will calculate the return.

So I was off base with my question, a `height` is indeed not contingent on `blocks` being an argument (or, the other way around: `blocks` is not contingent on `height`). But I think you'll have to look at the `zcash` code, to see particularly what's going on with `blocks` (and if the readthedocs are correct).

ala-mode:
> @ala-mode @nachog00 I think there's something else I'm missing...

@pacu one thing I see here right away is you're using `curl` to test the full node's endpoints. When I brought this up in previous Z3 discussions, Kris specifically told me to use `zcash-cli` as a reference, as opposed to `curl`.

I haven't been looking at things on this kind of level for maybe 5 or 6 weeks now, so I'm a little rusty.. and it may be the case that under the hood `zcash-cli` is using `curl`.. but for whatever reason that was some handed down wisdom I got from ECC. I was trying `curl` because that's what was in the online documentation (readthedocs), but I was asked to not do that and use their reference `cli` tool instead.

zaino-fetch/src/jsonrpsee/connector.rs
ala-mode: Did this comment make sense?

ala-mode (commented):
Looks pretty good so far! Please see the above comments.

Eventually we'll want some tests, too.

Thanks for working on this!

zaino-fetch/src/jsonrpsee/connector.rs
dorianvp: What do you mean? Are you saying we split each response type into their own separate file?

zaino-serve/src/rpc/jsonrpc/service.rs
```diff
@@ -312,8 +312,25 @@ pub trait ZcashIndexerRpc {
...
+    ///
+    /// zcashd reference: [`getnetworksolps`](https://zcash.github.io/rpc/getnetworksolps.html)
+    /// method: post
+    /// tags: blockchain
+    ///
+    /// # Parameters
+    ///
+    /// - `blocks`: (number, optional, default=120) Number of blocks, or -1 for blocks over difficulty averaging window.
+    /// - `height`: (number, optional, default=-1) To estimate network speed at the time of a specific block height.
+
```
dorianvp:
```suggestion
```

zaino-fetch/src/jsonrpsee/connector.rs
dorianvp: I think what @ala-mode is saying is the following, which is a bug:

If `blocks=None` and `height=Some(h)`, what gets passed through is `[]`. It may make sense to either move the `params.push(serde_json::json!(h));` outside of the outer if statement, or and/or set `params.blocks=120`.

integration-tests/tests/fetch_service.rs
```diff
@@ -1335,6 +1335,34 @@ async fn fetch_service_get_lightd_info(validator: &ValidatorKind) {
     test_manager.close().await;
 }
 
+async fn assert_fetch_service_getnetworksols_matches_rpc(validator: &ValidatorKind) {
```
dorianvp, line 1415: Nice! You should also add a similar test in `integration-tests/tests/state_service.rs`.

integration-tests/tests/fetch_service.rs
```diff
@@ -1527,6 +1555,11 @@ mod zcashd {
         pub(crate) async fn lightd_info() {
             fetch_service_get_lightd_info(&ValidatorKind::Zcashd).await;
         }
+
+        #[tokio::test]
+        pub(crate) async fn getnetworksops() {
```
dorianvp: Can you also add a similar function but under the `zebrad` mod? Only difference is the usage of `&ValidatorKind::Zebrad`.

dorianvp (changes requested):
Looking good! I requested some small changes.

zaino-fetch/src/jsonrpsee/connector.rs
AloeareV: I think this comment is about how the import statements exist, with some things being imported via `crate::...` and some things being imported via `super::...`

zaino-fetch/src/jsonrpsee/connector.rs
pacu: Well, actually that comment is inaccurate. The Zcash code is not canonical. In fact, at the RPC level, we are not replicating the functionality because there's a lot of tech debt in there. 

I'll add the reference with the URL but I will omit the paragraph above

zaino-fetch/src/jsonrpsee/connector.rs
pacu: oh yeah. good catch.

zaino-fetch/src/jsonrpsee/connector.rs
pacu: this in particular follows what's currently done. What shall I do?

dorianvp (dismissed):
Nice, thank you! All tests passing for me locally. Of course, we'll need to expand our test framework to be able to test mining-related endpoints more thoroughly.

zaino-fetch/src/jsonrpsee/connector.rs
ala-mode: On dev, only one dependency is currently added this way:
`use super::response::GetDifficultyResponse;`

Everything else is done via 
`use crate::jsonrpsee::response::{...}`

I believe these are all the same thing. I'm about to approve this PR if it passes tests for me, and I'll make a micro PR to follow on.

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -693,6 +693,39 @@ impl JsonRpSeeConnector {
...
+        height: Option<i32>,
+    ) -> Result<GetNetworkSolPsResponse, RpcRequestError<Infallible>> {
+        let mut params = Vec::new();
+
+        // check whether the blocks parameter is present
+        if let Some(b) = blocks {
+            params.push(serde_json::json!(b));
+        }
+
+        // check whether the height parameter is present
```
ala-mode, line 753: This line gave me a bit of trouble:
https://github.com/zcash/zcash/blob/d00fc6f4365048339c83f463874e4d6c240b63af/src/rpc/mining.cpp#L123C1-L123C122
But this is how zcashd handles the parameters.
If I'm reading it correctly (and I might not be!)... in order to specify `height` you DO need also to specify `blocks`. ie, if you specify only one argument, it MUST be blocks.

I know we were fussing around with this, but I think this is the right way to emulate what zcash is doing.

I think if there are more than two parameters passed, they are dropped, so it's fault tolerant in that way.

ala-mode (commented):
Overall it looks pretty good, but I didn't really dig into everything here yet.

The main thing that I uncovered is in the comment, about how params are passed.

I also made a few simple changes, and I am about to open a PR into your branch.

zaino-fetch/src/jsonrpsee/connector.rs
pacu: >  This line gave me a bit of trouble:
https://github.com/zcash/zcash/blob/d00fc6f4365048339c83f463874e4d6c240b63af/src/rpc/mining.cpp#L123C1-L123C122
But this is how zcashd handles the parameters.
If I'm reading it correctly (and I might not be!)... in order to specify height you DO need also to specify blocks. ie, if you specify only one argument, it MUST be blocks.

Yes, you are probably right. That's what the original commit was doing 😅. Although is conceptually wrong as well because it's assuming internal details of the validator's implementation. The safe way to do this is to actually enforce the default values ourselves.  Thank you!

> I think if there are more than two parameters passed, they are dropped, so it's fault tolerant in that way.

If more than two parameters are passed help is returned.

dorianvp (dismissed):

pacu:
> > @ala-mode @nachog00 I think there's something else I'm missing...
> 
> @pacu one thing I see here right away is you're using `curl` to test the full node's endpoints. When I brought this up in previous Z3 discussions, Kris specifically told me to use `zcash-cli` as a reference, as opposed to `curl`.
> 
> I haven't been looking at things on this kind of level for maybe 5 or 6 weeks now, so I'm a little rusty.. and it may be the case that under the hood `zcash-cli` is using `curl`.. but for whatever reason that was some handed down wisdom I got from ECC. I was trying `curl` because that's what was in the online documentation (readthedocs), but I was asked to not do that and use their reference `cli` tool instead.

Do you recall where that conversation was? I don't see how we should use zcashd's zcash-cli to actually see if a response passes through zaino from zebra

dorianvp (approved):

ala-mode:
@pacu 
> Do you recall where that conversation was? I don't see how we should use zcashd's zcash-cli to actually see if a response passes through zaino from zebra

It certainly was in a chat room. You could just ask him, too. I was also a little surprised but took it as the 'ECC-approved' way (which is the 'reference' as I understand it).

PR #577, opened 2025-09-24
Add db from file test

AloeareV:
## Motivation

We don't have a test that shows loading a pre-generated DB file functions.

## Solution

Adds a test that loads a pre-generated DB file.

### Follow-up Work

The DB file was manually generated. A procedural mechanism to regenerate the database could be a useful addition.


AloeareV:
I've also updated the names of two tests that claim to load a db from file, but are actually syncing with a mockchain.

zaino-state/src/chain_index/tests/finalised_state/v0.rs
```diff
@@ -195,7 +195,7 @@ async fn delete_blocks_from_db() {
 }
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-async fn load_db_from_file() {
+async fn snyc_db_from_mockchain() {
```
idky137: Hey I think there is a bit of confusion over this test (and the similarly named test in the v1 tests).

This test does in fact load ZainoDB v0 from file, the test which just syncs the database is called "sync_to_height".

This test currently works as follows:
1 - Database is launched.
2 - Database is synced to block height = 200.
3 - Database is closed but tempdir is kept in scope in BlockCacheconfig.
4 - Database is reopened using the tempdir kept in scope.
5 - Database height is checked *without* re syncing database, ensuring the database was loaded correctly from disk.

zaino-state/src/chain_index/tests/finalised_state/v1.rs
```diff
@@ -198,7 +200,7 @@ async fn delete_blocks_from_db() {
 }
 
 #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-async fn load_db_from_file() {
+async fn sync_db_from_mockchain() {
```
idky137: Similar to above I think this test was in fact named correctly before this change.

idky137 (changes requested):
This looks like a good new test but there was a test here already for this purpose, I think we should make sure the name of the old "load from disk" tests remains the same so we do not confuse what tests are doing.

I think this test is also useful as it will check that V1 schema never changes which is a very useful test to have!

zaino-state/src/chain_index/tests/finalised_state/v0.rs
AloeareV: Ahhhh...I see.

AloeareV: I have renamed to "save_db_to_file_and_reload" to distinguish between these, and the new test (which loads a pre-generated db)

idky137 (approved):
looks good!

PR #578, opened 2025-09-25
refactor/chain-index-methods

### LINKED ISSUE #548: Improve commitment tree API for semantic clarity and future-proofing
## Background

While working on BlockWithMetadata for issue #538, we identified opportunities to improve the commitment tree API for better semantic clarity and future-proofing.

## Current State vs Ideal

The codebase has well-designed structured types (CommitmentTreeData, CommitmentTreeRoots, CommitmentTreeSizes) but the interface uses primitive tuples and scattered default handling.

### Current Interface
```rust
async fn get_commitment_tree_roots(
    &self,
    id: BlockHash,
) -> Result<(
    Option<(zebra_chain::sapling::tree::Root, u64)>,
    Option<(zebra_chain::orchard::tree::Root, u64)>,
), Error>;
```

### Current Default Handling (Scattered)
```rust
// In multiple places throughout the codebase:
let ((sapling_root, sapling_size), (orchard_root, orchard_size)) = (
    sapling_root_and_len.unwrap_or_default(),  // Relies on primitive defaults
    orchard_root_and_len.unwrap_or_default(),  // Repeated in many places
);
```

### 1. Semantic Opacity
unwrap_or_default() doesn't convey intent. When reading the code, it's unclear that we want "proper empty commitment tree state" vs "arbitrary primitive defaults".

### 2. Future-Proofing Risk
If Zebra's NoteCommitmentTree evolves (different empty root calculation, metadata, etc.), we'd miss those updates by relying on primitive Default implementations rather than delegating to the authoritative tree type.

### 3. Repeated Logic
Every consumer has to handle the same default logic inline, leading to:
- Code duplication
- Inconsistent handling
- No single source of truth for "what is a proper empty tree"

### 4. Interface Mismatch
Method named get_commitment_tree_roots but returns both roots and sizes, and doesn't use the existing CommitmentTreeData type that was designed for this purpose.

## Proposed Direction

1. **Use existing structured types**: Return CommitmentTreeData instead of nested tuples
2. **Centralize semantic defaults**: Move unwrap_or_default() logic to one place in the interface/helper methods
3. **Delegate to proper tree types**: Use NoteCommitmentTree::default() for semantically correct empty states
4. **Future-proof**: Automatically benefit from any evolution in Zebra's tree handling

## Benefits

- **Semantic Clarity**: Code explicitly shows intent for proper empty tree handling
- **Future-Proof**: Automatically picks up improvements in Zebra's tree implementation
- **DRY Principle**: Single source of truth for empty tree logic instead of scattered defaults
- **Type Safety**: Use existing structured types designed for this data
- **Centralized Logic**: One place to handle edge cases, validation, etc.

## Investigation Needed

Before implementation, we should clarify what Zebra's None responses actually represent semantically (pre-activation vs missing data vs errors) to ensure the centralized defaulting logic handles all cases appropriately.

nachog00:
## Motivation

The NonFinalizedState's `initialize()` and `sync()` methods had grown to 200+ and 179 lines respectively, with complex inline logic that was difficult to test and maintain. This refactoring breaks down these monolithic methods into focused, testable components.

## Solution

- **Modularized initialize()**: Extracted helper methods for listener setup, initial block resolution, and genesis block fetching
- **Simplified sync()**: Broke down sync logic into dedicated methods for staging, reorg handling, and event processing
- **Improved IndexedBlock creation**: Added unified helper methods and cleaner intermediate types
- **Enhanced interfaces**: Replaced tuple-based TryFrom with BlockWithMetadata for better type safety

### Tests

Existing integration tests continue to pass, verifying that the refactoring preserves functionality while improving code organization.


idky137:
Hey looks like interesting work! 

Is this related to https://github.com/orgs/zingolabs/projects/19 at all or something new?

zaino-state/src/chain_index/types.rs
```diff
@@ -1362,195 +1377,305 @@ impl
...
 }
 
-impl
-    TryFrom<(
-        &zebra_chain::block::Block,
+/// Tree root data from blockchain source
+#[derive(Debug, Clone)]
+pub struct TreeRootData {
+    /// Sapling tree root and size
+    pub sapling: Option<(zebra_chain::sapling::tree::Root, u64)>,
```
idky137: Hey I think we have been very clear that this pattern is not allowed in this module. Types must have thier own internals and use from / into for zebra or librustzcash types.

zaino-state/src/chain_index/types.rs
```diff
@@ -1362,195 +1377,305 @@ impl
...
+            .ok_or_else(|| "Missing orchard tree roots".to_string())?;
+        Ok((sapling_root, sapling_size, orchard_root, orchard_size))
+    }
+}
+
+/// Intermediate type to hold block metadata separate from the block itself
+#[derive(Debug, Clone)]
+pub struct BlockMetadata {
+    /// Sapling commitment tree root
+    pub sapling_root: zebra_chain::sapling::tree::Root,
```
idky137: similar to above

zaino-state/src/chain_index/types.rs
```diff
@@ -1362,195 +1377,305 @@ impl
...
+    /// Sapling tree size
+    pub sapling_size: u32,
+    /// Orchard commitment tree root
+    pub orchard_root: zebra_chain::orchard::tree::Root,
```
idky137: similar to above

zaino-state/src/chain_index/types.rs
```diff
@@ -1362,195 +1377,305 @@ impl
...
+    /// Orchard tree size
+    pub orchard_size: u32,
+    /// Parent block's chainwork
+    pub parent_chainwork: ChainWork,
+    /// Network for block validation
+    pub network: zebra_chain::parameters::Network,
```
idky137: similar to above

idky137 (dismissed):
Hey this looks like it is breaking rules we have set out for the types module.

idky137:
After looking this looks like its making a lot of changes to the types.rs module without following the clear rules set out.

zaino-state/src/chain_index/types.rs
nachog00: This just bundles data into named types to pass onto `TryFrom<zebra's_block,...> for IndexedBlock`. It isn't meant to store data in Db nor does it alter any of the structs that read/write to database.

If you pay attention, you'll see it's no different from what was there before with the nameless tuple interface for the conversion.

Maybe you should reassess your rules ass it doesn't make much sense that they are set for the module but coherently for types that read/write to db.

nachog00:
> Hey looks like interesting work!
> 
> Is this related to [Zaino, The Zallet Release](https://github.com/orgs/zingolabs/projects/19) at all or something new?

It is related. The aim of this work is to enable more atomicity in the code, in order to enable currently impossible unit testing, which in turn is needed to showcase current unknown/silent bug classes.

AloeareV:
> After looking this looks like its making a lot of changes to the types.rs module without following the clear rules set out.

Hmm...those rules apply to the types used in the database. Currently, chain_index/types.rs is purely database types. These new types aren't database types, so these rules shouldn't need to apply to them...what if we make a new modules somewhere? (my inclination would be to suggest moving `chain_index/types.rs` to a `chain_index/types/database.rs` for the existing database types, but this would be a big move so maybe not for this PR).

How about for now a `chain_index/intermediate_types.rs`, or otherwise somewhere that makes it clear this type isn't being used by the database?

zaino-state/src/chain_index/types.rs
AloeareV: This rule makes sense if (and only if) `chain_index/types.rs` is purely for database types (which it currently is). If it is, where do we put chain_index types that aren't database types? We could rename `types` to `database_types`, or make a new `non_database_types` or something along those lines...

nachog00:
well, there's a PR that merged not long ago that starts the migration from the single massive types.rs module onto smaller ones. We could simply have types::db and types::anything else.

AloeareV:
> well, there's a PR that merged not long ago that starts the migration from the single massive types.rs module onto smaller ones. We could simply have types::db and types::anything else.

I'd be on-board with this

AloeareV (approved):

PR #579, opened 2025-09-26
add type module rules

idky137:
Type mod rules

## Motivation

Currently the module rules are not clearly defined, breaking these rule will break zainoDB so we must add these rules before any more changes happen.

## Solution

Add doc comment.

### PR Checklist

- [ ] The solution is tested.


Oscar-Pepper (approved):

fluidvanadium:
awesome!

fluidvanadium:
The only thing better than rules is writing rules!

Issue #522, opened 2025-09-04
Fix ChainIndex sync


idky137:
https://github.com/zingolabs/zaino/pull/496 adds several new tests which show that the ChainIndex is not currently working with zcashd. This is a top priority to fix for both Zallet and Zainod development. 

Currently the ChainIndex sync stalls at block 1 when running with zcashd, likely failing on transaction deserialisation.

The following test have been added to show this behaviour:
```
     Summary [ 284.739s] 10 tests run: 5 passed (1 slow), 5 failed, 229 skipped
        FAIL [  29.893s] integration-tests::chain_cache chain_query_interface::find_fork_point_zchashd
        FAIL [  28.900s] integration-tests::chain_cache chain_query_interface::get_block_range_zcashd
        FAIL [  29.896s] integration-tests::chain_cache chain_query_interface::get_raw_transaction_zcashd
        FAIL [  12.865s] integration-tests::chain_cache chain_query_interface::get_transaction_status_zcashd
        FAIL [  25.848s] integration-tests::chain_cache chain_query_interface::sync_large_chain_zcashd
```

idky137:
current failure:
```
[zaino-state/src/chain_index/non_finalised_state.rs:412:34] self.block_to_chainblock(prev_block, &block).await = Err(
    ZebradConnectionError(
        UnrecoverableError(
            InvalidData(
                "Block commitment could not be computed",
            ),
        ),
    ),
)
```

PR #582, opened 2025-09-30
Update infrastructure and LRZ with zingolib pre zebra patch

Oscar-Pepper:
Builds on a previous state of #565 and serves as a replacement to solve the same issue

## Motivation

Zaino testing needs to depend on latest infrastructure to fix issue #522  

## Solution

Update to latest infrastructure including necessary zingolib and LRZ updates for test crates, unifying across our projects.

Zingolib is updated to most recent commit without pulling in zebra patches causing some conflicts and complications.

### Follow-up Work

when zingolib has a non-patched zebra-chain it can be updated. also this will allow zaino testutils to correctly set activation heights for the lightclient which is currently forced to default (all heights at 1)

### PR Checklist

- [ ] The solution is tested.


Oscar-Pepper:
this PR is currently building but failing tests.
updating zebrad to latest main (9db326dacfd19ef0258cfb90424b234b013c128d) did solve a consensus validation error for lightclient sends. this means that tests such as `shield` will pass for zcashd and zebrad fetch service but fail for the state service. this leads me to believe - with discussion from @idky137 - that we need to update zaino to use this commit on zebra as a dependency as it is now mismatched with the binary version. I will start attmpeting to update zaino to latest zebra main now as I can't seem to find a solution using the v2.5.0 zebrad release

Oscar-Pepper:
in the current version (41cbbfebc769baed8c470b5576b07b24d430f0d3) the failing `shield` tests are passing with zebrad v2.5.0. there are other test fails to be investigated but this is a step forward!

Oscar-Pepper:
closed for #588

PR #583, opened 2025-09-30
define TestVector types in favor of large tuples

AloeareV:
## Motivation

More concrete types in the test vectors will allow changing elements of those types more easily, as unlike in tuples, field names don't change when new fields are added or removed.

## Solution

Use new structs in place of anonymous tuples. 
There's also a new helper function, to replace some code I was seeing in many places.

### Tests

Non-functional refactoring of test code


AloeareV:
Note: For ease of review, don't include the 'cargo format' commit

zaino-state/src/chain_index/finalised_state/migrations.rs
```diff
@@ -75,7 +75,7 @@ impl<T: BlockchainSource> MigrationManager<T> {
     }
 
     /// Return the next migration for the current version.
-    fn get_migration(&self) -> Result<impl Migration<T>, FinalisedStateError> {
+    fn get_migration(&self) -> Result<impl Migration<T> + use<T>, FinalisedStateError> {
```
idky137: What is the `+ use<T> ` for here?

idky137 (dismissed):
Much nicer!

zaino-state/src/chain_index/finalised_state/migrations.rs
AloeareV: This makes it so the `impl Migration` existential type will capture all lifetimes T contains. (see https://doc.rust-lang.org/std/keyword.use.html#precise-capturing)
Edit: Otherwise, the behaviour differs between 2021 edition rust and 2024 edition. I think I added this during an attempted update to 2024 rust, in which this is needed. I think this can be removed to no effect for the moment...although I would like to update our rust edition, at which point this would need to be added back again if removed.

nachog00 (approved):

PR #585, opened 2025-10-01
feat(rpc): add `getmininginfo` endpoint

### LINKED ISSUE #213: 1.17: Add Zcash RPC `getmininginfo`
Spec: https://zcash.github.io/rpc/getmininginfo.html

dorianvp:
fixes #213

## Motivation

Missing passthrough for `getmininginfo`.

## Solution

Implement `getmininginfo`.

### PR Checklist

- [x] The solution is tested.


dorianvp:
This endpoint and it's corresponding tests have been implemented as of [d33d1d1](https://github.com/zingolabs/zaino/pull/585/commits/d33d1d192b3df2fedcd77eabaa0575556d35b9c9).
However, there are some cases where `zainod` might act as a filter for `zcashd`. 

I will look into accounting for both response type variations and make it so that Zaino simply bubbles it up.

zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -1514,3 +1514,131 @@ pub struct GetMempoolInfoResponse {
...
     type RpcError = Infallible;
 }
+
+impl ResponseToError for GetMiningInfoWire {
+    type RpcError = Infallible;
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+enum Num {
```
fluidvanadium: What is the deal with this enum?

fluidvanadium:
I just grokked that this needs review. I will take a shot at it as soon as possible.

idky137 (dismissed):
Nice work on these!

zaino-fetch/src/jsonrpsee/response.rs
dorianvp: It's used to catch serialized upstream changes.

zaino-fetch/src/jsonrpsee/response/mining_info.rs
```diff
@@ -0,0 +1,135 @@
...
+        let network_hash_rate = w.networkhashps.map(|n| n.as_f64());
+        let network_solution_rate = w.networksolps.map(|n| n.as_f64());
+
+        Self {
+            tip_height: w.tip_height,
+            current_block_size: w.current_block_size,
+            current_block_tx: w.current_block_tx,
+            network_solution_rate,
+            network_hash_rate,
+            chain: if w.chain.is_empty() {
```
pacu: This field is supposed to be present in both implementation, why should it be empty? It's weird that falls back to main

pacu (changes requested):

zaino-fetch/src/jsonrpsee/response/mining_info.rs
dorianvp: Good catch, fixed!

idky137 (dismissed):
Hey looks really good! thanks for working on this!

Sorry to be pedantic, but i think it would be good for us to also add an test to integration_tests::json_server also, as this is the test module we have been building up to check the validity of the actual JsonRPC server in zainod.

I know you are quite busy with zingolib currently so im happy to add this test to this pr if helpful?

dorianvp:
> Hey looks really good! thanks for working on this!
> 
> Sorry to be pedantic, but i think it would be good for us to also add an test to integration_tests::json_server also, as this is the test module we have been building up to check the validity of the actual JsonRPC server in zainod.
> 
> I know you are quite busy with zingolib currently so im happy to add this test to this pr if helpful?

My bad, I'll add that test. Shouldn't be too extensive, as we're not currently setting up mining...

zaino-fetch/src/jsonrpsee/response/mining_info.rs
```diff
@@ -0,0 +1,131 @@
...
+
+use crate::jsonrpsee::connector::ResponseToError;
+
+impl ResponseToError for GetMiningInfoWire {
+    type RpcError = Infallible;
+}
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
+#[serde(untagged)]
+enum Num {
```
pacu: is this because Zebra returns a float and Zcashd an unsigned int?

Shouldn't Zaino *solve* this problem instead of forwarding it to the client instead?

zaino-fetch/src/jsonrpsee/response/mining_info.rs
```diff
@@ -0,0 +1,131 @@
...
+
+        Self {
+            tip_height: w.tip_height,
+            current_block_size: w.current_block_size,
+            current_block_tx: w.current_block_tx,
+            network_solution_rate,
+            network_hash_rate,
+            chain: w.chain,
+            testnet: w.testnet,
+
```
pacu: apply cargo fmt and clippy to get rid of blank spaces

pacu (dismissed):
I think this should actually make a decision on which is the final API. Otherwise we are just porting tech debt into Zaino and it deserves better

zaino-fetch/src/jsonrpsee/response/mining_info.rs
zancas: Are we constrained to represent values using "per second"?  Can we decrease the time delta?

Then we could say something like u64-of-solutions/nanosecond.

dorianvp: Maybe in an endpoint that does not specify the time delta in its name...

dorianvp: I ended up removing it.

zaino-fetch/src/jsonrpsee/response/mining_info.rs
```diff
@@ -104,28 +89,208 @@ pub struct MiningInfo {
...
+        assert_eq!(wire.pooledtx, Some(5));
+        assert_eq!(wire.generate, Some(false));
+
+        let s = serde_json::to_string(&wire).unwrap();
+        let wire2: GetMiningInfoWire = serde_json::from_str(&s).unwrap();
+        assert_eq!(wire, wire2);
+    }
+
+    #[test]
+    fn minimal_payload_defaults_zebrad() {
```
fluidvanadium: Maybe this is supposed to be `minimal_payload_defaults`? It doesnt have anything to do with zebrad.

zaino-fetch/src/jsonrpsee/response/mining_info.rs
```diff
@@ -104,28 +89,208 @@ pub struct MiningInfo {
...
+    #[test]
+    fn convert_to_internal_from_zebrad() {
+        let wire: GetMiningInfoWire = serde_json::from_str(&zebrad_json()).unwrap();
+        let mining_info: MiningInfo = wire.clone().into();
+
+        assert_eq!(mining_info.tip_height, wire.tip_height);
+        assert_eq!(mining_info.current_block_size, wire.current_block_size);
+        assert_eq!(mining_info.current_block_tx, wire.current_block_tx);
+
+        assert_eq!(mining_info.network_solution_rate, Some(1_234));
```
fluidvanadium: Why switch to checking values?
It would be the same test but drier if we continued comparing against `wire` instead of hardcoded values.

zaino-fetch/src/jsonrpsee/response/mining_info.rs
```diff
@@ -104,28 +89,208 @@ pub struct MiningInfo {
...
+        assert_eq!(mining_info.errors, None);
+        assert!(mining_info.extras.contains_key("somefuture"));
+    }
+
+    #[test]
+    fn convert_to_internal_from_zcashd() {
+        let wire: GetMiningInfoWire = serde_json::from_str(&zcashd_json()).unwrap();
+        let mining_info: MiningInfo = wire.clone().into();
+
+        assert_eq!(mining_info.tip_height, 765_432);
```
fluidvanadium: As above, lets compare against `wire`.

fluidvanadium (changes requested):
A couple of objectionitos.

zaino-fetch/src/jsonrpsee/response/mining_info.rs
fluidvanadium: The hardcoded values are already checked in another test

dorianvp: Fixed!

zaino-fetch/src/jsonrpsee/response/mining_info.rs
dorianvp: fixed

dorianvp:
Comments addressed!

fluidvanadium (approved):
This is ready to go.

zancas (approved):
Adding approval after @fluidvanadium and @idky137 .

zancas (approved):

PR #590, opened 2025-10-03
Correctly determine the best chain

### LINKED ISSUE #679: coinbase_height's optionality incorrectly used to determine presence on the best chain
We are using coinbase height's Optional return value as an indicator of presence on the best chain. This is incorrect.

AloeareV:
## Motivation
We had been using coinbase height's Optional return value as an indicator of presence on the best chain. This was incorrect.

## Solution

We now determine the best chain based off of the best chaintip and following it down to a known block. This also fixes an issue where the best tip was determined by height, instead of chainwork.

### Tests

Still draft, as there are no new tests which show the bugs this PR fixes. All current tests still pass.

### Follow-up Work

Tests are still in progress. 

### PR Checklist

- [ ] The solution is tested.


AloeareV:
Currently, block commitments are not being generated properly in the proptest. This is due to how network upgrades are set, via the network_upgrade_override field of the LedgerStateOverride. This makes it so that regardless of what network is arbitrarily generated, the blocks will have the supplied network upgrade active. Unfortunately, block commitment generation does not respect this NU override and will use the arbitrarily generated network. 

I've been thinking about workarounds (we could manually handle the block commitments, probably by mostly copy-pasting zebra code)...but this is already fixed in zebra 3.0.0-rc. The newer zebra, instead of overriding individual network upgrades, allows overriding the network itself...and this override is respected by the block commitment generation code, and so would correctly generate block commitments for whatever network we wish to provide.

AloeareV:
> Currently, block commitments are not being generated properly in the proptest. This is due to how network upgrades are set, via the network_upgrade_override field of the LedgerStateOverride. This makes it so that regardless of what network is arbitrarily generated, the blocks will have the supplied network upgrade active. Unfortunately, block commitment generation does not respect this NU override and will use the arbitrarily generated network.
> 
> I've been thinking about workarounds (we could manually handle the block commitments, probably by mostly copy-pasting zebra code)...but this is already fixed in zebra 3.0.0-rc. The newer zebra, instead of overriding individual network upgrades, allows overriding the network itself...and this override is respected by the block commitment generation code, and so would correctly generate block commitments for whatever network we wish to provide.

This will be addressed in #621

AloeareV:
Superseded by https://github.com/zingolabs/zaino/pull/771

PR #592, opened 2025-10-06
feat(rpc): add `getblocksubsidy` endpoint

dorianvp:
On top of #587.

## Motivation

#324. Also fixes a CI bug where workflows are not re-triggered when marking a PR as ready for review.

## Solution

Implement `getblocksubsidy`.
### PR Checklist

- [x] The PR name is suitable for the release notes.
- [x] The solution is tested.
- [x] The documentation is up to date.

PR #595, opened 2025-10-08
feat(rpc): add `gettxoutsetinfo` endpoint

### LINKED ISSUE #210: 1.14: Add Zcash RPC `gettxoutsetinfo`
Spec: https://zcash.github.io/rpc/gettxoutsetinfo.html

dorianvp:
On top of #592.

## Motivation

https://github.com/zingolabs/zaino/issues/210.

This PR also specifies the algorithm used to generate the UTXO set snapshot hash.

## Solution

Implement `gettxoutsetinfo`.

### PR Checklist

- [x] The solution is tested.


dorianvp:
This PR is mostly ready. I only left a couple of review to upgrade some types, but functionally it is ready for review.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+Any change to the encoding rules or semantics **MUST** bump the domain string (e.g., `…-V2\0`) and is out of scope of this document.
+
+## Inputs
+
+To compute the snapshot hash, the implementation needs:
+
+- `network`: ASCII string identifying the chain; recommended values: `"mainnet"`, `"testnet"`, `"regtest"`.
+- `best_height`: the best chain height at the time of the snapshot (unsigned 32-bit).
```
zancas: What's "best"..   "highest"?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+"ZAINO-UTXOSET-V1\0"
+```
+
+Any change to the encoding rules or semantics **MUST** bump the domain string (e.g., `…-V2\0`) and is out of scope of this document.
+
+## Inputs
+
+To compute the snapshot hash, the implementation needs:
+
+- `network`: ASCII string identifying the chain; recommended values: `"mainnet"`, `"testnet"`, `"regtest"`.
```
zancas: Can we consider "idiosyncratic" as a replacement for "regtest" in this context?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+- `best_height`: the best chain height at the time of the snapshot (unsigned 32-bit).
+- `best_block`: the 32-byte block hash of the best chain tip, in the node’s _canonical internal byte order_.
```
zancas, line 49: Can we have a reference for the _canonical internal byte order_?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+- `UTXO set`: a finite multimap keyed by outpoints `(txid, vout)` to outputs `(value_zat, scriptPubKey)`, where:
+
+  - `txid` is a 32-byte transaction hash (internal byte order).
```
zancas, line 52: Same as above?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+  - `vout` is a 32-bit output index (0-based).
+  - `value_zat` is a non-negative amount in zatoshis, range-checked to the node’s monetary bounds (e.g., `0 ≤ value_zat ≤ MAX_MONEY`).
+  - `scriptPubKey` is a byte string.
+
+Implementations **MUST** reject negative values or out-of-range amounts prior to hashing.
+
+## Canonical Ordering
```
zancas, line 59: Consider moving this section such that it precedes references to it.

zancas: OK, this section doesn't refer to what I thought that it might.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+
+This ordering **MUST** be used for serialization.
+
+## Serialization
+
+The byte stream fed to the hash is the concatenation of a **header** and **entries**:
+
+### Header
+
+- ASCII bytes: `"ZAINO-UTXOSET-V1\0"`
```
zancas, line 74: Why is this `\0` present?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+
+For each `(txid, vout, value_zat, scriptPubKey)`:
+
+- `txid` as 32 raw bytes.
+- `vout` as `u32` little-endian.
+- `value_zat` as `u64` little-endian.
+- `script_len` as CompactSize (Bitcoin/Zcash varint) of `scriptPubKey.len()`.
+- `scriptPubKey` raw bytes.
+
+**Note:** No per-transaction terminators or grouping markers are used; the format commits to _outputs_, not _transactions_.
```
zancas: Nice.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,153 @@
...
+## Error Handling
+
+- If any `value_zat` is negative or exceeds `MAX_MONEY`, the snapshot procedure **MUST** fail and **MUST NOT** produce a hash.
+- If the UTXO set changes during iteration (non-atomic read), the implementation **SHOULD** retry using a stable view (e.g., read lock or height-pinned snapshot).
+
+## Security and Interop Considerations
+
+- This hash is **not a consensus commitment** and **MUST NOT** be used to validate blocks or transactions.
+- The domain string prevents cross-protocol collisions.
+- Including `network`, `best_height`, and `best_block` prevents accidental equality across different nodes or heights.
```
zancas: Is "network" necessary?

Could a testnet utxo have the same form as a mainnet utxo?

Is "network" sufficient?

What is the network is none of "mainnet" "testnet" or "regtest"?   Ycash?

What is present in a utxo?

Doesn't it contain a height in which chase a

integration-tests/tests/fetch_service.rs
```diff
@@ -502,6 +504,53 @@ async fn fetch_service_get_address_tx_ids(validator: &ValidatorKind) {
     test_manager.close().await;
 }
 
+async fn fetch_service_get_txout_set_info() {
```
zancas, line 568: What does this test demonstrate?

integration-tests/tests/json_server.rs
```diff
@@ -726,6 +726,50 @@ mod zcashd {
             test_manager.close().await;
         }
 
+        #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
zancas: This demonstrates the equivalence of the txoutset across zaino and zcashd?

integration-tests/tests/state_service.rs
```diff
@@ -932,6 +934,59 @@ async fn state_service_get_address_tx_ids_testnet() {
     test_manager.close().await;
 }
 
+async fn state_service_get_txout_set_info() {
```
zancas, line 1698: state_service and a client?

zancas:
I see several equivalence tests, how about some non-equivalance tests proving sensitivity to variance in txoutsets?

zaino-fetch/src/jsonrpsee/response/common/block.rs
```diff
@@ -0,0 +1,229 @@
...
+// Helpers
+fn hex_lower(bytes: &[u8; 32]) -> String {
+    let mut s = String::with_capacity(64);
+    for &b in bytes.iter() {
+        use core::fmt::Write;
+        let _ = write!(&mut s, "{:02x}", b);
+    }
+    s
+}
+fn hex_nibble(b: u8) -> Option<u8> {
```
zancas: "b" is the parameter name here?

Hmm...

https://doc.rust-lang.org/book/appendix-02-operators.html#non-operator-symbols

Maybe choose a different parameter name?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
dorianvp: Network as a free-text could actually be removed. Probably it'd make more sense to use the network magic instead (or genesis?).

integration-tests/tests/json_server.rs
dorianvp: Yes. Please look at the module-level comments I have added.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,169 @@
...
+
+To compute the snapshot hash, the implementation needs:
+
+- `network`: ASCII string identifying the chain. Recommended values: `"mainnet"`, `"testnet"`, `"regtest"`.
+- `best_height`: the best chain height at the time of the snapshot (unsigned 32-bit).
+- `best_block`: the 32-byte block hash of the best chain tip, in the node’s _canonical internal byte order_.
+- `UTXO set`: a finite multimap keyed by outpoints `(txid, vout)` to outputs `(value_zat, scriptPubKey)`, where:
+
+  - `txid` is a 32-byte transaction hash (internal byte order).
+  - `vout` is a 32-bit output index (0-based).
```
dorianvp, line 53: Leaving a note here:

If we serialize per unspent as `txid` || `value` || `script` and a transaction contains two outputs with identical (`value`, `script`), then two different UTXO sets that differ only by which index is unspent will serialize to the same bytes (and hash).

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
dorianvp: Expanded it a bit

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
dorianvp: Replaced by `genesis_block_hash`

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
dorianvp: I ended up using `genesis_block_hash`

integration-tests/tests/state_service.rs
dorianvp: The module level docs specifies that these tests compare output from the `StateService` and the `FetchService`.

integration-tests/tests/fetch_service.rs
dorianvp: The top-level docs specify that these tests compare the output of `FetchService` with the output of `JsonRpSeeConnector`.

dorianvp: To be more precise, this test only asserts that the initial output for both returns the same shape.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
dorianvp: It only acts as a terminator/delimiter. It is not strictly necessary...

zaino-fetch/src/jsonrpsee/response/common/block.rs
dorianvp: I ended up deleting this module.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
nuttycom: `vout` is a misleading name for this; I would call it `output_index` because `vout` is generally used to refer to the vector of outputs of a transaction; I was confused by this name before I got to this line.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,168 @@
...
+  - `value_zat` is a non-negative amount in zatoshis, range-checked to the node’s monetary bounds (e.g., `0 ≤ value_zat ≤ MAX_MONEY`).
+  - `scriptPubKey` is a byte string.
+
+Implementations **MUST** reject negative values or out-of-range amounts prior to hashing.
+
+## Canonical Ordering
+
+The snapshot **MUST** be ordered as follows, independent of the node’s in-memory layout:
+
+1. Sort by `txid` ascending, comparing the raw 32-byte values as unsigned bytes.
```
nuttycom, line 63: This seems like a bad serialization, because it requires recomputation over the entire UTXO set whenever a new block is received. The UTXO set can be very large; it would be much better to choose a snapshot protocol where snapshot hashes can incrementally build on the snapshot hash prior to the addition of a new block.

nuttycom: If the UTXO set were stored in a B-tree data structure that internally kept Merkle hashes at the nodes, then it might be okay to use the Merkle root of that data structure for the snapshot identifier. It would need to be the case that the fanout of the B-tree and the insertion semantics were well-specified to ensure that everyone uses the same hashing approach.

One possibility that would allow for this to work as-specified would be to use a separate B-tree (implementing a set, rather than a map) for producing the hashes; since the txid commits to the effects of each transaction, one could build the snapshot identifier alongside the actual data, but building that identifier in parallel would have a risk of data inconsistencies with the primary store.

In general, I feel like the UTXO set would be best represented as a [persistent data structure](https://en.wikipedia.org/wiki/Persistent_data_structure) with good amortized append costs.

packages/zaino-state/src/backends/state.rs
```diff
@@ -853,6 +863,64 @@ impl StateServiceSubscriber {
...
 
+    /// Fetches all UTXOs from the state service, and returns them in a map.
+    pub(crate) async fn get_txout_set(
+        state: &ReadStateService,
+    ) -> Result<HashMap<transparent::OutPoint, transparent::Output>, StateServiceError> {
+        let mut utxos: HashMap<transparent::OutPoint, transparent::Output> = HashMap::new();
+
+        let mut state = state.clone();
+
+        let zebra_state::ReadResponse::Tip { 0: tip } = state
```
arya2, line 961: Nitpick: Why destructure it this way rather than as a tuple struct?

```suggestion
        let zebra_state::ReadResponse::Tip(tip) = state
```

packages/zaino-state/src/backends/state.rs
```diff
@@ -853,6 +863,64 @@ impl StateServiceSubscriber {
         }
     }
 
+    /// Fetches all UTXOs from the state service, and returns them in a map.
```
arya2, line 953: Loading every UTXO into memory shouldn't be necessary and could require more memory than is available on some systems.

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,168 @@
...
+
+- Verify that two nodes at the same best block have the same UTXO set across implementations and versions.
+- Pin failing test fixtures to a snapshot hash to reproduce issues.
+- Log periodic hashes to show continuity of state over time.
+
+The hash is _not_ input to consensus validation.
+
+## Motivation
+
+Different nodes (e.g., `zcashd`, Zebra, indexers) may expose distinct internals or storage layouts. Operators often need a cheap way to verify “we’re looking at the same unspent set” without transporting the entire set. A canonical, versioned snapshot hash solves this.
```
arya2, line 31: I see that zcashd had this method returning a `hash_serialized` already, but, why is it not enough to check that the block hashes match?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
```diff
@@ -0,0 +1,168 @@
...
+
+```
+"ZAINO-UTXOSET-V1\0"
+```
+
+Any change to the encoding rules or semantics **MUST** bump the domain string (e.g., `…-V2\0`) and is out of scope of this document.
+
+## Inputs
+
+To compute the snapshot hash, the implementation needs:
```
arya2, line 45: Why include anything other than the UTXOs as inputs in the snapshot hash? Shouldn't we already know that we're looking at the same UTXO set if the best block hashes match?

docs/json_rpc/gettxoutsetinfo/canonical_utxo_set_snapshot_hash.md
arya2: > The UTXO set can be very large; it would be much better to choose a snapshot protocol where snapshot hashes can incrementally build on the snapshot hash prior to the addition of a new block.

[`sparse-merkle-tree`](https://docs.rs/sparse-merkle-tree) could be useful here, Zaino could:
- Implement [`Value`](https://docs.rs/sparse-merkle-tree/latest/sparse_merkle_tree/traits/trait.Value.html) for a struct representing the transaction output data to which `hash_serialized` is committing, 
- Implement [`StoreReadOps`](https://docs.rs/sparse-merkle-tree/latest/sparse_merkle_tree/traits/trait.StoreReadOps.html)/[`StoreWriteOps`](https://docs.rs/sparse-merkle-tree/latest/sparse_merkle_tree/traits/trait.StoreWriteOps.html) for on-disk storage of the tree,
- Update the tree and a cache with the other fields in `TxOutSetInfo` when Zaino is indexing blocks, and
- Return the cached `TxOutSetInfo` from the RPC method.

packages/zaino-state/src/backends/state.rs
```diff
@@ -853,6 +863,64 @@ impl StateServiceSubscriber {
...
+        };
+
+        let (tip_h, _) = match tip {
+            Some(tip) => tip,
+            _ => return Ok(utxos),
+        };
+
+        for h in 0..=tip_h.0 {
+            let blk = state
+                .call(ReadRequest::Block(HashOrHeight::Height(Height(h))))
```
arya2, line 981: It could be faster to read just the unspent outputs by iterating over `TransactionLocation { height: 0, index: 0 }..={ height: target_height, index: OutputIndex::MAX }` in the [`utxo_by_out_loc`](https://zebra.zfnd.org/dev/state-db-upgrades.html#current-state-database-format) column family and then read any remaining blocks to add any UTXOs that were in blocks that were commited during the iteration.

packages/zaino-state/src/backends/state.rs
```diff
@@ -853,6 +863,64 @@ impl StateServiceSubscriber {
...
+                .await
+                .unwrap();
+            let block = match blk {
+                ReadResponse::Block(Some(b)) => b,
+                _ => continue,
+            };
+
+            for tx in block.transactions.clone() {
+                let txid = tx.hash();
```
arya2, line 990: It may be faster to lookup the transaction hashes in the [`tx_loc_by_hash`](https://zebra.zfnd.org/dev/state-db-upgrades.html#current-state-database-format) column family.

packages/zaino-state/src/backends/state.rs
```diff
@@ -1482,6 +1550,94 @@ impl ZcashIndexer for StateServiceSubscriber {
             .collect())
     }
 
+    async fn get_txout_set_info(&self) -> Result<GetTxOutSetInfo, Self::Error> {
+        let txouts = Self::get_txout_set(&self.read_state_service).await.unwrap();
+
+        let best_block_hash = self.get_best_blockhash().await.unwrap().hash();
```
arya2, line 1919: `get_txout_set()` may not (and on Mainnet, generally won't) read transactions from all blocks up to the best block hash being read here. It would be better to return the tip hash from `get_txout_set()`.

arya2 (commented):
There are some performance and concurrency issues with this design. It would be better to update a cached `TxOutSetInfo` as Zaino indexes blocks and to use a merkle tree to update the `hash_serialized` field.

dorianvp:
Hey folks, thanks for reviewing this! I'll be addressing these comments soon

zancas:
@dorianvp this RPC involves specification which is scheduled in the next Mile Stone.  I am requesting to move this RPC to that Mile Stone.

zancas:
@dorianvp is the original author of these commits, which I just rebased.

pacu:
closed. implemented #1112

PR #599, opened 2025-10-10
Linted clippy::pedantic autofixes

fluidvanadium:
## Motivation

Automatically added some really nice features like
- functional doc links
- fewer unnecessary casts
- specific imports instead of star
and other tidbits that dont really matter.

### PR Checklist

- [x] The solution is tested.


dorianvp (changes requested):
I see a bunch of applied lints, but I don't see the corresponding `clippy.toml` file.

fluidvanadium:
> I see a bunch of applied lints, but I don't see the corresponding `clippy.toml` file.

Ah, right. Better to do it that way.

PR #473, opened 2025-08-13
impl getblockheader

ala-mode:
fixes #203 

WIP / DRAFT

PR #603, opened 2025-10-15
feat(rpc): add `getblockheader` endpoint

### LINKED ISSUE #203: 1.7: Add Zcash RPC `getblockheader`
Spec: https://zcash.github.io/rpc/getblockheader.html

dorianvp:
On top of #595, supersedes #473.

## Motivation

#203.

## Solution

Implement `getblockheader`.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/integration-tests/tests/fetch_service.rs
+++ b/integration-tests/tests/fetch_service.rs
@@ -700,6 +700,54 @@ async fn fetch_service_get_block(validator: &ValidatorKind) {
     test_manager.close().await;
 }
 
+async fn fetch_service_get_block_header(validator: &ValidatorKind) {
+    let (test_manager, _fetch_service, fetch_service_subscriber) =
+        create_test_manager_and_fetch_service(validator, None, true, true, true, true).await;
+
+    const BLOCK_LIMIT: u32 = 10;
+
+    for i in 0..BLOCK_LIMIT {
+        test_manager.local_net.generate_blocks(1).await.unwrap();
+        tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+
+        let block = fetch_service_subscriber
+            .z_get_block(i.to_string(), Some(1))
+            .await
+            .unwrap();
+
+        let block_hash = match block {
+            GetBlock::Object(block) => block.hash(),
+            GetBlock::Raw(_) => panic!("Expected block object"),
+        };
+
+        let fetch_service_get_block_header = fetch_service_subscriber
+            .get_block_header(block_hash.to_string(), false)
+            .await
+            .unwrap();
+
+        let jsonrpc_client = JsonRpSeeConnector::new_with_basic_auth(
+            test_node_and_return_url(
+                test_manager.zebrad_rpc_listen_address,
+                false,
+                None,
+                Some("xxxxxx".to_string()),
+                Some("xxxxxx".to_string()),
+            )
+            .await
+            .unwrap(),
+            "xxxxxx".to_string(),
+            "xxxxxx".to_string(),
+        )
+        .unwrap();
+
+        let rpc_block_header_response = jsonrpc_client
+            .get_block_header(block_hash.to_string(), false)
+            .await
+            .unwrap();
+        assert_eq!(fetch_service_get_block_header, rpc_block_header_response);
+    }
+}
+
 async fn fetch_service_get_best_blockhash(validator: &ValidatorKind) {
     let (mut test_manager, _fetch_service, fetch_service_subscriber) =
         create_test_manager_and_fetch_service(validator, None, true, true, true, true).await;
@@ -1544,6 +1592,11 @@ mod zcashd {
         }
 
         #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+        pub(crate) async fn block_header() {
+            fetch_service_get_block_header(&ValidatorKind::Zcashd).await;
+        }
+
+        #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
         pub(crate) async fn difficulty() {
             assert_fetch_service_difficulty_matches_rpc(&ValidatorKind::Zcashd).await;
         }
@@ -1751,6 +1804,11 @@ mod zebrad {
         #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
         pub(crate) async fn block() {
             fetch_service_get_block(&ValidatorKind::Zebrad).await;
+        }
+
+        #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+        pub(crate) async fn block_header() {
+            fetch_service_get_block_header(&ValidatorKind::Zebrad).await;
         }
 
         #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
--- a/integration-tests/tests/fetch_service.rs
+++ b/integration-tests/tests/fetch_service.rs
@@ -715,7 +715,7 @@ async fn fetch_service_get_block(validator: &ValidatorKind) {
 
 async fn fetch_service_get_block_header(validator: &ValidatorKind) {
     let (test_manager, _fetch_service, fetch_service_subscriber) =
-        create_test_manager_and_fetch_service(validator, None, true, true, true, true).await;
+        create_test_manager_and_fetch_service(validator, None, true, true, true).await;
 
     const BLOCK_LIMIT: u32 = 10;
 
@@ -740,8 +740,7 @@ async fn fetch_service_get_block_header(validator: &ValidatorKind) {
 
         let jsonrpc_client = JsonRpSeeConnector::new_with_basic_auth(
             test_node_and_return_url(
-                test_manager.zebrad_rpc_listen_address,
-                false,
+                test_manager.full_node_rpc_listen_address,
                 None,
                 Some("xxxxxx".to_string()),
                 Some("xxxxxx".to_string()),
--- a/integration-tests/tests/fetch_service.rs
+++ b/integration-tests/tests/fetch_service.rs
@@ -760,12 +760,12 @@ async fn fetch_service_get_block_header(validator: &ValidatorKind) {
             .unwrap();
 
         let fetch_service_get_block_header_verbose = fetch_service_subscriber
-            .get_block_header(block_hash.to_string(), false)
+            .get_block_header(block_hash.to_string(), true)
             .await
             .unwrap();
 
         let rpc_block_header_response_verbose = jsonrpc_client
-            .get_block_header(block_hash.to_string(), false)
+            .get_block_header(block_hash.to_string(), true)
             .await
             .unwrap();
 
--- a/integration-tests/tests/fetch_service.rs
+++ b/integration-tests/tests/fetch_service.rs
@@ -758,7 +758,22 @@ async fn fetch_service_get_block_header(validator: &ValidatorKind) {
             .get_block_header(block_hash.to_string(), false)
             .await
             .unwrap();
+
+        let fetch_service_get_block_header_verbose = fetch_service_subscriber
+            .get_block_header(block_hash.to_string(), false)
+            .await
+            .unwrap();
+
+        let rpc_block_header_response_verbose = jsonrpc_client
+            .get_block_header(block_hash.to_string(), false)
+            .await
+            .unwrap();
+
         assert_eq!(fetch_service_get_block_header, rpc_block_header_response);
+        assert_eq!(
+            fetch_service_get_block_header_verbose,
+            rpc_block_header_response_verbose
+        );
     }
 }
 
--- a/integration-tests/tests/json_server.rs
+++ b/integration-tests/tests/json_server.rs
@@ -671,6 +671,8 @@ mod zcashd {
     use super::*;
 
     pub(crate) mod zcash_indexer {
+        use zebra_rpc::methods::GetBlock;
+
         use super::*;
 
         #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
@@ -774,6 +776,45 @@ mod zcashd {
         #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
         async fn z_get_block() {
             z_get_block_inner().await;
+        }
+
+        #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+        async fn get_block_header() {
+            let (
+                test_manager,
+                _zcashd_service,
+                zcashd_subscriber,
+                _zaino_service,
+                zaino_subscriber,
+            ) = create_test_manager_and_fetch_services(false, false).await;
+
+            const BLOCK_LIMIT: u32 = 10;
+
+            for i in 0..BLOCK_LIMIT {
+                test_manager.local_net.generate_blocks(1).await.unwrap();
+                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+
+                let block = zcashd_subscriber
+                    .z_get_block(i.to_string(), Some(1))
+                    .await
+                    .unwrap();
+
+                let block_hash = match block {
+                    GetBlock::Object(block) => block.hash(),
+                    GetBlock::Raw(_) => panic!("Expected block object"),
+                };
+
+                let zcashd_get_block_header = zcashd_subscriber
+                    .get_block_header(block_hash.to_string(), false)
+                    .await
+                    .unwrap();
+
+                let zainod_block_header_response = zaino_subscriber
+                    .get_block_header(block_hash.to_string(), false)
+                    .await
+                    .unwrap();
+                assert_eq!(zcashd_get_block_header, zainod_block_header_response);
+            }
         }
 
         #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
--- a/integration-tests/tests/json_server.rs
+++ b/integration-tests/tests/json_server.rs
@@ -796,7 +796,7 @@ mod zcashd {
                 zcashd_subscriber,
                 _zaino_service,
                 zaino_subscriber,
-            ) = create_test_manager_and_fetch_services(false, false).await;
+            ) = create_test_manager_and_fetch_services(false).await;
 
             const BLOCK_LIMIT: u32 = 10;
 
--- a/zaino-fetch/src/jsonrpsee/connector.rs
+++ b/zaino-fetch/src/jsonrpsee/connector.rs
@@ -25,12 +25,12 @@ use zebra_rpc::client::ValidateAddressResponse;
 use crate::jsonrpsee::{
     error::{JsonRpcError, TransportError},
     response::{
-        block_subsidy::GetBlockSubsidy, peer_info::GetPeerInfo, GetBalanceError,
-        GetBalanceResponse, GetBlockCountResponse, GetBlockError, GetBlockHash, GetBlockResponse,
-        GetBlockchainInfoResponse, GetInfoResponse, GetMempoolInfoResponse, GetSubtreesError,
-        GetSubtreesResponse, GetTransactionResponse, GetTreestateError, GetTreestateResponse,
-        GetUtxosError, GetUtxosResponse, SendTransactionError, SendTransactionResponse, TxidsError,
-        TxidsResponse,
+        block_header::GetBlockHeader, block_subsidy::GetBlockSubsidy, peer_info::GetPeerInfo,
+        GetBalanceError, GetBalanceResponse, GetBlockCountResponse, GetBlockError, GetBlockHash,
+        GetBlockResponse, GetBlockchainInfoResponse, GetInfoResponse, GetMempoolInfoResponse,
+        GetSubtreesError, GetSubtreesResponse, GetTransactionResponse, GetTreestateError,
+        GetTreestateResponse, GetUtxosError, GetUtxosResponse, SendTransactionError,
+        SendTransactionResponse, TxidsError, TxidsResponse,
     },
 };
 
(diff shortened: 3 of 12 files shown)


pacu (dismissed):
LGTM (untested)

zancas:
I see CI fails...

dorianvp:
> I see CI fails...

Same flakes we're seeing on dev.

zaino-serve/src/rpc/jsonrpc/service.rs
```diff
@@ -218,6 +219,24 @@ pub trait ZcashIndexerRpc {
...
 
+    /// If verbose is false, returns a string that is serialized, hex-encoded data for blockheader `hash`.
+    /// If verbose is true, returns an Object with information about blockheader `hash`.
+    ///
+    /// # Parameters
+    ///
+    /// - hash: (string, required) The block hash
+    /// - verbose: (boolean, optional, default=true) true for a json object, false for the hex encoded data
+    ///
+    /// zcashd reference: [`getblockheader`](https://zcash.github.io/rpc/getblockheader.html)
```
ala-mode, line 230: Question, are we currently okay with pointing to the online rpc referene and simply calling it 'zcashd reference' - when in fact we know that sometimes these documents are inaccurate or incomplete?

Do we have a full link to source code elsewhere in doc comments? If so, shouldn't we link to it or copy-paste?

zaino-fetch/src/jsonrpsee/response/common.rs
```diff
@@ -59,23 +67,29 @@ impl UnixTime {
     }
 }
 
+/// Duration in seconds.
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
 #[serde(transparent)]
 pub struct SecondsF64(pub f64);
+
 impl SecondsF64 {
+    /// Converts to a [`Duration`].
```
ala-mode, line 76: Are these doc comments now needed for some check? For example, several seem self-evident if one was to look at a documented function signature.

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+#[serde(untagged)]
+pub enum GetBlockHeader {
+    /// The verbose variant of the response. Returned when `verbose` is set to `true`.
+    Verbose(VerboseBlockHeader),
+
+    /// The compact variant of the response. Returned when `verbose` is set to `false`.
+    Compact(String),
+
+    /// An unknown response shape.
+    Unknown(serde_json::Value),
```
ala-mode, line 21: In what cases do we want a valid variant of this enum to be `Unknown`? My initial thought would be if there was something like this, it should be an error..

I see it used only once in testing...

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+        let out = serde_json::to_string(&block_header).unwrap();
+        assert_eq!(out, s);
+    }
+
+    #[test]
+    fn unknown_shape_falls_back_to_unknown_variant() {
+        let weird = r#"{ "weird": 1, "unexpected": ["a","b","c"] }"#;
+        let block_header: GetBlockHeader = serde_json::from_str(weird).unwrap();
+        match block_header {
+            GetBlockHeader::Unknown(v) => {
```
ala-mode, line 324: ... here. I think instead of this test being set how it is, it should not match a valid `GetBlockHeader` enum variant, and instead return an error.

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -541,6 +544,29 @@ impl JsonRpSeeConnector {
...
+    /// - hash: (string, required) The block hash
+    /// - verbose: (boolean, optional, default=true) true for a json object, false for the hex encoded data
+    ///
+    /// zcashd reference: [`getblockheader`](https://zcash.github.io/rpc/getblockheader.html)
+    /// method: post
+    /// tags: blockchain
+    pub async fn get_block_header(
+        &self,
+        hash: String,
+        verbose: bool,
```
ala-mode, line 561: so this is used every time in connector.rs
but it is optional for a request...

integration-tests/tests/fetch_service.rs
```diff
@@ -715,6 +715,53 @@ async fn fetch_service_get_block(validator: &ValidatorKind) {
...
+            .await
+            .unwrap();
+
+        let block_hash = match block {
+            GetBlock::Object(block) => block.hash(),
+            GetBlock::Raw(_) => panic!("Expected block object"),
+        };
+
+        let fetch_service_get_block_header = fetch_service_subscriber
+            .get_block_header(block_hash.to_string(), false)
```
ala-mode, line 739: This test only tests for the non-verbose mode. Can we build a test for verbose as well?

zaino-state/src/backends/state.rs
```diff
@@ -412,12 +413,12 @@ impl StateServiceSubscriber {
     ///
     /// This rpc is used by get_block(verbose), there is currently no
     /// plan to offer this RPC publicly.
```
ala-mode, line 415: Does this comment need a change, now?

zaino-state/src/backends/state.rs
```diff
@@ -412,12 +413,12 @@ impl StateServiceSubscriber {
...
-    async fn get_block_header(
+    async fn get_block_header_inner(
```
ala-mode, line 416: This helper function is only used in one place, and is in addition to several `get_block_header()` functions throughout the codebase. Is there a way we could iron these out without too much effort?

zaino-state/src/backends/state.rs
```diff
@@ -412,12 +413,12 @@ impl StateServiceSubscriber {
...
         state: &ReadStateService,
         network: &Network,
         hash_or_height: HashOrHeight,
         verbose: Option<bool>,
```
ala-mode, line 420: This currently is only used as `Option<true>`.

zaino-state/src/backends/state.rs
```diff
@@ -412,12 +413,12 @@ impl StateServiceSubscriber {
...
-    ) -> Result<GetBlockHeader, StateServiceError> {
+    ) -> Result<GetBlockHeaderZebra, StateServiceError> {
```
ala-mode, line 421: I might name this (though terrible it matches more closely the actual type in Zebra) something like `GetBlockHeaderZebraResponse`

How does this type relate to our own `GetBlockHeader` type?

zaino-state/src/backends/fetch.rs
```diff
@@ -389,6 +388,14 @@ impl ZcashIndexer for FetchServiceSubscriber {
             .try_into()?)
     }
 
+    async fn get_block_header(
```
ala-mode, line 391: I noticed a lack of doc comments here, while other rpcs have them.

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    /// The blockcommitments field of the requested block. Its interpretation changes
+    /// depending on the network and height.
+    #[serde(
+        with = "opthex",
+        rename = "blockcommitments",
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub block_commitments: Option<[u8; 32]>,
```
ala-mode, line 70: I don't see this in the online rpc specification, same as fields seen below...

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub final_sapling_root: Option<[u8; 32]>,
+
+    /// The root of the Orchard commitment tree after applying this block.
+    #[serde(
+        with = "opthex",
+        rename = "finalorchardroot",
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    final_orchard_root: Option<[u8; 32]>,
```
ala-mode: Here..

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    /// The block time of the requested block header in non-leap seconds since Jan 1 1970 GMT.
+    pub time: i64,
+
+    /// The nonce of the requested block header.
+    pub nonce: String,
+
+    /// The Equihash solution in the requested block header.
+    pub solution: String,
```
ala-mode, line 84: Here..

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    /// Floating point number that represents the difficulty limit for this block as a multiple
+    /// of the minimum difficulty for the network.
+    pub difficulty: f64,
+
+    /// Cumulative chain work for this block (hex).
+    ///
+    /// Present in zcashd, omitted by Zebra.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub chainwork: Option<String>,
```
ala-mode, line 97: Here..

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+
+    /// The next block hash after the requested block header.
+    #[serde(
+        rename = "nextblockhash",
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub next_block_hash: Option<String>,
+
+    /// Catch-all for any extra/undocumented fields.
```
ala-mode: and here... I am curious to know if these fields were discovered via manual use or testing? or inspecting the zcash source code? If it's the latter, I would like to see a link to the relevant files and lines.

ala-mode (changes requested):
Nice job overall!

A couple things.

I didn't see any links to zcashd source code like ZL was doing for some previous rpcs (e.g. `get_best_blockhash`). If this is still something we want to do, someone should add links to the relevant source code, and make sure the logic follows that pattern, or any differences are understood and documented. If we're not following that anymore, please disregard.

Today I found myself fretting about the multiple `fn get_block_header(...)` in fetch and state backends, state indexer and  jsonrpsee connector and service. This might be idiomatic to our project at this point, but I didn't love how this same name called itself on different levels, and then how there is also an `inner` function which is now 'legacy.'

I think we should at least understand but hopefully minimize this kind of name overloading.

Some other notes attached too.

I think I could probably pick up this PR and work with it in these ways, or I can leave these as requested changes and review again before long.

zaino-serve/src/rpc/jsonrpc/service.rs
dorianvp: Good catch. After looking at some other endpoints, it may be better to write proper docs per endpoints, mainly because a lot of them differ in how `zcashd` is implemented. Linking to a cpp implementation (or to the zcashd docs also) is not the best way to do so. For now, I'm putting doc comments at the top of the response types, and explicitly pointing when it differs from the zcashd impl.

ala-mode:
All tests passing locally with a few flakes (which I believe is a current good-standard).

zaino-fetch/src/jsonrpsee/response/block_header.rs
dorianvp: Having an unknown variant is useful when the consumer wants to be able to catch a response type with the wrong schema.

zaino-fetch/src/jsonrpsee/response/common.rs
dorianvp: Probably not... but I did not want to disable the clippy lint for that method

zaino-fetch/src/jsonrpsee/response/block_header.rs
dorianvp: Thing is, this will get consumed as a library too. It may make sense to include a fallback to allow the consumer how to react to it.

zaino-fetch/src/jsonrpsee/connector.rs
dorianvp: That's right. This connector is eventually called when doing passthrough.

zaino-state/src/backends/state.rs
dorianvp: I just noticed that this method was not created in this PR. We can follow up this PR with another one focused on these changes.

zaino-state/src/backends/state.rs
dorianvp: Same as above.

zaino-state/src/backends/fetch.rs
dorianvp: The trait definition already has doc comments. If you try getting the diagnostics, you'll see that it's inherited.

zaino-fetch/src/jsonrpsee/response/block_header.rs
dorianvp: Okay, I'll add a link to the zcashd code.

integration-tests/tests/fetch_service.rs
dorianvp: Sure. Although these tests are not ideal...

zaino-state/src/backends/state.rs
dorianvp: No. That is a different method, used only internally.

zaino-state/src/backends/state.rs
dorianvp: I think we should do this in a following PR. This PR did not introduce nor change that function.

dorianvp:
> All tests passing locally with a few flakes (which I believe is a current good-standard).

Extended the test to cover verbose responses & added a light reference to the zcashd implementation.

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+mod tests {
+    use std::str::FromStr;
+
+    use super::*;
+    use hex::FromHex;
+    use serde_json::{json, Value};
+    use zebra_chain::block;
+
+    /// Zcashd verbose response.
+    fn zcashd_verbose_json() -> &'static str {
```
ala-mode, line 144: Verbose returns in the zcash online rpc docs have 12 fields. 
https://zcash.github.io/rpc/getblockheader.html

Verbose return here has 16 fields here.

IMO it would be great, somewhere in zaino code, to document where these 4 extra fields are coming from, and how they are being produced as an output of zcashd.

zaino-fetch/src/jsonrpsee/response/block_header.rs
```diff
@@ -0,0 +1,432 @@
...
+          "chainwork": "0000000000000000000000000000000000000000000000000000000000001234",
+          "previousblockhash": "000000000053d2771290ff1b57181bd067ae0e55a367ba8ddee2d961ea27a14f",
+          "nextblockhash": "000000000053d2771290ff1b57181bd067ae0e55a367ba8ddee2d961ea27a14f",
+          "mediantime": 1700000500,
+          "nTx": 12
+        }"#
+    }
+
+    // Zebra verbose response
+    fn zebra_verbose_json() -> &'static str {
```
ala-mode, line 164: Verbose returns in the zcash online rpc docs have 12 fields.
https://zcash.github.io/rpc/getblockheader.html

Verbose return here has 14 fields here.

IMO it would be great, somewhere in zaino code, to document where these 2 extra fields are coming from, and how they are being produced as an output of zebrad.

zaino-fetch/src/jsonrpsee/response/block_header.rs
ala-mode: Having an extra field doesn't seem quite right to me.

IIUC we're getting rpc info from two full nodes (or some zaino-internal state derived from only one of these two full nodes) - I think we should only represent those fields in our struct that are presented as returns of those full nodes.

I think you have already set fields as options where we're getting different behavior from the two nodes, which I think is perfect - but I don't understand why we should have an `extra` placeholder. Is there any cases where we need it, or expect to need it very soon?

zaino-serve/src/rpc/jsonrpc/service.rs
```diff
@@ -218,6 +219,26 @@ pub trait ZcashIndexerRpc {
...
+    /// zcashd implementation [here](https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/rpc/blockchain.cpp#L668)
```
ala-mode, line 231:
```suggestion
    /// zcashd implementation [here](https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/rpc/blockchain.cpp#L668)
    /// The above function, if set to verbose, retrieves a pointer to a [CBlockIndex object](https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/chain.h#L197) for a particular block, formatted into JSON.
    /// [Here is the method used](https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/chain.h#L449).
    /// defined [here](https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/chain.h#L643) - but only has 8 fields there...
/// 37 fields [here] (https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/chain.h#L368)
/// then it's pulled back through [this function](https://github.com/zcash/zcash/blob/16ac743764a513e41dafb2cd79c2417c5bb41e81/src/rpc/blockchain.cpp#L107) which boils it down to 14 fields.
/// Is that correct? how does it all work in zcashd?
```
```

**Review by ala-mode — CHANGES_REQUESTED** [review]
Hi @dorianvp! Please see my attached comments.

I was not sure how things were really derived from the zcash source code versus what's present in your tests or in the struct.. If you think I might be onto something good with my concrete suggested changes, please do go ahead an merge them in and then make changes on top of them - or whatever is most fruitful.

File: zaino-fetch/src/jsonrpsee/response/block_header.rs (hunk 16 above)
Comment (dorianvp) [reply zaino-fetch/src/jsonrpsee/response/block_header.rs]: Yeah, I think it makes sense not to have the `extra` field. It's probably better to have the consumer match the `Unknown` variant, as it might imply a different meaning to the whole response.

File: zaino-fetch/src/jsonrpsee/response/block_header.rs (hunk 18 above)
Comment (dorianvp) [reply zaino-fetch/src/jsonrpsee/response/block_header.rs:164]: Those 2 extra fields were used for testing the `extras`. I'm now removing those...

File: zaino-fetch/src/jsonrpsee/response/block_header.rs (hunk 17 above)
Comment (dorianvp) [reply zaino-fetch/src/jsonrpsee/response/block_header.rs:144]: Fixed that instance. Just ended up removing the extra fields

File: zaino-serve/src/rpc/jsonrpc/service.rs (hunk 19 above)
Comment (dorianvp) [reply zaino-serve/src/rpc/jsonrpc/service.rs:231]: I ended up specifying which field was not documented but present

**Comment by dorianvp** [comment]
Comments addressed! Looks like I was adding a field that was present in `getblock` but not `getblockheader`. Good catch @ala-mode!

**Comment by zancas** [comment]
@ala-mode do you have any remaining concerns?

**Comment by ala-mode** [comment]
> @ala-mode do you have any remaining concerns?

I didn't get to re-review, but I am sitting down to do it right now.

File: zaino-fetch/src/jsonrpsee/response/block_header.rs [hunk 20]
```diff
@@ -0,0 +1,396 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub chainwork: Option<String>,
+
+    /// The previous block hash of the requested block header.
+    #[serde(
+        rename = "previousblockhash",
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub previous_block_hash: Option<String>,
```
ala-mode, line 105: Both full nodes return this field (in the verbose variant), but I do see that in a test in this file below there is a case where there it is not part of a raw string and is expected to be `None` because it is like a genesis block.

Have you observed this behavior, for example, in regtest, where there is no previous block, and the entire field is omitted?

Which full node has a response like this? Do both full nodes respond the same way?

If this is observed, using an `Option` is great! If it's not observed directly, we should probably look to see what the full nodes do when there's only 1 block.

integration-tests/tests/json_server.rs
```diff
@@ -789,6 +791,45 @@ mod zcashd {
...
         }
 
+        #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+        async fn get_block_header() {
+            let (
+                test_manager,
+                _zcashd_service,
+                zcashd_subscriber,
+                _zaino_service,
+                zaino_subscriber,
```
ala-mode, line 801: Below in a similar test this is called `state_service_sbcriber`
```
 let (
                test_manager,
                _fetch_service,
                fetch_service_subscriber,
                _state_service,
                state_service_subscriber,
            ) = create_test_manager_and_services(...
```

 I think in this case, we have two fetch service subscribers, so I might change the name of `zaino_subscriber` and `zcashd_subscriber` to include the word `fetch` or something similar for clarity.

ala-mode (approved):
I approve, but I couldn't pass all tests so didn't merge, see below. A couple notes:

I noticed the `Unknown` variant in `pub enum GetBlockHeader` stayed in - It's not worth blocking over, but I still don't love this.. even as a library, the call is going to query full nodes that have responses described in the variants and types you (nicely!) created. Are there other cases?

(Even if we query another zaino hypothetically, it will have ingested the same info in a `VerboseBlockHeader` or String.)

Other than this I attached a couple comments which are also not blocking, but could point toward a future issue.

Running tests now, I'll merge if they all pass for me locally (right now, allowing flakes)

ala-mode:
I had one test locally that was very slow:
`integration-tests::wallet_to_validator zebrad::fetch_service::monitor_unverified_mempool`

because I didn't complete this test, I didn't merge.

zaino-fetch/src/jsonrpsee/response/block_header.rs
dorianvp: Both zcashd and zebra omit this field if asked about the genesis block.

integration-tests/tests/json_server.rs
dorianvp: If you look at the module-level doc-comment, you'll see that it explicitly says all tests from this module rely on `FetchService`.

ala-mode:
Just for referece that test passed for me (fairly quickly) on the 3rd try.

PR #609, opened 2025-10-17
Cargo Hack Check in CI

fluidvanadium:
## Motivation

Ensure features work, but not spending too much time in CI.

## Solution

cargo hack `check´ is much faster than `build´

### Tests

Github CI

### Follow-up Work

Deal with flaky tests!

### PR Checklist

- [x] The solution is tested.


zancas (approved):
WOOT!   Assuming those flows are good!

PR #616, opened 2025-10-20
add "quick" cargo nextest profile

zancas:
## Motivation

When a developer is making changes to the code they need to be able to rapidly iterate.   That is they need to be able to change the source code, and then run a series of tests that demonstrate that their change has not caused an unexpected and incorrect behavior.

The "test suite" provides this assurance, but there are few unit-tests and the integration tests are impractically unreliable and slow.

We are adding unit-tests to better provide the necessary assurances quickly.

Until our unit-test suite is sufficiently comprehensive (one metric being code-coverage), a stop gap is to run a fast-reliable sub-set of the integration tests during development, and depend on CI to run the comprehensive set on a slower cadence.

## Solution

Add a non-invasive (cargo nextest profile) flow that allows rapid test-cycles during development:

_cargo nextest run --profile quick_

### Follow-up Work

By analyzing where this subset of tests is not providing feedback (e.g. by looking at coverage reports) a developer can plan for high-value unit-tests, that cover currently untested (by this set) code.   Such tests would be reliable and fast, and complement the tests that run under this profile.

### PR Checklist


.config/nextest.toml
```diff
@@ -15,3 +15,91 @@ default-filter = """
 
 [profile.ci.junit]  # this can be some other profile, too
 path = "junit.xml"
+
+[profile.rapid_superficial]
```
pacu: I'd tweak the name for a less quirky one like "quick_tests" or profile.quick

pacu: Also would actually document what's the criteria for adding a newly created test to this test suite.

pacu (dismissed):

.config/nextest.toml
zancas: OK..  profile `quick`.   Added criteria for inclusion.

pacu (approved):

pacu (approved):

PR #618, opened 2025-10-21
Rename ZainoVersionedSerialise to ZainoVersionSerde

nachog00:
## Motivation

`ZainoVersionedSerialise` reads and writes to db, so it deserialises too.

## Solution

Rename to `ZainoVersionSerde` seems appropiate

### Tests

Not needed, clean rename

### DIFF
--- a/zaino-state/src/chain_index/finalised_state/db/v0.rs
+++ b/zaino-state/src/chain_index/finalised_state/db/v0.rs
@@ -2,7 +2,7 @@
 //!
 //! WARNING: This is a legacy development database and should not be used in production environments.
 //!
-//! NOTE: This database version was implemented before zaino's `ZainoVersionedSerded` was defined,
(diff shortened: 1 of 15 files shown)


zaino-state/src/chain_index/finalised_state/db/v0.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! WARNING: This is a legacy development database and should not be used in production environments.
 //!
-//! NOTE: This database version was implemented before zaino's `ZainoVersionedSerialised` was defined,
+//! NOTE: This database version was implemented before zaino's `ZainoVersionedSerded` was defined,
```
zancas: Extra "d" on the end.

zancas (dismissed):
There's an extra "d".

zancas (approved):

Issue #620, opened 2025-10-23
Flakes in `chain_query_interface` tests


dorianvp:
A recent change made some flakes to appear more frequently. Upon further investigation (see #610) it turns out that, after removing the sleep and implementing a proper polling mechanism, the NFS drains when 100+ blocks are mined, and this causes a bunch of tests to fail if they hit this window.

See [this test](https://github.com/zingolabs/zaino/pull/610/files#diff-c36929304b43949616976abfd2f4c14280290ac5f5ee7fb9381afdfa4aa8f924R615) for a repro of the issue. The fix should make that test pass.

nachog00:
is this fixed? @dorianvp @zancas

dorianvp:
> is this fixed? [@dorianvp](https://github.com/dorianvp) [@zancas](https://github.com/zancas)

This issue and the linked PR mostly describe the weird behavior around the NFS. My understanding is that a workaround was implemented in #638... We can either rename this issue to refer to that more explicitly, or mark this as completed and open another one targeting the core issue.

nachog00:
would you expect #610 to pass if it got #638 merged into it?

nachog00:
> > is this fixed? [@dorianvp](https://github.com/dorianvp) [@zancas](https://github.com/zancas)
> 
> This issue and the linked PR mostly describe the weird behavior around the NFS. My understanding is that a workaround was implemented in [#638](https://github.com/zingolabs/zaino/pull/638)... We can either rename this issue to refer to that more explicitly, or mark this as completed and open another one targeting the core issue.

The second option seems best.

PR #622, opened 2025-10-24
use SOURCE_DATE_EPOCH for build date if present

antonleviathan:
This is to help move `zaino` towards deterministic builds. I would like to eventually move to work on making zaino full source bootstrapped and deterministic using [StageX](https://codeberg.org/stagex/stagex).

## Motivation

I am currently working on making [(z)wallet](https://github.com/zcash/wallet/pull/301) deterministic and `zaino` was causing non-determinism issues

## Solution

I updated the code to use the [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) if  it's present, in order to be able to set time to a static value. The fallback is the native rust date call which is more secure than the previously used Command. Overall there is no real need to inject the datetime into the binary but I didn't want to assume too much.

### Tests

I tried to build `zallet` with this updated code.

N/A

### Follow-up Work

N/A

### PR Checklist

- [ ] The solution is tested.


nachog00 (dismissed):
Merging this if it passes CI

fluidvanadium:
Hey, this is great code, but its hitting a few lints. `cargo fmt`  should fix it.

RyanSquared:
I am on Cargo 1.90 and I do not observe any difference when running with `cargo fmt`, and do not see any error when using `cargo fmt --all -- --check`. I also see no error message when using `--verbose`.

zancas (dismissed):

fluidvanadium:
How strange! I am getting a difference! Is it possible cargo isn't hitting the right package?

```
$cargo fmt --all -- --check
Diff in /home/cupress/ZINGO/zaino/zaino-state/build.rs:35:
     // at https://reproducible-builds.org/docs/source-date-epoch/
     let build_date = match env::var("SOURCE_DATE_EPOCH") {
         Ok(s) => s.trim().to_string(),
-        Err(_) => {
-            SystemTime::now()
-                .duration_since(UNIX_EPOCH)
-                .unwrap()
-                .as_secs()
-                .to_string()
-        }
+        Err(_) => SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_secs()
+            .to_string(),
     };
-
+
     println!("cargo:rustc-env=BUILD_DATE={}", build_date);

     // Set the build user
/home/cupress/ZINGO/zaino RC: 1
29;21:31_23 || (HEAD detached at anton/build-date-determinism)[ ] f594bfc4
$cargo --version
cargo 1.90.0 (840b83a10 2025-07-30)

```

> I am on Cargo 1.90 and I do not observe any difference when running with `cargo fmt`, and do not see any error when using `cargo fmt --all -- --check`. I also see no error message when using `--verbose`.

fluidvanadium:
You can merge this if it helps https://github.com/fluidvanadium/zaino/tree/build-date-determinism

fluidvanadium:
Try running cargo fmt -p zaino-state

RyanSquared:
Interesting - if I purposefully bodge the formatting to something that's _really_ wrong (indent 2x more), it finally reformats it. @antonleviathan throw a bunch of tabs in there, and `cargo fmt` should work.

antonleviathan:
There is one test failing but I'm not sure what this is about...

dorianvp (approved):

zancas:
We expect these tests to fail in CI until this issue is resolved:

https://github.com/zingolabs/zaino/issues/620

antonleviathan:
Any idea ~when this will make it into a release?

str4d:
We don't need this in a release to be able to use it in the Zallet alpha; we'll just update our existing pin. We obviously will need a release before Zallet can be published on crates.io.

PR #628, opened 2025-10-29
prefer traits from infrastructure to begin upgrading test code

zancas:
## Motivation

The test code needs to be DRYed.  By leveraging the modern traits provided in infrastructure significant boilerplate can be removed from the test code, including at least some 

## Solution
Use infra traits in the new version of infrastructure.  This code will need to land before this solution will work:  https://github.com/zingolabs/infrastructure/pull/152

### Tests

All extant tests

### Follow-up Work

More test code (that was originally copied and pasted) should be deletable.

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9087,7 +9087,7 @@ dependencies = [
 [[package]]
 name = "zcash_local_net"
 version = "0.1.0"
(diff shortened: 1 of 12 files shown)


nachog00:
Why is this still draft?? @zancas

PR #629, opened 2025-10-29
docs: mark `StateService` and `FetchService` types as deprecated

dorianvp:
## Motivation

`_Service` structs are going to be replaced. They should be marked as deprecated.

## Solution

Add attributes and `#[allow(deprecated)]` to keep track of where they're being used (and so that CI does not fail).

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zaino-state/src/backends/fetch.rs
+++ b/zaino-state/src/backends/fetch.rs
@@ -69,7 +69,7 @@ use crate::{
 ///       ServiceSubscribers are used to create separate chain fetch processes while allowing central state processes to be managed in a single place.
 ///       If we want the ability to clone Service all JoinHandle's should be converted to Arc\<JoinHandle\>.
 #[derive(Debug)]
-#[deprecated = "Will be eventually replaced by `BlockchainSource"]
+#[deprecated = "Will be eventually replaced by `BlockchainSource`"]
 pub struct FetchService {
     /// JsonRPC Client.
     fetcher: JsonRpSeeConnector,
--- a/zaino-state/src/backends/fetch.rs
+++ b/zaino-state/src/backends/fetch.rs
@@ -40,6 +40,7 @@ use zaino_proto::proto::{
     },
 };
 
+#[allow(deprecated)]
 use crate::{
     chain_index::{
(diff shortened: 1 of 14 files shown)


zaino-state/src/backends/fetch.rs
```diff
@@ -68,20 +69,27 @@ use crate::{
 ///       ServiceSubscribers are used to create separate chain fetch processes while allowing central state processes to be managed in a single place.
 ///       If we want the ability to clone Service all JoinHandle's should be converted to Arc\<JoinHandle\>.
 #[derive(Debug)]
+#[deprecated = "Will be eventually replaced by `BlockchainSource"]
```
dorianvp:
```suggestion
#[deprecated = "Will be eventually replaced by `BlockchainSource`"]
```

zaino-state/src/config.rs
```diff
@@ -80,6 +83,7 @@ impl StateServiceConfig {
 
 /// Holds config data for [crate::FetchService].
 #[derive(Debug, Clone)]
+#[deprecated]
```
zancas, line 86: Why is this deprecated?

zaino-state/src/error.rs
```diff
@@ -1,11 +1,13 @@
 //! Holds error types for Zaino-state.
+#![allow(deprecated)]
 
 use crate::BlockHash;
 
 use std::{any::type_name, fmt::Display};
 
 use zaino_fetch::jsonrpsee::connector::RpcRequestError;
 
+#[allow(deprecated)]
```
zancas, line 9: Why is this necessary if the whole mod is already `allow(deprecated` ?

zaino-state/src/error.rs
```diff
@@ -1,11 +1,13 @@
 //! Holds error types for Zaino-state.
+#![allow(deprecated)]
```
dorianvp:
```suggestion
```

zancas (approved):

zaino-state/src/config.rs
zancas: Can you add a "message"?

PR #632, opened 2025-10-30
feat(rpc): add `getaddressdeltas` endpoint

### LINKED ISSUE #197: 1.1: Add Zcash RPC `getaddressdeltas`
Spec: https://zcash.github.io/rpc/getaddressdeltas.html

dorianvp:
## Motivation

Closes #197.

## Solution

Implement `getaddressdeltas`.

### PR Checklist

- [x] The solution is tested.


zancas:
Hi @nachog00 we opened a new PR, to more accurately reflect the flow.  'looking forward to your review!

integration-tests/tests/state_service.rs
```diff
@@ -1034,7 +1035,71 @@ async fn state_service_get_address_utxos_testnet() {
...
+        state_service_subscriber
+            .get_address_deltas(simple_request)
+            .await
+    )
+    .unwrap();
+
+    assert_eq!(fetch_service_simple_deltas, state_service_simple_deltas);
+
+    // Test response with chain info
+    let chain_info_request =
```
nachog00: Not sure what this variable name is supposed to mean...? it looks like the value is a Param type, but i might be missing something?

zaino-fetch/src/jsonrpsee/connector.rs
```diff
@@ -399,6 +402,29 @@ impl JsonRpSeeConnector {
         Ok(request_builder)
     }
 
+    /// Returns all changes for an address.
+    ///
+    /// Returns information about all changes to the given transparent addresses within the given (inclusive)
```
nachog00: given "block range"?

integration-tests/tests/state_service.rs
dorianvp: Fixed

zaino-state/src/backends/fetch.rs
```diff
@@ -218,6 +219,28 @@ impl FetchServiceSubscriber {
 impl ZcashIndexer for FetchServiceSubscriber {
     type Error = FetchServiceError;
 
+    /// Returns all changes for an address.
+    ///
+    /// Returns information about all changes to the given transparent addresses within the given (inclusive)
+    ///
+    /// block height range, default is the full blockchain.
+    /// If start or end are not specified, they default to zero.
+    /// If start is greater than the latest block height, it's interpreted as that height.
```
nachog00: this applies to start, tho it seems like it makes more sense for end... ? also, if start >= lh, and end>=start, then start-end=lh right? seems like this collapses into an edge case worth of an error?

zaino-state/src/backends/state.rs
```diff
@@ -875,6 +878,82 @@ impl StateServiceSubscriber {
...
+            .read_state_service
+            .clone()
+            .ready()
+            .await?
+            .call(ReadRequest::BlockHeader(hash_or_height))
+            .await?;
+
+        match response {
+            ReadResponse::BlockHeader { hash, .. } => Ok(BlockInfo::new(
+                hex::encode(hash.bytes_in_display_order()),
```
nachog00, line 948: This line reminds me of the problem we have with txids being reversed in some places... we probably want either our won Hash type or the BlockInfo type to implement dedicated methods to construct from "order susceptible" sources, so that we don't depend on a line on an rpc endpoint for this critical bit.

I wouldn't block on this for this PR, but maybe leave a `Todo!` mentioning the problem?

zaino-fetch/src/jsonrpsee/response/address_deltas.rs
```diff
@@ -0,0 +1,523 @@
...
+
+use serde::{Deserialize, Serialize};
+use zebra_rpc::client::{Input, Output, TransactionObject};
+
+use crate::jsonrpsee::connector::ResponseToError;
+
+/// Request parameters for the `getaddressdeltas` RPC method.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum GetAddressDeltasParams {
```
nachog00, line 11: It seems like this type supports:
- multiple addresses + height range + chaininfo boolean flag
- single address + Implicit/no range (?) + no chaininfo 

what about: single address + range + chaininfo? i suppose that's covered by `filtered` with a single item vector?

zaino-fetch/src/jsonrpsee/response/address_deltas.rs
```diff
@@ -0,0 +1,523 @@
...
+    }
+}
+
+/// Response to a `getaddressdeltas` RPC request.
+///
+/// This enum supports both simple array responses and extended responses with chain info.
+/// The format depends on the `chaininfo` parameter in the request.
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(untagged)]
+pub enum GetAddressDeltasResponse {
```
nachog00, line 57: Love this type right here. Absolutely the way to go by typing the different answer variants to offload the logic getting the data from properly modelling this.

zaino-fetch/src/jsonrpsee/response/address_deltas.rs
dorianvp: Yeah, that's covered by a one-item vector.

zaino-state/src/backends/state.rs
dorianvp: I would instead open an issue so that it doesn't get buried with a lot of in-code nitpicks.

zaino-state/src/backends/fetch.rs
dorianvp: This comment was taken from the zcashd 6.2 docs. Both `start` and `end` are clamped down to the tip it they're above. In the case where both are `> tip`, the range ends up being `[tip, tip]`.

zaino-state/src/backends/state.rs
```diff
@@ -897,6 +976,75 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+
+        let transactions = self
+            .get_taddress_txs(addresses.clone(), start.0, end.0, true)
+            .await?;
+
+        let mut deltas =
+            GetAddressDeltasResponse::process_transactions_to_deltas(&transactions, &addresses);
+
+        // zcashd-like ordering
+        deltas.sort_by_key(|d| (d.height, d.block_index.unwrap_or(u32::MAX), d.index));
```
nachog00: I seems like this is sorting twice, right? we probably only need one of them

zaino-fetch/src/jsonrpsee/response/address_deltas.rs
```diff
@@ -0,0 +1,536 @@
...
+                            None,
+                        )
+                    }
+                });
+
+                input_deltas.chain(output_deltas)
+            })
+            .collect();
+        // zcashd-like ordering: (height ASC, blockindex ASC, index ASC)
+        deltas.sort_by_key(|d| (d.height, d.block_index.unwrap_or(u32::MAX), d.index));
```
nachog00, line 122: First sort

zaino-state/src/backends/state.rs
```diff
@@ -897,6 +976,75 @@ impl ZcashIndexer for StateServiceSubscriber {
...
+
+        if chain_info && start > Height(0) && end > Height(0) {
```
nachog00, line 1027: `chain_info=true` only means that we return the context on the response, right? shouldn't the height checks be performed whatever this value is?

zaino-fetch/src/jsonrpsee/response/address_deltas.rs
```diff
@@ -0,0 +1,536 @@
...
+impl GetAddressDeltasResponse {
+    /// Processes transaction objects into address deltas for specific addresses.
+    /// This is a pure function that can be easily unit tested.
+    pub fn process_transactions_to_deltas(
+        transactions: &[Box<TransactionObject>],
+        target_addresses: &[String],
+    ) -> Vec<AddressDelta> {
+        let mut deltas: Vec<AddressDelta> = transactions
+            .iter()
+            .filter(|tx| tx.height().unwrap_or(0) > 0) // Filtering out coinbase
```
nachog00: IIUC, this line filters both:
- height-less txs
- txs with height=0

Not sure which of them maps to coinbase tx? maybe it's worth it to split into 2 separate `.filter` calls, one per condition, to be clearer about what's being filtered

dorianvp: Actuall, that comment is outdated, good catch

zaino-state/src/backends/state.rs
dorianvp: According to the original implementation, whether `chainInfo` is present or not determines if the ranges also get returned. Note that it's always clamped beforehand.

nachog00: ah ok, i didn't realise that this conditional block was just to gather the start and end info 👌🏼

zancas:
@nachog00 what remains to be done to merge this PR?

nachog00:
I expanded on the tests by:
- having them make assertions about the contents of the response on top of the response variant
- splitting the sub-tests into dedicated functions while keeping a single test function
- splitting the chain setup logic into it's own function
- extracting constants 
- moving all addressdeltas inherent functions into it's own module
- moving the module into it's own file

nachog00:
tests are passing for me locally

dorianvp:
I see deprecated usages... let me fix that!

zancas:
Is this mergeable?

dorianvp:
> Is this mergeable?

Yeah, needs an approval.

nachog00 (approved):

PR #633, opened 2025-10-30
fix bug in json server logs

Oscar-Pepper:
### PR Checklist

- [ ] The solution is tested.


dorianvp (approved):

fluidvanadium (commented):
Having a more specific symbol name would have prevented this bug. We cannot change the symbol name right now cause it could debase other PRs. But we could add a toDO comment about changing `server` to `grpc_server` and `service` to `indexer_service`?

PR #636, opened 2025-10-31
fix(`zainod`): remove redundant `version` command

dorianvp:
## Motivation

We're deriving `version` AND defining the version command. This is redundant.

## Solution

Use only the derive macro provided.

### Tests

Manually tested with `cargo run -- --version` and `cargo run -- -V`.

### PR Checklist

- [x] The solution is tested.

### DIFF
--- a/zainod/src/main.rs
+++ b/zainod/src/main.rs
@@ -1,6 +1,6 @@
 //! Zingo-Indexer daemon
 
-use clap::{CommandFactory, Parser};
+use clap::Parser;
 use std::path::PathBuf;
 use tracing::{error, info};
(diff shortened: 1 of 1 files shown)


fluidvanadium:
How do we test this?

dorianvp:
> How do we test this?

For now a manual test should be enough. Later on we can include something like [assert_cmd](https://github.com/assert-rs/assert_cmd).

zancas (approved):
HOOOOORAY FOR CODE DELETION!!!

zancas:
I see CI fails...

PR #610, opened 2025-10-20
fix(chain_query_interface): flakes

dorianvp:
## Motivation

We're using the wrong polling strategies.

### Specifications & References

https://docs.rs/zebra-rpc/latest/src/zebra_rpc/sync.rs.html

### PR Checklist

- [ ] The PR name is suitable for the release notes.
- [ ] The solution is tested.
- [ ] The documentation is up to date.

PR #638, opened 2025-10-31
replace block generation delay with polling

Oscar-Pepper:
## Motivation

Chain index integration was held up by flaky and slow tests and increased failure rates due to underlying issues with test framework interacting with chain index. Fixing the test framework to use correct sleeping and to poll when generating blokcs instead of using fixed delays not only improves dev efficiency but fixes many failing tests in the chain integration (#596)

## Solution

- remove OS sleeps
- replace fixed delays with polling
- return service subscriber in the test manager to reduce backend duplication and out-of-sync bugs

### PR Checklist

- [ ] The solution is tested.


dorianvp:
Some issues we're seeing here are highly related to #620, and #610.

zaino-testutils/src/lib.rs
```diff
@@ -538,23 +563,132 @@ impl TestManager {
...
-        test_manager.generate_blocks_with_delay(1).await;
+        // NOTE: if this is removed when zebra fixes this issue we must replace with a generate_block_and_poll(0) when
+        // zaino is enabled to ensure its ready and not still syncing
+        if enable_zaino {
+            test_manager.generate_blocks_and_poll(1).await;
+        } else {
+            test_manager.local_net.generate_blocks(1).await.unwrap();
+        }
+
+        // FIXME: zaino's status can still be syncing instead of ready at this point
```
nachog00, line 581: IIUC there's a chance that this is caused by the subscriber status being a combination of sub-components statuses right?

Oscar-Pepper: i created an issue

nachog00 (approved):

PR #668, opened 2025-11-08
Propose a simple zainod.

zancas:
I started by finding bugs in the config logic...


zainod/src/config.rs
nachog00, line 1: if we are providing defaults, we probably should do so through implementing `Default` on the config types and using serde directives. We can grab the whole `ZainodConfig::default()` and serialise it into toml. us supporting that behaviour keeps the truth closer to code and also has the added benefit of making life easier for those testing with zainod, since they can easily construct configs from code.

Issue #502, opened 2025-09-02
God Object - IndexerConfig mixes concerns across multiple domains


nachog00:
# God Object - IndexerConfig mixes concerns across multiple domains

**Labels:** `configs`, `tech-debt`, `architecture`

## Problem

`IndexerConfig` is a **God Object** that violates the Single Responsibility Principle by mixing configuration for entirely different subsystems in a single struct.

### Current Problematic Code

```rust
pub struct IndexerConfig {
    // Backend selection
    pub backend: zaino_state::BackendType,
    
    // JSON-RPC server settings
    pub enable_json_server: bool,
    pub json_rpc_listen_address: SocketAddr,
    pub enable_cookie_auth: bool,
    pub cookie_dir: Option<PathBuf>,
    
    // gRPC server settings  
    pub grpc_listen_address: SocketAddr,
    pub grpc_tls: bool,
    pub tls_cert_path: Option<String>,
    pub tls_key_path: Option<String>,
    
    // Validator connection settings
    pub validator_listen_address: SocketAddr,
    pub validator_grpc_listen_address: SocketAddr,
    pub validator_cookie_auth: bool,
    pub validator_cookie_path: Option<String>,
    pub validator_user: Option<String>,
    pub validator_password: Option<String>,
    
    // Cache settings
    pub map_capacity: Option<usize>,
    pub map_shard_amount: Option<usize>,
    
    // Database settings
    pub zaino_db_path: PathBuf,
    pub zebra_db_path: PathBuf,
    pub db_size: Option<usize>,
    
    // Network and debug
    pub network: String,
    pub no_sync: bool,
    pub no_db: bool,
    pub slow_sync: bool,
}
```

### Impact

- **Unclear Responsibilities**: What is IndexerConfig actually responsible for?
- **Change Amplification**: Adding a new server option affects the "main" config
- **Documentation Overload**: Single struct needs to document 6+ different concepts
- **Testing Complexity**: Can't test server config without creating entire IndexerConfig
- **Poor Cohesion**: Fields that change together are scattered
- **Hidden Dependencies**: Which fields are used together isn't clear

## Solution

Split into focused, domain-specific configuration objects:

```rust
// Top-level config just orchestrates
pub struct IndexerConfig {
    pub network: Network,
    pub backend: BackendConfig,
    pub server: ServerConfig,
    pub storage: StorageConfig,
    pub debug: DebugConfig,
}

// Server owns its configuration
pub struct ServerConfig {
[...]

PR #669, opened 2025-11-09
zaino-common::config module

nachog00:
## Motivation

`zaino-common` has a few modules which are inherently config related. since the crate isn't called `zaino-config`, we should have a dedicated config module. This anticipates we adding a bunch of primitives that are meant to be used to conform the broader config sections.

what i mean with this terminology is the base structure of `zainodconfig` the config struct that maps to zainod's config file.
sections: server, backend, storage, etc.
primitives: CookieAuth, PassAuth, TlsConfig, etc

## Solution

We create the `zaino-common::config` module and add `use` reexports to not break current usage of the moved types, keeping the refactor PR clean and minimal. Those usages can be pruned incrementaly over time with more specific imports.

### Tests

None needed, clean code-moving only refactor.

### Follow-up Work

- The extraction of the mentioned primitives.
- The creation of the `backend` section with specific variants per possible backend (see comments on #502 for reference).
- The creation and impl of some traits like `ValidatedConfig` to standarise config sections and primitives self validating when possible.

### PR Checklist

- [X ] The PR name is suitable for the release notes.


AloeareV (approved):
So far, this is only thing in zaino-common. Do we know of anything else that wants to be in zaino-common? If so, I approve of this PR. If not...the better move may be to rename this crate to zaino-config.

nachog00:
> So far, this is only thing in zaino-common. Do we know of anything else that wants to be in zaino-common? If so, I approve of this PR. If not...the better move may be to rename this crate to zaino-config.

`common` is, conceptually, upstream of everything else on the zaino workspace, config is just an instance of something that belongs there. That alone lead me to belive that keeping the crate and grouping the config stuff is a good move.
Also, at some point on the config work I have been doing over the last few months (tho it's not 100% fresh in my mind), I remember conceptualising something else that would 100% belong on this crate but not on config.
That being said, keeping this as is will make the config work likely easier, but if it happens to no be the case and a long time goes by with config being the single module on `zaino-common`, I compromise to simplify the crate myself.

PR #571, opened 2025-09-22
Refurbish indexerconfig

ala-mode:
fixes #500 #506 relevant. #502 and #505 maybe also be partially solved.

PR #670, opened 2025-11-12
remove surplus if control flow

ala-mode:
Removing the superfluous `if` control flow in zainod/src/config.

Some of this nested if statement (inherited but not refined with my #571 PR) was unreachable, which @zancas pointed out should have probably been detected by cargo's various checks.

I made a minimal recreation on rust playground, but instead of posting that directly, I'll share [this open issue](https://github.com/rust-lang/rust-clippy/issues/5218) which seems to be related (ie, this is a known issue to the `cargo` project)


ala-mode:
Without demonstrably exhausting the ways this could be called, [here's the example on the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e98f733a93085dc1eb31cea4c861e586).

dorianvp:
I think we should be able to catch these dead code blocks by using something like:

```bash
cargo clippy -- -W clippy::all -W clippy::pedantic -W clippy::nursery
```

If this works, we should add it to a clippy file somewhere so that it applies for everyone.

ala-mode:
> I think we should be able to catch these dead code blocks by using something like:
> 
> ```shell
> cargo clippy -- -W clippy::all -W clippy::pedantic -W clippy::nursery
> ```
> 
> If this works, we should add it to a clippy file somewhere so that it applies for everyone.

I tested this on `dev`, and it underlined a LOT of things in Zaino... but it didn't complain about the triple if with unreachable code.

dorianvp:
> > I think we should be able to catch these dead code blocks by using something like:
> > ```shell
> > cargo clippy -- -W clippy::all -W clippy::pedantic -W clippy::nursery
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If this works, we should add it to a clippy file somewhere so that it applies for everyone.
> 
> I tested this on `dev`, and it underlined a LOT of things in Zaino... but it didn't complain about the triple if with unreachable code.

Oh, hmm. Then ignore my comment

zancas:
Hi @ala-mode there are CI fails.

zancas (approved):

PR #671, opened 2025-11-12
propose chain_index::Queryable as the name of the capability

### LINKED ISSUE #719: chain_index::ChainIndex is a repetitive name that doesn't hint at the item's kind
"ChainIndex" as a name doesn't hint what kind of thing it is.  In fact it is a trait, and as such should have a name that hints at the behavior implementors must have.

Additionally the context of the definition is inside a module of the same name.  This means that the module and the item it contains repeat the same information decreasing readability.

Here's a snapshot of the module structure on dev:

```Bash
cargo modules structure -p zaino-state --focus-on zaino_state::chain_index --no-fns --max-depth 1

crate zaino_state
└── mod chain_index: pub
    ├── trait ChainIndex: pub
    ├── struct NodeBackedChainIndex: pub
    ├── struct NodeBackedChainIndexSubscriber: pub
    ├── trait NonFinalizedSnapshot: pub
    ├── mod encoding: pub
    ├── mod finalised_state: pub
    ├── mod mempool: pub
    ├── mod non_finalised_state: pub
    ├── mod source: pub
    └── mod types: pub
```

zancas:
Fixes: #719 

Renaming the ChainIndex to Queryable has a few benefits.

(1) It hints to a reader that the thing is not a type.
(2) It conforms to a notion of a Capability that a thing can be made to offer by implementing it.
(3) It frees the name "ChainIndex" to be used less ambiguously on related types.
(4) It reduces a redundancy in the naming scheme when the mod name is also accounted for.

At least as a thought experiment, @nachog00 reconsidered related trait and type names and scopes and opened this PR as a proposal for an update to the naming scheme.


zaino-state/src/chain_index.rs
```diff
@@ -162,10 +295,10 @@ mod tests;
 /// If you were previously using `FetchService::spawn()` or `StateService::spawn()`:
 /// 1. Extract the relevant fields from your service config into a `BlockCacheConfig`
 /// 2. Create the appropriate `ValidatorConnector` variant (State or Fetch)
-/// 3. Call `NodeBackedChainIndex::new(source, config).await`
-pub trait ChainIndex {
+/// 3. Call `Index::new(source, config).await`
+pub trait Query {
```
AloeareV: A query is, intuitively, the thing that you pass as an argument to this type's methods and get back a response. Calling this thing itself a `Query` is misleading.

zaino-state/src/chain_index.rs
```diff
@@ -419,23 +419,23 @@ impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
...
         };
-        chain_index.sync_loop_handle = Some(chain_index.start_sync_loop());
+        index.sync_loop_handle = Some(index.start_sync_loop());
 
-        Ok(chain_index)
+        Ok(index)
     }
 
-    /// Creates a [`NodeBackedChainIndexSubscriber`] from self,
+    /// Creates a [`Subscriber`] from self,
```
AloeareV, line 436: A lot of information has been removed from this type name. Why? This name is now less informative to me.

zaino-state/src/chain_index/snapshot.rs
```diff
@@ -0,0 +1,61 @@
...
+pub mod non_finalised_state;
+use non_finalised_state::NonfinalizedBlockCacheSnapshot;
+
+use crate::chain_index::snapshot::non_finalised_state::BestTip;
+use crate::chain_index::types;
+use crate::IndexedBlock;
+
+use std::sync::Arc;
+
+impl<T> NonFinalized for Arc<T>
```
AloeareV: I think if we're to remove words from this trait name (again, less information? Why?) then it's more useful to call this thing a 'Snapshot'. The fact that it is of the non-finalized state is useful for understanding the underlying mechanics of the snapshot process, but it's irrelevant to the consumer.

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV, line 1: Discussed on call...The non-finalized state is used by the chain index a lot. If there's a module/submodule relation, than snapshot should be inside non_finalized_state and not the other way around...however, the snapshot being of the non-finalized is useful information for a zaino-developer...but at the public API level this is irrelevant, the snapshot is a component of the chain index. I think the `snapshot` and `non_finalized_state` mods should both be in the `chain_index` mod.

zaino-state/src/lib.rs
```diff
@@ -31,13 +31,14 @@ pub use backends::{
 pub mod chain_index;
 
 // Core ChainIndex trait and implementations
-pub use chain_index::{ChainIndex, NodeBackedChainIndex, NodeBackedChainIndexSubscriber};
+pub use chain_index::Query as ChainIndexQuery;
```
AloeareV: I said something about this in a previous comment, but to elaborate:

The trait `Read` means 'something that is readable'. `Read` as a trait is intuitively read as something readable...as there's no noun 'read' that you use in order to read (verb). 

Query, however, *is* a noun, meaning something like 'the request you make of a database in order to learn information from it', so when I see something called a `Query` I don't automatically think about it as something that can be queried...rather, the more obvious way to read this is that the thing called `Query` is, in fact, a `Query`, rather than 'something that is queryable'.

zaino-state/src/chain_index.rs
zancas: But..  a trait is not a "thing".

zaino-state/src/chain_index/snapshot.rs
AloeareV: (whoops. This comment is on the impl block, meant to put it on the NonFinalized trait itself)

zaino-state/src/chain_index.rs
zancas: How about "Queryable"?

AloeareV: Much clearer!

zaino-state/src/chain_index.rs
zancas: The "ChainIndex" component (especially to internal developers) is read from the mod name.

Per the `NodeBacked` qualifier..  are there other implementors/providers?

nachog00: even if there aren't any other ones, the main one that is aimed to be the one consumed in production is, to me, worthy of the context-less name

zaino-state/src/chain_index/non_finalised_state.rs
zancas: You prefer `snapshot` and `non_finalized_state` to be sibling mods within `chain_index`?

zaino-state/src/chain_index.rs
zancas: @nachog00 and I suspected that _NodeBacked_ might have been useful to discriminate from the trait named **ChainIndex**, but if we now name that trait **Queryable** then there's one less reason to prepend **NodeBacked**.

zaino-state/src/lib.rs
zancas: Heh...  great minds think alike.

zancas:
I have long been a fan of long explicit names..  this proposal shifts somewhat from that in favor of encouraging readers to account for the enclosing mod name to contextualize a given item.

zaino-state/src/chain_index/snapshot.rs
nachog00: Since this is the snapshot module, and it provides the impl to get a snapshot of the `non_finalised_state`, it seems to me like this trait should be called something along the lines of `Snapshottable` or plain `Snapshot`.

zancas:
I fixed a doc lint, and I think I addressed all of the current comments.   I didn't resolve any conversations, as I suppose it *might* make more sense for the initiator of conversations to click resolve.

I'm hoping that @AloeareV and/or @nachog00 are interested in continuing the conversation that this PR is derived from.

I'd be curious if @nachog00 had any comments to add here.

zaino-state/src/chain_index/non_finalised_state.rs
nachog00: Is it right to say that the chain index takes a snapshot of itself?

AloeareV: I do. Another possibility that I just thought of is to have the snapshot mod be inside the non_finalized_state mod, make sure that it's not public, and then pub use the snapshot mod in chain_index. This means that snapshot is clearly inside non_finalized_state from a developer perspective, but make them siblings at the public API level (as the snapshot being of the nonfinalized state is useful for developers to know, but unnecessary confusion to consumers)

zaino-state/src/chain_index/snapshot.rs
zancas: > again, less information? Why?

I think less information in the last part of the name, can stimulate the reader to consider the other parts of the name..  i.e. the mod path.    So..   often I think it's better to not replicate a mod name in the items it contains..   this practice can couple nicely with a practice of partial path imports..   so one could:

```Rust
use chain_index::snapshot;

let x = snapshot::NonFinalized
```

In that way the explicit and verbose naming information IS included but with the context that let's the reader immediately know about the definition's local mod hierarchy.

zaino-state/src/chain_index/non_finalised_state.rs
zancas: If this were the organization then the name of the trait might make more sense as "SnapShotable"?  Instead of "NonFinalized"?

zaino-state/src/chain_index.rs
```diff
@@ -97,13 +231,13 @@ mod tests;
...
-/// let chain_index = NodeBackedChainIndex::new(source, config).await?;
-/// let subscriber = chain_index.subscriber().await;
+/// let index = Index::new(source, config).await?;
+/// let subscriber = index.subscriber().await;
 ///
 /// // Take a snapshot for consistent queries
 /// let snapshot = subscriber.snapshot_nonfinalized_state();
 ///
-/// // Query blocks in a range using the subscriber
+/// // Queryable blocks in a range using the subscriber
```
nachog00, line 240: Query seemed right on this comment before the change

zaino-state/src/chain_index.rs
```diff
@@ -132,7 +266,7 @@ mod tests;
 ///     None,  // no cookie path
 /// ).await?;
 ///
-/// // Wrap the connector for use with ChainIndex
+/// // Wrap the connector for use with Queryable
```
nachog00, line 269: fully qualified `chain_index::Queryable` would make more sense on this doc line.

zaino-state/src/chain_index.rs
```diff
@@ -44,23 +45,156 @@ pub mod types;
...
+/// The primary implementation is [`Index`], which can be backed by either:
 /// - Direct read access to a zebrad database via `ReadStateService` (preferred)
 /// - A JSON-RPC connection to a validator node (zcashd, zebrad, or another zainod)
 ///
 /// # Example with ReadStateService (Preferred)
 ///
 /// ```no_run
 /// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
-/// use zaino_state::{ChainIndex, NodeBackedChainIndex, ValidatorConnector, BlockCacheConfig};
+/// use zaino_state::{Queryable, Index, ValidatorConnector, BlockCacheConfig};
```
nachog00, line 197: this import should be through the `chain_index` module path, for clarity

zaino-state/src/chain_index.rs
```diff
@@ -97,13 +231,13 @@ mod tests;
 /// );
 ///
 /// // Create the chain index and get a subscriber for queries
-/// let chain_index = NodeBackedChainIndex::new(source, config).await?;
-/// let subscriber = chain_index.subscriber().await;
+/// let index = Index::new(source, config).await?;
+/// let subscriber = index.subscriber().await;
 ///
 /// // Take a snapshot for consistent queries
 /// let snapshot = subscriber.snapshot_nonfinalized_state();
```
nachog00, line 238: is `.snapshot_non_finalized_state()` a good method name?
if we are providing it on the `Index` type, then should it not be `.snapshot()` and have consumers abstracted from the details?

AloeareV: Good call

zancas:
@nachog00 can you push this forward?   I don't want to fail to capture our collective mental model upgrade, but I can't make this a top-priority in the immediate term.

zancas:
I think we're all in agreement that snapshot can properly be private inside non_finalised_state, and only have a very specific entrypoint exported as pub?

zancas:
Here's the structure of the mods after the changes proposed (to this point) in this PR:

```BASH
cargo modules structure -p zaino-state --focus-on zaino_state::chain_index --no-fns --max-depth 1

crate zaino_state
└── mod chain_index: pub
    ├── struct Index: pub
    ├── trait Queryable: pub
    ├── struct Subscriber: pub
    ├── mod encoding: pub
    ├── mod finalised_state: pub
    ├── mod mempool: pub
    ├── mod non_finalised_state: pub
    ├── mod snapshot: pub
    ├── mod source: pub
    └── mod types: pub
```

By contrast see the start state in the tracking issue:

 https://github.com/zingolabs/zaino/issues/719#issue-3664123868

PR #672, opened 2025-11-13
Standardize test location

zancas:
TL;DR This moves a set of tests to a place that makes them run faster.

Unit tests are located in the package source in the standard pattern.

Because cargo nextest no longer builds a separate binary just to run integration tests, this saves some time during the test run.


utils/check_package.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
nachog00, line 1: what's up with this new script? looks familiar

nachog00 (approved):

PR #689, opened 2025-11-21
fix exclude list in get_mempool_tx to match behaviour of LWD

Oscar-Pepper:
## Motivation

a. Client RPC `get_mempool_tx` test is failing due to zaino not excluding mempool txs in the exclude list
b. fixes a bug where `get_mempool_tx` returns compact tx's with reversed txids (fixed in LWD but not yet released)

## Solution

remove the incorrect byte chunk reversal of the txid hex-encoded string, probably due to other reversal issues that are now solved.

### Tests

https://github.com/zingolabs/infrastructure/pull/168

### PR Checklist

- [ ] The solution is tested.

### DIFF
--- a/integration-tests/tests/fetch_service.rs
+++ b/integration-tests/tests/fetch_service.rs
@@ -1606,10 +1606,8 @@ async fn fetch_service_get_mempool_tx(validator: &ValidatorKind) {
     let mut sorted_fetch_mempool_tx = fetch_mempool_tx.clone();
     sorted_fetch_mempool_tx.sort_by_key(|tx| tx.hash.clone());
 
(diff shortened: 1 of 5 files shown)


AloeareV (approved):
Attempting to demonstrate LWD parity showcases bugs in LWD. This change makes zaino behave correctly wrt get_mempool_tx. It would be really useful to compare this with correct LWD behavior, but that is currently blocked on https://github.com/zcash/lightwalletd/pull/533...and I don't think we want to block this code landing on a LWD fix.

AloeareV:
Some tests are failing in CI for unclear reasons that may be CI related. I will merge assuming they pass locally!

PR #723, opened 2025-11-25
Finalized state

### LINKED ISSUE #728: As of now we use two spellings of "finalise" (and "finalize")
It's important for readability to unify our symbols.

zancas:
Fixes: https://github.com/zingolabs/zaino/issues/728

Prior to this there was a mix of finalised and finalized..

Move finalised --> finalized


zaino-state/src/chain_index/finalized_state/db/db_schema_v1_0.txt
```diff
@@ -1,5 +1,5 @@
 # ────────────────────────────────────────────────────────────────────────────────
-#  Zaino – Finalised State Database, on-disk layout (Schema v1, 2025-08-05)
+#  Zaino – Finalized State Database, on-disk layout (Schema v1, 2025-08-05)
```
nachog00, line 2: I belive our DB impl does some hashing with all db related stuff... I wonder the implicates of this change. I don't know if we have tests to assert db stability, for how delicate it seems to be (based on reactions to changing any of it in the past) I believe we might want to prioritize getting a handle on it.

zancas:
Sigh...  obsolete.

PR #746, opened 2025-11-29
get_raw_mempool  method migrated to use chain_index

### LINKED ISSUE #700: `get_raw_mempool`
## Overview
Migrate `get_raw_mempool` in `StateServiceSubscriber` to use `indexer` instead of deprecated fields.

## Details
- **Trait:** ZcashIndexer
- **Deprecated field(s):** `mempool`
- **Location:** `zaino-state/src/backends/state.rs` line 1490

## Current Implementation
Uses `self.mempool.get_mempool()`

nachog00:
fix #700

methods aren't called the same, but i believe it's the same data...


AloeareV (approved):
Great!

PR #750, opened 2025-12-02
Quick profile for ci

zancas:
Update CI to run the fastest and most stable 74% of our tests.


nachog00 (approved):

zancas:
I benchmarked the fastest most stable 74% of the tests as running 9x faster than the whole test suite.

PR #751, opened 2025-12-02
fetch service block subsidy fix

### LINKED ISSUE #693: block_subsidy zebra error
integration-tests/tests/fetch_service.rs:880:65:
```
called `Result::unwrap()` on an `Err` value: Critical("unhandled rpc-specific core::convert::Infallible error: RPC Error (code: 0): Zebra does not support founders' reward subsidies, use a block height that is after the first halving")
```

AloeareV:
Fixes #693, by adjusting the test to expect `get_block_subsidy` to fail below for heights below the first halving when run against zebrad. We also generate more blocks when run against zebrad, to ensure than some heights above the first halving are successful.


fluidvanadium (approved):
This looks right.

PR #757, opened 2025-12-07
get_block method migrated to use chain_index

### LINKED ISSUE #694: `get_block`
## Parent Issue
Part of #677

## Overview
Migrate `get_block` endpoint in `StateServiceSubscriber` to use `indexer` instead of deprecated fields.

## Details
- **Trait:** LightWalletIndexer
- **Deprecated field(s):** `block_cache`
- **Location:** `zaino-state/src/backends/state.rs` line 1969

## Current Implementation
Uses `self.block_cache.get_compact_block(hash_or_height.to_string())`

nachog00:
fix #694

- Added ChainIndex trait and chain_types imports
- Replaced block_cache.get_compact_block() with indexer methods
- Uses snapshot_nonfinalized_state() for consistent queries
- Handles both hash and height lookups via get_block_height()


zaino-state/src/backends/state.rs
```diff
@@ -1976,16 +1976,38 @@ impl LightWalletIndexer for StateServiceSubscriber {
                 "Error: Invalid hash and/or height out of range. Failed to convert to u32.",
             )),
         )?;
+
+        let snapshot = self.indexer.snapshot_nonfinalized_state();
+
+        // Convert HashOrHeight to chain_types::Height
+        let block_height = match hash_or_height {
```
AloeareV, line 1960: I don't understand this, and it seems incorrect. Lookup by height gets only blocks on the best chain, whereas lookup by hash can get blocks on the best chain. By converting a hash to a height, this means instead of getting the block with the provided hash, we 

1) get the height of the block with the provided hash
2) return the block on the best chain, at the height of the block of the provided hash.

This should pass a hash to get_compact_block...which I realize as I have typed this out, accepts a height as an argument. So the get_compact_block method also needs to accept a HashOrHeight. I now understand why you wrote it the way you wrote it.

On second thought, I am overcomplicating this. Lightwalletd does not support getting a block by hash at all, despite accepting one as an argument. We can from easiest to hardest:

1) Only accept blocks by height.
2) Check that the block we've gotten by hash is on the best chain, and return None if not.
2) Fully implement getting nonbest compact blocks as described above.

AloeareV (changes requested):

zaino-state/src/backends/state.rs
AloeareV: I've overcomplicated this. The indexer handles the reorg-safety, from this level we're fine to get block by height.

idky137 (approved):

PR #759, opened 2025-12-07
get_block_range(_nullfiers) migrated to use chain_index

### LINKED ISSUE #696: `get_block_range`
## Overview
Migrate `get_block_range` and `get_block_range_inner` in `StateServiceSubscriber` to use `indexer` instead of deprecated fields.

## Details
- **Trait:** LightWalletIndexer
- **Deprecated field(s):** `block_cache`
- **Location:** `zaino-state/src/backends/state.rs` lines 572, 2016

## Current Implementation
Uses `self.block_cache.get_chain_height()` and `self.block_cache.get_compact_block()`

### LINKED ISSUE #697: `get_block_range_nullifiers`
## Overview
Migrate `get_block_range_nullifiers` in `StateServiceSubscriber` to use `indexer` instead of deprecated fields.

## Details
- **Trait:** LightWalletIndexer
- **Deprecated field(s):** `block_cache` (via `get_block_range_inner`)
- **Location:** `zaino-state/src/backends/state.rs` line 2023

## Current Implementation
Calls `get_block_range_inner(request, true)` which uses `block_cache`

nachog00:
fixes: #696

- Added ChainIndex, NonFinalizedSnapshot traits and chain_types imports
- Simplified height parsing with chained map_err
- Use indexer.get_compact_block() in a loop for start..=end range
- Get chain height from snapshot.best_chaintip()
- Removed complex backwards-walking via prev_hash


nachog00:
This also fixes #697 as both endpoints share the same boolean-gated implementation

fluidvanadium:
Clippy...

nachog00:
> Clippy...

This work is in the context of migrating out of the old `local_cache` fields on StateService. We marked those fields as deprecated and it's been a really useful marker of what's left for this migration to be done.

It sure looks ugly that clippy spams all that bloat on this individual PR diff, i agree. But it's intended for it to merge without clippy fixed as those warning will be fixes before finishing the migration. (Notice this PR isn't against dev)

AloeareV (changes requested):
Still has the doubled imports!

devdotbo:
Related: #791 - Benchmark results showing `GetBlockRange` is currently 3x slower than lightwalletd.

The sequential loop approach used here (`indexer.get_compact_block()` in a loop) contributes to this. The batch LMDB cursor methods in `chain_index/finalised_state/db/v1.rs` could potentially be wired up for better performance after the migration is complete.

zaino-state/src/backends/state.rs
```diff
@@ -575,143 +575,93 @@ impl StateServiceSubscriber {
...
+            })?
+            .height
+            .try_into()
+            .map_err(|_| {
+                StateServiceError::TonicStatusError(tonic::Status::invalid_argument(
+                    "Error: End height out of range. Failed to convert to u32.",
+                ))
+            })?;
+
+        // Ensure start <= end
```
idky137: Hey I may be mistaken but it looks like this has removed the functionality to serve reverse block ranges? 

The behaviour we need to follow is:
- if start height is lower than end height -> serve in ascending order.
- if start height is greater than end height -> serve in descending order.

some wallets rely on this protocol for sync to function so we need to make sure we are keeping this behaviour the same (I believe zingolib uses this).

idky137 (approved):

PR #768, opened 2025-12-12
Remove container dependency that was added as an oversight.

fluidvanadium:
Revert #763 which was just me flailing around with CI.


AloeareV (approved):
Looks pretty straight-forward!

PR #776, opened 2025-12-19
Introduce Status trait for service readiness probing

### LINKED ISSUE #775: Flaky tests due to StateService readiness race condition
While working on the StateService migration to chain_index backend, I noticed some tests exhibiting flaky behavior. Tests like `zebra::get::raw_mempool_regtest` would sometimes stall indefinitely, and after investigation it became clear that the StateService was being queried before it was actually ready to serve requests.

The underlying problem is that the current status mechanism doesn't distinguish between a service being "live" (the process is running) and being "ready" (the service is fully synced and can reliably answer queries). Without this distinction, test code would proceed to make queries immediately after the service started, but before the chain index had finished initializing.

This manifests as tests that work most of the time but occasionally hang or fail, making CI unreliable and debugging difficult. The fix requires introducing proper readiness probing so that tests and consumers can wait for the service to be fully operational before sending requests.

nachog00:
Fixes #775

This PR introduces a `Status` trait that provides a proper mechanism to check service liveness and readiness, addressing the flaky test behavior described in #775.

## Summary

- Moved `StatusType` to `zaino-common/status` module for shared access across crates
- Added a probing module in `zaino-common` as foundation for health checks
- Introduced `Status` trait with blanket implementations for `Liveness` and `Readiness`
- Made `ZcashService` a supertrait of `Status` so all services must implement readiness checking
- Refactored `ZcashService::status()` to be sync and source status from ChainIndex
- Implemented `Status` trait for `AtomicStatus` and subscriber types
- Added `poll_until_ready` utility for tests to wait for service readiness

### DIFF
--- a/zaino-state/src/backends/fetch.rs
+++ b/zaino-state/src/backends/fetch.rs
@@ -697,7 +697,7 @@ impl LightWalletIndexer for FetchServiceSubscriber {
     /// Return the height of the tip of the best chain
     async fn get_latest_block(&self) -> Result<BlockId, Self::Error> {
         let tip = self.indexer.snapshot_nonfinalized_state().best_tip;
-        dbg!(&tip);
+        // dbg!(&tip);
 
         Ok(BlockId {
             height: tip.height.0 as u64,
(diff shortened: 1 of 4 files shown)


zancas (commented):
LGTM, needs review by @fluidvanadium and @AloeareV .

PR #778, opened 2025-12-19
Fix race condition in StateServiceSubscriber and add idempotent DB writes

### LINKED ISSUE #777: Race condition in StateServiceSubscriber::get_latest_block()
There's a race condition in `StateServiceSubscriber::get_latest_block()` that occurs when multiple subscribers share a `ZainoDB` instance. The method can return stale or inconsistent data when the chain index is being updated concurrently.

Additionally, the current database write implementation doesn't handle idempotent writes correctly, which can cause issues when multiple subscribers attempt to write the same block data simultaneously. This becomes particularly problematic in test scenarios where multiple state service instances might be operating on shared state.

nachog00:
Fixes #777

**Blocked by #776** - This PR depends on the Status trait changes from #776. Please review that PR first. The diff here will include changes from #776 until it is merged.

This PR addresses the race condition in `StateServiceSubscriber::get_latest_block()` and adds idempotent write support for shared `ZainoDB` instances.

## Summary

- Fixed race condition in `StateServiceSubscriber::get_latest_block()` that could return stale data during concurrent chain index updates
- Added idempotent write support for `ZainoDB` to handle concurrent writes from multiple subscribers gracefully


zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -1215,13 +1254,74 @@ impl DbV1 {
...
+                        })
+                    }
+                }
+            }
             Err(e) => {
-                warn!("Error writing block to DB.");
+                warn!("Error writing block to DB: {e}");
 
-                let _ = self.delete_block(&block).await;
+                // let _ = self.delete_block(&block).await;
```
idky137, line 1417: Hey, is there a reason we are no longer deleting the block here? what this was here for was to protect against writing part of the blocks data to the database but other parts of write block failing. As this would lead to database tables being out of sync and could create worse errors down the line (this could corrupt the transparent indexing tables without any error.). The clean way to handle this is to remove the whole block from the database and then try to write to the database again. 

This was also the reason that a critical error was not returned here. As we should be able to retry the write without crashing the whole program.

nachog00: Ah, this makes sense. I think the problem i was fighting against when messing with this code was that we, in testing, had both fetch and state services trying to interact with the same db instance.

idky137: Ah I see, I think that the code that would be touching the same database files as the legacy local cache database code would be the v0 implementation rather than v1 (I think this is still probably a good safety feature to have here), So we may need to add this behaviour to both the v0 and legacy local cache code if tests are breaking.

This should be completely resolved with the removal of the local cache though so unless we have a lot of test failures caused by this we should be ok as that code will be removed before any of this lands in dev.

zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -1215,13 +1254,74 @@ impl DbV1 {
 
                 Ok(())
             }
+            Err(FinalisedStateError::LmdbError(lmdb::Error::KeyExist)) => {
+                // Block write failed because key already exists - another process wrote it
```
idky137, line 1354: We need to be careful here. There is a chance that the block wasn't written properly or was partially written which could lead to the problem referenced below. We may be able to use validate_block to check that the block was written properly but I need to check that this would be sufficient. We would then want to delete the block if it was not properly written, similar to the catch-all below.

EDIT: After looking more closely I think the best solution would be to call `validate_block_blocking` in this branch of the match statement and delete the block if this returns an error.

idky137 (changes requested):
Overall looks good! Just a couple small changes I think we should add in the db::V1::write_block implementation to make sure block are not being partially written.

idky137:
One more addition I think we should add is updating the database CHANGELOG and the migration notes in migrations.rs with the change to write block (in the v1.0.0->1.1.0 migration).

PR #784, opened 2025-12-19
feat(config): add Docker DNS hostname resolution support

gustavovalverde:
## Summary

This PR adds support for Docker DNS hostname resolution in validator address configuration, allowing addresses like `zebra:18232` instead of requiring IP addresses. It also makes cookie authentication config-based rather than address-type-based.

### Changes

- **Add hostname resolution support**: Changed `validator_jsonrpc_listen_address` and related fields from `SocketAddr` to `String` to accept hostnames
- **Add `zaino-common::net` module**: New utility module with `resolve_socket_addr()` and `try_resolve_address()` functions that handle both IP:port and hostname:port formats
- **Add format validation**: Invalid address formats (missing port, garbage input) fail immediately at config time with descriptive errors, while valid hostnames that can't be resolved (Docker DNS) are deferred to connection time
- **Config-based cookie auth**: Cookie authentication is determined solely by whether `validator_cookie_path` is configured, not by address type. This gives operators control over their security choices in trusted network environments.
- **Update config conversions**: Changed `From<ZainodConfig>` to `TryFrom<ZainodConfig>` for `StateServiceConfig` and `FetchServiceConfig` to properly handle fallible address resolution

### Validator Address Validation

The validation rules are:
- **Resolved IPs**: Must be private (RFC1918/ULA)
- **Hostnames**: Validated at connection time (supports Docker/K8s service discovery)
- **Cookie auth**: Determined by `validator_cookie_path` config, not enforced by address type

The `AddressResolution` enum distinguishes between:
- `Resolved(SocketAddr)` - Successfully resolved, private IP validation applies
- `UnresolvedHostname` - Valid hostname:port format but DNS lookup failed (acceptable for Docker)
- `InvalidFormat` - Malformed input that fails immediately

### Rationale for Cookie Auth Change

Previously, cookie authentication was forced for non-loopback IPs regardless of configuration. This created friction for operators in trusted network environments (Docker, Kubernetes, internal networks) where cookie auth adds operational complexity without proportional security benefit.

The change makes behavior consistent: if you configure `validator_cookie_path`, cookie auth is used; if you don't, it's not required. This respects the operator's explicit configuration choice while maintaining the private IP requirement as the network security boundary.

### Testing

- Added unit tests for address format validation (DNS-independent)
- Added tests for config-based cookie auth behavior:
  - `test_cookie_auth_not_forced_for_non_loopback_ip` - validates non-loopback private IPs work without cookie auth
  - `test_public_ip_still_rejected` - validates private IP requirement still applies
- Marked DNS-dependent tests with `#[ignore]` to prevent CI flakiness
- All 14 config tests pass

### Breaking Changes

- `test_node_and_return_url()` now takes `&str` instead of `SocketAddr`
- Config conversion traits changed from `From` to `TryFrom`
- Non-loopback private IPs no longer require cookie authentication (operator's choice)

### DIFF
--- a/zainod/src/config.rs
+++ b/zainod/src/config.rs
@@ -122,34 +122,21 @@ impl ZainodConfig {
         let validator_addr_result =
             try_resolve_address(&self.validator_settings.validator_jsonrpc_listen_address);
 
+        // Validator address validation:
+        // - Resolved IPs: must be private (RFC1918/ULA)
+        // - Hostnames: validated at connection time (supports Docker/K8s service discovery)
+        // - Cookie auth: determined by validator_cookie_path config, not enforced by address type
         match validator_addr_result {
             AddressResolution::Resolved(validator_addr) => {
-                // Successfully resolved - perform full IP-based validation.
                 if !is_private_listen_addr(&validator_addr) {
                     return Err(IndexerError::ConfigError(
                         "Zaino may only connect to Zebra with private IP addresses.".to_string(),
                     ));
                 }
-
-                #[cfg(not(feature = "no_tls_use_unencrypted_traffic"))]
-                {
-                    // Require cookie auth for non-loopback addresses.
-                    if !is_loopback_listen_addr(&validator_addr)
-                        && self.validator_settings.validator_cookie_path.is_none()
-                    {
-                        return Err(IndexerError::ConfigError(
-                            "Validator listen address is not loopback, so cookie authentication must be enabled."
-                                .to_string(),
-                        ));
-                    }
-                }
             }
             AddressResolution::UnresolvedHostname { ref address, .. } => {
-                // Valid hostname format but DNS lookup failed (e.g., Docker DNS).
-                // Allow this - IP validation will happen at connection time.
                 info!(
-                    "Validator address '{}' is a hostname that cannot be resolved at config time. \
-                     IP validation deferred to connection time.",
+                    "Validator address '{}' cannot be resolved at config time.",
                     address
                 );
             }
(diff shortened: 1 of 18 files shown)


zainod/src/config.rs
```diff
@@ -139,16 +170,6 @@ impl ZainodConfig {
                     "TLS required when connecting to external addresses.".to_string(),
                 ));
             }
-
```
zancas, line 142: @nachog00 @AloeareV opinions?

This seems like a separable concern?  Something that could be its own PR?

zancas (commented):
I looked this over briefly.  Why is cookie auth for non-locahost checking removed in this PR?  Isn't it a separate concern to DNS host resolution?

zainod/src/config.rs
gustavovalverde: You're right that the cookie auth change is a separable concern conceptually. However, it's directly related to this PR because of a behavior inconsistency it exposed:

Before this PR, the code forced cookie auth for non-loopback IPs. With hostname support added, we had a drift:
- Resolved IPs → cookie auth forced (address-type-based)
- Hostnames → no enforcement (config-based)

This inconsistency meant the same logical configuration behaved differently depending on whether you used `192.168.1.10:18232` vs `zebra:18232`.

To fix this I made both behave the same way - cookie auth is now determined solely by whether `validator_cookie_path` is configured. This:
1. Respects operator configuration choices in trusted networks
2. Makes behavior consistent across address formats
3. Keeps the private IP requirement as the security boundary

gustavovalverde:
> I looked this over briefly. Why is cookie auth for non-locahost checking removed in this PR? Isn't it a separate concern to DNS host resolution?

Answered here: https://github.com/zingolabs/zaino/pull/784#discussion_r2668106272

gustavovalverde:
@nachog00 @AloeareV @zancas please let me know if this PR require further changes. This is the only missing piece to fix the Z3 stack deployment and have it working with the latest version of all the applications.

zancas:
This is top-priority!

nachog00 (approved):

Issue #786, opened 2025-12-19
Correctly determine the best chain


AloeareV:
### Usage mode

zainod (running the daemon)

### Describe the bug

We currently determine the best chain based on chain tip height rather than cumulative chainwork

### Steps to reproduce

#771 attempts to reproduce this bug, but fails due to an earlier bug triggered by reorgs and never actually triggers. A fix for #771 will likely fix this bug as well. 

### Expected behavior

Best chain to be determined by cumulative chainwork

### Environment

pr #771 

### Configuration

```toml

```

### Logs

```shell

```

idky137:
The majority of work for this should already be implemented. because we use 'IndexedBlock' in both the 'NonFinalizedState' and the 'FinalizedState' we already have access to the chainwork for any given block [ https://github.com/zingolabs/zaino/blob/f5ae337ece8169c5318740506548dfcca92be1ab/zaino-state/src/chain_index/types/db/legacy.rs#L1140 ]. We should be able to buid 'IndexedBlock's and then use this rather than block height.

AloeareV:
Fixed by #771

Issue #785, opened 2025-12-19
Sufficiently hostile reorg conditions, cause NFS sync fails


AloeareV:
### Usage mode

Library integration (using zaino-* crates)

### Describe the bug

As showcased by #771, chain_index sync is failing when the proptest mockchain simulates a large number of reorgs in a short amount of time.

### Steps to reproduce

run `cargo nextest run make_chain` on pr #771

### Expected behavior

"make_chain" test to pass

AloeareV:
Fixed by #771

PR #771, opened 2025-12-15
Proptest non best chains

AloeareV:
Fixes: #679 

This PR contains all proptest-related code originally part of #590, without the functional changes to zaino-state.

At the time of writing, this PR contains a test which asserts nothing. Once the correct assertions are made, this test should be able to show the presence of #679. After we have evidence of that bug's presence, the functional changes in #590 can be incorporated into this PR, which should fix #679.

PR #787, opened 2025-12-19
Fix sync

AloeareV:
Once this is merged into #771, the make_chain test will pass, and #679, #785, and #786 will all be fixed.

### DIFF
--- a/zaino-state/src/chain_index/finalised_state/db/v1.rs
+++ b/zaino-state/src/chain_index/finalised_state/db/v1.rs
@@ -843,9 +843,7 @@ impl DbV1 {
         self.status.store(StatusType::Syncing);
         let block_hash = *block.index().hash();
         let block_hash_bytes = block_hash.to_bytes()?;
-        let block_height = block.index().height().ok_or(FinalisedStateError::Custom(
-            "finalised state received non finalised block".to_string(),
-        ))?;
+        let block_height = block.index().height();
         let block_height_bytes = block_height.to_bytes()?;
 
         // check this is the *next* block in the chain.
@@ -883,12 +881,7 @@ impl DbV1 {
         })?;
 
         // Build DBHeight
-        let height_entry = StoredEntryFixed::new(
-            &block_hash_bytes,
-            block.index().height().ok_or(FinalisedStateError::Custom(
-                "finalised state received non finalised block".to_string(),
-            ))?,
-        );
+        let height_entry = StoredEntryFixed::new(&block_hash_bytes, block.index().height());
 
         // Build header
         let header_entry = StoredEntryVar::new(
@@ -1014,7 +1007,7 @@ impl DbV1 {
                     );
                 } else {
                     return Err(FinalisedStateError::InvalidBlock {
-                        height: block.height().expect("already  checked height is some").0,
+                        height: block.height().0,
                         hash: *block.hash(),
                         reason: "Invalid block data: invalid transparent input.".to_string(),
                     });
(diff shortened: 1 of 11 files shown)


zaino-state/src/chain_index/types/db/legacy.rs
```diff
@@ -685,7 +685,7 @@ impl ZainoVersionedSerde for BlockIndex {
         self.parent_hash.serialize(&mut w)?;
         self.chainwork.serialize(&mut w)?;
 
-        write_option(&mut w, &self.height, |w, h| h.serialize(w))
+        write_option(&mut w, &Some(self.height), |w, h| h.serialize(w))
```
idky137, line 688: is there  a reason we are still writing this as an option? I think the best approach would be to just write the inner height if we no longer use option<height> as adding the wrapper option would only increase database size and slow fetch.

However, if we have released a version of zaino with the chain index active it would require a new item version, and a minor db version bump and migration (I am happy to implement this if helpful).

Something I would like to do soon is to organise the V1 database code and add better documentation to explain this..

zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -843,9 +843,7 @@ impl DbV1 {
         self.status.store(StatusType::Syncing);
         let block_hash = *block.index().hash();
         let block_hash_bytes = block_hash.to_bytes()?;
-        let block_height = block.index().height().ok_or(FinalisedStateError::Custom(
-            "finalised state received non finalised block".to_string(),
-        ))?;
+        let block_height = block.index().height();
```
idky137, line 846: I think we used to use this to check whether the given block was actually part of the finalised state so we may want to add a different check to ensure this?

zaino-state/src/chain_index/types/db/legacy.rs
AloeareV: I think this is just due to me not wanting to figure out how to do a db migration for the sake of this field. A db version bump is the better way.

zaino-state/src/chain_index/finalised_state/db/v1.rs
AloeareV: This check didn't actually assert that! We wrote a few bits of code on the assumption that non-best blocks had no height, but this is not true..this PR and #771 were inspired by these issues being hard to test for. The block has a height for all valid blocks (i.e. the first transaction in the block is a coinbase transaction, maybe some other checks as well), any block capable of being accepted on chain has a height regardless of whether it's in the best chain or not. 

We check that this block is next in the chain later in this call, the checks that make sure we never send non-best blocks to the finalized state happen outside of this function, in non-finalized sync iirc.

zaino-state/src/chain_index/tests/proptest_blockgen.rs
```diff
@@ -196,71 +182,21 @@ impl BlockchainSource for ProptestMockchain {
...
+            // This implementation selects a block from a random branch instead
+            // of the best branch. This is intended to simulate reorgs
+            HashOrHeight::Height(height) => Ok(self
+                .genesis_segment
+                .iter()
+                .find(|block| block.coinbase_height().unwrap() == height)
+                .cloned()
+                .or_else(|| {
+                    self.branching_segments
+                        .choose(&mut rand::thread_rng())
```
fluidvanadium, line 194: Is there a Strategy way to pick this random?

fluidvanadium (commented):
Ready to merge after #793

PR #788, opened 2025-12-28
Feature/lightclient protocol v0.4.0

### LINKED ISSUE #808: Adopt lightclient-protocol v0.4.0
### Usage mode

zainod (running the daemon)

### Problem or use case

Adopt latest version of lightclient-protocol canonical files for light clients

https://github.com/zcash/lightwallet-protocol/tree/v0.4.0

### Proposed solution

Adopt following lightwalletd implementation as reference 
https://github.com/zcash/lightwalletd/pull/534/

### Alternatives considered

none

### Area

gRPC API (CompactTxStreamer)

### Additional context

_No response_

pacu:
Adding transparent data to Zaino light client protocol implementation by adopting lightclient-protocol v0.4.0

Closes #808 

## Motivation
provide better data for wallets to handle transparent ZEC 
## Solution
implement the new protoc changes 

### Tests

TBD

https://github.com/zcash/lightwalletd/pull/534/files

### PR Checklist

- [x] The solution is tested.


zaino-proto/lightwallet-protocol/walletrpc/service.proto
```diff
@@ -181,44 +219,63 @@ message GetAddressUtxosReplyList {
 service CompactTxStreamer {
     // Return the BlockID of the block at the tip of the best chain
     rpc GetLatestBlock(ChainSpec) returns (BlockID) {}
+
     // Return the compact block corresponding to the given block identifier
     rpc GetBlock(BlockID) returns (CompactBlock) {}
```
idky137, line 224: I notice here that transparent data is not available from GetBlock requests, only GetBlockRange requests. Should we add a comment specifying this? I think it would require changes in the upstream proto to change this behaviour so probably best to keep as is for now but add a clear comment?

zaino-proto/src/proto/utils.rs
```diff
@@ -0,0 +1,299 @@
...
+/// of `GetBlockRange` RPC.
+pub struct ValidatedBlockRangeRequest {
+    start: u64,
+    end: u64,
+    pool_types: Vec<PoolType>,
+}
+
+impl ValidatedBlockRangeRequest {
+    /// validates a BlockRange in terms of the `GetBlockRange` RPC
+    pub fn new_from_block_range(
```
idky137, line 73: In the majority of zcash codebases BlockHeight is actually a u32:

- https://github.com/zcash/librustzcash/blob/main/components/zcash_protocol/src/consensus.rs#L19

- https://github.com/ZcashFoundation/zebra/blob/067e0b4e8253560d022227afab0acda820096429/zebra-chain/src/block/height.rs#L28

- https://github.com/zingolabs/zaino/blob/b689a3173fc20254302cc2fb5d9ec117a8add1e3/zaino-state/src/chain_index/types/db/legacy.rs#L298

It is an annoying detail that height was implemented as a u64 in the lightwallet interface (possibly due to implementation within tonic build?). 

For this reason it may be a good idea to add a check to make sure the start and end fields are within u32 range, so we can be confident that this request is compatible with the rest of the codebase once validated.

zaino-state/src/local_cache.rs
```diff
@@ -278,9 +281,10 @@ async fn try_state_path(
                         block_hex.as_ref(),
                         Some(display_txids_to_server(txid_strings)?),
                     )?
-                    .into_compact(
```
idky137, line 281: I think there is a detail we should document somewhere with this change.

Specifically: 
If using the legacy database (local_cache), for this change to work properly the old database file would need to be deleted and the database rebuilt from scratch (a slow process). This is because the legacy database did not have a migration system implemented. 

Because prost deserialises missing fields as field::default (in this case an empty vec) trying to fetch from or write to the database without rebuilding will not fail, but transparent tx data would not be present for any blocks that were written to the database before this change.

Because the next version of zaino will not be released before the new ChainIndex (and ZainoDB) code lands, it may be simplest to document that this functionality will not be available until the database has fully migrated to V1?

zaino-proto/src/proto/utils.rs
```diff
@@ -0,0 +1,299 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct PoolTypeFilter {
+    include_transparent: bool,
+    include_sapling: bool,
+    include_orchard: bool,
+}
+
+impl PoolTypeFilter {
+    /// By default PoolType includes `Sapling` and `Orchard` pools.
+    pub fn default() -> Self {
```
idky137: Would it be possible for us to implement the std::default::Default::default trait instead of a custom method?

zaino-state/src/error.rs
```diff
@@ -199,6 +200,34 @@ pub enum FetchServiceError {
...
+    pub(crate) fn from_get_block_change_error(error: GetBlockRangeError) -> Self {
+        match error {
+            GetBlockRangeError::StartHeightOutOfRange => {
+                FetchServiceError::TonicStatusError(tonic::Status::invalid_argument(
+                    "Error: Start height out of range. Failed to convert to u32.",
+                ))
+            }
+            GetBlockRangeError::NoStartHeightProvided => {
+                FetchServiceError::TonicStatusError(tonic::Status::invalid_argument(
+                    "Error: Start height out of range. Failed to convert to u32.",
```
idky137, line 233: Should this read "Error: No start height given"?

zaino-state/src/error.rs
```diff
@@ -199,6 +200,34 @@ pub enum FetchServiceError {
...
+                ))
+            }
+            GetBlockRangeError::EndHeightOutOfRange => {
+                FetchServiceError::TonicStatusError(tonic::Status::invalid_argument(
+                    "Error: End height out of range. Failed to convert to u32.",
+                ))
+            }
+            GetBlockRangeError::NoEndHeightProvided => FetchServiceError::TonicStatusError(
+                tonic::Status::invalid_argument("Error: No start height given."),
```
idky137: I think should be "Error: No end height given".

zaino-state/src/backends/state.rs
```diff
@@ -550,50 +553,41 @@ impl StateServiceSubscriber {
...
-            None => {
-                return Err(StateServiceError::TonicStatusError(
-                    tonic::Status::invalid_argument("Error: No start height given."),
-                ));
-            }
-        };
-        let lowest_to_highest = if start > end {
-            (start, end) = (end, start);
+        let mut validated_request = ValidatedBlockRangeRequest::new_from_block_range(&request)
+            .map_err(|_| StateServiceError::Custom("fixme".to_string()))?;
```
idky137: It looks like this is missing the "from_get_block_change_error" method that has been implemented for FetchServiceError (and used in zaino-state::backends::fetch)?

zaino-state/src/backends/state.rs
```diff
@@ -550,50 +553,41 @@ impl StateServiceSubscriber {
...
+                            "PoolType::Invalid specified as argument in `BlockRange`.".to_string()
+                        ),
+                        PoolTypeError::UnknownPoolType(t) => StateServiceError::UnhandledRpcError(
+                            format!("Unknown value specified in `BlockRange`. Value '{}' is not a known PoolType.", t)
+                        )
+                    }
+                )
+            }
+        }?;
+        // FIX: find out why there's repeated code fetching the chain tip and then the rest
```
idky137, line 590: This is used to check whether a BlockRange request contains block heights greater than the best chain tip ( see line 611 ), as this error should be tonic::out_of_range rather than tonic::invalid_argument.

zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -2971,166 +2970,12 @@ impl DbV1 {
...
         height: Height,
     ) -> Result<zaino_proto::proto::compact_formats::CompactBlock, FinalisedStateError> {
-        let validated_height = self
-            .resolve_validated_hash_or_height(HashOrHeight::Height(height.into()))
-            .await?;
-        let height_bytes = validated_height.to_bytes()?;
-
-        tokio::task::block_in_place(|| {
-            let txn = self.env.begin_ro_txn()?;
+        let block = self.get_chain_block(height).await?;
```
idky137: I think this can be more efficient if we just fetch the data required for the compact block requested rather than fetching the full indexed block and then discarding data we do not need (Database tables have been designed so compact block can be built without fetching irrelevant data.

Maybe we should add PoolTypes input field to this method and then build the compact block here, only fetching data requested?

I am happy to implement this and PR into your branch? I have been wanting to implement the accompanying "get_compact_block_range" method also which should improve performance of this endpoint.

I think this will be useful for the NonFinalisedState though as we hold fully constructed IndexedBlocks for that portion of the chain.

zaino-proto/build.rs
```diff
@@ -0,0 +1,86 @@
+use std::env;
+use std::fs;
+use std::io;
+use std::path::{Path, PathBuf};
+
+const COMPACT_FORMATS_PROTO: &str = "proto/compact_formats.proto";
+const PROPOSAL_PROTO: &str = "proto/proposal.proto";
+const SERVICE_PROTO: &str = "proto/service.proto";
+
+fn main() -> io::Result<()> {
```
idky137, line 10: This all looks good. I think we should probably remove the old build.rs.bak file with this addition though.

zaino-proto/lightwallet-protocol/walletrpc/service.proto
pacu: This is a good catch. Thanks

zaino-state/src/local_cache.rs
pacu: Thanks! Will document this !

zaino-proto/src/proto/utils.rs
pacu: Good Idea.

I have a follow up questions though. 
1. Can Zaino fetch a block range from Block hashes? 
2. Shall ValidatedBlockRangeRequest have HashOrHeight type instead? 
3. should function `blockid_to_hashorheight` be moved to `zaino_proto::utils` instead?

zaino-proto/lightwallet-protocol/walletrpc/service.proto
pacu: deferred to the outcome of this issue https://github.com/zcash/lightwallet-protocol/issues/18

zaino-proto/src/proto/utils.rs
idky137: We can do this in with the ChainIndex, which should be landing very soon, we don't currently have "get block range by hash or height", but we do have a method for fetching block heights by hash, so it will be fairly easy for us to implement the hash side of BlockId with that update.

Ye using HashOrHeight is a good idea, we should be able to utilise the other (hash) side very soon, we could update this in a subsequent PR if needed though.

And ye sounds like the right place!

idky137:
Hey, all the changes look good! 

Ive got the database update PR ready ( https://github.com/pacu/zaino/pull/4 ).

idky137 (approved):
Looks brilliant!

PR #793, opened 2026-01-09
Fix sync review

fluidvanadium:
-Comments

-Changes the order of blocks in one iterator.

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -882,6 +882,7 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
                     .find_map(|(hash, block)| {
                         if *hash == mempool_tip_hash {
                             Some(block.height() + 1)
+                            // found the block that is the tip that the mempool is hanging on to
                         } else {
                             None
                         }
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -799,7 +799,6 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
             ConsensusBranchId::current(&self.non_finalized_state.network, dbg!(height))
                 .map(u32::from)
         });
-        dbg!(block_consensus_branch_id);
         full_block
             .transactions
             .iter()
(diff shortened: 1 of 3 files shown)


zaino-state/src/chain_index.rs
```diff
@@ -867,14 +872,17 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
                         Some(BestChainLocation::Mempool(snapshot.best_tip.height + 1));
                 }
             } else {
+                // what does it mean that the best chain and the mempool have divergent tip hashes -?
```
AloeareV: It means that the mempool is atop a block other than the best chain tip.

zaino-state/src/chain_index.rs
```diff
@@ -867,14 +872,17 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
                 let target_height = self
                     .non_finalized_state
                     .get_snapshot()
+                    // refetch the snapshot because -?
```
AloeareV: One way the mempool can be atop a block other than the best chain tip is if the snapshot we have is old. In this case, we want to find what height the mempool transaction was expected to be mined in, if possible

zaino-state/src/chain_index.rs
```diff
@@ -867,6 +871,8 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
                         Some(BestChainLocation::Mempool(snapshot.best_tip.height + 1));
                 }
             } else {
+                // the best chain and the mempool have divergent tip hashes
+                // refetch the snapshot to catch up to the mempool
```
AloeareV: I would phrase this as something like "get a fresh snapshot, and use it to find the mempool's height if it's beyond the provided snapshot's knowledge" to clarify that the new snapshot isn't replacing the old one

zaino-state/src/chain_index.rs
```diff
@@ -576,6 +576,14 @@ impl<Source: BlockchainSource> NodeBackedChainIndexSubscriber<Source> {
             .transpose()
     }
 
+    /**
+    Searches finalized and non-finalized chains for any blocks containing the transaction.
+    Ordered with finalized blocks first.
+
+    Warning: there might be multiple blocks containing the transaction.
+    In one case, diverging non-finalized chains might each confirm the transaction.
+    An uncertain case: If a transaction, which is already on a NonBest chain, becomes Finalized, it might show up in both places in a single return of this function.
```
AloeareV: This isn't uncertain. Could just rephrase the above line by saying `diverging chains might each confirm the transaction` without specifying that they have to be non-finalized. (with that said, I think that zaino doesn't track non-best chains that diverge below the finalization height...so this might be impossible). 

..Alright, there's some uncertainty there.

PR #795, opened 2026-01-14
updated finalised state documentation

### LINKED ISSUE #794: Update zaino-state::chain_index::finalized_state documentation.
### What's the smell?

The finalised state in zaino's chain index module is a large module that must follow certain use and development strategies to avoid creating bugs or breaking backwards compatibility with older database versions.

Currently documentation is insufficient for developers not familiar with this code to be able to understand these strategies and how this module should be developed.

Documentation should be added to describe:
- Code layout and sub-module.
- Versioning and migration strategy.
- Usage and development.

### Location

`zaino_state::chain_index::finalized_state::*`

### Category

Other

### Severity

Low - Nice to fix eventually

### Suggested fix

_No response_

idky137:
Fixes: #794

NOTE: This PR does not touch any code, it only updates documentation. Any finalised state / database updates discussed will be made in subsequent PRs.

### DIFF
--- a/zaino-state/src/chain_index/finalised_state.rs
+++ b/zaino-state/src/chain_index/finalised_state.rs
@@ -1,4 +1,171 @@
-//! Holds the Finalised portion of the chain index on disk.
+//! Finalised ChainIndex database (ZainoDB)
+//!
+//! This module provides `ZainoDB`, the **on-disk** backing store for the *finalised* portion of the
+//! chain index.
+//!
+//! “Finalised” in this context means: All but the top 100 blocks in the blockchain. This follows
+//! Zebra's model where a reorg of depth greater than 100 would require a complete network restart.
+//!
+//! `ZainoDB` is a facade around a versioned LMDB-backed database implementation. It is responsible
+//! for:
+//! - opening or creating the correct on-disk database version,
+//! - coordinating **database version migrations** when the on-disk version is older than the configured
+//!   target,
+//! - exposing a small set of core read/write operations to the rest of `chain_index`,
+//! - and providing a read-only handle (`DbReader`) that should be used for all chain fetches.
+//!
+//! # Code layout (submodules)
+//!
+//! The finalised-state subsystem is split into the following files:
+//!
+//! - [`capability`]
+//!   - Defines the *capability model* used to represent which features a given DB version supports.
+//!   - Defines the core DB traits (`DbRead`, `DbWrite`, `DbCore`) and extension traits
+//!     (`BlockCoreExt`, `TransparentHistExt`, etc.).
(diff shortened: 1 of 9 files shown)


zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -1,4 +1,171 @@
...
+//! # Core API and invariants
+//!
+//! `ZainoDB` provides:
+//!
+//! - Lifecycle:
+//!   - [`ZainoDB::spawn`], [`ZainoDB::shutdown`], [`ZainoDB::status`], [`ZainoDB::wait_until_ready`]
+//!
+//! - Writes:
+//!   - [`ZainoDB::write_block`]: append-only; **must** write `db_tip + 1`
+//!   - [`ZainoDB::delete_block_at_height`]/[`ZainoDB::delete_block`]: pop-only; **must** delete tip
```
AloeareV, line 124: Hang on, why does `delete_block` exist at all? This database is all finalized, it should be append-only.

idky137: I was actually thinking about this while updating the documentation, the original thought process was that it is useful for the controlling process to be sure about exactly what block is being deleted (so we don't accidentally delete the incorrect block height by just calling "pop_block"), however we could update the method to be called something along the lines of "pop_block" but still take an "expected_block_height" / "expected_block" (for "delete_block_at_height" and "delete_block" respectively)?

I didn't want to touch any of the actual database code in this PR as I believe there are open PRs that touch this module and didn't want to complicate the merge process as i think those PRs are higher priority. I could open issues for this and a couple other things I noticed while updating docs?

AloeareV: Yeah I don't think this is relevant for this PR just want to make a note of it. 

My querstion is more fundamental: Why do we have a `pop_block`/`delete_block` at all? The finalized database is append-only, if I understand this interface right we should never be removing blocks from it...I wouldn't expect a block to ever be deleted from the database.

idky137: Ahh I see what  you mean. Ye this is a good point. 

There is a place that the delete_block is currently used and a couple other places it could be useful but maybe we could privatise them to avoid misuse?

The place delete_block is currently used is in the write_block method, if a block is partially written before a write fails, this method is used to completely wipe any data written to the database to ensure the database tables do not go out of sync.

The areas this could be helpful are:
- If corrupt data is found in the database (which would be caught by the database validation / verification system), blocks can be deleted down to that height and the database rebuilt from there, meaning a full rebuild should never be necessary.
- If zebra decides to implement reorgs into the finalised-state (technically possible in the zcash spec), we can easily implement this in our finalised state with these methods.

AloeareV (approved):

PR #796, opened 2026-01-15
Chainindex passthrough when behind backing validator

### LINKED ISSUE #769: Passthrough queries when not fully synced
This is a study issue.

> It looks like Zaino's ChainIndex impl does not serve snapshots correctly. Data that should exist in the snapshot as of height X is returning Ok(None) when the background indexer state hasn't reached that part of the chain yet. The snapshot should instead be dynamically querying and indexing the relevant data if it hasn't been indexed yet, behaving as if it were already indexed (just slower). In cases where that isn't possible or efficient, the call should return an error state indicating to the caller that they need to wait and try again later.

AloeareV:
Fixes: #769 

This PR adds a `validator_finalized_height` field to a snapshot. If this height is higher than the snapshot's tip, most APIs will pass requests through to the backing validator.


zaino-state/src/chain_index.rs
```diff
@@ -638,32 +700,34 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
-    /// is greater than the snapshot's tip
+    /// is greater than the snapshot's tip and greater
+    /// than the validator's finalized height (100 blocks below tip)
     fn get_block_range(
         &self,
         nonfinalized_snapshot: &Self::Snapshot,
         start: types::Height,
         end: std::option::Option<types::Height>,
     ) -> Option<impl Stream<Item = Result<Vec<u8>, Self::Error>>> {
+        let max_servable_height = nonfinalized_snapshot
```
fluidvanadium: This inequality deserves a full explanation.

idky137:
Hey I've been mulling over the problem with testing this functionality and came up with a possible solution (you may have already fixed this but thought I would add this in case it has not been possible to implement in the "MockChainSource".

I think one way we could implement the test only "Stop finalised state syncing at block X" is to add a "test" feature flagged field ("pause_sync_at_height?) to the "BlockCacheConfig" struct (or we could define a "TestVariables" struct to be added as this new field?):

https://github.com/zingolabs/zaino/blob/31d0d85ef595c7ed3046a7b10686c08b9d4b77c2/zaino-state/src/config.rs#L136

Then under the same "test" feature flag stop writing to the finalised state at the given block height in "start_sync_loop":

https://github.com/zingolabs/zaino/blob/31d0d85ef595c7ed3046a7b10686c08b9d4b77c2/zaino-state/src/chain_index.rs#L520

It may still be cleaner to implement this in the "MockChainSource" if possible but that could be more intrusive than just adding here.

zaino-state/src/chain_index.rs
```diff
@@ -710,30 +779,62 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+    async fn find_fork_point(
         &self,
         snapshot: &Self::Snapshot,
         block_hash: &types::BlockHash,
     ) -> Result<Option<(types::BlockHash, types::Height)>, Self::Error> {
         let Some(block) = snapshot.as_ref().get_chainblock_by_hash(block_hash) else {
-            // No fork point found. This is not an error,
-            // as zaino does not guarentee knowledge of all sidechain data.
-            return Ok(None);
+            // If we don't have the block in our non-finalized state,
```
fluidvanadium: At this point in the code, we are certain that we dont have the block in our non-finalized state?

AloeareV:
> Hey I've been mulling over the problem with testing this functionality and came up with a possible solution (you may have already fixed this but thought I would add this in case it has not been possible to implement in the "MockChainSource".
> 
> I think one way we could implement the test only "Stop finalised state syncing at block X" is to add a "test" feature flagged field ("pause_sync_at_height?) to the "BlockCacheConfig" struct (or we could define a "TestVariables" struct to be added as this new field?):
> 
> https://github.com/zingolabs/zaino/blob/31d0d85ef595c7ed3046a7b10686c08b9d4b77c2/zaino-state/src/config.rs#L136
> 
> Then under the same "test" feature flag stop writing to the finalised state at the given block height in "start_sync_loop":
> 
> https://github.com/zingolabs/zaino/blob/31d0d85ef595c7ed3046a7b10686c08b9d4b77c2/zaino-state/src/chain_index.rs#L520
> 
> It may still be cleaner to implement this in the "MockChainSource" if possible but that could be more intrusive than just adding here.

I just thought of something, which I'm gonna try out...if we add a customizable delay to MockchainSource methods, we could do some experimentation with simulated network latency...and if we make the delay long enough, it could slow sync down to a crawl, and still allow passthroughs to go through (as a passthrough requires one or maybe two calls to the source, a one-second delay would mean that the passthrough takes a couple seconds, and syncing a block would also take seconds, so if we make the chain a few hundred blocks we won't catch up to the finalized boundary before the request comes in.)

AloeareV:
An open question: When a passthrough request is made, and the needed information is in the nonfinalized state, is this: 
1) An `Ok(None)` return, as if we and the validator were synced to the top of the finalized range
2) An `Error` return, indicating that the information requested exists but isn't synced yet
3) Something else?

If 2, do we just return this error whenever it's convenient (i.e., when we don't need to additional work to distinguish between the requested information being nonfinalized or nonexistent), or do we make the guarantee of correctly identifying this error in all cases?
Also, if 2, if the request covers multiple blocks and some of them are finalized, do we consider this a success case and return everything finalized, or an error case as the request covered information in the nonfinalized state that we can't safely serve until we've caught up to it? Possibly even vary by specific case/method?

AloeareV:
Still to be done:
* Make sure passthrough only triggers when the validator's finalized tip is ahead of ChainIndexes best (nonfinalized) tip.
* Some error handling
* Fix passthrough triggering before local lookup in some RPCs

zaino-state/src/chain_index.rs
```diff
@@ -641,32 +705,36 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
     ) -> Option<impl Stream<Item = Result<Vec<u8>, Self::Error>>> {
-        let end = end.unwrap_or(nonfinalized_snapshot.best_tip.height);
-        if end <= nonfinalized_snapshot.best_tip.height {
+        let max_servable_height = nonfinalized_snapshot
+            .validator_finalized_height
+            .max(nonfinalized_snapshot.best_tip.height);
+        let end = end
+            .unwrap_or(nonfinalized_snapshot.best_tip.height)
+            .min(max_servable_height);
+        if start <= max_servable_height.min(end) {
```
fluidvanadium, line 839: My rule of thumb is that if there is any arithmetical operation, the line needs a comment.

Can best_tip be above validator_finalized? Is calling MAX for situations where zaino already has a database but connects to a new validator?

The MIN and this inequality are in order to not try to get any blocks above best_tip.

zaino-state/src/chain_index.rs
```diff
@@ -760,22 +867,34 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+                            .to_vec(),
+                        ConsensusBranchId::current(&self.non_finalized_state.network, height)
+                            .map(u32::from),
+                    )));
+                }
+            }
+        }
+
+        // if the tranasction isn't finalized on the best chain
+        // check our indexes
```
fluidvanadium: should we check the index first?

zaino-state/src/chain_index.rs
```diff
@@ -710,30 +783,62 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
 
     /// Finds the newest ancestor of the given block on the main
     /// chain, or the block itself if it is on the main chain.
-    fn find_fork_point(
+    /// Returns Ok(None) if no fork point found. This is not an error,
+    /// as zaino does not guarentee knowledge of all sidechain data.
+    async fn find_fork_point(
```
fluidvanadium, line 1097: This function still needs another look.

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -112,6 +117,10 @@ impl From<UpdateError> for SyncError {
             UpdateError::DatabaseHole => {
                 SyncError::ReorgFailure(String::from("could not determine best chain"))
             }
+            UpdateError::ValidatorConnectionError => SyncError::ZebradConnectionError(todo!(
+                "don't have a request error. \
```
AloeareV: To any reviewers: I would appreciate help figuring out what to do with this error case.

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -213,11 +226,22 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
     ) -> Result<Self, InitError> {
         info!("Initialising non-finalised state.");
 
+        let validator_tip = source
+            .get_best_block_height()
+            .await
+            .map_err(|e| InitError::InvalidNodeData(Box::new(e)))?
+            .ok_or_else(|| {
+                InitError::InvalidNodeData(todo!("we need something that implements error"))
```
AloeareV: As well as this error case

zaino-state/src/chain_index.rs
AloeareV: In the process of commenting, I discovered that one of these mins was unneeded (and incorrect).

zaino-state/src/chain_index.rs
fluidvanadium: Proposed changes in #814

zaino-state/src/chain_index/non_finalised_state.rs
fluidvanadium: I will look into it.

fluidvanadium:
Why do self.get_fullblock_bytes_from_node and self.blockchain_source.get_block look so different? shouldnt they be sister functions?

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -112,6 +117,11 @@ impl From<UpdateError> for SyncError {
             UpdateError::DatabaseHole => {
                 SyncError::ReorgFailure(String::from("could not determine best chain"))
             }
+            UpdateError::ValidatorConnectionError => {
+                SyncError::ZebradConnectionError(NodeConnectionError::UnrecoverableError(Box::new(
```
nachog00: Why is this variant referencing zebra directly, shouldn't it be abstracted away from specific validators?

zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -484,6 +511,14 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
             .await
             .map_err(|_e| UpdateError::DatabaseHole)?;
 
+        let validator_tip = self
+            .source
+            .get_best_block_height()
+            .await
+            .map_err(|_| UpdateError::ValidatorConnectionError)?
```
nachog00: this sort of mapping i think is a bit troublesome, as it essentially hides any trace of the original error. I think the correct way of mapping it would be some way in the higher error can be traced back to the original through `.source()`.

AloeareV: Ah, I think this was meant to be a `todo!`

AloeareV:
> Why do self.get_fullblock_bytes_from_node and self.blockchain_source.get_block look so different? shouldnt they be sister functions?

How do you mean 'so different'? They're not sisters, one calls the other. `get_fullblock_bytes_from_node` calls `get_block` and then serializes it to a vec.

(Originally, I somehow edited your message with my answer. I've turned it into a reply now)

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: Renamed to ValidatorConnectionError

zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: Added a source error to the enum variant!

zaino-state/src/chain_index.rs
zancas: outdated.

fluidvanadium (approved):
Tests are testing correctness of behavior.

Logical flows are well-documented.

zaino-state/src/chain_index/non_finalised_state.rs
fluidvanadium: Discussion of errors moved into continuing scope.

zaino-state/src/chain_index/non_finalised_state.rs
fluidvanadium: Discussion of errors moved into continuing scope.

PR #389, opened 2025-06-19
feat(rpc): add `z_validateaddress` endpoint

nachog00:
## Motivation

#219.

## Solution

Implement `z_validateaddress`.

- [x] The PR name is suitable for the release notes.
- [x] The solution is tested.
- [x] The documentation is up to date.

PR #797, opened 2026-01-16
tonic 0.14

### LINKED ISSUE #749: Update Dependencies
### Task description

Will unlock upgrades from https://github.com/zingolabs/infrastructure/pull/178 after it merges.

### Context

See https://github.com/zingolabs/zaino/issues/748.

### Acceptance criteria

_No response_

### Area

_No response_

fluidvanadium:
prerequisite for https://github.com/zingolabs/zaino/pull/389

This update has difficult breaking changes in tonic

### DIFF
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -36,5 +36,5 @@ tower = { workspace = true, features = ["buffer", "util"] }
 tokio = { workspace = true }
 
 # Zingo-infra
-zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", rev = "20f3c206321825952fdf2d9d84947f5c9d2a1cf4" }
-zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", rev = "20f3c206321825952fdf2d9d84947f5c9d2a1cf4" }
+zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", branch = "dev" }
+zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", branch = "dev" }
--- a/zaino-proto/Cargo.toml
+++ b/zaino-proto/Cargo.toml
@@ -16,5 +16,5 @@ tonic = { workspace = true }
 prost = { workspace = true }
 
 [build-dependencies]
-tonic-build = { workspace = true, features = ["prost"] }
+tonic-build = { workspace = true }
 which = { workspace = true }
--- a/zaino-serve/Cargo.toml
+++ b/zaino-serve/Cargo.toml
@@ -10,7 +10,7 @@ version = { workspace = true }
 
 [features]
 state_service = []
-no_tls_use_unencrypted_traffic = ["tonic/tls-roots"]
+no_tls_use_unencrypted_traffic = []
 
 [dependencies]
 zaino-proto = { workspace = true }
@@ -26,7 +26,7 @@ tracing = { workspace = true }
 
 # Miscellaneous Workspace
 tokio = { workspace = true, features = ["full"] }
-tonic = { workspace = true, features = ["tls"] }
+tonic = { workspace = true }
 thiserror = { workspace = true }
 futures = { workspace = true }
 jsonrpsee = { workspace = true, features = ["server", "macros"] }
--- a/zaino-testutils/Cargo.toml
+++ b/zaino-testutils/Cargo.toml
@@ -22,7 +22,7 @@ zainod = { workspace = true }
 zcash_protocol = { version = "0.7.0", features = [
   "local-consensus",
 ] }
-zcash_client_backend = { version = "0.20", features = ["lightwalletd-tonic"] }
+zcash_client_backend = { version = "0.21", features = ["lightwalletd-tonic"] }
 
 # Zebra
 zebra-chain = { workspace = true }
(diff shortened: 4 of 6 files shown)


integration-tests/Cargo.toml
```diff
@@ -36,5 +36,5 @@ tower = { workspace = true, features = ["buffer", "util"] }
 tokio = { workspace = true }
 
 # Zingo-infra
-zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", rev = "20f3c206321825952fdf2d9d84947f5c9d2a1cf4" }
-zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", rev = "20f3c206321825952fdf2d9d84947f5c9d2a1cf4" }
+zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", branch = "dev" }
```
idky137, line 39: Hey we try to keep imports to at least using get revs so that libraries are not updated accidentally. 

We actually need to go further than this to support the upcoming crates.io release, for this all deps must be crates.io versions so if possible we should be looking at pushing releases for all of our repos that are used here.

integration-tests/Cargo.toml
```diff
@@ -36,5 +36,5 @@ tower = { workspace = true, features = ["buffer", "util"] }
...
+zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", branch = "dev" }
```
idky137, line 40: Similar to above.

zaino-testutils/Cargo.toml
```diff
@@ -22,19 +22,19 @@ zainod = { workspace = true }
...
+zcash_client_backend = { version = "0.21", features = ["lightwalletd-tonic"] }
 
 # Zebra
 zebra-chain = { workspace = true }
 zebra-state.workspace = true
 
 # Zingo-infra
-zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", rev = "20f3c206321825952fdf2d9d84947f5c9d2a1cf4" }
-zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", rev = "20f3c206321825952fdf2d9d84947f5c9d2a1cf4" }
+zcash_local_net = { git = "https://github.com/zingolabs/infrastructure.git", branch = "dev" }
```
idky137, line 32: Similar to above.

zaino-testutils/Cargo.toml
```diff
@@ -22,19 +22,19 @@ zainod = { workspace = true }
...
+zingo_test_vectors = { git = "https://github.com/zingolabs/infrastructure.git", branch = "dev" }
```
idky137, line 33: Similar to above.

zaino-testutils/Cargo.toml
```diff
@@ -22,19 +22,19 @@ zainod = { workspace = true }
...
 
 # Zingo-common
-zingo_common_components = { git = "https://github.com/zingolabs/zingo-common.git", tag = "zingo_common_components_v0.1.0", features = [ "for_test" ]}
-zingo_netutils = { git = "https://github.com/zingolabs/zingo-common.git", rev = "b64dfd6d6a2a597a5456d1cc7b2bc9b649328187"  }
+zingo_common_components = { git = "https://github.com/zingolabs/zingo-common.git", branch = "dev", features = [ "for_test" ]}
```
idky137, line 36: Similar to above.

zaino-testutils/Cargo.toml
```diff
@@ -22,19 +22,19 @@ zainod = { workspace = true }
...
+zingo-netutils = { git = "https://github.com/zingolabs/zingo-common.git", branch = "dev"  }
```
idky137, line 37: Similar to above.

zaino-proto/Cargo.toml
```diff
@@ -16,5 +16,5 @@ tonic = { workspace = true }
 prost = { workspace = true }
 
 [build-dependencies]
-tonic-build = { workspace = true, features = ["prost"] }
```
idky137, line 19: I think it may be useful to add a comment explaining that "prost" is now part of the tonic-build default-features. So it is clear that if we ever switch off default-features we will need to manually add prost back in.

zaino-serve/Cargo.toml
```diff
@@ -10,7 +10,7 @@ version = { workspace = true }
 
 [features]
 state_service = []
-no_tls_use_unencrypted_traffic = ["tonic/tls-roots"]
```
idky137, line 13: Im a little confused by this, going through the documentation ( https://docs.rs/tonic/0.14.2/tonic/ ) it looks like we do still need to enable tls in our import but it is not clear how this has changed / how it should now be used ( The legacy "TLS" feature has been removed and it looks like it has been split into several TLS backend choices ). 

Im unsure if we need to import these here or whether they are specifically for client side implementations. My feeling is we do still need to import here, probably using the "tls-ring" feature flag.

idky137 (changes requested):
Hey I think there are a few changes that would be good to make. 

I think we should keep to specifying with git revs or better still tagged versions for our imports so we know exactly which version we are importing and need to release to crates.io.

Also Im not certain but it looks like we have removed TLS with these changes currently. I think we will need to use one of the new features added in v0.14.

zancas:
Obsolete.

PR #805, opened 2026-01-19
minor fixes (updates) to rpc_api.md

LarryRuane:
I'm just trying to get up to speed on zaino (to possibly contribute to code dev), so while reading the documentation, I noticed a few minor problems, fixed by this PR.

- The lightwallet protocol definition files have moved to their own separate repo (the link here was out of date)
- The gRPC method `GetTaddressTxids` has been deprecated; actually, renamed to `GetTaddressTransactions` (change because the old name is misleading)
- I think we can remove `Ping` because it's test-only, and actually, I haven't used that for a very long time (and I very much doubt anyone else has).


pacu:
Oh,  I think I'll probably cherry pick this commit into my branch if you don't mind.

I'm working on adopting v0.4.0 here https://github.com/zingolabs/zaino/pull/788

LarryRuane:
Closing; it makes more sense to include this in #788.

PR #806, opened 2026-01-23
Added comments.

fluidvanadium:
Fixes:

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -778,8 +778,9 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
         }
     }
 
-    /// Finds the newest ancestor of the given block on the main
-    /// chain, or the block itself if it is on the main chain.
+    /// For a given block,
+    /// find its newest main-chain ancestor,
+    /// or the block itself if it is on the main-chain.
     /// Returns Ok(None) if no fork point found. This is not an error,
     /// as zaino does not guarentee knowledge of all sidechain data.
     async fn find_fork_point(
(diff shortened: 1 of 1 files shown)


zaino-state/src/chain_index.rs
```diff
@@ -788,9 +789,10 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
         block_hash: &types::BlockHash,
     ) -> Result<Option<(types::BlockHash, types::Height)>, Self::Error> {
         let Some(block) = snapshot.as_ref().get_chainblock_by_hash(block_hash) else {
-            // We don't have the block in our non-finalized state,
-            // we'll only be aware of it if it's main-chain.
-            // Find it from the source, and return its height and hash
+            // At this point, we know that
+            // the block is NOT PENDING in the INDEXER.
```
AloeareV, line 793: 'Pending' doesn't make a whole lot of sense in this context. These blocks are on-chain.
```suggestion
            // the block is NOT NONFINALIZED in the INDEXER.
```

zaino-state/src/chain_index.rs
```diff
@@ -788,9 +789,10 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+            // TODO check the INDEXER for finalized block
```
AloeareV, line 794: Checking the indexer isn't _needed_ here. A hit in the index means we don't need to check the validator. This means any index miss takes INDEX_QUERY_TIME plus VALIDATOR_QUERY_TIME, and an index hit takes INDEX_QUERY_TIME. If we don't check the indexer here, then it's always VALIDATOR_QUERY_TIME.

Checking the validator will most likely take a lot longer than local index lookup for fetch, and may be more comparable for state...we've done no benchmarking work here. It's likely more efficient to search our own cache first here, but I don't actually know. (there's another big optimization here, if the indexer has caught up to the validator we can do pure index lookup and not need to fallback to validator)

So...yeah, we should probably check the indexer.

zaino-state/src/chain_index.rs
```diff
@@ -799,26 +801,26 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+                // the block is NOT FINALIZED in the VALIDATOR.
+                // Return Ok(None) = no block found.
                 Ok(_) => Ok(None),
                 Err(e) => Err(ChainIndexError::backing_validator(e)),
             };
         };
-        // If we have the block in our heights_to_hashes set, it's main-chain
-        // Return it's hash and height
+        // At this point, we know that
+        // The block is PENDING in the INDEXER
```
AloeareV, line 823: In this rephrasing you've removed the part that specifies the block is main-chain. (also see above re. PENDING).

The block is non-finalized and in the indexer's main chain?

PR #809, opened 2026-01-30
Chain index passthrough behind zebra diagrams

fluidvanadium:
Fixes:

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -164,6 +164,9 @@ mod tests;
 /// 1. Extract the relevant fields from your service config into a `BlockCacheConfig`
 /// 2. Create the appropriate `ValidatorConnector` variant (State or Fetch)
 /// 3. Call `NodeBackedChainIndex::new(source, config).await`
+///
+/// When a call asks for info (e.g. a block), Zaino selects sources in this order:
+#[doc = simple_mermaid::mermaid!("chain_index_passthrough.mmd")]
 pub trait ChainIndex {
     /// A snapshot of the nonfinalized state, needed for atomic access
     type Snapshot: NonFinalizedSnapshot;
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -580,7 +580,7 @@ impl<Source: BlockchainSource> NodeBackedChainIndexSubscriber<Source> {
             .transpose()
     }
 
-    async fn get_snapshot_block_height(
+    async fn get_indexed_block_height(
         &self,
         nonfinalized_snapshot: &NonfinalizedBlockCacheSnapshot,
         hash: types::BlockHash,
@@ -705,7 +705,7 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
         hash: types::BlockHash,
     ) -> Result<Option<types::Height>, Self::Error> {
         let snapshot_block_height = self
-            .get_snapshot_block_height(nonfinalized_snapshot, hash)
+            .get_indexed_block_height(nonfinalized_snapshot, hash)
             .await?;
         match snapshot_block_height {
             Some(h) => Ok(Some(h)),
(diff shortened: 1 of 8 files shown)


zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -531,7 +531,7 @@ impl ZainoDB {
                 Some(block) => block,
                 None => {
                     return Err(FinalisedStateError::BlockchainSourceError(
-                        BlockchainSourceError::Unrecoverable(format!(
```
idky137, line 534: Hey is there a reason that unrecoverable is being changed to custom here?

Unrecoverable was chosen here in the zallet-zaino api meetings as a clear error type that zallet can react to.

This seams like a bit of a downgrade as we are losing clarity on what these errors mean.

zaino-state/src/chain_index.rs
```diff
@@ -787,6 +787,27 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+                    types::Height::from(height),
+                ))),
+                None => todo!(),
+            },
+            Ok(None) => {
+                todo!()
+            }
+            Err(e) => Err(ChainIndexError::backing_validator(e)),
+        };
+        todo!();
```
AloeareV, line 810: I don't follow this addition. We already check the blockchain_source later on in the control flow, as we should! This seems the same as that logic, earlier, and without the check to see if it's not still nonfinalized.

Is this intended to be checking the finalized state? If so, as an optimization it should probably be after the non-finalized state check. The most common reason to call this method is if the caller has a reason to suspect the block may be no longer on the best chain...so the most likely place to find the block is the non-finalized stat

zaino-state/src/chain_index.rs
```diff
@@ -568,16 +571,16 @@ impl<Source: BlockchainSource> NodeBackedChainIndexSubscriber<Source> {
...
-                    .map_err(ChainIndexError::backing_validator)
+                bk.zcash_serialize_to_vec().map_err(|e| {
+                    ChainIndexError::InvalidData(InvalidDataError::BlockSerialization(e))
+                })
             })
             .transpose()
     }
 
-    async fn get_snapshot_block_height(
+    async fn get_indexed_block_height(
```
AloeareV, line 583: Good name change!

zaino-state/src/chain_index.rs
```diff
@@ -589,10 +592,7 @@ impl<Source: BlockchainSource> NodeBackedChainIndexSubscriber<Source> {
                 .find(|h| **h == hash)
                 // Canonical height is None for blocks not on the best chain
                 .map(|_| block.index().height())),
-            None => match self.finalized_state.get_block_height(hash).await {
-                Ok(height) => Ok(height),
-                Err(_e) => Err(ChainIndexError::database_hole(hash)),
-            },
+            None => self.finalized_state.get_block_height(hash).await?,
```
AloeareV, line 595: Oh. Whoops. Nice fix!

AloeareV:
I remember we (some subgroup of us. I don't remember exactly who was part of this conversation...at least myself and @nachog00, I think @zancas, not sure who if anyone else) made the decision to turn `ChainIndexError` from an enum to a struct containing an ErrorKind enum after talking in detail about how we wanted to handle errors, and I am hesitant to walk back this pattern without first reconstructing the reason we wanted to pattern it this way in the first place.

EDIT: I personally was convinced by this article https://matklad.github.io/2020/10/15/study-of-std-io-error.html

Essentially, we don't want to expose all these variants in our public API. However, it's useful to keep track of which error variant happened internally, in order to match on. By having an internal ErrorKind wrapped by a more public-facing error type, we get the best of both worlds! Notably, the changes you've made to the error variants don't appear tied to this change...how straight-forward would it be to take all these variants and make them into `ChainIndexErrorKind` variants, restoring the `ChainIndexError` structure?

zaino-state/src/chain_index/finalised_state.rs
fluidvanadium: Right. In the new system based on the diagram I put in the Signal Chat, I will leave BlockchainSourceError alone.

AloeareV (changes requested):
Requested changes are in comments instead of in review, due to a github UI whoosie.

PR #811, opened 2026-01-30
exclude openssl

### LINKED ISSUE #810: openssl is a known source of critical security vulns, because it is implemented in C it's always going to be vulnerable
We need to ensure that openssl is never a dependency.

https://www.lesswrong.com/posts/7aJwgbMEiKq5egQbd/ai-found-12-of-12-openssl-zero-days-while-curl-cancelled-its

zancas:
Fixes:  #810


zancas:
@idky137 what remains to be done for this?

dorianvp (approved):

PR #814, opened 2026-01-31
Ciptbz by diagram 3

fluidvanadium:
Add flowchart.
Reweave find_fork_point
Replace unwraps.
Change symbols.
Add comments.

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -785,8 +785,9 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
         }
     }
 
-    /// Finds the newest ancestor of the given block on the main
-    /// chain, or the block itself if it is on the main chain.
+    /// For a given block,
+    /// find its newest main-chain ancestor,
+    /// or the block itself if it is on the main-chain.
     /// Returns Ok(None) if no fork point found. This is not an error,
     /// as zaino does not guarentee knowledge of all sidechain data.
     async fn find_fork_point(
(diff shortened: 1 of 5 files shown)


zaino-state/src/chain_index.rs
```diff
@@ -642,21 +646,33 @@ impl<Source: BlockchainSource> NodeBackedChainIndexSubscriber<Source> {
...
         {
-            Ok(Some(block))
-                if types::Height::from(block.coinbase_height().expect("block to have height"))
-                    <= nonfinalized_snapshot.validator_finalized_height =>
-            {
-                Ok(Some(block.coinbase_height().unwrap().into()))
+            Ok(Some(block)) => {
+                // At this point, we know that
+                // the block is in the VALIDATOR.
+                match block.coinbase_height() {
```
AloeareV, line 660: This is not how coinbase_height works. If coinbase height is `None`, the block has no coinbase transaction...and is therefore an invalid block. (by 'invalid' I mean 'will be rejected by the validator'). Receiving a block with a coinbase_height of `None` from the validator is an unrecoverable error.

zaino-state/src/chain_index.rs
```diff
@@ -642,21 +646,33 @@ impl<Source: BlockchainSource> NodeBackedChainIndexSubscriber<Source> {
...
+                    None => {
+                        // the block is in the VALIDATOR. but doesnt have a height.
+                        // Therefore, it's non-best chain!
+                        Ok(None)
+                    }
+                    Some(height) => {
+                        // The validator returned a block with a height, implying that this block is on the best chain.
```
AloeareV: As above, this height does not mean the block is on the best chain. To clarify, a height of `none` meaning `non-best chain` is a feature offered by the `zainod`/`lightwalletd` RPC interface, and is not true in most contexts.

zaino-state/src/chain_index.rs
```diff
@@ -781,54 +789,81 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+                        // Now, we ask the VALIDATOR.
+                        match self
+                            .blockchain_source
+                            .get_block(HashOrHeight::Hash(zebra_chain::block::Hash::from(*hash)))
+                            .await
+                        {
+                            Ok(Some(block)) => {
+                                // At this point, we know that
+                                // the block is in the VALIDATOR.
+                                match block.coinbase_height() {
```
AloeareV, line 847: Again, this will always be `Some` from the validator, unless something has gone really wrong. 

To distinguish best-chain from non-best chain in this case, well...we're only 'safe' to passthrough blocks from the validator's finalized state, and as far as I know zebra tracks only best-chain blocks in the finalized state.

AloeareV (changes requested):
Some changes needed, mostly involving `coinbase_height` not meaning what you think it means! I made this same mistake in the original design of this code, which was what inspired the proptest PR (the rest of our test framework is best-chain only, and so didn't catch issues with non-best chain handling).

fluidvanadium:
Handled comments, confused by rebase.

zaino-state/src/chain_index.rs
```diff
@@ -642,21 +646,32 @@ impl<Source: BlockchainSource> NodeBackedChainIndexSubscriber<Source> {
...
         hash: types::BlockHash,
     ) -> Result<Option<types::Height>, ChainIndexError> {
         match self
             .blockchain_source
             .get_block(HashOrHeight::Hash(hash.into()))
             .await
         {
-            Ok(Some(block))
-                if types::Height::from(block.coinbase_height().expect("block to have height"))
-                    <= nonfinalized_snapshot.validator_finalized_height =>
```
AloeareV, line 655: We need this check, it's not reorg-safe to return a height from above the validator_finalized_height

zaino-state/src/chain_index.rs
```diff
@@ -724,112 +733,106 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
-                                None => self
-                                    .get_fullblock_bytes_from_node(HashOrHeight::Height(
-                                        zebra_chain::block::Height(height),
-                                    ))
-                                    .await?
-                                    .ok_or(ChainIndexError::database_hole(height)),
-                            }
-                        }
-                    }
+                    self.get_fullblock_bytes_from_node(HashOrHeight::Height(
```
AloeareV: For blocks above validator_finalized_height, it's not reorg-safe to get blocks by height. It _is_ reorg-safe to get blocks by hash. What we need to do in this case is use our snapshot index to look up the hash at a given height, and then get that hash from the validator.

zaino-state/src/chain_index.rs
```diff
@@ -724,112 +733,106 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
-            // Otherwise, it's non-best chain! Grab its parent, and recurse
-        } else {
-            // gotta pin recursive async functions to prevent infinite-sized
-            // Future-implementing types
-            Box::pin(self.find_fork_point(snapshot, block.index().parent_hash())).await
+            }
+            None => {
+                // At this point, we know that
+                // the block is NOT non-FINALIZED in the INDEXER.
+                // TODO check the INDEXER for finalized block
```
AloeareV: Stale TODO, we check the indexer for finalized block on the next line

zaino-state/src/chain_index.rs
```diff
@@ -724,112 +733,106 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+                        {
+                            Ok(Some(block)) => {
+                                // At this point, we know that
+                                // the block is in the VALIDATOR.
+                                match block.coinbase_height() {
+                                    None => {
+                                        // the block is in the VALIDATOR. but doesnt have a height. That would imply a bug.
+                                        Err(ChainIndexError::validator_data_error_block_coinbase_height_missing())
+                                    }
+                                    Some(height) => {
```
AloeareV, line 852: Again, we need a check that this height isn't above the validator_finalized_tip.

zaino-state/src/chain_index.rs
```diff
@@ -942,7 +945,9 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
             .blocks_containing_transaction(snapshot, txid.0)
             .await?
             .collect::<Vec<_>>();
-        let start_of_nonfinalized = snapshot.heights_to_hashes.keys().min().unwrap();
+        let Some(start_of_nonfinalized) = snapshot.heights_to_hashes.keys().min() else {
```
AloeareV, line 990: Non-blocking (out of the scope of this PR) but we instantiate this field with a value and never remove the last item from it. Would be great to replace this with some sort of NonEmpty type...

AloeareV (changes requested):

zaino-state/src/chain_index.rs
fluidvanadium: Removed.

fluidvanadium (commented):
reviewed comments

zaino-state/src/chain_index.rs
fluidvanadium: Re-activated on line 668

zaino-state/src/chain_index.rs
fluidvanadium: Error now propagated on line 850.

zaino-state/src/chain_index.rs
fluidvanadium: Error is now propagated on line 663

zaino-state/src/chain_index.rs
fluidvanadium: Re-activated check on line 855

zaino-state/src/chain_index.rs
fluidvanadium: Deleted.

AloeareV (approved):
Woohoo!

PR #690, opened 2025-11-24
state service chain index integration

Oscar-Pepper:
fix #677 

This deprecates all fields from `StateService` and `StateServiceSubscriber` without deleting them, so that we can update the rpc implementations (`ZcashIndexer` and `LightwalletdIndexer` traits) while still compiling fine.

Once those fields are completely un-relied upon by any endpoint, we can remove them and call this done.

NOTE: this work is done on top of PR #650, and it's expected to land before it

PR #819, opened 2026-02-04
Update state service get block

idky137:
Pullis updates in https://github.com/zingolabs/zaino/pull/690 and fixes conflicts.


idky137:
all tests pass locally.

Note 
8a82a99 is the only "real" commit (all others are from the parent pr).

PR #822, opened 2026-02-04
improve example config

### LINKED ISSUE #742: zainod config example is outdated
We need to update the example `zainod` config file.

Outdated examples lead (partially) onto issues like #737 

Implementing #741 would heavily streamline the maintenance of the config examples.

Moreover, should we even have a tracked example file or rely on the command providing it instead?

AloeareV:
Fixes: #742

example config should now be accurate to how ZainoConfig is now actually structured.


nachog00:
#854 completely removes the example config as it introduces a generate command. Documentation on the config fields themselves is delegated to cargo docs, that way there's a single source of truth and lesser maintenance burden.

If you agree, i would close this PR en favor of the other one.

AloeareV:
Agreed!

PR #825, opened 2026-02-09
Ciptbz by diagram 6

fluidvanadium:
Fixes: comments only

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -767,6 +767,8 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
         if start <= max_servable_height.min(end) {
             Some(
                 futures::stream::iter((start.0)..=(end.0)).then(move |height| async move {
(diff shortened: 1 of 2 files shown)


zaino-state/src/chain_index.rs
```diff
@@ -755,6 +766,7 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
         if start <= max_servable_height.min(end) {
             Some(
                 futures::stream::iter((start.0)..=(end.0)).then(move |height| async move {
+                    // For blocks above validator_finalized_height, it's not reorg-safe to get blocks by height. It is reorg-safe to get blocks by hash. What we need to do in this case is use our snapshot index to look up the hash at a given height, and then get that hash from the validator.
```
AloeareV, line 769: This is true, but is probably better served as a comment on line 786 (`Ok(None)) => match ...`). This line is getting heights from the finalized state.

AloeareV (approved):
There's a comment that I think would be clearer in a different place. Not sure if this is a blocking concern, I'll leave that up to you.

zaino-state/src/chain_index.rs
AloeareV: In fact, this line might be unnecessarily complex...we should be able to trust any finalized blocks zebra serves us without having to check the finalized state at all..but for that I'd prefer to just...land passthrough, and open an issue for a follow-up simplification, at this point.

PR #828, opened 2026-02-11
BUG FIX: BlockIndex V2 serde

idky137:
Fixes:
https://github.com/zingolabs/zaino/pull/787 updated the height field of the BlockIndex struct but did not implement this change as a V2 serialisation schema.

This PR brings this change into line with the database versioning strategy by:
- Adding V2 BlockHeaderData and V2 BlockIndex serde (while keeping V1 serde valid).
- Updates v1.0.0 -> v1.1.0 database migration and changelog entry
- Adds new db migration test.
- Updates ZainoVersionedSerde to enable construction of historical versions (required for checksum validation of legacy database formats)
- Updates StoredEntry verification to work with legacy versions.
- Adds encoding and StoredEntry unit tests.

NOTE: 

Because BlockHeaderData is stored in a [`StoredEntryVar`], and the headers lmdb table does not use fixed length optimisations, it is possible for us to just provide the new serde implementation. 

Where the new chain index database has been partially built before this update, the headers table will hold a mix of V1 and V2 BlockHeaderData structs'.

Where update to chainindex has not yet taken place, or where the database will be built fresh, V2 BlockHeaderData serde will be used.


idky137:
NOTE: If we release zaino before this PR lands it should be updated so that the database updates go into the v1.1.0->v1.2.0 migration (instead of the v1.0.0->1.1.0 migration currently on dev).

idky137:
NOTE: This PR is code complete but will require a couple updates once release lands. For this reason I have converted it back to draft.

idky137:
this work has been moved to #989

PR #831, opened 2026-02-12
Remove redundant fields

fluidvanadium:
Redundant fields add developer friction. Sometimes they can get tangled and become difficult to remove. Not this time.


AloeareV (approved):
Yeah this is likely better.

PR #857, opened 2026-02-20
BUG FIX: Database partial write safeguard and docs

### LINKED ISSUE #836: Potential for partially written blocks to corrupt db
We need to be careful here. There is a chance that the block wasn't written properly or was partially written which could lead to the problem referenced below. We may be able to use validate_block to check that the block was written properly but I need to check that this would be sufficient. We would then want to delete the block if it was not properly written, similar to the catch-all below.

EDIT: After looking more closely I think the best solution would be to call `validate_block_blocking` in this branch of the match statement and delete the block if this returns an error.

_Originally posted by @idky137 in https://github.com/zingolabs/zaino/pull/778#discussion_r2769596537_

### LINKED ISSUE #837: db modifications from pr #778 need documentation
One more addition I think we should add is updating the database CHANGELOG and the migration notes in migrations.rs with the change to write block (in the v1.0.0->1.1.0 migration).

_Originally posted by @idky137 in https://github.com/zingolabs/zaino/issues/778#issuecomment-3854738042_

idky137:
Fixes bug in database write leading to partially written blocks, corrupting database tables. Updates database changelog and migration documentation. (Also ran cargo fmt --all)

Closes #836 
Closes #837 

NOTE: The changes in `zaino_proto/src/proto/service.rs` and `zaino_state/src/error.rs` are all from running `cargo fmt --all`

### DIFF
--- a/zaino-proto/src/proto/service.rs
+++ b/zaino-proto/src/proto/service.rs
@@ -380,10 +380,10 @@ pub mod compact_tx_streamer_client {
         dead_code,
         missing_docs,
         clippy::wildcard_imports,
-        clippy::let_unit_value,
+        clippy::let_unit_value
     )]
-    use tonic::codegen::*;
     use tonic::codegen::http::Uri;
+    use tonic::codegen::*;
     #[derive(Debug, Clone)]
     pub struct CompactTxStreamerClient<T> {
         inner: tonic::client::Grpc<T>,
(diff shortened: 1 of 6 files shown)


idky137:
NOTE: If we release zaino before this PR lands it should be updated so that the database updates go into the v1.1.0->v1.2.0 migration (instead of the v1.0.0->1.1.0 migration currently on dev).

nachog00 (approved):

PR #864, opened 2026-02-24
Release Blocker Bug hot fixes

### LINKED ISSUE #865: Zaino tries to store the complete chain in memory.
### Usage mode

Both / Not sure

### Describe the bug

Currently the NodeBackedChainIndex tries to fully sync the non-finalised before it starts syncing the finalised state.

This is made worse because the non-finalised state currently tries to hold blocks down to the finalised state's tip, meaning on first time start up the non-finalised state will try to hold the whole chain in memory, leading to excessive memory requirements.

There are a few options for fixing this:
- Sync finalised state first: This will fix the memory leak but will mean that the non-finalised state doesn't sync the non-finalised state until the finalised state is fully synced, meaning the pass-through will not operate properly.
- Sync non-finalised state first: this will fix the memory error but required several other changes in the non-finalised state; specifically nfs should only hold a maximum of 101 blocks, nfs should not sync from origin as this is extremely slow. however, if the nfs does not sync from genesis we will need a way to calculate the cumulative chain work at the point the nfs starts syncing from  (as this currently comes from the finalised state).
- Sync both concurrently. NOTE the nfs should still be limited to 101 block in this solution.

### Steps to reproduce

Start Zainod with default config.

### Expected behavior

Zainod syncs correctly

### Environment

Dev latest

### Configuration

```toml

```

### Logs

```shell

```

### Additional context

_No response_

idky137:
**UPDATED PR COMMENT**

This PR fixes several bugs in the ChainIndex that are blocker to release. The following is a summary of the bugs encountered testing ZainoD with testnet and the fixes implemented.

**BUG 1**
Bug:
 NodeBackedChainIndex::start_sync_loop tries to fully sync the non-finalised state before syncing the finalised state. This leads to ZainoD trying to store the whole chain in memory before starting to move block to disk, this is because the non-finalised state sync loop is implemented to always hold blocks down to the finalised tip.

Fixes:
This PR changes this behaviour so that the finalised_state is synced first, then the non_finalised state is synced.

Additional changes:
- finalised state assumed sapling / orchard commitment trees exist, this is not the case before the relevant network upgrades are active. the finalised state now stores empty commitment tree data when below the relevant height.
- NodeBackedChainIndex now uses ZainoDB::sync_to_height instead of write block.

**BUG 2**
Bug:
Non-finalised state initialises ether from the finalised tip or genesis if not present, then in sync it starts syncing from this point. This is an issue on first time startup, leading to the non-finalised state syncing from genesis, not the finalised tip.

Fixes:
Non-finalised state sync now checks the finalised tip in loop.

**BUG 3**
Bug: 
ZainoDB is intermittently hitting an error when writing blocks to the DB: The database is not finding the correct output that a tx in the current block spent. This leads to block write failing due to the block looking invalid from the batadase's point of view. This was not causing write failure however, just a warning.

This was actually made up of several bugs:
Bug 3a:
Write_block was not failing and deleting corrupt data from the DB, leading to a corrupt DB fle.

Cause:
In write_block the write failures were not properly propagated and were mixed with tokio task errors leading to missed failed writes.

Fix:
Properly handle tokio task error.

Bug 3b:
Write_block was randomly failing when given valid data.

Cause: 
The write_block method used multiple LMDB transactions to write the given block to the DB, the problem was that one of these writes sometimes required data from the current block to already be written to the DB, and there was a race condition happening where sometimes the previous write transaction that included the required data had not yet properly synced to the DB, meaning that the required data was not present.

This race condition could happen when a tx in a block spent an output that was created in the same block.

Fix:
Write_block (and delete_block) now uses a single LMDB transaction to write (or delete) all block data, removing this race.

Additional updates:
In several places explicit errors are now returned where before they were not, hardening the database implementation.

Several extra warn debugs were added (on deleting corrupt block data).

The mark_spent and mark_unspent were updated to use more explicit byte matching.

### DIFF
--- a/zaino-state/src/chain_index/finalised_state/db/v1.rs
+++ b/zaino-state/src/chain_index/finalised_state/db/v1.rs
@@ -1192,7 +1192,6 @@ impl DbV1 {
             config: self.config.clone(),
         };
         let join_handle = tokio::task::spawn_blocking(move || {
-            // let post_result = tokio::task::spawn_blocking(move || {
             // Write block to ZainoDB
             let mut txn = zaino_db.env.begin_rw_txn()?;
 
@@ -1245,11 +1244,7 @@ impl DbV1 {
                 WriteFlags::NO_OVERWRITE,
             )?;
 
-            txn.commit()?;
-
             // Write spent to ZainoDB
-            let mut txn = zaino_db.env.begin_rw_txn()?;
-
             for (outpoint, tx_location) in spent_map {
                 let outpoint_bytes = &outpoint.to_bytes()?;
                 let tx_location_entry_bytes =
@@ -1262,11 +1257,7 @@ impl DbV1 {
                 )?;
             }
 
-            txn.commit()?;
-
             // Write outputs to ZainoDB addrhist
-            let mut txn = zaino_db.env.begin_rw_txn()?;
-
             for (addr_script, records) in addrhist_outputs_map {
                 let addr_bytes = addr_script.to_bytes()?;
 
@@ -1294,8 +1285,6 @@ impl DbV1 {
                 }
             }
 
-            txn.commit()?;
-
             // Write inputs to ZainoDB addrhist
             for (addr_script, records) in addrhist_inputs_map {
                 let addr_bytes = addr_script.to_bytes()?;
@@ -1317,14 +1306,12 @@ impl DbV1 {
                 for (_record, record_entry_bytes, (prev_output_script, prev_output_record)) in
                     stored_entries
                 {
-                    let mut txn = zaino_db.env.begin_rw_txn()?;
                     txn.put(
                         zaino_db.address_history,
                         &addr_bytes,
                         &record_entry_bytes,
                         WriteFlags::empty(),
                     )?;
-                    txn.commit()?;
 
                     // mark corresponding output as spent
                     let prev_addr_bytes = prev_output_script.to_bytes()?;
@@ -1334,7 +1321,8 @@ impl DbV1 {
                         })?;
                     let prev_entry_bytes =
                         StoredEntryFixed::new(&prev_addr_bytes, packed_prev).to_bytes()?;
-                    let updated = zaino_db.mark_addr_hist_record_spent_blocking(
+                    let updated = zaino_db.mark_addr_hist_record_spent_in_txn(
+                        &mut txn,
                         &prev_output_script,
                         &prev_entry_bytes,
                     )?;
@@ -1353,6 +1341,8 @@ impl DbV1 {
                     }
                 }
             }
+
+            txn.commit()?;
 
             zaino_db.env.sync(true).map_err(|e| {
                 FinalisedStateError::Custom(format!("LMDB sync failed before validation: {e}"))
@@ -1385,7 +1375,7 @@ impl DbV1 {
                 tokio::task::block_in_place(|| self.env.sync(true))
                     .map_err(|e| FinalisedStateError::Custom(format!("LMDB sync failed: {e}")))?;
                 self.status.store(StatusType::Ready);
-                if block.index().height().0 % 100 == 0 {
+                if block.index().height().0.is_multiple_of(100) {
                     info!(
                         "Successfully committed block {} at height {} to ZainoDB.",
                         &block.index().hash(),
@@ -1753,7 +1743,6 @@ impl DbV1 {
                     Err(e) => return Err(FinalisedStateError::LmdbError(e)),
                 }
             }
-            let _ = txn.commit();
 
             // Delete addrhist input data and mark old outputs spent in this block as unspent
             for (addr_script, records) in &addrhist_inputs_map {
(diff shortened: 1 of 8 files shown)


idky137:
Possible fix for https://github.com/zingolabs/zaino/issues/865

AloeareV:
This is a partial fix, as syncing non finalized state currently goes from where non-finalized state left off and not from the top of finalized state. (on first load it's _initialized_ from the finalized tip, but this is before fs sync triggers). I have a fix for this, which I will push assume it doesn't break any tests!

idky137:
Im not sure if you already added the fix but from what I can see the nfs (non-finalised state) updates the point it starts syncing from in the sync method based on the fs's (finalised state) tip. So even though the nfs stores the origin block on initialisation when the fs is empty it resets to the fs's tip it before ever starting to sync. I may be wrong but it now does look safe.

https://github.com/idky137/zaino/blob/63f876f1644b67098075074fc7fb6edc85208c27/zaino-state/src/chain_index/non_finalised_state.rs#L340

I still havent managed to test running on testnet as my zebrad testnet node is taking a very long time to sync but all tests pass locally for me on this branch now. I will carry on with further testing once my testnet node syncs..

EDIT: I just re-examined the diff and this is new code. this all looks good to me now! I think someone other than me should probably approve and land this though..

idky137:
Current error I am seeing:

`
2026-02-26T16:28:37.453543751Z  INFO zaino_state::chain_index::finalised_state::db::v1: Successfully committed block 000155416b4b1fa3300a56128afc6b5799d14335e5c640c775a67d6ea133f42d at height 4900 to ZainoDB.

2026-02-26T16:28:39.337431775Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.338302444Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.338992313Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.339677863Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.340379364Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.340745954Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.341147951Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.341595673Z  WARN zaino_state::chain_index::finalised_state::db::v1: invalid block @ height 4988 (hash 0000ad1907f6140cd5121b37e26cdf2c0270786983c4c08ce930c49df93f91d8): missing addrhist input record

2026-02-26T16:28:39.667964375Z  INFO zaino_state::chain_index::finalised_state::db::v1: Successfully committed block 00001192e17b94da26728c17479fe9b842742eeb097ac7a7b31ce0e0735dba68 at height 5000 to ZainoDB.
`

this looks like a block is being written incorrectly to the finalised state but the fallback block deletion is never triggering. 

The worry is that these blocks are being incorrectly written and not deleted is that the data zainodb uses to serve downstream clients may be incorrect.

There is also a chance that these blocks are never being written incorrectly, but there is a race going on in the db between write_block and validate_block, so the validation cannot see the new block that has been written.

However even if this is the case then the block that fails validation should be deleted, not accepted.

idky137:
With the updates to the DB the following unit test is now failing:

`chain_index::tests::finalised_state::v1::delete_blocks_from_db`

It would be worth running the full suite of unit tests to check this behaviour.

zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -1245,11 +1244,7 @@ impl DbV1 {
                 WriteFlags::NO_OVERWRITE,
             )?;
 
-            txn.commit()?;
-
             // Write spent to ZainoDB
```
AloeareV, line 1247: Is this comment now stale?

AloeareV (approved):
Tests all pass locally (with some flakes, as normal)!  Testnet seems to now be syncing smoothly. All the new changes look good.

PR #871, opened 2026-02-25
fix(docker): use ENTRYPOINT/CMD split with privilege dropping

### LINKED ISSUE #872: Docker image CMD doesn't work with new CLI subcommand requirement
The CLI now requires a subcommand (e.g. `zainod start`), but the Docker image only has `CMD ["zainod"]` which just prints help and exits.

Fixed by #871

### LINKED ISSUE #873: Docker volume mounts fail due to permission issues
When mounting volumes to the container, permission errors occur because:
- Container runs as non-root user (UID 1000)
- Mounted volumes may be owned by different UIDs
- No mechanism to fix ownership at startup

The zebra image solves this by starting as root, fixing permissions, then dropping privileges.
https://github.com/ZcashFoundation/zebra/blob/main/docker/entrypoint.sh

Fixed by #871

### LINKED ISSUE #874: Cookie path default uses ad-hoc XDG handling instead of xdg module
`default_ephemeral_cookie_path()` in `zainod/src/config.rs` manually checks `XDG_RUNTIME_DIR` instead of using the centralized xdg module. This duplicates logic and could lead to inconsistencies.

The xdg module should support `XDG_RUNTIME_DIR` for runtime files like cookies.

Fixed by #871

### LINKED ISSUE #875: Document Docker image usage and volume permissions
The Docker image needs documentation covering:

## Basic usage
- Running with default settings
- Using subcommands (start, generate-config, --help)
- Configuration via environment variables

## Volume mounts
- Default XDG paths the image expects
- How to mount config files (read-only supported)
- How to persist database/cache data

## Permission handling
- The entrypoint runs as root, creates directories, then drops to UID 1000
- Default paths are handled automatically
- Custom paths (set via TOML config) require user to ensure directories exist with appropriate permissions
- Read-only mounts are supported (entrypoint skips chown gracefully)

## Examples
- docker run
- docker-compose
- Kubernetes ConfigMap/Volume patterns

Blocker for #842

### LINKED ISSUE #876: Docker: XDG paths are not ergonomic for volume mounting
## Problem

Zaino follows the XDG Base Directory specification for configuration and data paths:
- Config: `~/.config/zaino/`
- Data: `~/.cache/zaino/`

This makes sense for **native installations** (running the binary directly on a host machine without containers), where XDG paths provide standard, predictable locations that integrate well with the user's environment.

However, when running Zaino in Docker, these paths become awkward for volume mounts:

```yaml
volumes:
  - ./config:/home/container_user/.config/zaino:ro
  - zaino-data:/home/container_user/.cache/zaino
```

### Issues with XDG paths in Docker

1. **Long, verbose paths** - `/home/container_user/.config/zaino` is unwieldy
2. **Hidden directories** - `.config` and `.cache` are hidden by convention, which is odd for explicit mount points
3. **Coupled to user name** - `container_user` in the path couples the compose file to internal container details
4. **Not intuitive** - Users expect simple paths like `/app/config` or `/data` for containers

### Desired compose experience

```yaml
volumes:
  - ./config:/app/config:ro
  - zaino-data:/app/data
```

## Related

Part of Docker integration improvements.

nachog00:
## Summary

The CLI now requires a subcommand (e.g. `zainod start`). This PR:

- Adds `entrypoint.sh` that creates writable directories with correct ownership before dropping privileges to the non-root user
- Uses ENTRYPOINT for the script and CMD for the default subcommand
- Follows zebra's pattern for handling volume mount permissions
- Gracefully handles read-only mounts (skips chown instead of failing)
- Adds `XDG_RUNTIME_DIR` support to the xdg module for cookie paths
- Refactors `default_ephemeral_cookie_path()` to use the xdg module
- Creates symlinks for ergonomic Docker mount paths (`/app/config`, `/app/data`)
- Adds Docker usage documentation

## Docker usage

```bash
docker run zaino                  # runs zainod start
docker run zaino --help           # shows help
docker run zaino generate-config  # generates config
```

## Volume mounts

The container provides simple mount points that symlink to XDG paths internally:

| Purpose | Mount Point | Symlinks to |
|---------|-------------|-------------|
| Config | `/app/config` | `~/.config/zaino` |
| Data | `/app/data` | `~/.cache/zaino` |

Example docker-compose:
```yaml
services:
  zaino:
    image: zaino:latest
    volumes:
      - ./config:/app/config:ro
      - zaino-data:/app/data
```

## Volume permissions

The entrypoint handles permissions automatically for `/app/*` paths. Docker users just mount to these paths and everything works.

Custom XDG paths configured via TOML are only relevant for native installations—Docker users don't need to care about internal paths since the symlinks handle it transparently.

## Testing

Local permission tests:
```bash
./test_environment/test-docker-permissions.sh zaino:latest
```

Tests cover:
- Basic commands (help, version, generate-config, start --help)
- Volume mounts to `/app/config` and `/app/data`
- File ownership verification (UID 1000)
- Read-only config mounts
- Root-owned directories (if sudo available)

### Why local-only tests?

Testing Docker volume permissions in CI is impractical because permission edge cases (root-owned mounts, UID mismatches) require privileged operations that CI runners typically don't allow or behave differently across environments. The test script is designed to run locally and will automatically run additional root-owned directory tests when executed with sudo access available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,7 +40,7 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
     fi
 
 ############################
-# Runtime (slim, non-root)
+# Runtime
 ############################
 FROM debian:bookworm-slim AS runtime
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
@@ -50,33 +50,35 @@ ARG GID
 ARG USER
 ARG HOME
 
-# Only the dynamic libs needed by a Rust/OpenSSL binary
+# Runtime deps + setpriv for privilege dropping
 RUN apt-get -qq update && \
     apt-get -qq install -y --no-install-recommends \
-      ca-certificates libssl3 libgcc-s1 \
+      ca-certificates libssl3 libgcc-s1 util-linux \
     && rm -rf /var/lib/apt/lists/*
 
-# Create non-root user
+# Create non-root user (entrypoint will drop privileges to this user)
 RUN addgroup --gid "${GID}" "${USER}" && \
     adduser  --uid "${UID}" --gid "${GID}" --home "${HOME}" \
              --disabled-password --gecos "" "${USER}"
 
+# Make UID/GID available to entrypoint
+ENV UID=${UID} GID=${GID} HOME=${HOME}
+
 WORKDIR ${HOME}
 
-# Copy the installed binary from builder
+# Copy binary and entrypoint
 COPY --from=builder /out/bin/zainod /usr/local/bin/zainod
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+RUN chmod +x /usr/local/bin/entrypoint.sh
 
-RUN mkdir -p .cache/zaino
-RUN chown -R "${UID}:${GID}" "${HOME}"
-USER ${USER}
-
-# Default ports (adjust if your app uses different ones)
+# Default ports
 ARG ZAINO_GRPC_PORT=8137
 ARG ZAINO_JSON_RPC_PORT=8237
 EXPOSE ${ZAINO_GRPC_PORT} ${ZAINO_JSON_RPC_PORT}
 
-# Healthcheck that doesn't assume specific HTTP/gRPC endpoints
 HEALTHCHECK --interval=30s --timeout=5s --start-period=15s --retries=3 \
   CMD /usr/local/bin/zainod --version >/dev/null 2>&1 || exit 1
 
-CMD ["zainod"]
+# Start as root; entrypoint drops privileges after setting up directories
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
+CMD ["start"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,6 +66,14 @@ ENV UID=${UID} GID=${GID} HOME=${HOME}
 
 WORKDIR ${HOME}
 
+# Create ergonomic mount points with symlinks to XDG defaults
+# Users mount to /app/config and /app/data, zaino uses ~/.config/zaino and ~/.cache/zaino
+RUN mkdir -p /app/config /app/data && \
+    mkdir -p ${HOME}/.config ${HOME}/.cache && \
+    ln -s /app/config ${HOME}/.config/zaino && \
+    ln -s /app/data ${HOME}/.cache/zaino && \
+    chown -R ${UID}:${GID} /app ${HOME}/.config ${HOME}/.cache
+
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -50,33 +50,43 @@ ARG GID
...
 ARG ZAINO_JSON_RPC_PORT=8237
 EXPOSE ${ZAINO_GRPC_PORT} ${ZAINO_JSON_RPC_PORT}
 
-# Healthcheck that doesn't assume specific HTTP/gRPC endpoints
 HEALTHCHECK --interval=30s --timeout=5s --start-period=15s --retries=3 \
   CMD /usr/local/bin/zainod --version >/dev/null 2>&1 || exit 1
 
-CMD ["zainod"]
+# Start as root; entrypoint drops privileges after setting up directories
+ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
```
pacu, line 91: Idk if this is usual practice but it doesn't make sense to me to see a bash file in a bin directory

zaino-common/src/xdg.rs
```diff
@@ -39,6 +39,12 @@ pub enum XdgDir {
     ///
     /// Default: `$HOME/.cache`
     Cache,
+
+    /// `XDG_RUNTIME_DIR` - Runtime files (sockets, locks, cookies).
```
pacu, line 43: I don't have a suggestion for this but probably there's a library that already handles XDG paths for us and most of this code would not be needed

pacu (approved):
Untested but LGTM

zancas (approved):
This is sufficient.

Future potential improvements:

* use standard xdg tools
* migrate more to podman
* separate concerns such that different usage patterns are cleanly supported.

zaino-common/src/xdg.rs
nachog00: that's possible yes... code wasn't all that complex so i wiped it out and avoided an extra dep, but a dedicated library would probably consider caveats i didn't even consider.
See #878 wher this is mentioned.

Dockerfile
nachog00: in my experience that directory has been a place to put executable meant for the current user, and it's usually findable in PATH... but now that you mention it, i did a quick search, and it seems like a more common practice in these cases is to put the entrypoint script right at `/entrypoint.sh`, right at `/`.
At first glance it seems odd, but it is true that it makes it clear that it's not a thing meant to be used by the container itself, but more of an infrastructure thing (if you start an interactive session in the container, you wont be able to ever run that script unless it's running as root).

PR #877, opened 2026-02-25
remove in-house user permissions, doc explicit run requirement

zancas:
This simpler version of a container for zainod doesn't manage host filesystem permissions (assuming it's properly run with `--userns=keep-id`).


Containerfile
```diff
@@ -43,50 +36,21 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
...
-EXPOSE ${ZAINO_GRPC_PORT} ${ZAINO_JSON_RPC_PORT}
+EXPOSE 8137 8237
 
 HEALTHCHECK --interval=30s --timeout=5s --start-period=15s --retries=3 \
   CMD /usr/local/bin/zainod --version >/dev/null 2>&1 || exit 1
 
-# Start as root; entrypoint drops privileges after setting up directories
-ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
-CMD ["start"]
+# Run as non-root. The caller MUST map their host user:
```
nachog00, line 58: what happens if the user forgets to do this mapping? I think the container runs completely as root doesn't it? Not sure it's better than the script which ensures privilege dropping.
It also complicates the UX.

zancas: I'm using podman, I'll have to do more research, but I don't think it runs as root by default.

Dockerfile
```diff
@@ -43,50 +36,21 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
...
 HEALTHCHECK --interval=30s --timeout=5s --start-period=15s --retries=3 \
   CMD /usr/local/bin/zainod --version >/dev/null 2>&1 || exit 1
 
-# Start as root; entrypoint drops privileges after setting up directories
-ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
-CMD ["start"]
+# Run as non-root. The caller MUST map their host user:
+#   podman: podman run --userns=keep-id zaino
+#   docker: docker run --user "$(id -u):$(id -g)" zaino
+CMD ["zainod", "start"]
```
nachog00: this breaks the documented usage. The entrypoint script, besides all the directories stuff, also replaces itself with `zainod` in the end. the  `CMD ["start"]` sets `start` as the default command to `zainod` but also enables usage like:
```
docker run zaino generate-config
```
instead of 
```
docker run zaino zainod generate-config
```
which is overly unintuitive.

zancas:
Here's the sequence of commands I ran to use your Dockerfile:

```BASH
  podman build -t zaino .

  podman run --userns=keep-id -v ./zainod_config:/home/zaino/.config/zaino zaino generate-config

  podman run --userns=keep-id -v ./zainod_config:/home/zaino/.config/zaino -p 8137:8137 -p 8237:8237 zaino
```

This flow isn't intended to support complex deployments, and perhaps could be separated as a different build-stage, or otherwise differentiated from flows that wire in privilege de-escalation.

I think this example is useful for simpler, low-privilege, non-docker containerization strategies.

Dockerfile
nachog00: if the whole script was to be removed in the end, the `ENTRYPOINT` line should remain and the whole thing should be:
```Dockerfile
ENTRYPOINT ["zainod"]
CMD ["start"]
```

Containerfile
nachog00: That could discard podman as susceptible to this. What about docker?

Dockerfile
zancas: I think I ran like this:

```BASH
podman run --userns=keep-id -v ./zainod_config:/home/zaino/.config/zaino zaino generate-config
```

So there was no `zaino zainod` repetition.

Containerfile
zancas: I waved this wand:

```
podman info --format '{{.Host.Security.Rootless}}'
```
and pulled this rabbit out of a hat:

```
true
```

Dockerfile
zancas: But.. that shouldn't have worked..

zancas: Huh..   that *is* what it is locally..  checking commits..

zancas: OK, good catch, commit pushed.

PR #884, opened 2026-02-26
ignore generated file lints

### LINKED ISSUE #883: The service.rs file generated by proto wasn't linted
### What's the smell?

Somehow this unlinted file was checked in, possibly the lint rules have been updated upstream.

### Location

zaino-proto

### Category

Other

### Severity

High - Actively causing problems

### Suggested fix

Once fixed commits won't fail CI due to this.

zancas:
Fixes: https://github.com/zingolabs/zaino/issues/883

clippy and fmt ignore autogenerated files


pacu (changes requested):
Autogenerated files must to be auto formatted neither "clippy'd"

zaino-proto/src/proto.rs
```diff
@@ -1,9 +1,12 @@
 //! Holds tonic generated code for the lightwallet service RPCs and compact formats.
```
pacu, line 1: I think this file will be regenerated when running cargo build as well

pacu (dismissed):

zaino-proto/src/proto.rs
zancas: I think it's these three lines that overwrite the Rust:

https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-proto/build.rs#L35

https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-proto/build.rs#L77

https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-proto/build.rs#L83

fluidvanadium (approved):
send

Issue #856, opened 2026-02-19
Zaino Observability Metrics discussion


pacu:
### Usage mode

zainod (running the daemon)

### Problem or use case

- What do we need to track? 
- What metrics do we want?
- Zaino doesn’t have any sort of metrics framework so it will be helpful to align those with whatever Zebra folks have in mind. 
- Str4d put together these wallet syncing metrics https://bencher.dev/perf/zcash-speed/plots which are run on CI.   can we do something similar to this on Zaino? 
- What's the goal of these metrics? is it catching regressions? optimizations? just health metrics?

### Proposed solution

No proposal. Just discuss this!

### Area

Other

nachog00:
At first thought, i think there are 2 clear metrics categories: static/benchmark and live runtime metrics.

For daemon systems like zebra and zaino, that is, ever running, reporting on the go is key to keep awareness of the alive beast. They would serve extra metrics endpoints for this.

Static/benchmark could be ran periodically in CI, but are less of an operational feature i think.

Issue #598, opened 2025-10-10
Uncontrollable test printout.


idky137:
Zaino print the following line multiple times throughout test runs, polluting test debug outputs.

`TEST_BINARIES_DIR environment variable is not set. It should contain the necessary executables for a test run.`

idky137:
Stale

Issue #640, opened 2025-10-31
Inline handling of "order susceptible" fields


nachog00:
This line reminds me of the problem we have with txids being reversed in some places... we probably want either our won Hash type or the BlockInfo type to implement dedicated methods to construct from "order susceptible" sources, so that we don't depend on a line on an rpc endpoint for this critical bit.

I wouldn't block on this for this PR, but maybe leave a `Todo!` mentioning the problem?

_Originally posted by @nachog00 in https://github.com/zingolabs/zaino/pull/632#discussion_r2481803648_

nachog00:
#634 is the intended fix to the "reversed txids" problem

nachog00:
Trying to recapitulate. The general case of this problem seems to be:

Type `Foo` is some hash/string and it has a canonical orientation. But it also has a common reversed representation which is used in some contexts, typically associated with a **"display order"** idea.

The problem seems to be that conversions between the two orientations are common in interfaces, and it's hard to be sure of which is the right one in any given context, as well as telling which order IS being assumed in a given context.

My hypothesis is that this problem is exacerbated by the fact that the decision of flipping the order or not is being made ad-hoc on the interfaces, with little formal definition of `ORDER(layerA)` vs `ORDER(layerB)`.

The typical case, AFAIK looks something like this:

```rust
pub struct Foo(pub [u8; 32]);

impl Foo {
    pub fn bytes_in_display_order(&self) -> [u8; 32] { ... }

    pub fn from_bytes_in_display_order(bytes_in_display_order: &[u8; 32]) -> Foo { ... }
}
```

Then when we get `Foo` data from sources A and B, and maybe send `Foo` data to sources C and D, we do:

```rust
let foo_from_A: Foo = Foo::new(response_from_source_A);
let foo_from_B: Foo = Foo::from_bytes_in_display_order(&response_from_source_B);

let foo_from_db: Foo = read_foo(...);
let foo_to_C: [u8; 32] = foo_from_db.0;
let foo_to_D: [u8; 32] = foo_from_db.bytes_in_display_order();
```

This implicitly encodes that source A deals with `Foo` in the "normal" order and source B does so in the "reversed" order. Likewise for C and D.

## Why This Approach Is Problematic

This seems fine and sufficient, but once we need to perform these transformations coherently on multiple interfacing points, then it's really easy to end up with ad-hoc encoding which is:

[...]

PR #888, opened 2026-02-26
feature/improved logging

nachog00:
## Summary

Adds centralized logging infrastructure to `zaino-common` with configurable output formats (stream, tree, JSON) controlled via environment variables (`RUST_LOG`, `ZAINO_LOG_FORMAT`, `ZAINO_LOG_COLOR`). The logging module is now used consistently across `zainod` and `zaino-testutils`.

This PR also introduces `NamedAtomicStatus` for observable component status (logged at debug level), adds tracing instrumentation to key functions in `zaino-state` and `TestManager`, and improves log readability with compact timestamps, truncated hash displays, and structured tracing fields.

Container tests now pass through `RUST_LOG` and `ZAINO_LOG_FORMAT` so developers can control log verbosity and format during test runs.

See `docs/logging.md` for usage examples.

## Related Issues

This is foundational work toward #856 (Observability Metrics) by establishing consistent, structured logging patterns. It also partially addresses the concerns in #598 (Uncontrollable test printout) by giving developers control over log output during tests via environment variables.


zaino-state/src/status.rs
```diff
@@ -39,3 +42,54 @@ impl Status for AtomicStatus {
...
 }
+
+/// Thread-safe status wrapper with component name for observability.
+///
+/// Unlike [`AtomicStatus`], this variant logs all status transitions,
+/// making it easier to trace component lifecycle during debugging.
+#[derive(Debug, Clone)]
+pub struct NamedAtomicStatus {
+    name: &'static str,
+    inner: Arc<AtomicUsize>,
```
nachog00, line 53: This might be wrong, now that i see it, I think it should be using `AtomicStatus` as an inner, not just `AtomicUsize` ...?

nachog00: well actually, it looks more like NamedAtomicStatus is a full on replacement of AtomicStatus, it could have been a simple feature added to it... now i think we should remove one and probably keep the old name (AtomicStatus) as it having a name is quite irrelevant if there's a single variant of this

zaino-state/src/chain_index/types/db/legacy.rs
```diff
@@ -208,7 +214,13 @@ impl TransactionHash {
 
 impl fmt::Display for TransactionHash {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.write_str(&self.encode_hex::<String>())
+        // Show truncated hash for cleaner logs (first 8 hex chars = 4 bytes)
```
pacu, line 239: I would double check that no other part of Zaino depends on fmt::Display before applying this system-wide change for such a specific requirement.

zaino-state/src/chain_index/types/db/legacy.rs
```diff
@@ -73,7 +73,13 @@ impl BlockHash {
 
 impl fmt::Display for BlockHash {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.write_str(&self.encode_hex::<String>())
+        // Show truncated hash for cleaner logs (first 8 hex chars = 4 bytes)
```
pacu, line 87: I would double check that no other part of Zaino depends on fmt::Display before applying this system-wide change for such a specific requirement.

pacu (dismissed):
I think these changes look OK. Modulo the fmt::Display changed is verified.

zaino-state/src/chain_index/types/db/legacy.rs
AloeareV: Nothing _should_ be depending on exact display implementations.

AloeareV (commented):
Unfortunately, there's at least one load-bearing `to_string()`: https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-state/src/chain_index/source.rs#L178

AloeareV:
> Unfortunately, there's at least one load-bearing `to_string()`:
> 
> https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-state/src/chain_index/source.rs#L178

If I merge this PR into #864, zainod fails on 
```
thread 'main' (910691) panicked at zaino-state/src/backends/fetch.rs:144:14:
called `Result::unwrap()` on an `Err` value: InvalidNodeData(Unrecoverable("unexpected error response from server: RPC Error (code: -8): parse error: could not convert the input string to a hash or height"))
``` 
which I believe to be due to sending a truncated block hash to the validator.

nachog00:
> > Unfortunately, there's at least one load-bearing `to_string()`:
> > https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-state/src/chain_index/source.rs#L178
> 
> If I merge this PR into #864, zainod fails on
> 
> ```
> thread 'main' (910691) panicked at zaino-state/src/backends/fetch.rs:144:14:
> called `Result::unwrap()` on an `Err` value: InvalidNodeData(Unrecoverable("unexpected error response from server: RPC Error (code: -8): parse error: could not convert the input string to a hash or height"))
> ```
> 
> which I believe to be due to sending a truncated block hash to the validator.

I think this is being called on zebra types, not ours... whether that's the right contract to rely on them or not for this usage... not sure. Do they advertise Display stability? sounds like we might be looking at a similar problem just not related to the Display changes in this PR.

nachog00:
> > Unfortunately, there's at least one load-bearing `to_string()`:
> > https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-state/src/chain_index/source.rs#L178
> 
> If I merge this PR into #864, zainod fails on
> 
> ```
> thread 'main' (910691) panicked at zaino-state/src/backends/fetch.rs:144:14:
> called `Result::unwrap()` on an `Err` value: InvalidNodeData(Unrecoverable("unexpected error response from server: RPC Error (code: -8): parse error: could not convert the input string to a hash or height"))
> ```
> 
> which I believe to be due to sending a truncated block hash to the validator.

As i stated above, that `id.to_string()` is called on zebra, but i did identify a bunch of callsites of `to_string` on the affected type on this PR. they were all in `chain_index/tests/finalised_state/v1.rs`

nachog00:
I updated all `.to_string` calls to `.hex_encode`. Display on thus type should be free to change now.

as i mentioned in #896 , taking on the work planned in #640 should come right after and will make us way more robust against this sort of situations.

AloeareV:
> > > Unfortunately, there's at least one load-bearing `to_string()`:
> > > https://github.com/zingolabs/zaino/blob/20d28c0e81b88a746631fa77de1dc4060aa53ca6/zaino-state/src/chain_index/source.rs#L178
> > 
> > 
> > If I merge this PR into #864, zainod fails on
> > ```
> > thread 'main' (910691) panicked at zaino-state/src/backends/fetch.rs:144:14:
> > called `Result::unwrap()` on an `Err` value: InvalidNodeData(Unrecoverable("unexpected error response from server: RPC Error (code: -8): parse error: could not convert the input string to a hash or height"))
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which I believe to be due to sending a truncated block hash to the validator.
> 
> I think this is being called on zebra types, not ours... whether that's the right contract to rely on them or not for this usage... not sure. Do they advertise Display stability? sounds like we might be looking at a similar problem just not related to the Display changes in this PR.

Yeah, this is a zebra type, not a zaino type!

pacu (dismissed):
utACK!

AloeareV (approved):
New code looks great!

PR #897, opened 2026-03-02
sync dev with latest Rc/0.2.0 hotfixes

nachog00:


nachog00:
there's a problem with this flow i just realised about. pushing the `Update Branch` button forcibly merged dev into the rc branch... that means, applying hotfixes onto the rc branch to allow dev to keep accepting non rc changes is limited, cause the moment we want to update dev with the latest fixes, we then need to dump ALL of dev into the rc...

idky137:
> there's a problem with this flow i just realised about. pushing the `Update Branch` button forcibly merged dev into the rc branch... that means, applying hotfixes onto the rc branch to allow dev to keep accepting non rc changes is limited, cause the moment we want to update dev with the latest fixes, we then need to dump ALL of dev into the rc...

Ye this is why I push for having a more straightforward freeze on dev while we are in the release process as it gets complicated trying to maintain multiple branches and often extra things get added that were not meant to be (and may not be as tested as we would like for a stable release). It is also easy for subtle bug fixes to get lost when dealing with conflicts between the two branches.

pacu (approved):
utAck

AloeareV (approved):

PR #901, opened 2026-03-04
HOTFIX: Update / fix `addr_hist_records_by_addr_and_index_blocking`

### LINKED ISSUE #900: `addr_hist_records_by_addr_and_index_blocking` is CPU-bound (mdb_cursor_next hot loop) and cripples initial sync
**Current implementation**: https://github.com/zingolabs/zaino/blob/7da17274bfec7dc8bf7d99c1caa084a240710e2f/zaino-state/src/chain_index/finalised_state/db/v1.rs#L5420

### Problem
`addr_hist_records_by_addr_and_index_blocking` iterates every duplicate entry for an address (cursor.iter_dup_of(&addr_script_bytes)) and filters by tx_location in Rust. In practice this causes extremely large numbers of mdb_cursor_next steps and becomes the dominant CPU cost during initial sync.

### Symptoms observed
- perf shows ~70% of cycles in mdb_cursor_next and another ~15% in mdb_cursor_get / the LMDB cursor iterator wrapper; the function addr_hist_records_by_addr_and_index_blocking itself shows up in the top hotspots.
- strace -c shows heavy futex time (≈92% syscall time) and some fdatasync calls; overall behaviour indicates lock/coordination around very expensive cursor traversal rather than IO saturation.
- Process RSS is large (~50 GB) from touched mmapped pages (expected for LMDB), but disk IO rates and IO wait are low — the bottleneck is CPU work in LMDB cursor walking, not disk bandwidth or RPC.

### Why this is bad
The current pattern is effectively O(rows_scanned) per call, and when called repeatedly from the ingest loop it becomes multiplicative (N×M), drastically slowing initial sync and wasting CPU. It also increases contention and thread waiting.

### Proposed solution (high level only)
Change the function to perform a seek + bounded iteration instead of scanning all duplicates:

Seek once using a B-tree seek (SET_RANGE / CursorOp::SetRange) to address || block_height || tx_index (or the minimal key that lands at the first candidate).

Iterate only while the key prefix matches the requested address and while the block/tx index is within the small target range — stop as soon as the prefix changes or the index passes the target.

Ensure callers do not invoke a full scan per item: reuse a single read transaction & cursor across a block’s processing, coalesce lookups per block (in-block caching), or maintain a small “last seen” / UTXO index to avoid scanning old history repeatedly.

If the current key encoding doesn’t permit lexicographic address||height||tx_index ordering, migrate/address the schema so targeted seeks are possible.

### Acceptance criteria
- perf top during initial sync no longer shows mdb_cursor_next consuming the majority of cycles.
- Initial sync wall-clock time drops significantly (measurable speedup).
- Function behaviour/return values remain unchanged.

Priority
High — fixes a major initial-sync hotspot and reduces CPU cost for large DBs.

idky137:
Fix: make `addr_hist_records_by_addr_and_index` efficient + allow txn reuse

Context / issue
`addr_hist_records_by_addr_and_index_blocking` previously scanned the entire duplicate list for an address and filtered values in Rust (`cursor.iter_dup_of(&addr)`), causing excessive `mdb_cursor_next` walking, enormous page-cache pressure and very high CPU during initial sync. See issue: https://github.com/zingolabs/zaino/issues/900

What this PR does
- Algorithmic change:
  - Replace the full-duplicate scan with a B-tree seek + bounded iteration:
  - Build a minimal value-prefix: `[StoredEntryVersion, RecordVersion, height_be, tx_index_be]`.
  - Call `RoCursor::get(key, data, MDB_SET_RANGE)` to jump to the first duplicate ≥ `(height,tx_index)`.
  - Iterate duplicates with `MDB_NEXT_DUP`, collecting exact matches and stop when the duplicate ordering passes the target.

This makes lookups **O(log N + K)** instead of **O(N)**.

Transaction reuse
- Provide `addr_hist_records_by_addr_and_index_in_txn(&self, txn: &lmdb::RoTransaction<'_>, addr: &[u8], tx_location)` so callers can reuse an existing RO txn instead of creating a txn per lookup. This reduces txn/cursor overhead and reader-slot churn.

Preserve correctness
- Length checks and version checks are retained; the function still returns the same serialized `StoredEntryFixed<AddrEventBytes>` bytes for matches.

Testing
- Reproduced an initial-sync workload and captured the same diagnostics used during the original investigation:
  - `smaps_rollup` (memory residency)  
  - `vmstat` (system memory / IO)  
  - `pidstat` (per-process CPU / IO / faults)  
  - `iostat` (device IO)  
  - `/proc/<pid>/io` (process IO counters)  
  - `strace -c` (syscall time breakdown)  
  - `perf record` + `perf report` (hot functions with call-graph)

Results
- `smaps_rollup`: RSS / Private_Clean reduced from ~48 GiB → ~2.6 GiB — large reduction in touched mapped pages (NOTE: this is not fully representative as the DB had been syncing for an extended period before initial testing).
Clarification: the original very high memory numbers were primarily file-backed mmapped pages (Private_Clean), not heap allocations: the initial smaps_rollup showed Private_Clean ≈ 50,592,364 kB (≈48.2 GiB). That means the earlier high RSS was mostly safe — the process had touched many DB pages and the kernel retained them as file-backed clean pages — rather than a large anonymous/malloc footprint. The change now drastically reduces the number of touched mapped pages, which lowers page-pressure and page-fault churn even though the prior memory usage was not an unsafe heap growth.
- `perf` (before → after):
  - `mdb_cursor_next`: **~66.8% → 40.2%** (big drop)
  - `mdb_cursor_get`: **~7.0% → 19.9%** (increase, expected)
  - `addr_hist_records_by_addr_and_index_*`: **~7.85% → 13.8%** (function now concentrates work)
- `strace -c`: `futex` **~92% → ~98%** (higher fraction), `fdatasync` dropped (**~7.6% → ~2%**).
- `zainod` CPU: still **~190–195%**. Overall CPU remained high because the sync still performs many targeted seeks/gets (we removed the worst O(N) scanning but not the per-lookup work).

Interpretation
**Success:** The change removes the pathological full-scan behaviour (major memory and CPU win). The DB is no longer being touched at huge scale per lookup.

**Remaining work:** The cost shifted into many small seeks / cursor ops and into lock contention (`futex`). To reduce overall wall-clock time further, we must remove per-lookup overhead and contention by:

- **Cursor reuse:** add a `_with_cursor` helper or require callers to reuse a single `RoCursor` and `RoTransaction` for whole-block processing. This removes repeated cursor/txn setup and reduces futex contention.
- **Caller-side batching/caching:** during block processing, coalesce repeated lookups for the same address in a short-lived in-memory cache.
- **Profile futex stacks:** run `perf report --call-graph` to identify the exact mutex/futex hot frames and remove/relax the lock(s).
- **Optional index change:** add an auxiliary exact-key DB keyed by `addr || height_be || tx_index_be` if per-lookup seeks remain the dominant cost; that makes lookups true exact-key operations at low per-lookup cost (requires migration/backfill).

NOTE: This change seams to have sped up initial sync dramatically, with sync no longer stalling on individual blocks.

- Closes #900


AloeareV (approved):
Code looks good, I'm starting to get a feel for how this database works. Testing locally, I haven't noticed a significant sync speed increase as of testnet height 762785. I'm not sure why this would be different between machines, unless the speed increase only gets noticeable higher up the chain? Looking at the code, this improvement seems like it will matter more the larger the blockchain gets. 

Are you running a release build or a debug build?

idky137:
The slowdowns started happening for me between 1.5M-2.5M blocks into syncing. 

This problem gets worse as sync goes on as the function in question searches the transaction history of the given transparent address to validate and write the transparent indexes in that block.

Currently on dev this method searches the full history for the given address from oldest to newest, meaning that, as the histories of the transparent addresses we are dealing with grow, the inefficiency of this method grows.

It may be that I am testing on lighter weight hardware so the issue is more pronounced for me, it would definitely be nice to have a uniform testing environment for this purpose!

My experience before this change: (while syncing in the range of ~1.7M-2M blocks) individual blocks kept on massively slowing sync, with some blocks taking tens of minutes to sync.

My experience after this change: I have never had a block take longer than 1 minute (and this has been very rare, with the majority of "slow" blocks taking under 10 seconds to sync).

I have been running a release build.

Issue #907, opened 2026-03-12
There's more than one source of truth for the RPC spec.


zancas:
### Observation

The correct architecture is a single source of truth ecosystem wide.

I have a PR in flight that addresses this issue by reducing the dependencies in zaino-proto.

This will make it easy for zingolib to depend on zaino-proto as it's protobuf source of truth, eliminating skew.

### Location in codebase

zaino-proto

### Type of concern

Coupling / Dependencies

### Impact

The possibility of RPC skew increases the projects complexity.

### Suggested direction

I am opening a PR now.

zancas:
Fixed by #1259 as that eliminates zainos dependence on zingolib.

PR #908, opened 2026-03-12
makes optional deps MORE optional and default

zancas:
Addresses:  #907 

This change is backward compatible because optional dependencies are included by default.  The first non-zaino consumer, a test server, will depend on zaino-proto with default-features=false.

The upshot is unification of interface in a back-compatible way.

### DIFF
--- a/zaino-proto/src/proto/utils.rs
+++ b/zaino-proto/src/proto/utils.rs
@@ -2,9 +2,9 @@ use crate::proto::{
     compact_formats::{ChainMetadata, CompactBlock, CompactOrchardAction},
     service::{BlockId, BlockRange, PoolType},
 };
-#[cfg(feature = "zebra")]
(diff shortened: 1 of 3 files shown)


zaino-proto/src/proto/utils.rs
```diff
@@ -2,7 +2,9 @@ use crate::proto::{
     compact_formats::{ChainMetadata, CompactBlock, CompactOrchardAction},
     service::{BlockId, BlockRange, PoolType},
 };
+#[cfg(feature = "heavy")]
```
pacu, line 5: I think the feature name itself is not auto-explanatory. 
Also this should have a CHANGELOG entry so it's notes on release notes

pacu (changes requested):

zancas:
Obsolete.

PR #909, opened 2026-03-12
Minimal grpc server

zancas:
Continues to address: #907
Builds on #908

This allows a consumer to re-export types from zaino proto.

Because the re-exports are feature gated they only affect consumers that opt in.

Because they re-export they allow consumers from outside the workspace/repo to match the transitive dependency versions.


pacu (approved):
utACK

PR #911, opened 2026-03-12
Add triggering of zcash integration tests to PRs and merges to `dev`

nachog00:
same as #910

opening as a zingolabs owned branch PR so the workflow has access to repo secrets


zancas (approved):
LGTM!

PR #913, opened 2026-03-13
Fix slow get block range init

### LINKED ISSUE #912: DB get_compact_block_stream slow init time with very large block ranges.
Currently the `get_compact_block_stream` in DbV1 uses `validate_block_range`, checking that all heights in the range hold valid data.

Currently DbV1 does run this check on startup, once heights marked as valid this is a very efficient check:
```
fn is_validated(&self, h: u32) -> bool {
        let tip = self.validated_tip.load(Ordering::Acquire);
        h <= tip || self.validated_set.contains(&h)
    }
```
However, if `get_compact_block_stream` is called before the initial block scan has completed, then `validate_block_range` will call `validate_block_blocking` for every height not yet in the DB's validated set or above validated tip. This can be very slow for large un-validated block ranges and currently stops `get_compact_block_stream` returning any blocks until complete.

This should be updated to call `validate_block_blocking` for each height separately, so that  `get_compact_block_stream` does not stall for long periods, just has slightly slower performance when working with un-validated database heights.

idky137:
- Closes https://github.com/zingolabs/zaino/issues/912

Blocked by: (built on top of these Prs)
- https://github.com/zingolabs/zaino/pull/914
- https://github.com/zingolabs/zaino/pull/915

### DIFF
--- a/zaino-state/src/chain_index/finalised_state/db/v1.rs
+++ b/zaino-state/src/chain_index/finalised_state/db/v1.rs
@@ -3690,8 +3690,13 @@ impl DbV1 {
         end_height: Height,
         pool_types: PoolTypeFilter,
     ) -> Result<CompactBlockStream, FinalisedStateError> {
-        let (validated_start_height, validated_end_height) =
-            self.validate_block_range(start_height, end_height).await?;
+        // Do NOT validate the whole requested range up-front here.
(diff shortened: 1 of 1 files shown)


idky137:
tests pass locally

AloeareV (dismissed):

AloeareV (approved):

PR #914, opened 2026-03-13
removed DB v1.0.0 -> v1.1.0 migration

idky137:
Removes the DB v1.0.0->v1.1.0 migration.

This migration is not required as we have never released 1.0, and is making fixing bugs in DbV1 more complicated than it needs to be.


idky137:
tests pass locally

AloeareV (approved):

PR #921, opened 2026-03-18
move best_chaintip to chainindex method, implement passthrough to validator

AloeareV:


zancas (dismissed):
Let's add a regression test, shall we?

pacu (dismissed):
I would be ideal that this also creates an integration test that actually reproduces the scenario and verifies the fix. But if that's too costly we can resolve that later

fluidvanadium (approved):
Code reads correctly. Test runs.

zaino-state/src/chain_index/tests/proptest_blockgen.rs
```diff
@@ -229,6 +229,24 @@ fn passthrough_get_raw_transaction() {
...
+        assert_eq!(
+            tip.height.0,
+            mockchain
+                .best_branch()
+                .last()
+                .unwrap()
+                .coinbase_height()
+                .unwrap()
+                .0
+                .saturating_sub(100)
```
fluidvanadium, line 245: Wait, why do we subtract here??

Issue #318, opened 2025-05-09
Local Cache rework [ChainIndex].


idky137:
This issue will document the changes (planned and made) in the upcoming local cache rework.

The goals of this rework are:
- Address #249
- Fix #295.
- Implement block ranges for non-best chain data (eg. [`fork_tip`] -> [`fork_point`] -> [`best_chain_tip`].
- Maintain data to build wallet / complex indexes locally - The new ChainIndex will be designed to enable Zaino efficiently locate the blocks and transactions required to build indexes locally. Full blocks and transactions will continue to be fetched from the backed validator (eg. through Zebra's ReadStateService).
- (possible inclusion of FlyClient data).
- Subtree (shard) root indexing.

- *Make all structs used by ZainoDB explicitly versioned and separate from internal chain structs (Maintained by Zaino)

Tasks:
Zaino:
\- Add `ChainBlock` struct: https://github.com/zingolabs/zaino/pull/330
&nbsp;&nbsp;\- Add struct and implement getters / setters.
&nbsp;&nbsp;\- Implement conversion from / to full blocks / compact blocks.
&nbsp;&nbsp;\- Implement safe serialisation for ZainoDB.
\- Fix pre-overwintered hex block parsing (in Zaino-fetch::chain::{Block, Transaction}) (Required to sync pre-sapling activation blocks):
&nbsp;&nbsp;\- Use default values for sapling and orchard block header fields.
&nbsp;&nbsp;\- Add parse_v1, parse_v2, parse_v3 methods for legacy transaction formats.
&nbsp;&nbsp;\- Update ParseFromSlice for Transaction.
\- Re-implement non-finalised state: https://github.com/zingolabs/zaino/pull/327
&nbsp;&nbsp;\- Implement new structure for non-finalised state.
&nbsp;&nbsp;\- Implement safe reorg logic for non-finalised state.
&nbsp;&nbsp;\- Implement non-best chain block ranges.
&nbsp;&nbsp;\- Implement efficient subscriber logic.
\- Re-implement finalised state: https://github.com/zingolabs/zaino/pull/342
&nbsp;&nbsp;\- Implement version safe type schema.
&nbsp;&nbsp;\- Implement new database structure.
&nbsp;&nbsp;\- Implement database validation.
&nbsp;&nbsp;\- Implement efficient subscriber logic.i

(\- *Posslible updates for NU7)

(Zaino Continuation Grant - MileStone 2)
[...]

PR #922, opened 2026-03-19
`FlyClient` support for light clients

ordian:
Related to #318 (mentions `FlyClient`).

# Motivation

https://zips.z.cash/zip-0221 has been implemented, but there's no support for light clients to receive these proves currently

Depends on https://github.com/zcash/lightwallet-protocol/pull/21.

# Verification

A proof of concept verification client utilizing this code is available here:
* https://github.com/ordian/zflyclient

#### Disclaimer

The changes here were made in a half-day for a hackathon: https://ns.com/zcash.
It's likely that the proposed interface is suboptimal or does not make sense, so please review with a grain of salt.


zancas:
We'll circle back.

PR #924, opened 2026-03-19
Update dockerignore to ignore testing dirs

nachog00:
Without this, the docker build passes those massive dirs as context and they aren't really needed.

The dirs are:

```
/docker_cargo/
/container-target/
```

these are for `makers` container testing to persist a cache in between runs, we don't need them passed around in zainod docker builds

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -16,3 +16,7 @@ target/
 
 # Misc
 **/*.log
+
+docker_cargo/
+container-target/
+


nachog00:
these don't affect CI builds since those dirs are empty there, but they are a pain when building locally for testing on arbitrary refs.

nachog00:
These changes don't affect source code and shouldn't be bounded by the regular CI testsuite.

AloeareV (approved):

PR #927, opened 2026-03-19
Zebra 4 2 support

pacu:
Fixes:

### DIFF
--- a/zaino-common/Cargo.toml
+++ b/zaino-common/Cargo.toml
@@ -18,3 +18,6 @@ serde = { workspace = true, features = ["derive"] }
 # Error handling
 thiserror = { workspace = true }
 
+# Zingo 
+zingo_common_components = { workspace = true }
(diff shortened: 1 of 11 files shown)


zaino-common/Cargo.toml
```diff
@@ -18,3 +18,6 @@ serde = { workspace = true, features = ["derive"] }
 # Error handling
 thiserror = { workspace = true }
 
+# Zingo 
+zingo_common_components = { workspace = true }
+zcash_local_net = { workspace = true }
```
fluidvanadium: zcash_local_net is supposed to be a test-only dependency

fluidvanadium (changes requested):
Mostly great, but one dependency that should not be added to that crate. I am going to investigate.

fluidvanadium (approved):
ready to go!

zancas (commented):
LGTM!

PR #821, opened 2026-02-04
Update Zaino to Zebra 4.1.0

pacu:
Closes #816

PR #928, opened 2026-03-19
Enable building with Rust 1.85

str4d:
This matches the current MSRV of Zallet (and the Zebra crates that Zaino depends on after #821).


pacu (approved):

daira (commented):
Post-hoc ACK

PR #930, opened 2026-03-19
Add MSRV

fluidvanadium:
Has a coded MSRV. I am not 100% sure if this PR is an advantage.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ homepage = "https://www.zingolabs.org/"
 edition = "2021"
 license = "Apache-2.0"
 version = "0.2.0"
+rust-version = "1.85.0" # This MSRV is known to work, but has not been completely automatically tested.
 
 
 [workspace.dependencies]


Cargo.toml
```diff
@@ -22,6 +22,7 @@ homepage = "https://www.zingolabs.org/"
 edition = "2021"
 license = "Apache-2.0"
 version = "0.2.0"
+rust-version = "1.85.0" # This MSRV is known to work, but has not been completely automatically tested.
```
str4d, line 25: The way we handle MSRVs in our repos is:
- Repos have a `rust-toolchain.toml` file in their root that specifies the MSRV version.
- Developers use `rustup` by default for their Rust installations.
- We check in the `Cargo.lock` file even for library-only workspaces.
- CI tests "both ends" of our MSRV range:
  - One job tests using `rustup` and the repo as-is, for testing MSRV plus known-working concrete set of dependencies for that MSRV.
  - One job tests by overriding with stable Rust and first deleting the lockfile, for testing with latest Rust plus latest dependency versions.

With this approach:
- Developers don't accidentally use newer Rust features that would break MSRV.
- We detect incompatibilities caused by recent dependency updates.

zancas:
We'll implement this policy, as tracked here:  https://github.com/zingolabs/zaino/issues/976

Issue #933, opened 2026-03-25
GetBlockRange gRPC method returns OutOfRange status as startup check rather than in stream


idky137:
Currently the GetBlockRange gRPC method returns OutOfRange status as startup check rather than in the returned stream, as it should. This propagates from ChainIndex behaviour:
- https://github.com/zingolabs/zaino/blob/93a9495336e7ee6f28ab1b02d1959a23b459f035/zaino-state/src/chain_index.rs#L959

This should be updated to return blocks that are available and return the OutOfRange error if / when an out of range block is requested.

This will require updating behaviour in:
- (possibly) zaino-state::backends::{fetch, state}
- zaino-state::chain_index
- zaino-state::chain_index::finalised_state::db::v1

Zk-nd3r:
We hit this in production with a compact block adapter that scans anchor blocks via GetBlockRange. When requesting a range that includes blocks Zaino hasn't indexed yet, the call fails immediately instead of streaming available blocks and signaling OutOfRange at the boundary.

Current workaround: we cap the end height at the tip reported by GetLatestBlock before calling GetBlockRange. Works but adds an extra round trip and a race condition if the chain advances between the two calls.

The fix described here (stream available blocks, then return OutOfRange when hitting the gap) would let clients handle partial results cleanly without the tip-check workaround.

For reference, the adapter code that exercises this path: https://github.com/Frontier-Compute/zap1/blob/main/src/bin/zaino_adapter.rs

idky137:
We have now updated this behaviour (PR: https://github.com/zingolabs/zaino/pull/934), this will be included in the next zaino release but is also available for testing on dev if required earlier.

PR #937, opened 2026-03-26
update NodeBackedChainIndex::get_compact_block_stream() out_of_range behaviour to match lwd

idky137:
Updates NodeBackedChainIndex::get_compact_block_stream() out_of_range behaviour to match lwd.

- Required for https://github.com/zingolabs/zaino/issues/933

### DIFF
--- a/zaino-state/src/chain_index.rs
+++ b/zaino-state/src/chain_index.rs
@@ -934,7 +934,9 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
 
     /// Streams *compact* blocks for an inclusive height range.
     ///
-    /// Returns `None` if either requested height is greater than the snapshot's tip.
+    /// Returns `Ok(None)` if the request is descending and `start_height` exceeds the chain tip.
+    /// For ascending requests that exceed the tip, returns a stream that ends with an
+    /// `out_of_range` error after all available blocks have been sent.
     ///
     /// - The stream covers `[start_height, end_height]` (inclusive).
     /// - If `start_height <= end_height` the stream is ascending; otherwise it is descending.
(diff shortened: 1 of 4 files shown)


zaino-state/src/chain_index.rs
```diff
@@ -962,14 +964,28 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
 
         let is_ascending = start_height <= end_height;
 
+        // Descending: the first block we'd try to return is already above the tip → error immediately.
```
pacu, line 967: I don't think that existing or non existing blocks at chain tip can be determined at the time of calling this function. Reorgs can happen at any time and a valid range also cease to exist or viceversa. I think the correct thing to do is to fail when the backing data fails to retrieve the intended block since streams can be long living and wide.

pacu (changes requested):

zaino-state/src/chain_index.rs
idky137: Because the `NodeBackedChainIndex` serves data from snapshots (`NonFinalizedSnapshot`) for the non-finalised state rather than directly from the "live" non-finalised chain this meathod will never hit a reorg.

This snapshot based design was a choice made in the Zallet-Zaino API meetings and was chosen for several reasons, but the main reason was: a library user (zallet) can chain multiple calls on the same snapshot, so it can assume that the backing data will not change between calls. 

This also helped fix a bug where, if a reorged occurred during the time that get_block_range was running, the client could be served data partially from the old chain, and partially from the new chain, which would lead to a corrupted chain state.

The behaviour is now as follows:
- Zainod recieves GetBlockRange request (or any other RPC request that touches the non-finalised state)
  - RPC worker task fetches non-finalised snapshot to use for duration of RPC call (locking chain state)
  - RPC is served with snapshotted data (+ data from finalised state / mempool)
  - non-finalised snapshot is dropped
 
 Because of this I *think* we are safe to use this implementation.

pacu (approved):

pacu (approved):

PR #939, opened 2026-03-26
depend on release versions of other zaino crates temporarily during release process

AloeareV:
This PR begins the process of using versioned dependencies on other zaino crates. This is needed to publish, and each 'layer' of version updates needs to go in order, a crate can't depend on an unpublished version of another crate, even during the publishing process. A change is needed to publish zaino-state, the removal of a zebra version check which relies on access to the lockfile, which isn't available in the cargo publish flow. That part of this PR needs review. All of these version dependencies may wish to revert to path dependencies before this PR merges, but the temporary presence of the version dependencies in our git history, corresponding to the commits which were published to crates.io, will be useful. 

Edit: A lot of this no longer applies, as this has needed to be rebased due to stable/dev conflicts. We still need to land these version number changes, but their introduction no longer accurately reflects the exact commits published. Fortunately, that's effectively preserved in our version-tagged commit.


AloeareV:
Another change which _will_ persist after this PR is the addition of individual version numbers to each zaino crate. As these version numbers expect to diverge as we add breaking changes to different crates at different times, a workspace version number no longer makes sense.

packages/zaino-common/Cargo.toml
```diff
@@ -6,7 +6,7 @@ repository = { workspace = true }
 homepage = { workspace = true }
 edition = { workspace = true }
 license = { workspace = true }
-version = { workspace = true }
+version = "0.1.0"
```
nachog00, line 9: is this correct with the addition of `path` and `version` on the workspace declarations?

AloeareV: the combination of `path` and `version` should mean that path is prefered, and version is used in contexts like `cargo publish` where path is disallowed.

zancas (dismissed):
This is fine!

packages/zaino-common/Cargo.toml
AloeareV: Hang on. This is declaring the version of this crate as distinct from other crate versions. Unrelated to the path and version combination.

zancas (approved):

PR #942, opened 2026-03-30
Support Zebra 4.3.0

pacu:
`zingo_common_components` points to tag `zingo-netutils-v4.0.0` `zingo_netutils` points to tag `zingo-netutils-v4.0.0`

Zingolib points to commit `823a8f77b85592064ecd8daaa584a027766ee1ab` which needs to be merged and released.

zingo_test_vectors and zcash_local_net point to infrastructure branch `support-zebra-4-3-0` PR https://github.com/zingolabs/infrastructure/pull/238

Fixes:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ version = "0.2.0"
 
 # Zingo 
 
-zingo_common_components =  { git = "https://github.com/zingolabs/zingo-common", tag = "zingo-netutils-v4.0.0" }
+zingo_common_components =  "0.3.0"
 zingo-netutils = { git = "https://github.com/zingolabs/zingo-common", tag = "zingo-netutils-v4.0.0", features = ["back_compatible"] }
 zingolib = { git = "https://github.com/zingolabs/zingolib.git", rev = "823a8f77b85592064ecd8daaa584a027766ee1ab", features = [
     "testutils",
@@ -154,7 +154,6 @@ lazy_static = "1.5.0"
 #zcash_primitives = { git = "https://github.com/zcash/librustzcash", tag = "zcash_client_backend-0.21.0"}
 #zcash_protocol = { git = "https://github.com/zcash/librustzcash", tag = "zcash_client_backend-0.21.0"}
 #zcash_transparent = { git = "https://github.com/zcash/librustzcash", tag = "zcash_client_backend-0.21.0"}
-zingo_common_components =  { git = "https://github.com/zingolabs/zingo-common", tag = "zingo-netutils-v4.0.0" }
 
 [profile.test]
 opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,14 +28,14 @@ version = "0.2.0"
 
 # Zingo 
 
-zingo_common_components = "0.3.0"
-zingo-netutils = "2.0.2"
-zingolib = { git = "https://github.com/zingolabs/zingolib.git", rev = "14a69853a8bd2e473dee8a433004c4c06aaf5308", features = [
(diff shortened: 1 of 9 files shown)


idky137 (dismissed):
Looks good! Waiting on tests to pass to merge.

idky137:
all tests now pass for me with the latest commit.

integration-tests/tests/state_service.rs
```diff
@@ -1506,7 +1506,7 @@ async fn state_service_get_address_transactions_regtest<V: ValidatorExt>(
 
     if matches!(validator, ValidatorKind::Zebrad) {
         test_manager.local_net.generate_blocks(100).await.unwrap();
-        tokio::time::sleep(std::time::Duration::from_millis(1000)).await;
+        tokio::time::sleep(std::time::Duration::from_millis(3000)).await;
```
AloeareV, line 1509: Glad this 'fixes' the test...add another tally to the box of 'we really need deterministic health/status checks for this sort of stuff.'

zaino-common/Cargo.toml
```diff
@@ -29,3 +29,4 @@ time = { workspace = true }
 
 # Zingo 
 zingo_common_components = { workspace = true }
+zingolib = { workspace = true }
```
AloeareV: Why do we need this? I don't see any changes to zaino-common to indicate this is needed as a zaino-common dependency, and I believe non-test zaino crates depending on zingolib has caused hard-to-resolve dependency snarls in the past.

AloeareV (changes requested):
I think (well, hope) that zaino-common depending on zingolib isn't actually needed. Other than that, looks great!

Cargo.toml
```diff
@@ -154,6 +154,7 @@ lazy_static = "1.5.0"
 #zcash_primitives = { git = "https://github.com/zcash/librustzcash", tag = "zcash_client_backend-0.21.0"}
 #zcash_protocol = { git = "https://github.com/zcash/librustzcash", tag = "zcash_client_backend-0.21.0"}
 #zcash_transparent = { git = "https://github.com/zcash/librustzcash", tag = "zcash_client_backend-0.21.0"}
+zingo_common_components =  { git = "https://github.com/zingolabs/zingo-common", tag = "zingo-netutils-v4.0.0" }
```
AloeareV: Also, this patch is the same exact version as the specified dependency.

AloeareV: Oh...this is setting some dependency's version of this to match.

Cargo.toml
```diff
@@ -28,14 +28,14 @@ version = "0.2.0"
 
 # Zingo 
 
-zingo_common_components = "0.3.0"
-zingo-netutils = "2.0.2"
-zingolib = { git = "https://github.com/zingolabs/zingolib.git", rev = "14a69853a8bd2e473dee8a433004c4c06aaf5308", features = [
+zingo_common_components =  { git = "https://github.com/zingolabs/zingo-common", tag = "zingo-netutils-v4.0.0" }
```
AloeareV: As far as I can tell this change is no longer needed.

Cargo.toml
AloeareV: This can be removed hand-in-hand with the version change.

AloeareV (changes requested):
I mentioned this in signal, but this should probably be official review comments!

Cargo.toml
AloeareV: Note: Still confirming that tests pass. In-progress locally. EDIT: Confirmed.

AloeareV (approved):
Great!

PR #943, opened 2026-03-31
Zallet regtest fixes

str4d:
Fixes:


nuttycom (approved):
utACK a2d2f35fc14729356745b02c21b57b63246491c9

idky137 (dismissed):
As this changes the API of `JsonRpSeeConnector`, which is public, it will require a changelog entry (and version bump on release).

Apart from that it looks good!

pacu (dismissed):

pacu:
utACK 

Modulo changelog additions

str4d:
> As this changes the API of `JsonRpSeeConnector`, which is public, it will require a changelog entry (and version bump on release).

@idky137 I didn't make a changelog entry because I don't know which changelog is the correct one:
- There is no changelog file inside the `zaino-fetch` folder, where the change is.
- There is a changelog file inside `zaino-state`, suggesting there should be a separate one for `zaino-fetch`
- There is a changelog file at the root, which contains some `zaino-fetch` changes, but none of the version numbers line up at all with `zaino-fetch`.

idky137:
> > As this changes the API of `JsonRpSeeConnector`, which is public, it will require a changelog entry (and version bump on release).
> 
> @idky137 I didn't make a changelog entry because I don't know which changelog is the correct one:
> - There is no changelog file inside the `zaino-fetch` folder, where the change is.
> - There is a changelog file inside `zaino-state`, suggesting there should be a separate one for `zaino-fetch`
> - There is a changelog file at the root, which contains some `zaino-fetch` changes, but none of the version numbers line up at all with `zaino-fetch`.

Sorry this was more a note to other reviewers.

There is an ongoing internal conversation at zingolabs regarding how this should be recorded, with some arguing traditional changelogs are not the best form, which is why I have not added this myself.

I have raised this in zingolabs and can update it when I know the direction chosen.

zaino-fetch/src/jsonrpsee/response.rs
```diff
@@ -1035,11 +1043,15 @@ impl TryFrom<GetTreestateResponse> for zebra_rpc::client::GetTreestateResponse {
             zebra_chain::serialization::SerializationError::Parse("negative block height")
         })?;
 
-        let sapling_bytes = value.sapling.commitments().final_state().clone();
-        let sapling = Treestate::new(Commitments::new(None, sapling_bytes));
+        let sapling = value
+            .sapling
+            .clone()
+            .unwrap_or_else(|| Treestate::new(Commitments::new(None, None)));
```
daira, line 1049: Was it considered to just store the empty Treestate rather than None?

daira (approved):
utACK

zaino-fetch/src/jsonrpsee/response.rs
nuttycom: I don't like that solution, because then you don't have the invariant that every persisted tree state is known to be valid and not the default value. This seems reasonable to me.

pacu (approved):

idky137:
Do not merge - test run failed: (`integration-tests::state_service zebra::get::z::treestate_regtest`)

There is now differing behaviour between the StateService and FetchService (I am looking into this failure)

Output from StateService and FetchService:
```
[integration-tests/tests/state_service.rs:1214:35] fetch_service_subscriber.z_get_treestate("2".to_string()).await.unwrap() = GetTreestateResponse {
        hash: block::Hash(
            "bc941c9b0fbb623556e034b109294de9bc4755a23ab5e8e536ed77fff5d4e1d8",
        ),
        height: Height(
            2,
        ),
        time: 1296699402,
        sprout: None,
        sapling: Treestate {
            commitments: Commitments {
                final_root: Some(
                    [
                        62,
                        73,
                        181,
                        249,
                        84,
                        170,
                        157,
                        53,
                        69,
                        188,
                        108,
                        55,
                        116,
                        70,
                        97,
                        238,
                        164,
                        141,
                        124,
                        52,
                        227,
                        0,
                        13,
                        130,
                        183,
                        240,
                        1,
                        12,
                        48,
                        244,
                        194,
                        251,
                    ],
                ),
                final_state: Some(
                    [
                        0,
                        0,
                        0,
                    ],
                ),
            },
        },
        orchard: Treestate {
            commitments: Commitments {
                final_root: Some(
                    [
                        174,
                        41,
                        53,
                        241,
                        223,
                        216,
                        162,
                        74,
                        237,
                        124,
                        112,
                        223,
                        125,
                        227,
                        166,
                        104,
                        235,
                        122,
                        73,
                        177,
                        49,
                        152,
                        128,
                        221,
                        226,
                        187,
                        217,
                        3,
                        26,
                        229,
                        216,
                        47,
                    ],
                ),
                final_state: Some(
                    [
                        0,
                        0,
                        0,
                    ],
                ),
            },
        },
    }
    [integration-tests/tests/state_service.rs:1219:35] state_service_subscriber.z_get_treestate("2".to_string()).await.unwrap() = GetTreestateResponse {
        hash: block::Hash(
            "bc941c9b0fbb623556e034b109294de9bc4755a23ab5e8e536ed77fff5d4e1d8",
        ),
        height: Height(
            2,
        ),
        time: 1296699402,
        sprout: None,
        sapling: Treestate {
            commitments: Commitments {
                final_root: None,
                final_state: Some(
                    [
                        0,
                        0,
                        0,
                    ],
                ),
            },
        },
        orchard: Treestate {
            commitments: Commitments {
                final_root: None,
                final_state: Some(
                    [
                        0,
                        0,
                        0,
                    ],
                ),
            },
        },
    }
```

zancas:
@idky137 if these CI fails are always on false negatives, can we please deactivate the jobs so we can at least get the benefit other jobs?

idky137:
Update:
This is not a bug in the code in this PR but this change has surfaced it and a fix needs to be implemented before this PR can land safely.

Currently the treestate RPC methods (the z_get_treestate JsonRPC service and the get_tree_state lightclient gRPC serve) have differing implementations in the `StateService` and `FetchService`. This most likely happened during the switchover from the old `local_cache` code to the new `ChainIndex`, with different devs updating the implementations differently.

The issue is that the fixes in this PR are being picked up by one path but not the other, leading to inconsistent behaviour in ZainoD.

While trying to fix this I have run into several issues with the ChainIndex that will need to be addressed before a clean solution can be implemented ( https://github.com/zingolabs/zaino/issues/964 ). 

As there are several tasks that need to be completed to close this issue and implement the proper fix for the treestate RPCs I will look into whether a temporary fix can be implemented so this code can land.

idky137:
> @idky137 if these CI fails are always on false negatives, can we please deactivate the jobs so we can at least get the benefit other jobs?

This is an issue with our CI setup, currently the zcash-integration-test jobs cannot run on non-zingolabs branches. It would be good if we could find a fix for this as my PRs are also now failing.

zancas:
I've refactored the locations of the zaino-* packages to [clean up the top-level directory](https://github.com/zingolabs/zaino/issues/957).

These git mvs show up as conflicts.   To fix them I have merged dev and pushed to branches in this repo, but the feature for this PR is in @str4d 's fork.

Folks can use this duplicated (except for lints and a new CHANGELOG) PR:  https://github.com/zingolabs/zaino/pull/971 .

Or resolve the conflicts themselves, which ever is easiest.

idky137:
I have added the required downstream zaino fixes, and rebased all commits on top of dev to fix conflicts, in #973.

That PR should be merged instead of this one, it is now ready for review, the additional fixes did not touch the ChainIndex fix here that has been reviewed, only updated the downstream `StateService`/`FetchService` code to properly and uniformly pick up there changes.

zancas:
Superceded by #973 .

PR #948, opened 2026-04-06
use test vectors from crates.io

### LINKED ISSUE #947: test vectors are mixed into code
### Observation

The zaino-testvectors data should be available across the ecosystem for reference from all interested projects.

### Location in codebase

zaino-testvectors

### Type of concern

Coupling / Dependencies

### Impact

test data is confused for unit-or-integration test specific logic, and isn't useful as a reference for outside parties.

### Suggested direction

Publish the vectors.

zancas:
Fixes: #947 

I like the net change in lines of code.

I don't trust CI so I have kicked off a local test run.


pacu (approved):
Removing code is good.

PR #954, opened 2026-04-08
divide workspaces by promoting integration-tests to become its own *excluded* cargo workspace

### LINKED ISSUE #946: test and production code are unnecessarily combined in a single workspace
### Observation

Because dependencies used by integration tests are managed in the same cargo workspace as the source dependencies, it's possible to block upgrades to source dependencies because of packages that are only needed in test.

### Location in codebase

The  whole repo.

### Type of concern

Separation of concerns

### Impact

Build times are long, dependencies are harder to analyze than necessary, and test-only behavior is sometimes not properly understood to BE test-only.

### Suggested direction

_No response_

### LINKED ISSUE #955: integration testing in the divide workspaces feature fail
### Usage mode

Both / Not sure

### Describe the bug

The work to separate integration tests doesn't have a passing integration test suite, but that's not due to the migration.

Here's claude's analysis:

```bash
● Found it. The issue is in the sync loop at chain_index.rs:556-571. When get_best_block_height() fails,
   the ? propagates the error out of the loop, and line 599 sets StatusType::CriticalError. The
  indexer's serve loop detects this on the next tick and shuts down.

  This is a startup race: the sync loop starts before the validator connection is fully ready, the first
   RPC call fails, and the entire indexer dies.
```

### Steps to reproduce

cargo nextest run

### Expected behavior

Many tests fail

### Environment

0.2.0

### Configuration

```toml

```

### Logs

```shell

```

### Additional context

_No response_

### LINKED ISSUE #956: CI needs to be updated to reflect new directories and workspaces

zancas:
Fixes:  #946 

TODO: Fix #955
TODO: Fix #956

The integration testing code is needed to validate correct behavior of zainod, and other zaino-*, public interfaces *AFTER* detailed development and unit-testing of the source code has validated logic at a higher granularity of abstraction.

For the developer these tasks are separate:

* conceptually
* as a part of practical development flows
* in terms of the tooling they require

A useful way to represent these separate responsibilities is to map them cleanly into cargo tooling.

By making "integration-tests" into it's own workspace that is *excluded* from the source code workspace, cargo is informed that the contained logic is not by default, to be evaluated.

Practically speaking this means that a developer that executes a clean clone, and cargo nextest run, of this repo:

* does not build integration-test only dependencies
* does not run integration tests

This is an efficient an reasonable default flow for development.   Rather than continue to explain why, I will argue by example:

Try "cargo nextest run" in a "clean" top level directory on the "dev" branch.

Now try it against a "clean" top level directory of this feature branch.

Which experience did you enjoy more?  Which was more useful to you as a developer?


zancas:
Leaving a sanity test of:

`cd integration-tests && cargo nextest run -j1`

running overnight.

zancas:
Running sanity check against `dev` integration-tests now.

zancas:
I see [a similar test failure pattern on dev](https://github.com/zingolabs/zaino/issues/955#issuecomment-4207858286), indicating that this PR is not introducing a regression.

nachog00 (approved):

nachog00:
I think there are some clear following steps including: moving over all the "testing","docker_cargo", etc etc... over here.
putting all src code crates into a `crates` or `src` directory is probably also the best.

PR #959, opened 2026-04-08
Add opt-in system RocksDB linking to skip slow librocksdb-sys builds

### LINKED ISSUE #958: librocksdb when bundled and statically linked significantly increases the time for clean builds
For example here's a point estimate of runtime for `cargo check`:

```bash
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 09s

real    2m9.214s
user    27m17.242s
sys     1m16.443s

### PR DESCRIPTION
**Pull request by zancas**
Fixes:  #958

Adds cargo-make tasks that let developers link against their system-installed RocksDB instead of compiling it from C++ source every clean build. A source code scan verifies that any symbols missing from the system library are not reachable at runtime, making it safe despite version mismatches. An audit fingerprint tracks Cargo.lock and system version changes to re-check automatically when needed.

New tasks: use-system-rocksdb, use-bundled-rocksdb, check-system-rocksdb, audit-system-rocksdb

The point estimates for build time speed-up:

Bundled:

`cargo clean && makers use-bundled-rocksdb && time cargo check`

```BASH

Finished `dev` profile [unoptimized + debuginfo] target(s) in 2m 10s

real    2m10.653s
user    28m12.568s
sys     1m18.086s
```

System:

`cargo clean && makers use-system-rocksdb && time cargo check`

```Bash
Finished `dev` profile [unoptimized + debuginfo] target(s) in 46.56s

real    0m46.618s
user    6m11.808s
sys     0m36.191s
```

In other words it takes about 84 fewer seconds per clean build.


juanky201271 (approved):
great!

PR #961, opened 2026-04-08
Optional nfs

### LINKED ISSUE #960: ChainIndex's NonFinalizedState should be optional
### What's the smell?

Currently, we initialize a NonFinalizedState right when the ChainIndex comes into being. For initial sync, this contains a single block, the genesis block. Then, we sync the FinalizedState. Once that's done, we blow away the state NFS, initialize a new one from the finalized tip, and go from there. 

This means that during the entire process of syncing the FinalizedState, we have a 'dummy' NFS containing only the genesis block. This is in no way useful, and would be more correctly represented as a None value.

### Location

https://github.com/zingolabs/zaino/blob/2c657847aec0de360b430b014e01a3a56d4b11d1/zaino-state/src/chain_index.rs#L485-L490

### Category

Missing abstraction

### Severity

Medium - Should fix when touching this code

### Suggested fix

_No response_

AloeareV:
Fixes: #960


zancas:
I opened a new PR with a simplified commit structure and the same logical changes #1010 .

PR #973, opened 2026-04-10
Zallet regtest fixes + downstream indexer service fixes

idky137:
This PR contains both the treestate regtest fixes in the ChainIndex required by Zallet (originally in #943 ) and the downstream fixes required to properly propagate this change and keep zainod's behaviour consistent (see below).

Changes added on top of #943 (which has already been reviewed and approved):
- StateService and FetchService implementations of `z_get_treestate`, `get_tree_state` and `get_latest_tree_state` diverged, all implementations have been updated to use the chain index rather than fetching data directly with the `JsonRpSeeConnector` or `ReadStateService`, keeping implementations consistent across backends, readying them for the planned merge of the `StateService` and `FetchService.`
- Updatd tests to be less flaky: they now use the actual chain height to query data rather than a set height that may not have synced through the stack at call time.
- Rebases commits on top of dev to avoid merge conflicts.

** Replaces #943 **

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ and this library adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- [943] Zallet regtest fixes
+
+### Added 
+### Changed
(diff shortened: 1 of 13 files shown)


zancas:
Superceded by #1000 

That PR skips CI integration-testing while we're setting up the new server.

PR #975, opened 2026-04-10
0.2.0 donation address feature

alexxiy:
For 0.2.0 tag version
Add an abilty to configure a donation_address in Zaino config to make it possible for Zaino node operators define their donation address which used on node monitoring services, such as https://nodus.alexxiy.top

I've made changes on tag v0.2.0 (take a look only 2nd commit "Add donation_address feature").


zancas:
Superceded by #977

PR #1014, opened 2026-04-17
add new mempool tests (stale requested chain tip)

idky137:
Adds mempool tests.

Zallet has been having issues with zaino's mempool interface. this PR adds new unit and integration tests to validate the expected behaviour of the API and to try to model the conditions zallet makes.

TLDR: No bugs could be found in zaino but these are useful tests to have (particularly the new unit tests.)


zancas (approved):
YAY testing!   Thanks!

PR #1023, opened 2026-04-21
● Sync README Project Structure and enforce it via pre-push check

zancas:
Fills in tracked paths that had drifted out of the Project Structure
  block — top-level manifests (Cargo.toml/lock, rust-toolchain.toml,
  deny.toml), contrib files (CHANGELOG, CLAUDE.md, CONTRIBUTING.md,
  LICENSE, README.md, .gitignore, .dockerignore), plus the
  integration-tests entries that weren't documented (src/,
  Cargo.toml, .config/nextest.toml, and the test_environment
  permission-test helper).

  Adds tools/scripts/check-readme-structure.sh, a bidirectional
  check that asserts (a) every path listed under ## Project
  Structure exists and (b) every tracked path at a depth that
  block documents is listed. Directories mentioned only as a
  prefix of a deeper file path don't trigger interior enforcement
  — only the leading component at the parent's level counts. The
  check is wired into .githooks/pre-push so the README and
  filesystem stop diverging silently.


AloeareV (approved):
Nicely done!

PR #1024, opened 2026-04-21
propose removal of workspace version

zancas:
Fixes:   Unused workspace version.


AloeareV (approved):
Workspace version is no-longer needed. Great!

PR #1025, opened 2026-04-21
sketch of release process

zancas:
Fixes:  Some communication issues about release.


nachog00:
We should have at leat 3 tiers of testing:
- unit/leightweight: runs in a few minutes and gates PR's onto dev
- integration: can take as much as hours, runs nightly
- long deploys: can take days, performs full syncs on production chains

then the release flow can be:

1. something merges in dev, so it passes tier 1, no guarantee that it passes the other 2, although we hope it does (the better tier 1 is done, the more likely this happy path will be)
2. github workflow machinery keeps track of the latest commit from dev (between stable and dev HEAD) that passes tier 2 tests. every night it tries to advance that boundary. if it suceeds, it cuts a release candidate from the new commit it finds
3. as soon as there's a new rc, tier 3 tests are kicked off.
5. once the release day (end of release period) comes, whatever commit made it past step 3. and passed those tier 3 tests, is then eligible for releasing on that date.

PD: to make all of this as streamlined as possible, using cangests on every PR is best, so that the machinery can produce semver bumps autonomously. So, come release heartbeat day, devs are presented with an already bundled and versioned commit to sign off on and release happens switfly. changests also streamline changelog keeping.

nachog00:
a more detailed explanation of the flow i propose https://hackmd.io/@zingoistas/H1t3aFBabg

docs/decision_records/release/periodic.md
```diff
@@ -1,62 +1,335 @@
...
+| ------------ | ------- | ----- | ------------------------------ |
+| zaino-state  | 0.1.0   | 0.2.0 | new sync mode, fix #987        |
+| zaino-serve  | 0.1.0   | 0.1.1 | fix RPC edge case              |
+| zaino-fetch  | 0.1.0   | 0.1.0 | (unchanged)                    |
+| zaino-proto  | 0.1.0   | 0.1.1 | new message types              |
+| zaino-common | 0.1.0   | 0.1.0 | (unchanged)                    |
+| zainod       | 0.2.0   | 0.3.0 | new sync mode exposed          |
+```
+
+## Changesets: Per-Crate Version Tracking
```
AloeareV, line 203: This description of changesets is sufficient for determining what number to bump the new crates to. It's not, however, sufficient for creating a public-facing comprehensive changelog.

AloeareV: The changeset, therefore, must include a comprehensive list of all public facing changes (unless we are keeping changelogs separately from changesets, which seems more fiddly than just keeping changelogs), in order to generate a per-crate comprehensive changelog from the set of changesets.

AloeareV (dismissed):

docs/decision_records/release/periodic.md
nachog00: I think we can easily accomodate for richer changelog entries right on the changeset. Is your concern regarding the size of the changelog entry? formatting?

nachog00: isntead ot toml, markdown with yaml like headers can be used. after the structured header, the rest of the file can be free formatted text, perfect for long and expressive changelog entries.

nachog00: like this: https://github.com/knope-dev/changesets#changesetnew_feature_to_fix_bugmd

zancas: This seems like a good-enough iteration.

zancas:
@nachog00 @AloeareV what remains to approve this PR?

docs/decision_records/release/periodic.md
AloeareV: > isntead ot toml, markdown with yaml like headers can be used. after the structured header, the rest of the file can be free formatted text, perfect for long and expressive changelog entries.

My concern is purely to make the requirement to document all public-facing changes explicit, as the current description of changesets doesn't actually specify the need for each individual change to be documented.

idky137 (dismissed):
This currently deviates from the agreed release process in the zingolabs adr 3: https://github.com/zingolabs/zingo-adrs/blob/dev/ADR%20003-Zaino%20Branching%2C%20Versioning%2C%20Documentation%2C%20Public%20Interfaces%2C%20and%20Release%20Strategy.md

We must ensure we do not have multiple, conflicting, release processes.

If we want to update zainos release process it should be done by proposing a change to the current adr, rather than introducing a second, conflicting document.

zancas:
The release process should be documented in the zaino repo, where the versioning system disambiguates what code the policy is relevant.

When I created zingo-adrs I defined it in a context that's too broad/general, because it loses this explicit versioned binding. 

Managing conflicts between documents is something that should be framed with appropriate tooling (i.e git).

AloeareV:
> The release process should be documented in the zaino repo, where the versioning system disambiguates what code the policy is relevant.
> 
> When I created zingo-adrs I defined it in a context that's too broad/general, because it loses this explicit versioned binding.
> 
> Managing conflicts between documents is something that should be framed with appropriate tooling (i.e git).

I think this is a very good argument for moving that document from zingo-adrs to this repo.

zancas:
Instead of moving it I plan to cross-reference for two reasons:

(1)  I think the we do want to have the more general ADRs repo, specifically for for more generic, lower-granularity records
(2)  I think the model of in-repo specific superceding out-of-repo general is a cake and eat it too approach

I have pushed proposed changes, that places a premium on the designs that are favored by the likely implementer.

juanky201271 (approved):

Issue #833, opened 2026-02-12
proto identi-code


fluidvanadium:
### What's the smell?

The code is identical to librustzcash. It has unnecessary elements.

Blocking dependency upgrade!

### Location

zaino-proto/src/proto/service.rs

### Category

Duplication

### Severity

Medium

### Suggested fix

Depend on librustzcash, delete all copy-pasted declarations.

AloeareV:
If you look at the zaino-proto readme, the `.proto` files are explicitly kept identical with the ones in librustzcash via `git subtree`. To the best of my knowledge the `service.rs` file is _not_ copy-pasted. It looks identical because it's procedurally generated via prost-build, from an identical proto file. I don't think we can depend on librustzcash any more directly than this, iirc having librustzcash's proto-generated types instead of our own has been the cause of significant problems in the past.

Issue #730, opened 2025-11-26
DB and Business orthogonalization


nachog00:
### Discussed in https://github.com/zingolabs/zaino/discussions/617

<div type='discussions-op-text'>

<sup>Originally posted by **nachog00** October 21, 2025</sup>
There's mixing of concerns between DB and business layers.</div>

idky137:
Im not sure I understand what you mean by this. Could you expand on your concerns?

As I mentioned in Zaino dev signal channel I would like to do a final pass of this code as I think there are several places that are not properly understood currently.

PR #1028, opened 2026-04-22
Package height hash into index

zancas:
Addresses: #730
Addresses: #617
Progress on: #516
Progress on: #833

### DIFF
--- a/packages/zaino-state/src/chain_index/tests/mempool.rs
+++ b/packages/zaino-state/src/chain_index/tests/mempool.rs
@@ -1,17 +1,69 @@
 //! Zaino-State ChainIndex Mempool unit tests.
 
-use std::{collections::HashMap, io::Cursor, str::FromStr as _};
-use tokio::time::{sleep, timeout, Duration};
+use std::{
+    collections::{HashMap, HashSet},
+    io::Cursor,
+    str::FromStr as _,
+};
+use tokio::time::{timeout, Duration};
 use zebra_chain::serialization::ZcashDeserialize as _;
 
 use crate::{
     chain_index::{
         mempool::MempoolSubscriber,
         source::test::MockchainSource,
-        tests::vectors::{build_active_mockchain_source, load_test_vectors, TestVectorBlockData},
+        tests::{
+            poll::poll_until,
+            vectors::{build_active_mockchain_source, load_test_vectors, TestVectorBlockData},
+        },
     },
-    BlockchainSource as _, Mempool, MempoolKey, MempoolValue,
+    BlockHash, BlockchainSource as _, Mempool, MempoolKey, MempoolValue,
 };
+
+/// Poll the subscriber until its mempool keys exactly match `expected_txids`.
+///
+/// Mockchain mempool is deterministic — non-coinbase txs of block[tip+1] — so
+/// once the Mempool sync task has caught up to a mined tip, the subscriber's
+/// keys equal the expected set. Waiting for that exact match replaces fixed
+/// `sleep` pads and fails loud on drift.
+async fn wait_for_mempool_to_reflect(
+    subscriber: &MempoolSubscriber,
+    expected_txids: impl IntoIterator<Item = String>,
+) {
+    let expected: HashSet<String> = expected_txids.into_iter().collect();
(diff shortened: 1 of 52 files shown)


packages/zaino-state/src/chain_index/types/db/legacy.rs
```diff
@@ -719,39 +720,41 @@ impl FixedEncodedLen for Outpoint {
...
+/// This is the type business code operates on; it deliberately has no serde
+/// impl. Persistence is the sole responsibility of the module-private
+/// `PersistentBlockContext`, which is strictly mapped to this type via
+/// `From` conversions.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-#[cfg_attr(test, derive(serde::Serialize, serde::Deserialize))]
-pub struct BlockIndex {
-    /// The hash identifying this block uniquely.
-    pub hash: BlockHash,
+pub struct BlockContext {
```
AloeareV: This module is explicitly for types which implement SerdeVersionedSerde. Furthermore, this is a 'legacy' module, for types which have not yet been sorted into more appropriate modules. No new types should go into this file...and this type should go somewhere outside the `db` module as it's not a database type

zancas: Agreed — this was fixed later in the branch by 77ec0ebf. BlockContext now lives in chain_index/types/block_context.rs (outside db/, no serde impl), and PersistentBlockContext (the one db-serde helper for this family) lives in chain_index/types/db/block.rs as pub(super). The From/TryFrom trait impls between them were also replaced with named inherent methods (from_business / into_business) per the convention that was then codified in CLAUDE.md + the new makers lint-boundary-conversions lint.

packages/zaino-state/src/chain_index/types/block_context.rs
```diff
@@ -0,0 +1,58 @@
+//! Business-layer container pairing a [`BlockIndex`] with the block's
+//! parent hash and cumulative chainwork.
+//!
+//! `BlockContext` deliberately has no serde impl — persistence is the sole
+//! responsibility of a module-private helper in `types/db/block.rs`
+//! (`PersistentBlockContext`), and the two types round-trip via `From`
+//! conversions defined alongside that helper.
```
AloeareV:
```suggestion
//! (`PersistentBlockContext`), and the two types round-trip via `from_business`/
//! `to_business` conversion methods defined on that type.
```

packages/zaino-state/src/chain_index/types.rs
AloeareV, line 1: These additions to the public API need to be changelogged.

packages/zaino-state/src/chain_index/tests/sync_loop.rs
```diff
@@ -47,6 +47,7 @@ async fn survives_transient_source_failure() {
 /// Uses [`SyncTimings::fast`] (10× shrunk) so the full backoff schedule fits
 /// in a few seconds instead of ~40 s.
 #[tokio::test(flavor = "multi_thread")]
+#[ignore = "slow"]
```
AloeareV: I think the unit-test speedup PR which just landed makes it such that we don't need these ignores anymore.

AloeareV (dismissed):
This is great! Couple things:

packages/zaino-state/src/chain_index/tests/mockchain_tests.rs
```diff
@@ -66,6 +66,7 @@ async fn get_block_range() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
+#[ignore = "slow"]
```
AloeareV: This ignore

packages/zaino-state/src/chain_index/tests/mempool.rs
```diff
@@ -88,6 +88,7 @@ async fn spawn_mempool_and_mockchain() -> (
 }
 
 #[tokio::test(flavor = "multi_thread")]
+#[ignore = "slow"]
```
AloeareV: and this ignore

juanky201271 (approved):

PR #1031, opened 2026-04-23
Speed up unit tests

### LINKED ISSUE #1030: unit tests take about 60 seconds (point estimate)
If they were faster that would be better!

### LINKED ISSUE #1032: `DbV0::shutdown` / `DbV1::shutdown` unconditionally sleep 5s instead of awaiting the background handle
### Usage mode
  Library integration (using `zaino-*` crates)

  ### Describe the bug
  Both `DbV0::shutdown` and `DbV1::shutdown` contain identical code that *unconditionally* sleeps for 5 seconds before aborting the
  background `db_handler` task, regardless of whether the task has already finished:

  `packages/zaino-state/src/chain_index/finalised_state/db/v1.rs:150-164`
  ```rust
  async fn shutdown(&self) -> Result<(), FinalisedStateError> {
      self.status.store(StatusType::Closing);

      if let Some(handle) = &self.db_handler {
          let timeout = tokio::time::sleep(Duration::from_secs(5));
          timeout.await;
          // TODO: Check if handle is returned else abort
          handle.abort();
      }
      let _ = self.clean_trailing().await;
      if let Err(e) = self.env.sync(true) {
          warn!("LMDB fsync before close failed: {e}");
      }
      Ok(())
  }
  ```

  The identical pattern exists in `db/v0.rs:183-197`. The `TODO` comment confirms the intent was a `tokio::select!` between handle
  completion and a 5 s timeout, but the select was never written — so every caller of `shutdown()` pays a flat 5 seconds even when the
  background task finished instantly.

  ### Steps to reproduce
  ```
  makers container-test chain_index::tests::finalised_state::migrations::v0_to_v1_partial
  ```
  The test runs ~18.5 s wall. Breakdown:
  - 3 × `shutdown()` → 15 s of unconditional padding
  - 2 × 1 s test-level inter-step sleeps → 2 s
  - Real work → ~1.5 s

  Same picture for `v0_to_v1_interrupted` (~18.7 s) and `v0_to_v1_full` (~10 s floor from 2 shutdowns).

  ### Expected behavior
  `shutdown()` should return as soon as the background handle completes (the common case), falling back to abort only after a bounded wait:

  ```rust
  if let Some(handle) = self.db_handler.take() {
      match tokio::time::timeout(Duration::from_secs(5), handle).await {
          Ok(_) => {}                     // task finished cleanly
          Err(_) => { /* aborted on drop via JoinHandle */ }
      }
  }
  ```

  or equivalently with `tokio::select!` against a sleep future, aborting only on the timeout branch.

  ### Environment
  - Zaino: `dev` branch, commit `1ca45181`
  - Backend: N/A (pure library, reproduced against the in-process mockchain tests)
  - OS: Linux x86_64

  ### Configuration
  N/A — reproduced in unit tests.

  ### Logs
  N/A — the symptom is wall-clock time.

  ### Additional context
  This bug is effectively the biggest contributor to the migration tests' ~55 s combined runtime, but affects **every** `ZainoDB` caller —
  production `zainod` shutdown, integration tests, wallet teardown, anything that calls `ZainoDB::shutdown()`.

  Fix is ~5 lines in each of `db/v0.rs` and `db/v1.rs` and the intent is already documented in the `TODO` comment.

  **Related:**
  - #1030 — umbrella "unit tests take about 60 seconds"; this bug is one concrete contributor.
  - #676 — "chain index tests are ignored due to slow/hanging tests"; same family of symptoms.
  - #498 (closed) — adjacent race condition in the outer `NodeBackedChainIndex::shutdown`, different code path.

zancas:
Fixes: #1030 
Fixes: #1032

### DIFF
--- a/packages/zaino-state/src/chain_index/finalised_state/db/v0.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state/db/v0.rs
@@ -69,7 +69,10 @@ use lmdb::{Cursor, Database, DatabaseFlags, Environment, EnvironmentFlags, Trans
 use prost::Message;
 use serde::{Deserialize, Serialize};
 use std::{fs, sync::Arc, time::Duration};
-use tokio::time::{interval, MissedTickBehavior};
+use tokio::{
+    sync::Notify,
+    time::{interval, MissedTickBehavior},
+};
 use tracing::{info, warn};
 
 // ───────────────────────── ZainoDb v0 Capabilities ─────────────────────────
@@ -178,17 +181,37 @@ impl DbCore for DbV0 {
 
     /// Requests shutdown of background tasks and syncs the LMDB environment before returning.
     ///
-    /// This method is best-effort: background tasks are aborted after a timeout and the LMDB
-    /// environment is fsync’d before exit.
+    /// Signals the background task via `shutdown_notify` so it wakes out of
+    /// `zaino_db_handler_sleep` immediately, then awaits its join handle with
+    /// a 5 s timeout (aborting only if the handle fails to exit in time).
     async fn shutdown(&self) -> Result<(), FinalisedStateError> {
         self.status.store(StatusType::Closing);
+        self.shutdown_notify.notify_waiters();
 
-        if let Some(handle) = &self.db_handler {
+        let taken = self
+            .db_handler
+            .lock()
+            .expect("db_handler mutex poisoned")
+            .take();
+        if let Some(mut handle) = taken {
             let timeout = tokio::time::sleep(Duration::from_secs(5));
-            timeout.await;
-            // TODO: Check if handle is returned else abort
-            handle.abort();
+            tokio::pin!(timeout);
+
+            tokio::select! {
+                res = &mut handle => {
+                    match res {
+                        Ok(_) => {}
+                        Err(e) if e.is_cancelled() => {}
+                        Err(e) => warn!("background task ended with error: {e:?}"),
+                    }
+                }
+                _ = &mut timeout => {
+                    warn!("background task didn’t exit in time – aborting");
+                    handle.abort();
+                }
+            }
         }
+
         let _ = self.clean_trailing().await;
         if let Err(e) = self.env.sync(true) {
             warn!("LMDB fsync before close failed: {e}");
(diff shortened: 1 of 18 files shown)


packages/zaino-state/src/chain_index/tests/mempool.rs
```diff
@@ -1,18 +1,53 @@
...
     BlockchainSource as _, Mempool, MempoolKey, MempoolValue,
 };
 
+/// Poll the subscriber until its mempool keys exactly match `expected_txids`.
+///
+/// Mockchain mempool is deterministic — non-coinbase txs of block[tip+1] — so
+/// once the Mempool sync task has caught up to a mined tip, the subscriber's
+/// keys equal the expected set. Waiting for that exact match replaces fixed
+/// `sleep` pads and fails loud on drift.
+async fn wait_for_mempool_to_reflect(
```
AloeareV, line 29: This is called in enough places with the same arguments that there's probably some room to dry out these mempool tests.

packages/zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -150,12 +150,30 @@ impl DbCore for DbV1 {
     async fn shutdown(&self) -> Result<(), FinalisedStateError> {
         self.status.store(StatusType::Closing);
 
-        if let Some(handle) = &self.db_handler {
+        let taken = self
+            .db_handler
+            .lock()
+            .expect("db_handler mutex poisoned")
+            .take();
+        if let Some(mut handle) = taken {
```
AloeareV, line 158: This isn't added by this PR, so it's not in-scope, but we should probably have handling for the None case. In theory, we should only ever hit the None case if shutdown has already been called elsewhere, but this isn't a compiler-level invariant or anything.

packages/zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -386,36 +408,6 @@ impl DbV1 {
         Ok(zaino_db)
     }
 
-    /// Try graceful shutdown, fall back to abort after a timeout.
```
AloeareV, line 389: This was never used. Why didn't this hit a dead code warning?

packages/zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -566,6 +578,7 @@ impl DbV1 {
                     warn!("clean_trailing failed: {}", e);
                 }
             }
+            _ = self.shutdown_notify.notified() => {},
```
AloeareV, line 581: There are lot of similar-looking tokio selects in this code. I wonder if we want a helper called here and also in other places.

packages/zaino-state/src/chain_index/finalised_state/db/v0.rs
```diff
@@ -186,7 +186,11 @@ impl DbCore for DbV0 {
     /// a 5 s timeout (aborting only if the handle fails to exit in time).
     async fn shutdown(&self) -> Result<(), FinalisedStateError> {
         self.status.store(StatusType::Closing);
-        self.shutdown_notify.notify_waiters();
+        // `notify_one` stores a permit if no waiter is currently registered,
```
AloeareV, line 189: What happens if there is more than one task caught in the race window? The first task will get the single permit, and then the second one will hit the same problem this alleges to solve. I think this is a wild edge-case, I don't know if we'll ever have more than one background task registering itself as a waiter...but for deterministic correctness, we want something that, once notify() is called, notified().await will _always_ return immediately for all future callers.

AloeareV (approved):
Unit test run locally has gone from ~55s to ~14s. Some places where we had long waits to ensure races were run are now deterministic. Big improvement!

PR #1045, opened 2026-04-24
reboot integration-tests

### LINKED ISSUE #1044: Move `chain_cache.rs` whitebox tests from `integration-tests/` into `zaino-state`
## Problem

`integration-tests/tests/chain_cache.rs` (~600 lines) is a set of tests that directly inspect `zaino-state`'s internal non-finalized block cache. Representative accesses:

- `snapshot.blocks.len()` — cache size
- `snapshot.blocks.values().flat_map(|block| block.transactions())` — enumerate all cached transactions
- `snapshot.heights_to_hashes.values()` — enumerate indexed height→hash entries
- `snapshot.best_tip.hash` / `.best_tip.height` — tip fields

These are assertions *about `zaino-state`'s internal data structure*, not about externally-observable behavior. They were placed in `integration-tests/` originally because that's where the test fixture (live validator via `TestManager`) lives, not because they exercise a cross-crate integration surface.

## Why this matters now

Recent changes to `packages/zaino-state/src/chain_index/non_finalised_state.rs`:

- `ChainIndex::snapshot_nonfinalized_state` migrated from sync `-> Self::Snapshot` to `async -> Result<ChainIndexSnapshot, Self::Error>`.
- `NonfinalizedBlockCacheSnapshot` dropped from `pub` to `pub(crate)`; `IndexedBlock` is likewise `pub(crate)`.
- The `ChainIndexSnapshot` enum wrapping the cache uses `#[allow(private_interfaces)]` to keep the variant shape declarable publicly while the inner type stays private.

As a result, `chain_cache.rs` stopped compiling entirely — its field accesses cross a crate boundary into types that are intentionally private. ~35 distinct compile errors were exposed the first time someone ran `cargo test --no-run --manifest-path integration-tests/Cargo.toml` after the change landed.

Patching in place would require adding a fresh pub API surface on `zaino-state` (`blocks_iter`, `block_count`, `heights_to_hashes_iter`, …) solely to let an external test poke at internals. That's API pressure for the wrong reason: the code was correctly made private; the fix isn't to un-privatize it for a single test file.

## Proposed fix

Relocate the tests into `packages/zaino-state/src/chain_index/tests/chain_cache.rs`, alongside the existing `mockchain_tests.rs`. Gated by `#[cfg(test)]`, they gain full legitimate access to `pub(crate)` cache internals and no longer require any public API expansion.

## Fixture question

`chain_cache.rs` currently uses `TestManager` + `generate_blocks_and_poll_chain_index`, which depend on `zaino-testutils` and `zcash_local_net` (live regtest validator, podman, etc.). Moving the tests across crates raises the fixture question:

- **Port to `MockchainSource`** (the harness already used by `src/chain_index/tests/mockchain_tests.rs`). Likely the right answer: the mock has deterministic block generation and no podman dependency, and the tests fundamentally ask *"given this known block sequence, does the cache hold the expected shape?"* — which the mock can answer. Requires re-authoring each `chain_cache.rs` test against the mock harness.
- **Pull `zaino-testutils` in as a dev-dep of `zaino-state`**. Mechanically works, but inverts the workspace dep graph (testutils currently depends *on* zaino-state) and would invite circular dep risk. Not recommended.

Option 1 is the intended target.

## Scope estimate

- ~600 lines of test code to port.
- N test functions (count TBD — see the `mod zebra` entry points in the file) re-authored against `MockchainSource`.
- Delete `integration-tests/tests/chain_cache.rs` once the port is complete.
- No public-API changes to `zaino-state` required.

## Relationship to #1043

#1043 tracks the broader friction from `integration-tests/` being a separately-versioned workspace that drifts silently from `zaino-state` API changes. This issue is one concrete remedy: `chain_cache.rs` was always misfiled, and relocating it both fixes the current breakage and removes one of the largest future-drift risks.

## Blocking / blocked-by

- Currently blocking: `feat/rpc-getblockhash` (PR #642) integration tests — `chain_cache.rs` fails to compile, which prevents any integration test binary from building. That branch is using a targeted workaround; the proper fix lives here.

zancas:
Fixes:  #1044


integration-tests/tests/chain_cache.rs
```diff
@@ -268,187 +263,17 @@ mod chain_query_interface {
...
-
-    // #[ignore = "prone to timeouts and hangs, to be fixed in chain index integration"]
-    #[tokio::test(flavor = "multi_thread")]
-    async fn find_fork_point_zebrad() {
-        find_fork_point::<Zebrad, StateService>(&ValidatorKind::Zebrad).await
-    }
-
-    #[ignore = "prone to timeouts and hangs, to be fixed in chain index integration"]
-    #[tokio::test(flavor = "multi_thread")]
-    async fn find_fork_point_zcashd() {
```
AloeareV, line 305: find fork point is tested in the state_service.rs/fetch_service.rs mods, this removal is acceptable.

integration-tests/tests/chain_cache.rs
```diff
@@ -268,187 +263,17 @@ mod chain_query_interface {
...
-                    .unwrap()
-                    .unwrap()
-                    .0
-            )
-        }
-    }
-
-    // #[ignore = "prone to timeouts and hangs, to be fixed in chain index integration"]
-    #[tokio::test(flavor = "multi_thread")]
-    async fn get_raw_transaction_zebrad() {
```
AloeareV, line 343: Also covered in other test directories

integration-tests/tests/chain_cache.rs
```diff
@@ -268,187 +263,17 @@ mod chain_query_interface {
...
-
-            let correct_txid = zebra_txn.hash().0;
-
-            assert_eq!(txid, correct_txid);
-        }
-    }
-
-    // #[ignore = "prone to timeouts and hangs, to be fixed in chain index integration"]
-    #[tokio::test(flavor = "multi_thread")]
-    async fn get_transaction_status_zebrad() {
```
AloeareV, line 408: Also covered.

AloeareV (approved):

PR #1048, opened 2026-04-24
Add test of coldboot passthrough behavior

### LINKED ISSUE #1047: best_chaintip misclassifies passthrough miss as "hole in validator database" during boot window
### Security check

### Usage mode

zainod (running the daemon)

### Describe the bug

When `snapshot_nonfinalized_state()` falls into its `StillSyncingFinalizedState` branch (the non-finalized state builder has not yet produced its first snapshot), `best_chaintip()` attempts a passthrough fetch of `source.get_block(Height(validator_finalized_height))`. If that fetch returns `Ok(None)` — which happens when zebrad's `ReadStateService` doesn't currently serve the block *and* the JSON-RPC fallback also returns `None`/500 — the code emits `ChainIndexError::database_hole(...)`.

A `database_hole` error asserts that the indexer's own on-disk state contains a recorded height it can't materialize — i.e., corruption. That is not what happened here: there is no indexer-owned record of the block yet. A passthrough miss during the boot window is a benign retryable condition, not corruption. Surfacing it as `Status::Unknown: Chain index error: internal server error: InternalServerError: hole in validator database, missing block N` is a categorical misclassification and gives operators and wallet clients a scary, incorrect signal.

Source location: `packages/zaino-state/src/chain_index.rs:1834-1857` (the `StillSyncingFinalizedState` arm of `best_chaintip`). The offending `database_hole` calls are at lines 1845 and 1850.

Related sites that share the same shape (passthrough miss → `database_hole`) and should be reviewed together:

- `chain_index.rs:948` (`snapshot_nonfinalized_state` — "validator has no best block" labelled as hole)
- `chain_index.rs:1610` (`get_transaction_status` — empty `heights_to_hashes` labelled as hole)
- `backends/state.rs:1848` (`get_raw_transaction` verbose path — `get_compact_block` returning `None` after a tx read labelled as hole)

The primary bug-hit site is the `best_chaintip` one above; the others are sibling misclassifications worth resolving in the same fix.

### Steps to reproduce

1. Check out `dev` at `f807095f` (or any commit where the `wallet_to_validator` integration test exists).
2. Run the shield test against zebrad:
   ```
   makers container-test --manifest-path integration-tests/Cargo.toml -E 'binary(wallet_to_validator) & test(=zebrad::state_service::shield)'
   ```
3. The test fails at `integration-tests/tests/wallet_to_validator.rs:335` on the first `clients.faucet.sync_and_await().await.unwrap()` call, with the server returning `"Chain index error: internal server error: InternalServerError: hole in validator database, missing block 1"`.

The trigger path:

- zebrad has mined ~101 blocks on a fresh regtest chain.
- zaino is still in the boot window: its NFS builder has not produced a first snapshot, so `snapshot_nonfinalized_state()` returns `StillSyncingFinalizedState { validator_finalized_height: Height(1) }` (`tip.saturating_sub(100) == 1`).
- The wallet issues `GetLatestBlock`, which goes through `best_chaintip`.
- `source.get_block(Height(1))` returns `Ok(None)` — `ReadStateService` doesn't serve block 1 yet (still in zebrad's own NFS) and the JSON-RPC fallback misses for the same race window.
- `best_chaintip` emits `ChainIndexError::database_hole(Height(1), None)`; this surfaces to the client as `Unknown: ... missing block 1`.

### Expected behavior

`best_chaintip` (and the other passthrough sites listed above) must distinguish:

- **Indexer-owned corruption** — a height is recorded in the indexer's finalized state index but its bytes are unreachable. This is `database_hole`, a real invariant violation.
- **Passthrough not ready / transient** — the indexer is asking zebrad for data and zebrad can't produce it right now (startup race, NFS-only block, JSON-RPC warming up). This should be a distinct error kind (e.g. `PassthroughNotReady` / `StillSyncing`) that clients can retry on, *or* `best_chaintip` should delay serving until one of the snapshot preconditions is actually satisfiable.

Concretely: a wallet's `GetLatestBlock` against a zaino that is still booting should either block briefly, return a retryable `Unavailable`, or return the best height zaino can defend — but it must not tell the caller that the validator's database has a hole.

### Environment

```
Zaino: commit f807095f (dev)
Backend: Zebra 4.3.1 (per the CI container tag RUST_1.92-ZCASH_6.12.1-ZEBRA_4.3.1-CONTAINER_cb8026a091a4bd)
OS: Linux x86_64 (reproduced in the zaino-ci container image)
```

### Configuration

```toml
# default regtest config used by `integration-tests` TestManager; no custom overrides
```

### Logs

```
START [         ] (1/1) integration-tests::wallet_to_validator zebrad::state_service::shield

thread 'zebrad::state_service::shield' (2571) panicked at tests/wallet_to_validator.rs:335:43:
called `Result::unwrap()` on an `Err` value: SyncError(ServerError(RequestFailed(Status {
    code: Unknown,
    message: "Chain index error: internal server error: InternalServerError: hole in validator database, missing block 1",
    source: None
})))

test zebrad::state_service::shield ... FAILED

 FAIL [  27.922s] (1/1) integration-tests::wallet_to_validator zebrad::state_service::shield
```

### Additional context

Full trace of every `ChainIndexError::database_hole` call site in the tree, classified:

| Location | Trigger | Real corruption? |
|---|---|---|
| `chain_index.rs:813` (`get_indexed_block_height`) | `finalized_state.get_block_height(hash)` → `Err` | yes |
| `chain_index.rs:948` (`snapshot_nonfinalized_state`) | `get_best_block_height` → `None` | **no** — empty chain, not a hole |
| `chain_index.rs:1119` (`get_block_range`) | Indexed hash present, node won't produce bytes | yes |
| `chain_index.rs:1134` (`get_block_range`) | Snapshot has block, node missing bytes | yes |
| `chain_index.rs:1142` (`get_block_range`) | Height ≤ finalized tip, node returns `None` | yes |
| `chain_index.rs:1193` (`get_compact_block`) | `finalized_state.get_compact_block` → `Err` | yes |
| `chain_index.rs:1448` (`find_fork_point`) | `finalized_state.get_block_height` → `Err` | yes |
| `chain_index.rs:1610` (`get_transaction_status`) | `heights_to_hashes` empty | **no** — ambiguous boot/invariant |
| `chain_index.rs:1845 / 1850` (`best_chaintip`, StillSyncing arm) | `source.get_block(Height(validator_finalized_height))` errored / None | **no — this is the bug** |
| `backends/state.rs:1848` (`get_raw_transaction`) | `get_compact_block(tx.height)` → `None` after tx served | **no** — NFS-vs-tx race |

`get_block(tip+1)` sync path — only real call site is `packages/zaino-state/src/chain_index/non_finalised_state.rs:382-393`, which handles `Ok(None)` correctly by exiting the sync loop (not treated as a hole). That path is not implicated in this bug.

**Observed frequency:** reliably reproducible on the cited integration test; in mainnet steady state the `StillSyncingFinalizedState` branch is bypassed once the NFS builder produces its first snapshot, so production wallets will only see this during the first few seconds of a cold zaino+zebrad boot. Still wrong — just narrow.

**Suggested acceptance criteria:**

1. Introduce a distinct `ChainIndexErrorKind::PassthroughNotReady` (or equivalent) and route the `best_chaintip`, `snapshot_nonfinalized_state`, `get_transaction_status`, and `get_raw_transaction` sites through it.
2. Add a regression test in `packages/zaino-state/src/chain_index/tests/` that builds a `NodeBackedChainIndexSubscriber` against a mock source where `get_best_block_height` returns `Some(h)` but `get_block(Height(h - 100))` returns `Ok(None)`, manually constructs a `ChainIndexSnapshot::StillSyncingFinalizedState`, calls `best_chaintip`, and asserts the returned error is **not** `database_hole` (kind is not the corruption kind).
3. Add a companion pinning test that exercises the `non_finalised_state.rs:382` sync loop against a mock where `get_block(Height(tip+1))` returns `Ok(None)`; assert the loop exits cleanly with zero `database_hole` emissions.
4. Client-facing gRPC surface for the new error kind is `Unavailable` (retryable), not `Unknown`.

zancas:
Adds test for #1047


zancas:
Superceded see e.g. #1099

Issue #862, opened 2026-02-23
ZainoDB DbV1 refactor


idky137:
The DbV1 code (https://github.com/zingolabs/zaino/blob/dev/zaino-state/src/chain_index/finalised_state/db/v1.rs) is currently all in a single file which is very large, hindering development on this code. This module should be refactored into sub-modules, divided by capability. 

This will require no code changes, just moving the current code into sub modules. 

Documentation for this code should also be improved in the process to ease development further.

Issue #860, opened 2026-02-23
Implement ZainoDB DbV2


idky137:
Zaino should have a small footprint database option for zallet users that do not require the service improvements that the large DbV1 offers.

This version should only build / save the minimal set of indexes to provide basic chain indexing functionality and serve the extra indexes required in zaino (not provided by zebrad), most notably block / chain cumulative chain work.

### DbV1 Schema / Functionality
Tables:
- Block headers: `Height` -> `StoredEntryVar<BlockHeaderData>`
- Txids: `Height` -> `StoredEntryVar<TxidList>`
- Heights: `Hash` -> `StoredEntryFixed<Height>`

Capability:
- DbRead
- DbWrite
- DbCore
- BlockCoreExt

Migrations:
- V0->V2.0.0 (closely follows V0->V1.0.0)
- V1->V2.0.0 
- V2->V1.0.0

### Note
Currently zebra serves the transparent address history indexes but it has been suggested that Zaino take responsibility for offering this service (DbV1 holds a full transparent address history indexing implementation that is yet to be utilised in zaino), if this is confirmed and the transparent address indexing service removed from zebra then the schema and Capability this new db option would need to be updated to handle this extra functionality.

cartesien:
Real-world bootstrap experience + prefetch optimization patch
Hi, I'm running a self-hosted ZEC node (zebrad 5.0.0 + zainod 0.3.1) on a Windows PC (i7, 32GB DDR5) using WSL2 with the ZainoDB on a native ext4 filesystem (VHD on NVMe SSD — ~5.9 GB/s write speed confirmed).
Observed bootstrap performance:

Sprout zone (17k–120k): ~4 blocs/10s originally, bottleneck confirmed as single-threaded RPC fetching via htop (one core at 55.9%, zebrad at 0.7%)
After applying a buffered(64) prefetch patch to sync_to_height: improved to ~350 blocs/min in that zone (roughly x8–x10 speedup)
Current zone (120k–200k): ~70–80 blocs/min — CPU now saturated at ~99% on one core, disk and RAM are not the bottleneck
Estimated total bootstrap time at this rate: several days on high-end hardware

The patch replaces the sequential for loop in sync_to_height with a futures::stream::iter(...).map(|h| tokio::spawn(...)).buffered(PREFETCH_DEPTH) pipeline, parallelizing block + commitment tree fetches while keeping writes strictly sequential (chainwork integrity preserved).
[...]

PR #1050, opened 2026-04-24
● refactor: extract shared DbLifecycle trait from DbV0/DbV1

### LINKED ISSUE #1033: a background task that completes, may not be registered for shutdown notification in a race condition
Per this comment the solution implemented for #1032 may miss a case where multiple background tasks finished in the gap that's currently covered by a single permit produced via notify_one.

### LINKED ISSUE #1049: DRY: extract shared DbCore lifecycle scaffolding from DbV0/DbV1 (~70 LOC)
Parent: #862
Related: #1033 (bug lives inside the duplicated block), #858 / #860 (DbV2 will triplicate it if left alone)

## Problem

`packages/zaino-state/src/chain_index/finalised_state/db/v0.rs` (1062 LOC) and
`packages/zaino-state/src/chain_index/finalised_state/db/v1.rs` (723 LOC)
contain byte-identical blocks that carry no V0- vs V1-specific logic. Verified
by `diff` (zero output):

| Item | v0.rs | v1.rs |
|---|---|---|
| `impl DbCore::shutdown` body (incl. `notify_one` + `take()` + `select!{handle\|timeout}`) | 187–224 | 153–190 |
| `zaino_db_handler_sleep` (sleep / maintenance.tick / shutdown_notify) | 439–449 | 577–587 |
| `clean_trailing` | 455–459 | 690–694 |
| `open_or_create_db` | 464–476 | 697–709 |
| inherent `status()` + trait forwarder | 178–180, 368–370 | 149–151, 427–429 |
| `wait_until_ready` | 376–386 | 436–446 |

Total: ~70 LOC of pure copy-paste. Adding `DbV2` (#860) would triplicate it.

## Proposed solution

Per CLAUDE.md ("DRY: deduplicate with functions first") extract to the parent
`db.rs` module:

1. A `pub(super) trait DbLifecycle` with required getters exposing the six
   shared struct fields (`env`, `db_handler`, `shutdown_notify`, `status`) and
   **provided methods** for `status()`, `wait_until_ready()`, `shutdown()`,
   `clean_trailing()`, `zaino_db_handler_sleep()`. No macros needed — a
   provided-method trait expresses the abstraction.
2. `pub(super) async fn open_or_create_db(env, name, flags)` as a free
   function (no `self`, no need for a trait).

Both `DbV0` and `DbV1` become implementors by forwarding getters; the
duplicated bodies disappear.

## Must resolve alongside the extraction

**#1033** — the duplicated `shutdown()` uses `shutdown_notify.notify_one()`,
which only stores a single permit. If multiple background tasks need to be
woken, the extra wakes are lost. The DRY'd helper must fix this (widen to
e.g. `Notify::notify_waiters`, a per-task `oneshot`, or a count-gated
barrier) rather than ossify the bug into a shared method.

## Explicit non-goals

- **Table-layout sharing / struct-field sharing.** The table sets of v0 and
  v1 are disjoint in type, not just count:

  | Relation | v0 | v1 |
  |---|---|---|
  | height → hash | `heights_to_hashes` (direct) | derived from `headers[height].context.index.hash` |
  | hash → height | not stored | `heights` (direct, inverse direction from v0) |
  | hash → compact block | `hashes_to_blocks` (JSON blob) | assembled on read from `headers/txids/transparent/sapling/orchard` (five height-keyed tables, `StoredEntry{Var,Fixed}<T>` with checksums) |

  v0's information is a strict subset of v1's, but the *tables* are not —
  keys, values, serialization format (JSON wrappers vs `StoredEntry*`) and
  direction of the hash↔height map all differ. There is no inner-struct
  seam to factor out. This issue is scoped to behavior/lifecycle DRY, not
  schema or table-layout DRY.
- **`spawn_handler` scaffolding.** The outer shape (clone-fields, spawn,
  Ready, 60 s interval, closing-check loop) is parallel, but the inner body
  diverges completely (v0: empty; v1: three initial scans + per-iteration
  validation). No clean seam.
- **Other `tokio::select!` sites.** Five other `select!` sites exist in the
  tree (`finalised_state.rs:555`, `jsonrpc.rs:103`); they are one-offs with
  distinct semantics and are out of scope here.

## Acceptance criteria

- [ ] `DbV0` and `DbV1` implement `DbLifecycle`; the bodies listed in the
      table above are removed from both files.
- [ ] The shutdown race from #1033 is fixed in the shared helper (close
      #1033 as part of this PR or cleanly separate it).
- [ ] `open_or_create_db` is a single free function in `db.rs`, called by
      both backends.
- [ ] Trait visibility is `pub(super)` (per the minimum-visibility rule).
- [ ] No new `tokio::select!` sites introduced; existing duplicated ones
      collapse into one.
- [ ] Behavior unchanged: existing `zaino-state` tests pass, including the
      shutdown/ready sequencing.

## Out of scope / follow-ups

- DbV2 adoption of `DbLifecycle` — handled naturally when #860 lands.
- Status-system unification (#889, #923, #1040) — orthogonal; the trait
  just forwards to whatever `NamedAtomicStatus` becomes.

zancas:
Fixes: #1049
Fixes: #1033 

Move the ~70 LOC of byte-identical lifecycle scaffolding (status,
  wait_until_ready, clean_trailing, zaino_db_handler_sleep, shutdown)
  and the open_or_create_db helper from db/v0.rs and db/v1.rs into a
  single pub(super) trait DbLifecycle and a free fn in the parent db.rs.

  DbV0 and DbV1 now impl DbLifecycle via four field getters; their
  DbCore::{status, shutdown} delegate to the trait. The duplicated
  tokio::select! shutdown/abort block and the three-way
  sleep-or-maintenance body now live in one place.

  DbBackend::{status, shutdown} in db.rs use fully-qualified DbCore::
  paths to disambiguate the two traits now in scope on DbV0/DbV1.

  #1033 (notify_one race) preserved verbatim on DbLifecycle::shutdown —
  fixing it is now a single-call-site change.

  Parent: #862


zancas:
How to fix #1033, I chose tokio_util "option C" below.
 
Analysis

  B (watch::channel<bool>) is the minimum-dependency choice. The pattern is already in use a few hundred
   lines away in finalised_state.rs:547, so adopting it here makes the codebase more uniform. Waiter
  ergonomics (changed() + borrow() or wait_for) are slightly clunkier than C's flat cancelled().await,
  but not much. No new direct crate edge.

  C (tokio_util::sync::CancellationToken) is the semantic fit — purpose-built for exactly this scenario.
   Waiter and notifier APIs are each one line. The decisive feature is child_token(): as DbV2 lands
  (#860) and likely spawns multiple sub-tasks (validator, compactor, metrics), each can carry a child
  token. Cancelling a child does not propagate up, which contains worker-bug blast radius — a property
  watch does not offer without manual scoping. tokio-util is a near-universal companion to tokio; adding
   it as a direct edge is cheap.

zancas:
@idky137 I have not carefully evaluated the claims in "C", I would appreciate your insights worker threads management at that layer.

packages/zaino-state/src/chain_index/finalised_state/db.rs
```diff
@@ -626,3 +732,80 @@ impl TransparentHistExt for DbBackend {
...
+    /// Regression for #1033 — every task awaiting cancellation must observe shutdown,
+    /// not just one. Originally written against the `Notify::notify_one` implementation
+    /// (which strands N-1 waiters); now passes against `CancellationToken::cancel`,
+    /// which wakes all current waiters and persists state for late subscribers.
+    #[tokio::test]
+    async fn wakes_every_shutdown_waiter() {
+        let tmp = tempfile::tempdir().unwrap();
+        let env = Arc::new(
+            lmdb::Environment::new()
+                .set_map_size(1 << 20)
```
zancas, line 777: @idky137 I don't understand why this is necessary.

idky137:
utACK modulo one architectural concern for the follow-up discussion.

This refactor looks fine for the current LMDB-backed `DbV0`/`DbV1` implementations, and I do not think it blocks the planned graph-based migration work.

However, one thing worth clearing up is that `DbLifecycle` now bakes LMDB-specific details into the shared `DbVx` lifecycle abstraction via `env()` and `clean_trailing()`. The versioned DB model was intended to allow each `DbVx` to choose its own underlying storage implementation — LMDB today, but potentially RocksDB or a future Rust-native store later. This trait moves us slightly further toward “every `DbVx` has an LMDB `Environment`”.

Not a blocker for this PR, but before this abstraction spreads further we may want to make a small adjustment so the generic lifecycle remains storage-agnostic, and LMDB-specific cleanup/sync behaviour is provided by the LMDB-backed implementations only (Enabling moving to a pure rust DB impl when one is available in the future).

integration-tests/.config/nextest.toml
AloeareV, line 1: Not a blocker, but why do we need this manual list of tests? It seems like a bit of a footgun to have this list we need to add/remove tests from by hand from lest...something goes out of alignment? I may be misunderstanding what this file does.

packages/zaino-state/src/chain_index/finalised_state/db.rs
```diff
@@ -71,19 +71,125 @@ use crate::{
...
 use super::capability::Capability;
 
+/// Lifecycle scaffolding shared by every `DbVx` finalised-state backend.
+///
+/// Implementors expose the four shared struct fields via required getters;
+/// provided methods cover the duplicated `status()`, `wait_until_ready()`,
+/// `shutdown()`, `clean_trailing()`, and the background task's per-iteration
+/// `zaino_db_handler_sleep()`.
+#[async_trait]
+pub(super) trait DbLifecycle: Sync {
```
AloeareV: Renaming this to `LmdbLifecycle`, as that leaves us a couple breadcrumbs. 
* Makes it clear that this ties the database implementation to LMBD.
* Implies the possible future existence of a non-Lmdb DbLifecycle.

packages/zaino-state/src/chain_index/finalised_state/db.rs
```diff
@@ -71,19 +71,125 @@ use crate::{
...
+use tokio::{
+    task::JoinHandle,
+    time::{interval, sleep, MissedTickBehavior},
+};
+use tokio_util::sync::CancellationToken;
+use tracing::warn;
 
 use super::capability::Capability;
 
+/// Lifecycle scaffolding shared by every `DbVx` finalised-state backend.
```
AloeareV, line 96: Also alter this comment to make clear the LMDB connection here.

AloeareV (dismissed):

packages/zaino-state/src/chain_index/finalised_state/db.rs
zancas: Let's issue this?  And call it nonblocking?

packages/zaino-state/src/chain_index/finalised_state/db.rs
zancas: Again...   if we're not generalizing now, then we should open a non-blocking issue document the intent to generalize.

AloeareV: We have many old issues about things like this that are not in my regular awareness. I've added this document to the code, as that has a much higher chance of being seen when we go to make this sort of change.

AloeareV (approved):
Reapproving after adding non-functional changes

PR #1054, opened 2026-04-24
helperize shutdown signal

### LINKED ISSUE #1052: refactor(zaino-serve): replace shutdown_signal polling with CancellationToken
Parent: #1051

## Today

Both `packages/zaino-serve/src/server/jsonrpc.rs:86-95` and `…/grpc.rs:49-58` build an explicit polling future to integrate with their underlying server's shutdown plumbing (jsonrpsee's `stop` race, tonic's `serve_with_shutdown`):

```rust
let shutdown_check_status = status.clone();
let mut shutdown_check_interval = interval(Duration::from_millis(100));
let shutdown_signal = async move {
    loop {
        shutdown_check_interval.tick().await;
        if shutdown_check_status.load() == StatusType::Closing {
            break;
        }
    }
};
```

This block exists *only* because `NamedAtomicStatus` lacks a wake primitive. Observed shutdown latency from `close()` to signal completion: 0–100 ms (~50 ms typical).

## Proposed change

- Add `cancel_token: CancellationToken` to both `JsonRpcServer` and `TonicServer` (`zaino-serve` already gets `tokio-util` transitively via the new dep added on #1049's branch — verify).
- Replace each polling closure with:
  ```rust
  let token = self.cancel_token.clone();
  let shutdown_signal = async move { token.cancelled().await };
  ```
- `close()` calls `self.cancel_token.cancel()` AND still stores `StatusType::Closing` (the status flag remains the externally observable state indicator and gates `status()` reads).
- `Drop` continues to `abort()` as a backstop for the "dropped without close" path.

## Acceptance criteria

- The `shutdown_check_interval` polling closure is gone from both `jsonrpc.rs` and `grpc.rs`.
- New unit test asserts the shutdown-signal future completes within ~20 ms after `close()`. The test fails against the current polling pattern and passes after migration.
- Existing server-shutdown integration tests remain green.

## Out of scope

- Token parent linkage to a zainod-level root — that's #1051. Land each server with its own owned `CancellationToken`; child-token wiring follows when zainod owns the root.

zancas:
Pre-requires:  #1050 

Fixes: #1052


AloeareV (approved):
Nice!

PR #1055, opened 2026-04-25
● refactor(status): back NamedAtomicStatus with tokio::sync::watch

### LINKED ISSUE #1037: fix: flake in get_mempool_stream_for_stale_snapshot
Parent: #1035

## Describe the bug

`chain_index::tests::mockchain_tests::get_mempool_stream_for_stale_snapshot` fails ~0.1% of the time (**1 / 1000 runs**) under `cargo nextest run` with no extra args.

Observed with `tools/thousand_runs.sh` (`N_RUNS=1000`) against commit `cd5d8adc` on branch `speed_up_unit_tests`.

## Location

`packages/zaino-state/src/chain_index/tests/mockchain_tests.rs:503`

## Likely root cause

Matches the race category flagged during the 50 ms `SYNC_LOOP_INTERVAL` experiment: the mempool's own chain-tip watch channel can lag the indexer's, so `get_mempool_stream(Some(stale))` occasionally returns `Some` before the staleness check fires. At the current default 500 ms `SYNC_LOOP_INTERVAL` (`SyncTimings::default()`, commit `8a011430`) this is rare but non-zero.

## Expected behavior

Deterministic: 1000 / 1000 runs pass.

## Steps to reproduce

```bash
N_RUNS=1000 tools/thousand_runs.sh
```

Note: with a 0.1% rate, a 1000-run sweep will typically see 0–3 failures — plan for multiple sweeps (or a larger `N_RUNS`) when verifying a fix.

## Acceptance criteria

- [ ] Close the tip-channel skew between mempool and indexer at the source, OR restructure the test so it doesn't depend on the relative observation order of the two watch channels
- [ ] At least one `N_RUNS=1000` sweep on `dev` shows 0 failures for this test after the fix (prefer 2+ sweeps given the low base rate)

zancas:
Subscribe to the chain-index sync loop's `NamedAtomicStatus` via
  `tokio::sync::watch` so test setup wakes on the first `Ready`
  transition instead of after the next 2 s polling tick. A single
  `sleep(sync_timings.interval)` after rendezvous preserves the
  load-bearing teardown alignment at ~500 ms (down from 2 s); that shim
  disappears when the indexer gains `CancellationToken`-based shutdown.
  `wait_for_indexer_tip` is rewired against the same watch with a
  `tokio::time::timeout` budget.

Fixes: #1037


packages/zaino-state/src/status.rs
```diff
@@ -70,10 +74,11 @@ impl NamedAtomicStatus {
 
     /// Loads the value held in the NamedAtomicStatus.
     pub fn load(&self) -> StatusType {
-        StatusType::from(self.inner.load(Ordering::SeqCst))
+        *self.inner.borrow()
     }
 
-    /// Sets the value held in the NamedAtomicStatus, logging the transition.
+    /// Sets the value held in the NamedAtomicStatus, logging and broadcasting
+    /// the transition. Storing the current value is a no-op.
```
idky137: `NamedAtomicStatus::store` should use `watch::Sender::send_if_modified` rather than `load()` + `send_replace()`. The current code is thread-safe, but the equality check is not atomic with the update, so concurrent identical stores can still double-log/double-notify. `send_if_modified` matches the documented behaviour in the comment.

possible solution:
```
pub fn store(&self, status: StatusType) {
    self.inner.send_if_modified(|current| {
        if *current == status {
            false
        } else {
            debug!(
                component = self.name,
                from = %*current,
                to = %status,
                "[STATUS] transition"
            );
            *current = status;
            true
        }
    });
}
```

packages/zaino-state/src/chain_index/source.rs
```diff
@@ -874,6 +910,14 @@ pub(crate) mod test {
         >,
         active_chain_height: Arc<AtomicU32>,
         force_requests_against_source_to_fail: Arc<std::sync::atomic::AtomicBool>,
+        /// Pings every subscriber registered via
+        /// [`BlockchainSource::change_subscribe`] when [`Self::mine_blocks`]
+        /// advances the active height, so each can wake from its interval
+        /// timer immediately. `broadcast` (over `Notify`) gives every
+        /// subscriber its own buffered receiver, so a `mine_blocks` that
+        /// fires while one subsystem is mid-iteration is preserved on its
+        /// receiver and consumed on the next `recv().await`.
```
idky137: This overstates the `broadcast` guarantee. `broadcast` gives each subscriber its own receiver, but it is still bounded: if a receiver falls behind far enough, older messages are dropped and `recv()` returns `RecvError::Lagged`.

That is probably fine here because this is only being used as a coalescible “source changed” wakeup and the loop re-reads the current source state after waking. But the comment should not claim that every `mine_blocks` notification is preserved and consumed on the next `recv().await`.

Could we reword this to say that `broadcast` gives each loop an independent wakeup path, but notifications are not a lossless event log.

packages/zaino-state/src/chain_index.rs
```diff
@@ -1761,13 +1780,31 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+            }
+        }
+
         let mut subscriber = self.mempool.clone();
 
-        match subscriber
-            .get_mempool_stream(expected_chain_tip)
-            .now_or_never()
-        {
+        match subscriber.get_mempool_stream(None).now_or_never() {
```
idky137: I don't think we should drop the `expected_chain_tip` check here.

The previous call passed the snapshot tip into the mempool subscriber:

`subscriber.get_mempool_stream(expected_chain_tip)`

This PR now validates the snapshot against the chain-index's current non-finalized tip, then calls:

`subscriber.get_mempool_stream(None)`

That fixes the flake by avoiding the lagging mempool-tip comparison, but it also weakens the `get_mempool_stream(snapshot)` contract. The stream can now be created without checking that the mempool stream is aligned to the caller's snapshot tip.

I think the correct fix is to keep passing the expected tip to the mempool layer, but make the mempool layer handle tip skew correctly. In other words: snapshot freshness should be checked against the chain-index's authoritative snapshot state, but the returned mempool stream should still be guarded by the expected mempool chain tip so that the stream terminates/rejects consistently when the mempool's observed tip does not match the snapshot.

Concretely, I think this should remain structurally closer to:

```rust
let expected_chain_tip = non_finalized_snapshot.map(|snapshot| snapshot.best_tip.hash);

// optional: validate expected_chain_tip against current chain-index NFS here

match subscriber
    .get_mempool_stream(expected_chain_tip)
    .now_or_never()
{
    ...
}

**Review by idky137 — CHANGES_REQUESTED** [review]
this is mostly a good. I think the watch-based status subscription and event-driven test waits are an improvement over fixed polling sleeps.

I left three comments:
1. `NamedAtomicStatus::store` should use `send_if_modified()` so duplicate suppression is atomic with the update.
2. The `broadcast` docs should describe these notifications as coalescible wakeups, not preserved events, since lagging receivers can skip messages.
3. I do not think `get_mempool_stream(expected_chain_tip)` should be replaced with `get_mempool_stream(None)`. We should keep the expected-tip guard and fix the skew handling in the mempool layer rather than bypassing that check.

With those addressed, this looks good to me.

File: packages/zaino-state/src/chain_index/tests.rs [hunk 4]
```diff
@@ -119,18 +98,35 @@ async fn load_with_settings(
...
+    // parked in its `tokio::time::sleep(sync_timings.interval)` (see
+    // chain_index.rs sync loop) at teardown so runtime shutdown doesn't
+    // race a mid-iteration DB write. The watch-based rendezvous above
+    // wakes one statement *before* that sleep is polled, so under the
+    // multi-thread runtime there is a sub-millisecond window where the
+    // sync loop has not yet yielded. One full `interval` here lets the
+    // sync loop reach the sleep statement and stay there. Removing this
+    // requires explicit indexer shutdown (`CancellationToken`-based,
+    // tracked separately).
+    tokio::time::sleep(sync_timings.interval).await;
```
AloeareV: Leaving a breadcrumb comment here, hasn't this CancellationToken based API been added recently?

packages/zaino-state/src/chain_index/tests/mockchain_tests.rs
```diff
@@ -1,61 +1,53 @@
...
-        "indexer tip to match expected height",
-        Duration::from_secs(10),
-        Duration::from_millis(25),
-        || async {
-            let tip = index_reader
-                .snapshot_nonfinalized_state()
+    let mut status = index_reader.status_subscribe();
+    let work = async {
+        loop {
+            if *status.borrow_and_update() == StatusType::Ready {
```
AloeareV, line 36: I don't understand the use of `borrow_and_update` here, what are we updating? Maybe I misunderstand this API.

packages/zaino-state/src/status.rs
zancas: Suggestion applied.

packages/zaino-state/src/chain_index/source.rs
zancas: Addressed.

packages/zaino-state/src/chain_index.rs
zancas: Addressed.

zancas:
Comments to this point have been addressed and all tests pass locally.

packages/zaino-state/src/chain_index/source/mockchain_source.rs
```diff
@@ -184,6 +203,8 @@ impl MockchainSource {
             force_requests_against_source_to_fail: Arc::new(std::sync::atomic::AtomicBool::new(
                 false,
             )),
+            change_broadcast: tokio::sync::broadcast::channel(16).0,
```
AloeareV: Why 16?

packages/zaino-state/src/chain_index/source/mockchain_source.rs
```diff
@@ -244,6 +267,30 @@ impl MockchainSource {
...
+        }
+    }
+
+    /// Like [`Self::mine_blocks`] but does *not* fire the source's
+    /// change-notify. Lets the chain-index sync loop fall through to its
+    /// timer instead of waking immediately — the only way to put the
+    /// chain-index *behind* the mempool in tests, since the mempool's
+    /// serve loop polls `get_best_block_hash` directly and always
+    /// notices, notify or not.
+    pub(crate) fn mine_blocks_silent(&self, blocks: u32) {
```
AloeareV, line 303: This is duplicate logic of almost all of `mine_blocks`

AloeareV (changes requested):

packages/zaino-state/src/chain_index/tests.rs
AloeareV: Looks to me like this has since been added to this PR, or at least, this comment no longer exists.

packages/zaino-state/src/chain_index/source/mockchain_source.rs
zancas: Wrong idiom, fixed here:  https://github.com/zingolabs/zaino/pull/1055/changes/eed2200f15116e4c8a07dc941d5ba2f35e2f13b9

packages/zaino-state/src/chain_index/source/mockchain_source.rs
zancas: Good catch!  Fixed by:  b578fa80f2a0334f3d2711ae33e99dc5fcb48a51

packages/zaino-state/src/chain_index/tests/mockchain_tests.rs
AloeareV: Right, this is this tokio thing with the confusing method names. This is 'read and mark read' as opposed to borrow, which means 'peek without marking read'. Uhg.

packages/zaino-state/src/chain_index.rs
```diff
@@ -60,6 +62,33 @@ pub mod types;
...
+/// derived from the chain's current best tip.
+///
+/// Blocks at height `<= finalized_height(chain_height)` are finalized and live
+/// in the finalized DB; blocks above are non-finalized and live in the NFS.
+/// The seam overlaps by exactly one block (the block at this height appears
+/// in both layers).
+///
+/// Saturates at zero on early chain (`chain_height < NON_FINALIZED_DEPTH`),
+/// meaning the finalized seam is at genesis.
+pub(crate) fn finalized_height(chain_height: u32) -> types::Height {
```
AloeareV: There's a scenario where a new best tip comes in during a reorg, with a greater cumulative chainwork, but a lower height, than the current best tip. In this case, the finalized height will be <100 blocks below the chaintip. We can't rely on this function to accurately identify the finalized height in all cases. Outside of the scope of this PR, as this logic all existed, this is a previously unknown bug that is not introduced in this PR. See #1128 for complete issue description.

packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -244,6 +244,21 @@ impl NonfinalizedBlockCacheSnapshot {
...
+
+    /// Predicate: is this block already represented in the NFS?
+    ///
+    /// True iff the block's coinbase height is at or below the current
+    /// best tip — meaning the iter (or a prior iter) has already processed
+    /// the height. Used by `sync` to skip pre-fetched window entries that
+    /// don't extend the chain. Blocks without a coinbase height (malformed
+    /// input) are not considered already-in-NFS; the caller's parent-hash
+    /// chain check will surface the problem.
+    fn block_is_already_in_nfs(&self, block: &zebra_chain::block::Block) -> bool {
```
AloeareV, line 256: This name will see non-best-chain blocks with a height the same as a block on the best chain as already being in the nfs. This method is only sound to call on blocks in the best chain. If this is only intended to be used for best-chain blocks, can that be made more clear in the name?

packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -349,9 +364,26 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
             .map(Mutex::new)
     }
 
-    /// sync to the top of the chain, trimming to the finalised tip.
-    #[instrument(name = "NonFinalizedState::sync", skip(self, finalized_db))]
-    pub(super) async fn sync(&self, finalized_db: Arc<ZainoDB>) -> Result<(), SyncError> {
+    /// Sync the NFS to the iter's committed chain tip, trimming to the
+    /// finalised tip.
+    ///
+    /// `window` is the iter's pre-fetched view of the non-finalized region —
```
AloeareV, line 370: Currently, we can start work as soon as the first block is fetched. This delays the start of work until the entire window is fetched. Could we make this a collection of futures? A `futures::stream::FuturesOrdered` may be the best way...although a `Vec<F>` may get the job done fine. Not too clear on the advantages of the more specialized type here.

packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -379,19 +411,13 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
...
-                SyncError::ValidatorConnectionError(NodeConnectionError::UnrecoverableError(
-                    Box::new(e),
-                ))
-            })?
-        {
+        for block in window {
+            // Window may start below `working_snapshot.best_tip` (when iter
+            // N-1 already advanced past `finalized_height_floor(chain_height)`); skip
+            // those entries.
+            if working_snapshot.block_is_already_in_nfs(&block) {
```
AloeareV, line 418: I think this has the potential to skip blocks not on the best chain, if there's already a block on the best chain with the same height. See comment on the `block_is_already_in_nfs` method.

AloeareV (changes requested):
Partial review: 14 of 19 files read.

packages/zaino-state/src/chain_index.rs
```diff
@@ -836,20 +907,62 @@ impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
...
+                    // committed to (#1126).
+                    //
+                    // Range is `(finalized_height.0 + 1)..=chain_height.0` —
+                    // exactly `NON_FINALIZED_DEPTH` blocks in steady
+                    // production, fewer when the chain itself is shorter
+                    // than the depth (early chain / regtest).
+                    let window_first = finalized_height.0 + 1;
+                    let mut window: Vec<Arc<zebra_chain::block::Block>> =
+                        Vec::with_capacity(chain_height.0.saturating_sub(finalized_height.0) as usize);
+                    for h in window_first..=chain_height.0 {
```
AloeareV, line 926: A couple things. 

First of all, this is not reorg-aware. If a reorg happens between two block fetches, we could end up serving some blocks from one chain, and then others from another chain.

(also, out of the scope of this PR, but maybe block fetch could be parallelized).

packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -768,3 +794,85 @@ impl Block for zebra_chain::block::Block {
         nfs.block_to_chainblock(prev_block, self).await
     }
 }
+
+#[cfg(test)]
+mod block_is_already_in_nfs_predicate {
+    //! Reproducer for [#1129](https://github.com/zingolabs/zaino/issues/1129):
```
AloeareV, line 800: We were talking about this test just as we split. I think there was something about it you are changing, that I didn't fully follow.

zancas:
I am factoring this work into a series of more coherent follow up commits.

PR #1059, opened 2026-04-26
[#1058] CompactBlock in GetBlockRange reports different version from LightWalletD

### LINKED ISSUE #1058: CompactBlock in GetBlockRange reports different version from LightWalletD
### Security check

### Usage mode

zainod (running the daemon)

### Describe the bug

the problem is this field specifically.

message CompactBlock {
    uint32 protoVersion = 1;         // the version of this wire format, for storage

lightwalletd returns 0, Zaino 4. Neither of those numbers make much sense but given there is no versioning for storage it makes more sense that it's zero instead of of 4

### Steps to reproduce

run integration tests grpc comparison tests. https://github.com/zcash/integration-tests/pull/84/

### Expected behavior

this should be equivalent 

### Environment

all platforms

### Configuration

```toml

```

### Logs

```shell

```

### Additional context

_No response_

pacu:
closes #1058

### DIFF
--- a/packages/zaino-fetch/src/chain/block.rs
+++ b/packages/zaino-fetch/src/chain/block.rs
@@ -409,7 +409,7 @@ impl FullBlock {
         let header = Vec::new();
 
         let compact_block = CompactBlock {
-            proto_version: 4,
+            proto_version: 0,
             height: self.height as u64,
             hash: self.hdr.cached_hash.clone(),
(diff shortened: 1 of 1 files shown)


zancas (changes requested):
Can we add a sanity/regression test here?   How could this skew have happened?

zancas (approved):

PR #1067, opened 2026-04-27
Align CompactTxStreamer responses with lightwalletd

### LINKED ISSUE #1066: Align Zaino CompactTxStreamer responses with lightwalletd
The strict gRPC comparison test in `zcash/integration-tests` surfaced several CompactTxStreamer response divergences between Zaino and lightwalletd.

Zaino should match lightwalletd for the compact block and nullifier response shapes covered by the test, so the comparison suite can detect real protocol or implementation regressions instead of known differences.

Required changes:

- `GetBlock` should return full compact transaction data, including transparent inputs/outputs, matching lightwalletd.
- `GetBlockNullifiers` should clear transparent `vin`/`vout`, matching lightwalletd’s nullifier-only response.
- Compact blocks should use the same `protoVersion` value returned by lightwalletd.
- Tree-state lookups should be served from the backing validator instead of depending on the local chain-index cache containing the block first.
- Recent non-finalized compact block lookups should fall back to the backing validator when the local snapshot/finalized DB does not contain the block yet, instead of returning a database-hole error.

Acceptance criteria:

- Zaino matches lightwalletd for the compact block/nullifier fields checked by the gRPC comparison test.
- `GetBlock` and `GetBlockRange` work for serviceable recent blocks while Zaino’s local cache is still catching up.
- `uv run ./qa/zcash/grpc_comparison_tests.py` passes in `zcash/integration-tests` against the updated Zaino binary.

pacu:
Note: this is a draft PR since the `protoVersion` part is yet to be defined. 

Fixes:
## Summary

Align Zaino CompactTxStreamer compact block responses with lightwalletd for the fields covered by the strict integration-tests gRPC comparison suite.

This PR:

- returns all pool data for single-block `GetBlock` responses
- matches lightwalletd `protoVersion` values in compact blocks
- clears transparent `vin`/`vout` in `GetBlockNullifiers`
- delegates tree-state lookups to the backing validator
- falls back to the backing validator when a serviceable recent compact block is absent from the local non-finalized snapshot/finalized DB

## Verification

- `cargo build --release --bin zainod`
- `timeout 600s uv run ./qa/zcash/grpc_comparison_tests.py` from `zcash/integration-tests`

Closes #1066


pacu:
This branch succeeds against this branch of integration tests https://github.com/zcash/integration-tests/pull/84

zancas:
Can you please run:

`makers container-test && makers integration-test --profile stable`

There might be flakes that need more than one run in the integration tests..   stable.....

pacu:
> Can you please run:
> 
> `makers container-test && makers integration-test --profile stable`
> 
> There might be flakes that need more than one run in the integration tests.. stable.....

I found some errors within the containerized setup which I'm fixing

packages/zaino-state/src/chain_index.rs
```diff
@@ -801,6 +804,78 @@ pub struct NodeBackedChainIndexSubscriber<Source: BlockchainSource = ValidatorCo
...
+            ))
+        })?,
+        orchard_root,
+        orchard_size.try_into().map_err(|_| {
+            ChainIndexError::backing_validator(std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                "orchard commitment tree size overflow",
+            ))
+        })?,
+        ChainWork::from_u256(0.into()),
```
AloeareV, line 1057: Maybe not in scope for this PR, but if we're going to have a dummy value of 0 here we should _really_ turn this into an Option

AloeareV: (If we use an `Option<NonZero>` type, that lets us convert safely between an `Option<NonZeroU256>` and a `U256` (where 0 represents the None case) for ease of serializing it still as a regular number and representing it in our code as an option.)

idky137 (dismissed):

packages/zaino-state/src/chain_index.rs
idky137: I think this would be a good update to make but, as ZainoDB uses the IndexedBlock code that would need updating for this, maybe should be done in a follow up PR so we can land the LWD replacement?

pacu: That sounds like a good enhancement but it seems a bit out of the scope of this specific PR

AloeareV (approved):

idky137 (approved):

PR #1071, opened 2026-04-27
Handle empty upgrade metadata in GetLightdInfo

### LINKED ISSUE #1070: GetLightdInfo should handle empty upgrade metadata without panic
Summary

`GetLightdInfo` assumes validator `getblockchaininfo.upgrades` has at least one entry. If that metadata is empty, both the fetch and state backends reach `.last().expect(...)` and panic instead of returning a service error.

Affected paths

- `packages/zaino-state/src/backends/fetch.rs`
- `packages/zaino-state/src/backends/state.rs`

Impact

A malformed, incompatible, or misconfigured validator response can make the lightwalletd info path panic. I am framing this as low severity robustness, not funds risk.

Proposed fix

Return `failed_precondition` when upgrade metadata is empty, and keep serving normal responses when metadata is present.

PR incoming with the smallest patch I could make for this one concern.

Zk-nd3r:
Summary

This replaces the `GetLightdInfo` empty-upgrades panic path with a `failed_precondition` service error.

Changes

- add a small helper for selecting the latest validator network upgrade
- use it from both fetch and state backends
- add a regression test for empty upgrade metadata

Context

This is the first single-concern issue/PR pair from the Zaino findings list. Kept narrow on purpose so it can fit your normal review and test policy before I split out the rest.

Testing

- `cargo +stable fmt --all --check`
- `git diff --check`

Could not run the targeted Rust test on this Windows host because the local GNU toolchain is missing `dlltool.exe`; build stops in dependency compile before reaching `zaino-state`.

Closes #1070


idky137:
utACK (Thanks for finding and fixing this, and helping to make Zaino stronger!)

Zk-nd3r:
CI note: \\CI - PR\\ is green. The failed \\Trigger Integration Tests\\ run dies before repo code runs, in \\Generate app token\\:

\\Secret source: None\\
\\Error: [@octokit/auth-app] appId option is required\\

So this looks like integration-test dispatch secret/app-token wiring for the workflow context, not a failure in this branch.

Zk-nd3r:
Rebased onto current `dev` via @idky137, thanks.

Looks ready for formal review whenever convenient. The current red `Trigger integration tests` check appears to fail before tests run, at `Generate app token` with `appId option is required`, so I don't think it is branch-specific. `compute-tag` is green.

Happy to re-sync again if anything else lands on `dev` before merge.

idky137 (approved):
Thanks for this! And sorry it took so long to review!

We have a few issues with CI currently so checking locally while things are down

PR #1075, opened 2026-04-28
Update containers against worktrees

zancas:
This eliminates unnecessary builds in worktrees that don't contain standard .git sub-directories.


zancas:
This is blocked on publication of infrastructure.

PR #1078, opened 2026-04-28
Run container tests without final-stage root

### LINKED ISSUE #1077: Container test image requires root-owned setup paths
## Problem

Running `makers container-test` exposed two related container-test issues:

1. Some cargo-make tasks are executed by `/bin/sh` even though their scripts use Bash-only syntax (`[[ ... ]]`) or inherit a prelude that uses `source`. This causes failures like:

```text
[[: not found
source: not found
```

2. The test image final stages switch back to `USER root` to copy/chmod test artifacts and install helper tools under root-owned paths. A maintainer clarified that root-requiring instructions are against project policy. The final image setup should not require root-only instructions, and runtime tests should not depend on write access to root-owned paths.

A related failure appeared in the containerized test suite: `chain_index::tests::finalised_state::v1::load_db_backend_from_file` attempted to open a checked-in LMDB fixture in place. LMDB creates lock metadata next to the database, which fails in the container when the fixture path is effectively read-only or not writable by the test user.

## Expected behavior

`makers container-test` should run under the non-root `container_user` policy, without final-stage root setup, and without tests mutating checked-in fixtures.

## Observed behavior

Before the fix:

```text
/tmp/fsio_...sh: [[: not found
/tmp/fsio_...sh: source: not found
```

After getting past the shell issue, the image also depended on root-owned setup paths, and the LMDB fixture test failed with:

```text
LmdbError(Other(13))
```

where `13` is `EACCES`.

## Proposed fix

- Mark Bash-dependent cargo-make tasks with `script_runner = "bash"`.
- Keep final image targets on `container_user` instead of switching to `USER root`.
- Use `COPY --chown=container_user:container_user --chmod=0755` for test artifacts and the entrypoint instead of copying as root and chmodding later.
- Install cargo helper tools into the configured cargo home, which is made writable before switching to `container_user`.
- Copy the LMDB fixture to a temporary directory before opening it in tests.

## Validation

Verified locally with:

```text
makers container-test
```

Result:

```text
232 tests run: 232 passed, 2 skipped
```

pacu:
## Summary

Fix `makers container-test` so the test container setup follows the non-root container policy and the test suite no longer depends on mutating checked-in LMDB fixtures.

This PR:

- Adds `script_runner = "bash"` to cargo-make tasks that use Bash syntax or inherit a Bash-only prelude.
- Keeps the final test-image targets on `container_user` instead of switching back to `USER root`.
- Copies test artifacts and the entrypoint with `COPY --chown=container_user:container_user --chmod=0755` instead of copying as root and chmodding afterwards.
- Installs `cargo-nextest` and `rust-script` into the configured cargo home, made writable before switching to `container_user`.
- Copies the v1 LMDB fixture to a temp directory before opening it, avoiding writes beside checked-in fixture files.

## Validation

```text
makers container-test
```

Result:

```text
232 tests run: 232 passed, 2 skipped
```

closes #1077

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -132,6 +132,7 @@ script.post = ""
 
 [tasks.init-podman-volumes]
 description = "Initialize named podman volumes for container builds"
+script_runner = "bash"
 script = '''
 for vol in zaino-container-target zaino-cargo-git zaino-cargo-registry; do
     if ! podman volume inspect "$vol" >/dev/null 2>&1; then
@@ -178,6 +179,7 @@ echo "CONTAINER_DIR_HASH=$HASH"
 [tasks.ensure-image-exists]
 description = "Ensure the image exists locally, building if needed"
 dependencies = ["init-podman-volumes"]
+script_runner = "bash"
 extend = "base-script"
 script.main = '''
 if ! podman image inspect ${IMAGE_NAME}:${TAG} > /dev/null 2>&1; then
--- a/integration-tests/test_environment/Containerfile
+++ b/integration-tests/test_environment/Containerfile
@@ -67,6 +67,7 @@ ARG CARGO_HOME
 
 ENV CARGO_HOME=${CARGO_HOME}
 ENV CARGO_TERM_COLOR=always
+ENV PATH="${CARGO_HOME}/bin:${PATH}"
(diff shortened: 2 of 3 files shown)


zancas (approved):
Way better!

PR #1082, opened 2026-04-29
● [#1081] fix TonicServer::spawn bind race

### LINKED ISSUE #1081: TonicServer::spawn returns Ok before bind; AddrInUse and other startup failures silently swallowed
## Summary

`TonicServer::spawn` (`packages/zaino-serve/src/server/grpc.rs:62-69`) returns `Ok(TonicServer { ... })` *before* its inner `server_future.await` does the actual TCP bind. If that bind fails — `EADDRINUSE`, `EACCES`, anything — the error is swallowed inside the spawned task, the status field stays cached at `StatusType::Ready` from a `store` call placed *before* the await, and zainod's main process keeps running indefinitely with the status board claiming everything is fine.

Downstream consumers (test fixtures, real wallets) get `ConnectionRefused` on the configured gRPC port with no upstream signal that anything went wrong.

## The bug

```rust
// packages/zaino-serve/src/server/grpc.rs:62-69
let server_handle = tokio::task::spawn(async move {
    task_status.store(StatusType::Ready);   // ← status = Ready BEFORE bind
    server_future.await?;                    // ← bind happens here; on Err, `?` exits the task
    task_status.store(StatusType::Offline);  // ← only reached on success path
    Ok(())
});

Ok(TonicServer {
    status,                                  // ← caller now holds a TonicServer
    server_handle: Some(server_handle),      // ←   reporting Ready
})                                           // ←   on a possibly-dead task
```

`Server::serve_with_shutdown(addr, _)` builds a future whose `.await` does the actual `TcpListener::bind(addr)`. If the bind returns `Err`, the `?` propagates inside the spawned task and the task terminates — but `task_status` was already `store(Ready)` on the line above, and is never updated to `Offline` on the failure path.

The parent serve loop in `packages/zainod/src/indexer.rs::launch_inner` polls `check_for_critical_errors()`, which derives from `status_int()`, which reads `self.server.as_ref().map(|s| s.status())` — the same stale `Ready`. So critical-error detection never fires, and zainod stays alive forever with a dead gRPC server.

## Production impact

Operator runs \`zainod\`. Some other process on the host already holds the gRPC port (e.g., a leftover instance, an unrelated service, a stale port mapping). zainod logs:

\`\`\`
INFO Starting Zaino
INFO Zaino Indexer started successfully.
\`\`\`

The gRPC bind silently fails inside the spawned task. No error log, no exit, no status change. Wallets connect → \`ConnectionRefused\`. Operator reads logs: \"Started successfully.\" Files a bug against the wallet rather than against zainod.

This is also the root cause of intermittent \`ConnectionRefused\` flakes in zaino's own integration tests under nextest parallelism — when sibling test subprocesses race for kernel-ephemeral ports, the loser's zainod silently runs without serving, and the test client gets refused.

## Suggested fix

\`TonicServer::spawn\` should bind synchronously before returning, then hand the bound listener to tonic via \`Server::serve_with_incoming_shutdown\`:

\`\`\`rust
pub async fn spawn<...>(...) -> Result<Self, ServerError> {
    // Bind synchronously — propagates AddrInUse / EACCES / etc. to the caller.
    let listener = tokio::net::TcpListener::bind(server_config.listen_address)
        .await
        .map_err(|e| ServerError::ServerConfigError(format!(\"gRPC bind failed: {e}\")))?;

    let incoming = tokio_stream::wrappers::TcpListenerStream::new(listener);

    let server_handle = tokio::task::spawn(async move {
        task_status.store(StatusType::Ready);
        server_builder
            .add_service(svc)
            .serve_with_incoming_shutdown(incoming, shutdown_signal)
            .await?;
        task_status.store(StatusType::Offline);
        Ok(())
    });

    Ok(TonicServer { status, server_handle: Some(server_handle) })
}
\`\`\`

After this, \`Ok(TonicServer { ... })\` is a true \"bind succeeded, server is reachable\" signal. A bind failure surfaces as \`Err(ServerError::...)\` (or a new \`ServerError::Bind\` variant) on the line \`Indexer::launch_inner\` calls \`TonicServer::spawn\`, propagating naturally up through \`start_indexer\` → \`zainodlib::run\` → the caller.

## Smaller follow-on

Even with the bind fix, the spawned serve task can still die *later* (panics, tonic-internal errors, etc.). \`Indexer::launch_inner\`'s serve loop should also poll the spawned task — \`server_handle.is_finished()\` or a similar mechanism — and treat a finished gRPC task as \`StatusType::Offline\` so \`check_for_critical_errors\` triggers. ~2-line change next to the existing status check.

## Repro

zaino's own integration tests under nextest parallelism reproduce the bug intermittently. Failure shape is \`ConnectionRefused\` from \`zaino-testutils/src/lib.rs:540\` (\`build_client(...).await.unwrap()\`) on a port that the test launched zainod on. With portpicker as the port allocator the rate is ~10%; even with kernel-ephemeral allocation it stays at a few percent because of the bind/drop/spawn/bind window.

zancas:
Pre-binds the gRPC port and calls TonicServer::spawn against the
  occupied address, asserting the call returns Err. Today the call
  returns Ok and the bind error is swallowed inside the spawned
  serve task — this assertion fails, demonstrating the bug. After
  the issue's suggested fix (synchronous bind), the test passes.

  The StubSubscriber covers ZcashIndexer + LightWalletIndexer with
  unimplemented!() bodies; none are reached because the bind fails
  before any RPC dispatches.

Fixes:  #1081

### DIFF
--- a/packages/zaino-serve/src/server/grpc/tests/spawn.rs
+++ b/packages/zaino-serve/src/server/grpc/tests/spawn.rs
@@ -1,286 +1,25 @@
 //! Regression test for zingolabs/zaino#1081.
 //!
-//! `TonicServer::spawn` (in the parent module) currently returns
-//! `Ok(TonicServer { ... })` before its inner `server_future.await` does
-//! the actual TCP bind. If that bind fails (e.g. `EADDRINUSE`), the
-//! error is swallowed inside the spawned serve task, and the parent
-//! serve loop in `zainod::indexer::Indexer::launch_inner` keeps running
-//! indefinitely with the status board claiming everything is fine.
-//!
-//! The fix per the issue: bind synchronously before constructing the
-//! spawn so a bind failure surfaces as `Err` to the caller.
+//! `TonicServer::spawn` must perform its TCP bind synchronously so a
+//! bind failure (e.g. `EADDRINUSE`) surfaces as `Err` to the caller
+//! instead of being swallowed inside the spawned serve task — without
+//! the synchronous bind the parent serve loop in
+//! `zainod::indexer::Indexer::launch_inner` keeps running indefinitely
+//! with the status board claiming everything is fine.
 //!
 //! This test pre-binds the gRPC port, then calls `TonicServer::spawn`
-//! against the occupied address and asserts that the call returns `Err`.
-//! Today the call returns `Ok` and this test fails — that's the bug.
-//! After the fix lands, the synchronous bind surfaces `AddrInUse` as
-//! `Err`, and the test passes.
-//!
-//! `StubSubscriber` exists only to satisfy the
-//! `IndexerSubscriber<S: ZcashIndexer + LightWalletIndexer>` type
-//! parameter on `TonicServer::spawn`. Its method bodies are never
-//! invoked: the bind fails before any RPC dispatches, both in the
-//! current buggy implementation (the spawned serve task dies before
-//! accepting a connection) and after the fix (the synchronous bind
-//! returns `Err` before the spawn).
+//! against the occupied address with an empty `Routes` (no real RPC
+//! dispatcher — the bind fails before any request would be routed) and
+//! asserts that the call returns `Err`. Before the fix the call returned
+//! `Ok` and the assertion failed; after the synchronous bind landed,
+//! `AddrInUse` is propagated and the assertion passes.
 
 use std::net::{Ipv4Addr, SocketAddr};
 
 use tokio::net::TcpListener;
-use tonic::async_trait;
-use zaino_fetch::jsonrpsee::response::{
-    address_deltas::{GetAddressDeltasParams, GetAddressDeltasResponse},
-    block_deltas::BlockDeltas,
-    block_header::GetBlockHeader,
-    block_subsidy::GetBlockSubsidy,
-    chain_tips::GetChainTipsResponse,
-    mining_info::GetMiningInfoWire,
-    peer_info::GetPeerInfo,
-    z_validate_address::ZValidateAddressResponse,
-    GetMempoolInfoResponse, GetNetworkSolPsResponse, GetSpentInfoRequest, GetSpentInfoResponse,
-    GetTxOutResponse,
-};
-use zaino_proto::proto::{
-    compact_formats::CompactBlock,
-    service::{
-        AddressList, Balance, BlockId, BlockRange, Duration, GetAddressUtxosArg,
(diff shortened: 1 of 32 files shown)


packages/zaino-serve/src/server/grpc/tests/spawn.rs
```diff
@@ -0,0 +1,304 @@
...
+    () => {
+        unimplemented!(
+            "StubSubscriber methods must not be invoked: TonicServer::spawn \
+             returns before any RPC is dispatched"
+        )
+    };
+}
+
+#[async_trait]
+impl ZcashIndexer for StubSubscriber {
```
nachog00: The 200-line `StubSubscriber` in the test is a bit of a code smell worth flagging: `TonicServer::spawn` requires `S: ZcashIndexer + LightWalletIndexer` (~35 methods), but its body is pure transport — bind, TLS, shutdown signal, task spawn. It never calls a single RPC method; it just passes the subscriber through to `CompactTxStreamerServer::new(GrpcClient { service_subscriber })` and hands the result to tonic's `add_service`.

From what I can see, there's no circular dependency between the two concerns. The dispatcher (`CompactTxStreamerServer<GrpcClient<S>>`) doesn't need anything from the transport layer to construct itself — it's just `Arc::new(inner)` plus compression defaults. So in principle, `spawn` could take a pre-built tonic service (`S: Service<Request<Body>> + NamedService + ...`) instead of knowing about the indexer traits, and the caller in `indexer.rs` would assemble the service before passing it in:

```rust
let svc = CompactTxStreamerServer::new(GrpcClient { service_subscriber });
let grpc_server = TonicServer::spawn(svc, grpc_config).await?;
```

I haven't dug into the rest of the codebase enough to be sure there aren't subtler reasons for the current shape, but if this holds up it would make the transport layer independently testable (the bind-race test here would need a trivial no-op service instead of 200 lines of trait stubs) and keep the RPC wiring concern in the caller where it belongs.

nachog00 (dismissed):
The potential decoupling i spotted could be explored in a next PR. This one looks correct as it stands.

zancas:
The last commit merges in the fix in #1107 so that the full test suite runs clean.

packages/zaino-serve/src/server/grpc/tests/spawn.rs
zancas: This has been fixed.

AloeareV (approved):

PR #1084, opened 2026-05-01
Implement gettxoutsetinfo native txoutset backing data

### LINKED ISSUE #210: 1.14: Add Zcash RPC `gettxoutsetinfo`

pacu:
## Summary
- Adds `gettxoutsetinfo` RPC passthrough plumbing through fetch, state, and serve.
- Adds native txoutset storage types and LMDB tables that preserve full transparent `scriptPubKey` bytes for zcashd-compatible behavior.
- Adds a validator-assisted v1.1.0 -> v1.2.0 migration to backfill txoutset data without rebuilding existing Zaino state.
- Adds migration sanity checks for best height/hash, UTXO checksums, txout aggregates, and tx-count consistency.
- Keeps the txoutset index updated during ongoing sync after migration.

## Tests
- `cargo check -p zaino-fetch -p zaino-state -p zaino-serve`
- `cargo test -p zaino-fetch get_tx_out_set_info_tests`
- `cargo check -p zaino-state`
- `cargo check -p zaino-state --features transparent_address_history_experimental`
- `cargo test -p zaino-state txoutset_tests`

## Notes
Integration test compilation for `integration-tests` is currently blocked by the yanked `core2 = "=0.3.3"` dependency.

Closes #210


idky137:
Please do not merge this yet.

I have not finished review but notice a couple issues that we need to fix:
- we already have the data required in the "transparent" / "spent" database tables so do not need the new tables added in this PR. (Although we could add a small "accumulator" table for performance)
- Currently this is not implemented for `DbReader`, and has not been added to the capability traits, meaning any downstream users would need access to the actual database rather than the (safe) read only type, and cannot rely on trait impls.


I will have a look at updating this / recommending changes.

Issue #858, opened 2026-02-23
ZainoDB V2 — small-footprint DB option for Zallet users


idky137:
## Summary / motivation

Zaino’s current `DbV1` is optimised for service efficiency (fast, feature-complete reads) but has a large on-disk footprint and slow initial build (chain sync). Many Zallet users prefer a small, fast-building database with fewer on-disk resources.  

We will add a lightweight `DbV2` backend that stores only the minimal indexes required for basic chain indexing and the additional, Zaino-specific indexes that Zebra does not provide. `DbV2` trades runtime efficiency for smaller size and much faster build times so users can choose the best trade-off for their needs.

## Design principles

- **Capability driven.** DB functionality is exposed through capability traits. A DB that doesn’t implement a capability returns a canonical `FeatureUnavailable` error.  
- **Router + abstraction.** The router layer already lets multiple DB implementations be used behind the same API; callers remain unaware of which major is active.  
- **Major co-existence & switching.** Majors live in separate directories. We support building/promoting a new major (shadow/promote) and optionally keeping the old major so users can switch back cheaply. Deleting old majors is configurable.  
- **Migration planning.** Migrations move from a strict linear ladder to a version-graph planner so branching (e.g., v1 → v2, v1 → v1.x) is supported without fragile `match` logic.  
- **Transparent passthrough.** When a capability is unavailable in `DbV2`, the system should fetch the requested data from the backing validator so callers remain functional.

## Behavioral overview

- `DbV2` implements a reduced capability set. When finalised-state returns `FeatureUnavailable`, finalised-state will delegate the request to the backing validator and return the validator response to callers (configurable timeout/enable flag).  
- Adding new migrations requires registering version edges in the migration registry; a planner computes a path from current → target. Existing migration bodies remain usable with minimal changes.  
- Operators may choose to keep old major directories after promotion (fast switch back) or remove them to conserve disk.

## Acceptance criteria

- `DbV2` implemented and selectable by config; builds much faster and has a smaller disk footprint than `DbV1`.  
- Migration system uses a version-graph planner; adding new migrations does not require editing a `match` dispatch.  
- Finalised-state passthrough delegates `FeatureUnavailable` requests to the backing validator transparently, with a configurable timeout and metric for passthroughs.  (**Possible Solution**)
- Startup honours configured major selection even when multiple majors exist on disk.  
[...]

PR #1094, opened 2026-05-07
Add getblockhashes JSON-RPC support

### LINKED ISSUE #202: 1.6: Add Zcash RPC `getblockhashes`
Spec: https://zcash.github.io/rpc/getblockhashes.html

LOW PRIORITY. Potential deprecation.

pacu:
## Summary
- add typed getblockhashes options and response handling
- expose getblockhashes through Zaino's JSON-RPC service and indexer trait
- serve results from Zaino's finalized headers and non-finalized snapshot data
- support noOrphans and logicalTimes options

## zcashd compatibility notes
- zcashd 6.12.2 filters on its monotonic timestamp index, so Zaino filters on logical timestamps rather than raw block time
- timestamp range is low inclusive and high exclusive
- logicalTimes=true returns objects with blockhash and logicalts
- noOrphans=true limits results to the main chain; noOrphans=false also includes known non-finalized non-best blocks

## Testing
- cargo test -p zaino-fetch -p zaino-state -p zaino-serve --no-run

Closes https://github.com/zingolabs/zaino/issues/202


zancas:
Follow-up cleanup, perf, and correctness items have been split out into #1105 so this PR can stay focused on landing the `getblockhashes` index + migration + consumer wiring.

Highlights tracked there:
- **A1**: orphan iteration in `get_block_hashes` is HashMap-nondeterministic (open from the original review; still present after the rewrite).
- **A2**: JSON-RPC error mapping should be `InternalError`, not `InvalidParams`.
- **B1**: `hashes_by_logical_ts_range` falls back to a full table scan because `lmdb-0.8.0::iter_from` panics on empty; needs the `MDB_SET_RANGE` rewrite to recover `O(matches)`.
- **B2**: migration runs in a single rw transaction; should chunk + checkpoint via `DbMetadata.migration_status` for mainnet scale.
- **C1**: `z_get_treestate` is duplicated between `state.rs` and `fetch.rs`; extract the shared response-assembly half.
- Plus visibility tightenings, doc/README updates for `makers parity-test`, and the lingering `#1102` / `#1103` cleanups.

See #1105 for the full list with file/line pointers and severity ordering.

packages/zaino-state/src/chain_index/finalised_state/capability.rs
```diff
@@ -787,6 +788,37 @@ pub trait BlockCoreExt: Send + Sync {
...
+    ///
+    /// Matches the `[low, high)` semantics of zcashd's `getblockhashes` RPC.
+    /// Results are returned in ascending `logical_ts` order. If `low >= high`
+    /// the result is empty without touching the database.
+    ///
+    /// The index is currently a structural slot — no writes yet (see
+    /// zingolabs/zaino#1101) — so on a freshly-built v1 DB this method
+    /// returns `Ok(vec![])` for any input. Once the write-path hook lands
+    /// the same method body serves real results without API changes.
+    async fn hashes_by_logical_ts_range(
```
idky137, line 803: I dont think we want this in "BlockCoreExt"?

This will  lock the ChainIndex into always having to index this extra data in all deployment situations as this trait is one of the non-optional capability traits.

Please see https://github.com/zingolabs/zaino/issues/858 as this implementation is in direct conflict with that planned work.

It should be easy to move this to an optional capability trait, or, add a new "ChainMetaExt" capability trait to hold this new functionality

packages/zaino-state/src/chain_index/finalised_state/capability.rs
```diff
@@ -787,6 +788,37 @@ pub trait BlockCoreExt: Send + Sync {
...
+    /// Returns the logical timestamp stored for `hash` in the
+    /// `logical_ts_by_hash` reverse index, or `None` if the hash has no
+    /// entry (block was written before the index hook landed, or simply
+    /// does not exist in the finalised state).
+    ///
+    /// Used to seed `LogicalTimestamp::next` at the finalized →
+    /// non-finalized boundary: the first non-finalized block's parent is
+    /// the finalized tip, and the recurrence needs that parent's
+    /// logical_ts to produce a correct result.
+    async fn logical_ts_for_hash(
```
idky137, line 818: Similar to above, this is in direct conflict with https://github.com/zingolabs/zaino/issues/858.

packages/zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -220,6 +220,26 @@ pub(crate) struct DbV1 {
...
+    hash_by_logical_ts: Database,
+
+    /// Logical timestamp by hash: `BlockHash` -> `StoredEntryFixed<LogicalTimestamp>`
+    ///
+    /// Reverse index of [`hash_by_logical_ts`]. Needed so `delete_block`
+    /// can find which forward-index entry to remove without recomputing
+    /// `logical_ts` from genesis.
+    ///
+    /// **Currently a structural slot** — see zingolabs/zaino#1101.
+    logical_ts_by_hash: Database,
```
idky137, line 241: This database table does not make sense.

We already have this data in the table added above. It is easily possible to use either range or cursor search and return both the key and value data from the hash_by_logical_ts table. This is adding database complexity for no reason.

packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
```diff
@@ -1008,6 +1095,102 @@ impl DbV1 {
...
+            Some(h) => h,
+            None => return Ok(()), // empty DB — nothing to backfill
+        };
+
+        // Read all headers in height order in one pass before opening
+        // the rw_txn. Keeping the read out of the rw_txn lets the
+        // checksum-verifying decode happen without LMDB ro+rw borrow
+        // entanglement. We go through the `BlockCoreExt` trait method
+        // because the inherent equivalent is private to `block_core`.
+        let headers = BlockCoreExt::get_block_range_headers(self, GENESIS_HEIGHT, tip).await?;
```
idky137, line 1124: This currently fetches block headers for the complete chain.

For Testnet this is now over 4M blocks and would require a massive amount of memory.

This should be done either block by block, or, in small batches, to minimise the memory requirement to build this table.

Alternatively, cursor search could be used for this rather than range search, this is possible as block headers are stored in block height order, so a cursor can be opened at the required height and iterated until complete. this is very efficient and has very low memory requirements.

packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
```diff
@@ -1008,6 +1095,102 @@ impl DbV1 {
...
+    /// write-path hook uses for new blocks — walked here in height order
+    /// over already-stored headers.
+    ///
+    /// The whole pass runs in a single LMDB rw transaction so the index
+    /// is either fully populated or untouched after a crash. With ~3M
+    /// blocks each contributing ~80 bytes (two entries * (5-byte key +
+    /// 32-byte hash + checksum overhead)) the working-set cost is
+    /// proportional to the chain height; well within LMDB's working
+    /// envelope for the configured `map_size`.
+    pub(crate) async fn backfill_logical_ts_index(&self) -> Result<(), FinalisedStateError> {
```
idky137, line 1113: This method does not seam to be crash safe or resumable, what is the intended behaviour if zaino crashes or is closed mid migration?

packages/zaino-state/src/chain_index/types/primitives.rs
```diff
@@ -21,3 +21,148 @@ pub struct BlockIndex {
...
+/// boundary in `helpers::extract_block_data` (where `chrono::DateTime::timestamp()`
+/// returns `i64`) and at the on-disk serde layer (encoded as 8 bytes LE
+/// for backward compatibility with the v1 schema — see
+/// zingolabs/zaino#1102 for the plan to narrow that too).
+///
+/// Distinct from [`Height`] and [`LogicalTimestamp`] at the type level so
+/// callers cannot silently confuse the three despite all three wrapping
+/// `u32`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub(crate) struct MinerTime(u32);
```
idky137, line 54: This must implement ZainoVersionedSerde as it has been used as a database type.

We must ensure we are not using bespoke serde that could easily fail if changed.

packages/zaino-state/src/chain_index/types/db/legacy.rs
```diff
@@ -958,7 +961,10 @@ impl ZainoVersionedSerde for BlockData {
         let mut w = w; // re-borrow
 
         write_u32_le(&mut w, self.version)?;
-        write_i64_le(&mut w, self.time)?;
+        // Disk format stays 8-byte LE i64 for backward compatibility with
+        // existing v1 records; narrowing the on-disk encoding is tracked
+        // in zingolabs/zaino#1102.
+        write_i64_le(&mut w, i64::from(self.time.as_u32()))?;
```
idky137, line 967: time is no longer a bare i64 and should not use bespoke serde.

packages/zaino-state/src/chain_index.rs
```diff
@@ -1166,6 +1178,137 @@ impl<Source: BlockchainSource> ChainIndex for NodeBackedChainIndexSubscriber<Sou
...
+            entries.push((hash.to_rpc_hex(), logical_ts.as_u32()));
+            seen_hashes.insert(hash);
+        }
+
+        // *** Boundary seed ***
+        // The first non-finalized block's parent is the finalized tip,
+        // and the recurrence needs that parent's logical_ts to produce
+        // a correct result. We no longer walk finalized blocks, so look
+        // up the tip's logical_ts directly.
+        let mut previous_logical_ts: Option<LogicalTimestamp> =
```
idky137, line 1225: I dont think we need this extra fetch. we should be able to look at the finalised entries and use the hash associated with the greatest logical timestamp. 

(and if empty only fetch the latest logical timespamp - hash pair from the finalised state to build from).

packages/zaino-state/src/chain_index/finalised_state/migrations.rs
```diff
@@ -592,3 +598,74 @@ impl<T: BlockchainSource> Migration<T> for Migration1_0_0To1_1_0 {
...
+///   text in `db_schema_v1.txt`.
+///
+/// Safety and resumability:
+/// - The backfill runs in a single LMDB rw transaction so a crash
+///   leaves the index tables empty and the next launch retries.
+/// - The `txn.put` calls use `WriteFlags::empty` (idempotent
+///   overwrite), so a retry that finds partial-but-committed entries
+///   from a prior aborted attempt re-writes them with the same
+///   deterministic value.
+struct Migration1_1_0To1_2_0;
```
idky137, line 626: I may be wrong but as far as I can see this migration is not crash resistant or resumable.

All migrations must abide by these constraints.

idky137 (changes requested):
There are currently several efficiency and safety critical bugs that must be addressed before this can land.

packages/zaino-state/src/chain_index/finalised_state/capability.rs
zancas: Logic is moving to a new ChainMetadataExt trait per the suggestion..  I believe that this addresses the #858 concern.

packages/zaino-state/src/chain_index/finalised_state/capability.rs
zancas: Also to be defined in non-core Ext.

zancas:
This RPC requires either the addition of the logical_ts abstraction to the zainoDB, or a deviation from the RPCs specification.   Since the RPC's only known requester is a possibly deprecated block-explorer we will close this as not-supported.

We might consider adding support in the future if there's sufficient demand.

Issue #1098, opened 2026-05-09
chain_index: sync loop status writes overwrite shutdown's Closing flag, breaking cooperative exit


zancas:
## Summary

`NodeBackedChainIndex::shutdown()` sets the sync loop's status to
`Closing` (`packages/zaino-state/src/chain_index.rs:744`), expecting
the loop's top-of-iteration check (`chain_index.rs:782`) to observe
`Closing` and return `Ok(())`. In practice, the loop's in-iteration
`status.store(Syncing | Ready | RecoverableError)` calls overwrite the
`Closing` flag, and the cooperative exit path only fires when
`shutdown()` lands while the loop is in its post-success `interval`
sleep.

## Reproduction

The race window:

1. Loop iter starts: `if status.load() == StatusType::Closing { return Ok(()); }` — sees `Ready`, proceeds.
2. Loop body: `status.store(StatusType::Syncing)` (`chain_index.rs:786`).
3. `shutdown()` runs: `status.store(StatusType::Closing)` (`chain_index.rs:744`).
4. Iter completes: `status.store(StatusType::Ready)` (or `RecoverableError`) — **overwrites `Closing`**.
5. Sleep `interval`.
6. Loop top: `Closing` flag is gone. Iteration continues.

After `shutdown()`, `finalized_db.shutdown().await` makes every
subsequent `fs.*` call fail, so the iteration in step 4 produces an
`Err` on the next pass. The loop falls into the `RecoverableError`
branch with exponential backoff and only terminates once
`max_consecutive_failures` (10) is reached and the loop returns
`Err(SyncError::...)` with `CriticalError`.

With default `SyncTimings` (initial 250 ms backoff, 8 s cap, 10
failures), the cumulative window before exit is ~47.75 s
(`SyncTimings::max_backoff_window()`). With `SyncTimings::fast` it
collapses to ~5.5 s but is still order-of-magnitude longer than the
intended interval-sleep window.

## How it was found

Adding a regression test that calls `indexer.shutdown().await` and
joins `sync_loop_handle` (`chain_index/tests/non_finalised_state.rs::shutdown_terminates_sync_loop_cleanly`,
introduced alongside #1096) timed out at 5 s with fast timings. The
harness comment at `chain_index/tests.rs:48-56` already alludes to
this race:

> *"most callers ... drop the indexer without calling `shutdown()`,
> relying on the background sync loop being in its post-success
> `interval` sleep at teardown to avoid racing with runtime
[...]

Issue #1096, opened 2026-05-09
collapse lazy NonFinalizedState instantiation


zancas:
## Summary

`NodeBackedChainIndex` stores its non-finalized state as
`Arc<ArcSwapOption<NonFinalizedState<Source>>>`
(`packages/zaino-state/src/chain_index.rs:624`, `:877`). The slot starts
empty and is populated only after the sync loop has driven the
finalized DB up to `validator_best_height − 100`. Until that moment,
`snapshot_nonfinalized_state` returns
`ChainIndexSnapshot::StillSyncingFinalizedState { validator_finalized_height }`
and every read path falls back to validator passthrough.

This issue tracks the complexity that lazy instantiation has imposed.

## Symptoms / costs

1. **Two-level `Option`/`Arc` wrapping.** `Arc<ArcSwapOption<NFS>>` exists
   purely so the slot can flip from `None` → `Some` once. The inner
   `NFS` already holds an `ArcSwap<NonfinalizedBlockCacheSnapshot>` for
   atomic snapshot rotation; the outer `ArcSwapOption` only encodes
   "not built yet."

2. **Special-case branch in the sync loop.**
   `chain_index.rs:809-827` carries a `match *intermediate_nfs_for_scoping`
   to lazily call `NonFinalizedState::initialize` and `nfs.store(...)`.
   The branch contains two `.expect("todo")` calls, in violation of
   `CLAUDE.md` — no `.unwrap()` rule.

3. **Stop-the-world load contortion.** The same block must keep an
   `Arc<NFS>` borrow live across the init branch, hence
   `let intermediate_nfs_for_scoping = nfs.load(); … &nfs.load_full().expect("just set to Some")`
   followed by a literal `std::mem::drop(intermediate_nfs_for_scoping)`
   to release the guard.

4. **Implicit single-writer invariant.** "Load `None` → call
   `initialize` → `store(Some)`" is non-atomic. The code is correct
   only because the sync loop is the sole writer; this constraint is
   nowhere documented or enforced.

5. **12+ consumer match arms.** Every `ChainIndex` query in
   `chain_index.rs` repeats the
   `match snapshot { NonFinalizedStateExists { … } | StillSyncingFinalizedState { … } }`
   shape (sites at lines 1025, 1063, 1103, 1128, 1304, 1608, 1668,
   1763, 1845, 2023, 2133, 2143, 2153). Many of the
   `StillSyncingFinalizedState` arms simply re-implement validator
[...]

PR #1099, opened 2026-05-09
add tests that enforce invariants across the snapshot extended-lifecycle upgrade

zancas:
Pins observable behavior of the NonFinalizedState lifecycle today,
  against the refactor in #1096 (collapse the lazy
  `Arc<ArcSwapOption<NFS>>` slot into an always-present `Arc<NFS>` with
  per-field Provisional/Resolved availability):

  - NFS lowest-height block overlaps the finalized-DB tip at the seam.
  - Blocks are evicted from the NFS once the finalized DB crosses their height (trim invariant).
  - The slot is monotonic post-init: snapshots never observe its absence after the first publish.
  - `shutdown()` terminates the sync loop cleanly. Uses default timings to navigate the race in #1098 (the sync loop's in-iteration status writes overwrite shutdown's `Closing` flag) — papering over, not fixing.

Adds tests that protect against regressions from #1096 .

### DIFF
--- a/packages/zaino-state/src/chain_index/tests/non_finalised_state.rs
+++ b/packages/zaino-state/src/chain_index/tests/non_finalised_state.rs
@@ -1,0 +1,197 @@
+//! Regression tests pinning observable lifecycle behavior of the
+//! [`NonFinalizedState`](crate::NonFinalizedState) as it stands today,
+//! against the refactor tracked in
+//! https://github.com/zingolabs/zaino/issues/1096 (collapse the lazy
+//! `Arc<ArcSwapOption<NFS>>` slot into an always-present `Arc<NFS>`
+//! with per-field Provisional/Resolved availability).
+//!
+//! Pinned invariants — each must survive the refactor under whatever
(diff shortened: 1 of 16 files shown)


zancas:
Apart from imports and lints there are no non-test changes.

packages/zaino-state/src/chain_index/tests/non_finalised_state.rs
```diff
@@ -0,0 +1,197 @@
...
+    let target_hash = *initial_nfs
+        .heights_to_hashes
+        .get(&initial_seam_height)
+        .expect("NFS retains the block at the finalized-DB tip height");
+    assert!(
+        initial_nfs.blocks.contains_key(&target_hash),
+        "precondition: block at seam height is in NFS",
+    );
+
+    mockchain.mine_blocks(20);
```
zancas, line 88: Note..  maybe we should tighten this case to check the that a single block increase is sufficient to remove the (now finalized?) block from the snapshot?

AloeareV (approved):

PR #1095, opened 2026-05-07
moved finalised state spent mapping out of experimental feature

idky137:
- Required for https://github.com/zingolabs/zaino/issues/210
- Built on top of #1111 

Moves the spent mapping functionality in ZainoDB out of the "experimental-features" feature. Add database migration.

PR #1111, opened 2026-05-13
Simplify migration testing and metadata only migration implementation

idky137:
- Adds test only helpers to simplify migration testing.
- Adds default metadata only migration to simplify metadata only migrations.

### DIFF
--- a/packages/zaino-state/src/chain_index/finalised_state.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state.rs
@@ -745,13 +745,226 @@ impl ZainoDB {
     pub(crate) async fn get_metadata(&self) -> Result<DbMetadata, FinalisedStateError> {
         self.db.get_metadata().await
     }
+}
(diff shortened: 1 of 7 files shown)


pacu (approved):
UtACK! 

These are good additions!

It makes me think a bit on whether Zaino is enforcing guardrails to common human-induced persistent state f'ups like:

- Loading a backup from an older version than intended 
- network mismatches 
- unreleased db versions (can this even happen?)

idky137:
This PR should not be merged on its own as squashing will cause many conflicts in https://github.com/zingolabs/zaino/pull/1095  and https://github.com/zingolabs/zaino/pull/1112.

These have been opened as separate PRs to ease review but https://github.com/zingolabs/zaino/pull/1112 should be merged if all are approved to avoid large rebuilds.

Alternatively, this could be merged without squashing to not break the commit history.

PR #1120, opened 2026-05-14
Database naming conventions

zancas:
This proposal is intentionally LMDB centric, if in the future ZainoDB moves to other lower layer DB backing this policy would have to be deprecated.


zancas:
We're thinking of migrating to the zebra_db crate.

PR #1132, opened 2026-05-22
test artifact bumps

zancas:
Causes the release candidate to test against correct binaries.


zancas:
This passes local tests.

zancas:
**makers container-test && makers integration-test**

pass against this commit

AloeareV (approved):

PR #1133, opened 2026-05-22
zaino --> zainod on dockerhub

zancas:


AloeareV (approved):
Looks good! Let's look into getting rid of zaino, but I like publishing to both in favor of dropping the old name with little warning.

idky137 (approved):
LGTM

utACK

Issue #1128, opened 2026-05-20
chain_index: finalized_height(chain_height) conflates chain-derived reorg floor with monotonic finalized-DB tip


zancas:
## Subtlety

`chain_index::finalized_height(chain_height: u32) -> types::Height` computes
`chain_height.saturating_sub(NON_FINALIZED_DEPTH)`. The name suggests
"zaino's authoritative finalized tip," but the value is a **chain-tip-
derived lower bound**, not the actual finalized-DB tip.

The two agree in steady state — each iter advances `fs.db_height` to track
this value. They **diverge after a reorg that shortens the best-known-
chain tip below the prior finalization boundary**: zaino's finalized DB
is monotonic (once a block is added, it stays), so `fs.db_height` can
sit *above* `chain_height - NON_FINALIZED_DEPTH`.

We do assume reorgs cannot pass `NON_FINALIZED_DEPTH` blocks deep
(Zcash protocol bound via coinbase maturity), so every block zaino has
finalized was safe by protocol at the time of finalization. The
divergence is purely about the chain-derived *lower bound* falling
below the (monotonic) actual tip after the chain shortens — not about
"reorg-unsafe finalization."

## Scenario

1. Best-known-chain tip = 200; zaino has finalized to `fs.db_height = 100`.
   `finalized_height(200) = 100`. Function value == fs tip. Consistent.
2. Validator reorgs ~50 blocks (within the protocol bound). New
   best-known-chain tip = 180.
3. `finalized_height(180) = 80`. **But `fs.db_height` is still 100** —
   those finalized blocks at heights 1..=100 weren't touched by the
   reorg (it only reached heights 151..=200).

## Why current callers tolerate it

- `fs.sync_to_height(finalized_height(chain_height), &source)` iterates
  `(db_height + 1)..=target`. If `target < db_height`, the range is empty
  → no-op. Correct outcome by accident; the name suggests the call
  conveys "the actual finalized tip is now this value," but really it's
  "advance the DB toward this value if we haven't already passed it."

- `ChainIndexSnapshot::StillSyncingFinalizedState { validator_finalized_height }`
  is used for read-routing — queries at heights ≤ this value are passed
  through to the finalized DB. Under-reporting (returning 80 when fs is
  at 100) means heights 81..=100 get routed to the non-existent NFS
  instead of the finalized DB. Probably manifests as "no result" rather
  than wrong data, but the snapshot is misleading about what zaino has
  actually finalized.

## Proposed rename

[...]

PR #1136, opened 2026-05-23
Fix race in nfs sync

### LINKED ISSUE #1126: chain_index sync iter: NFS sync's unbounded while loop overshoots iter's chain_height
## Per-iter invariant

Each pass of the chain_index sync worker (one iteration of the `loop { ... }`
inside `start_sync_loop`'s spawned task) is supposed to commit to a single
chain tip and keep all three layers consistent with it:

```
chain_height        = source.get_best_block_height() at iter start
finalized_db.tip()  == chain_height - 100
nfs.lowest_height() == chain_height - 100           (one-block seam overlap)
nfs.best_tip()      == chain_height
```

The 100-block window is the non-finalized zone where reorgs can happen; deeper
blocks are final. NFS and finalized DB overlap at exactly one block — the
seam at `chain_height - 100` — enforced by `remove_finalized_blocks`'s
`retain(|height >= finalized_height|)` (the >= keeps the seam block).

## The bug

The iter currently honors this invariant for `finalized_db`, but **not** for
NFS. The two consumers of `source.get_block` inside the iter have asymmetric
terminators:

**`fs.sync_to_height(finalised_height, ...)`** is bounded by the iter's
snapshotted target (`finalised_state.rs:576`):

```rust
for height_int in (db_height.0)..=height.0 {  // <- iter's snapshotted target
    let block = source.get_block(...).await?;
    self.write_block(chain_block).await?;
}
```

**`non_finalised_state::sync`'s while loop** is unbounded
(`non_finalised_state.rs:382-432`):

```rust
while let Some(block) = self.source.get_block(
    HashOrHeight::Height(zebra_chain::block::Height(
        u32::from(working_snapshot.best_tip.height) + 1,
    ))
).await? {
    ...add_block_new_chaintip(chainblock)...
}
```

The terminator is *the source returned None* — the validator's **current**
view, not the iter's snapshotted view. If the source advances mid-loop, the
loop keeps fetching past the iter's `chain_height`, and the final `update()`
trims with the **stale** `finalized_height` (the iter-start value, from a
finalized DB that was synced to the old chain_height).

## Sequence (iter where the source advances)

1. `chain_height = source.get_best_block_height() = 150`
2. `finalised_height = 50`
3. `fs.sync_to_height(50, ...)` — no-op (DB already at 50).
4. `non_finalised_state::sync(fs)` begins; working snapshot grows 51..150
   inside the while loop.
5. **External event** — validator produces blocks 151..170 (in the failing
   test, `mockchain.mine_blocks(20)` from a sibling task; in production, a
   new block arrives from the network). The source's `active_chain_height`
   advances atomically.
6. Next iteration of the while loop: `get_block(151)` returns the block
   (active is now 170). Loop continues to 170, exits at 171.
7. `update()` reads `finalized_db.db_height() = 50` (still 50;
   `fs.sync_to_height` was called with finalised_height=50 at iter start).
   `remove_finalized_blocks(50)` removes nothing. CAS swap publishes
   `{50..170}` — NFS bottom at 50, NFS top at 170, finalized tip at 50.

The seam-overlap invariant
(`nfs.lowest_height == finalized_db.db_height()`) is violated:
`nfs.lowest_height == 50` but the invariant says it should be 70 once the
chain has moved to height 170.

## Self-heal

On the next iter, `fs.sync_to_height(70)` runs, `update()` trims with
`finalized_height=70`, snapshot becomes `{70..170}` as designed. The
inconsistency persists for one iter (~one sync interval, default 500 ms).

## Production impact

Yes — this can occur in production any time the validator produces a block
during a sync iter. On Zcash mainnet (~75 s block time, ~1–2 s per iter),
the per-iter probability is ~1.3–2.7%. Severity is moderate:

- Block **data** fetched during the window is correct.
- The seam **layout** invariant is briefly violated: callers observe
  `nfs.lowest_height < finalized_db.db_height() + 1` (NFS extends below
  where it should, finalized DB tip falls behind NFS bottom).
- Consumers that assume the seam invariant (e.g., the
  `nfs_lowest_block_matches_finalized_db_tip` regression test, or any
  caller routing reads by depth) can return stale or duplicated data for
  the ~one-iter window.
- Self-heals on the next iter.

## Why the regression test only fails on one branch

- On `dev`, the inter-iter wait is `tokio::time::sleep(timings.interval)`
  (uninterruptible). `mine_blocks` updates `active_chain_height` atomically
  but the worker's sleep continues. When iter N+1 wakes, the test has
  already finished mining; `get_best_block_height` returns 170 at iter
  start and the iter is consistent end-to-end.
- On `accelerate_unit_tests_replace_sleeps`, the harness's post-Ready
  `sleep(sync_timings.interval)` and the worker's
  `wait_or_source_change(interval)` both expire at ~t=500 ms. If iter N+1
  starts by timer expiry **just before** `mine_blocks` fires, iter N+1 sees
  stale `chain_height=150` but its NFS while loop observes the post-mine
  source.

## Reproducer

`packages/zaino-state/src/chain_index/tests/non_finalised_state.rs::block_is_evicted_from_nfs_when_finalized_advances_past_it`
deterministically fails on `accelerate_unit_tests_replace_sleeps` and passes
on `dev`. Tracing in `update()` shows
`finalized_height=50, pre_trim_min=50, pre_trim_max=170` at the failing CAS
swap.

## Fix options

1. **Cap NFS sync's while loop at the iter's `chain_height`** (recommended).
   Pass `chain_height` into `non_finalised_state::sync`; break the while
   loop when `working.best_tip.height + 1 > chain_height`. Mirrors
   `fs.sync_to_height`'s bounded shape. New blocks get picked up on the
   next iter. Minimal change; preserves the per-iter invariant.

2. **Re-sync finalized after NFS sync, before trim.** After the NFS while
   loop ends, recompute `target = working.best_tip.height - 100`, sync
   finalized forward to that target, *then* update/trim. Catches the new
   blocks in the same iter; requires plumbing `source` into `update()`.

3. **Pre-fetch all NFS blocks at iter top.** Fetch
   `finalised_height + 1..=chain_height` up front into a `Vec<Block>`,
   hand the vec to `non_finalised_state::sync`. Removes the runtime
   dependency on `source` from inside NFS sync entirely. Bigger refactor
   (`handle_reorg` also fetches parents on demand) but makes iter atomicity
   structural rather than a discipline the loop has to maintain.

Option 1 looks cleanest.

zancas:
Builds on #1135 review that first.
Fixes: #1126

### DIFF
--- a/packages/zaino-state/src/chain_index.rs
+++ b/packages/zaino-state/src/chain_index.rs
@@ -858,8 +858,13 @@ impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
                         }
                     };
 
-                    // Sync nfs to chain tip, trimming blocks to finalized tip.
-                    non_finalized_state.sync(fs.clone()).await?;
+                    // Sync nfs to the iter-committed `chain_height`, trimming
+                    // blocks to finalized tip. Passing `chain_height` rather
+                    // than letting NFS extend until `get_block` returns None
+                    // bounds the iter against mid-iter source advances (#1126).
+                    non_finalized_state
+                        .sync(fs.clone(), chain_height.into())
+                        .await?;
                     std::mem::drop(intermediate_nfs_for_scoping);
 
                     Ok(())
--- a/packages/zaino-state/src/chain_index/non_finalised_state.rs
+++ b/packages/zaino-state/src/chain_index/non_finalised_state.rs
@@ -349,9 +349,22 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
             .map(Mutex::new)
     }
 
-    /// sync to the top of the chain, trimming to the finalised tip.
+    /// Sync to the iter-committed `chain_height`, trimming to the finalised
+    /// tip.
+    ///
+    /// `chain_height` is the worker's snapshot of the source's best block
+    /// height at the start of this iter (the same value `fs.sync_to_height`
+    /// was called against). NFS extension is bounded by that height, so a
+    /// source advance mid-iter — the validator producing new blocks while
+    /// the worker's NFS-sync loop is still running — is deferred to iter
+    /// N+1, which will read a fresh `chain_height` and trim the published
+    /// snapshot with the correct finalised floor. Closes #1126.
     #[instrument(name = "NonFinalizedState::sync", skip(self, finalized_db))]
-    pub(super) async fn sync(&self, finalized_db: Arc<ZainoDB>) -> Result<(), SyncError> {
+    pub(super) async fn sync(
+        &self,
+        finalized_db: Arc<ZainoDB>,
+        chain_height: Height,
+    ) -> Result<(), SyncError> {
         let mut initial_state = self.get_snapshot();
         let local_finalized_tip = finalized_db.to_reader().db_height().await?;
         if Some(initial_state.best_tip.height) < local_finalized_tip {
@@ -392,6 +405,15 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
                 ))
             })?
         {
+            // Bail before applying any block that lies above the iter's
+            // committed `chain_height`. The speculative `get_block` above
+            // can return a block that wasn't yet on the source when the
+            // worker committed (the mid-iter source-advance race in
+            // #1126); applying it would silently widen this iter's
+            // publish past its iter-start `fs.sync_to_height` floor.
+            if u32::from(working_snapshot.best_tip.height) + 1 > u32::from(chain_height) {
+                break;
+            }
             let parent_hash = BlockHash::from(block.header.previous_block_hash);
             if parent_hash == working_snapshot.best_tip.hash {
                 // Normal chain progression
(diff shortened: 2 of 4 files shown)


zancas:
makers container-test && makers integration-test pass locally.

There were 4 flakes in integration-test, all passed on a second run:

```bash
     Summary [ 369.533s] 169 tests run: 165 passed (45 slow), 4 failed, 21 skipped
        FAIL [  12.584s] (104/169) integration-tests::state_service zebra::get::taddress_transactions_regtest
        FAIL [  69.724s] (142/169) integration-tests::state_service zebra::get::z::get_block_range_default_request_returns_all_pools_regtest
        FAIL [  69.075s] (143/169) integration-tests::state_service zebra::get::z::subtrees_by_index_regtest
        FAIL [  68.906s] (144/169) integration-tests::state_service zebra::get::z::treestate_regtest
```

AloeareV (approved):
New regression tests!

AloeareV (approved):
As I outline in https://github.com/zingolabs/zaino/issues/1126#issuecomment-4536489053, this PR partially-mitigates #1128. I have some questions about the overall approach, but this is an improvement on previous behavior that's much simpler to implement than what I am thinking about.

PR #1137, opened 2026-05-23
Extend use of cancellation token

### LINKED ISSUE #1098: chain_index: sync loop status writes overwrite shutdown's Closing flag, breaking cooperative exit

zancas:
Builds on #1136 review that first!
Fixes:  #1098

### DIFF
--- a/packages/zaino-state/src/chain_index.rs
+++ b/packages/zaino-state/src/chain_index.rs
@@ -29,6 +29,7 @@ use hex::FromHex as _;
 use non_finalised_state::NonfinalizedBlockCacheSnapshot;
 use source::{BlockchainSource, ValidatorConnector};
 use tokio_stream::StreamExt;
+use tokio_util::sync::CancellationToken;
 use tracing::{info, instrument};
 use zaino_fetch::jsonrpsee::response::{
     address_deltas::{GetAddressDeltasParams, GetAddressDeltasResponse},
@@ -660,6 +661,12 @@ pub struct NodeBackedChainIndex<Source: BlockchainSource = ValidatorConnector> {
     network: ZebraNetwork,
     source: Source,
     sync_timings: SyncTimings,
+    /// Signals the sync worker to exit cooperatively. `shutdown()` fires
+    /// `cancel_token.cancel()` *before* tearing down `finalized_db`, so the
+    /// worker wakes from any in-flight `tokio::time::sleep` and returns
+    /// `Ok(())` instead of cycling through the failure-escalation ladder
+    /// once `fs.*` calls start failing. Closes the race tracked in #1098.
+    cancel_token: CancellationToken,
 }
 
 /// Timing parameters for the ChainIndex sync loop.
@@ -750,6 +757,7 @@ impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
             network: config.network.to_zebra_network(),
             source,
             sync_timings,
+            cancel_token: CancellationToken::new(),
         };
         chain_index.sync_loop_handle = Some(chain_index.start_sync_loop());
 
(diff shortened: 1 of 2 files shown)


zancas:
Tests pass again, but again there are a set of flakes:

```bash
     Summary [ 372.528s] 169 tests run: 164 passed (44 slow), 5 failed, 21 skipped
        FAIL [  12.844s] (105/169) integration-tests::state_service zebra::get::taddress_transactions_regtest
        FAIL [  68.512s] (139/169) integration-tests::state_service zebra::get::get_mempool_info
        FAIL [  69.117s] (143/169) integration-tests::state_service zebra::get::z::subtrees_by_index_regtest
        FAIL [  69.056s] (144/169) integration-tests::state_service zebra::get::z::treestate_regtest
        FAIL [  89.927s] (152/169) integration-tests::state_service zebra::get::z::get_block_range_default_request_returns_all_pools_regtest
```

AloeareV (commented):
Changes individual to this PR approved. Blocked on reevaluation of #1136

AloeareV (approved):

PR #1138, opened 2026-05-23
Add wake channel to mock chain source

zancas:
Builds on #1137 review that first.

Note that this PR doesn't close a proper issue but as documented in the commit message it does address several comments originally made against PR #1055.


AloeareV (commented):
Approve of changes individual to this PR. Blocked on discussion about #1136

AloeareV (approved):

PR #1150, opened 2026-05-27
Instantiate nfs at launch

zancas:
Builds on #1152 review that first.
Addresses:  #1096

This work eliminates the class of bugs [surfaced as the database hole exemplar (bullet point 5)](https://github.com/zingolabs/zaino/issues/1066#issue-4337617991).

### DIFF
--- a/packages/zaino-state/src/chain_index.rs
+++ b/packages/zaino-state/src/chain_index.rs
@@ -70,16 +70,28 @@ mod tests;
 /// zingolabs/zaino#1130.
 pub(crate) const NON_FINALIZED_DEPTH: u32 = zebra_state::MAX_BLOCK_REORG_HEIGHT + 1;
 
-/// Lower bound on zaino's finalized-DB tip, derived from the current
-/// best-known chain tip.
+/// The ceiling of the finalized (immutable) chain for a given chain tip:
+/// `NON_FINALIZED_DEPTH` below `best_tip`. A block is finalized — and therefore
+/// reorg-safe — exactly when its height is at or below this value.
 ///
-/// After a chain-shortening reorg this floor can move backwards while
-/// the on-disk `finalized_height` does not — finalized blocks are
-/// never evicted. Callers comparing this floor against
-/// `finalized_height` should account for the asymmetry
-/// (see zingolabs/zaino#1128).
-pub(crate) fn finalized_height_floor(chain_tip: u32) -> crate::Height {
-    crate::Height(chain_tip.saturating_sub(NON_FINALIZED_DEPTH))
+/// Reorg-safety and finality are the same notion: a reorg is bounded to the
+/// non-finalized window and can never evict a finalized block, so the ceiling
+/// monotonically increases and is unaffected by reorgs.
+///
+/// This is the single boundary the passthrough path needs: a height at or below
+/// it is reorg-safe to fetch from the backing validator by height (see
+/// [`is_finalized`]). It also drives the finalized DB's sync target.
+pub(crate) fn finalization_ceiling(best_tip: u32) -> crate::Height {
+    crate::Height(best_tip.saturating_sub(NON_FINALIZED_DEPTH))
+}
+
+/// Whether `height` is finalized given the chain `best_tip` — at or below the
+/// [`finalization_ceiling`], in the immutable range, and therefore reorg-safe
+/// to fetch from the backing validator by height. Heights above it are in the
+/// (reorg-mutable) non-finalized window and must be served only from the NFS's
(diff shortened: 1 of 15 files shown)


packages/zaino-state/src/chain_index/finalised_state.rs
```diff
@@ -509,6 +509,14 @@ impl ZainoDB {
     where
         T: BlockchainSource,
     {
+        // Honour a source-imposed sync cap (a test/backpressure hook): the
+        // finalized DB may be held below the chain's finalization ceiling,
+        // which — since the non-finalized state leads — holds the snapshot
+        // Provisional with a deterministic catch-up gap.
+        let height = match source.finalized_sync_cap() {
```
AloeareV, line 517: I am somewhat concerned about adding a line like this to production code purely for testing purposes. Does this make any sense in a non-test context? If not, can this be feature-gated and/or made `#[cfg(test)]`-only?

packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -42,43 +41,75 @@ pub struct NonFinalizedState<Source: BlockchainSource> {
...
-        /// The height the validater had last finalized as of snapshot creation.
-        validator_finalized_height: Height,
-    },
+/// The non-finalized state always exists — it is built eagerly at chain-index
+/// creation — so a snapshot always carries a [`NonfinalizedBlockCacheSnapshot`].
+/// Whether that window has been validated against the finalized chain yet (the
+/// finalized DB has reached its seam) is its [`SnapshotAvailability`]: while
+/// `Provisional`, reads that need finalized data pass through to the validator.
+pub struct ChainIndexSnapshot {
+    #[allow(private_interfaces)]
```
AloeareV: No longer needed as this field is private.

packages/zaino-state/src/chain_index.rs
```diff
@@ -321,6 +332,12 @@ impl ChainBlock {
 ///
 /// When a call asks for info (e.g. a block), Zaino selects sources in this order:
 #[doc = simple_mermaid::mermaid!("chain_index_passthrough.mmd")]
+// Interim (#1096): this pub trait's surface references in-crate types — the
+// `Snapshot: NonFinalizedSnapshot` bound and the `ChainBlock` returned by
+// `get_indexed_block_by_*` are `pub(crate)`. Keeping those types in-crate is
+// the intended minimum visibility; the lint resolves when the resolution
+// promotion turns `ChainBlock` into `IndexedBlock`.
+#[allow(private_interfaces, private_bounds)]
```
AloeareV: This trait is intended to be publicly implementable. Notably, we want zallet to be able to mock this trait for their own testing. We can't have private interfaces or bounds in this trait. I'm not sure what you mean by the lint resolving when the resolution promotion turns chainblock into indexedblock.

AloeareV: 'Interim' makes me think this attribute is intended to go away at some point, but I don't understand how without making ChainBlock public.

packages/zaino-state/src/chain_index.rs
```diff
@@ -696,7 +708,7 @@ pub trait ChainIndex {
 pub struct NodeBackedChainIndex<Source: BlockchainSource = ValidatorConnector> {
     #[allow(dead_code)]
```
AloeareV: Hang on. This wasn't added by this PR so it's totally not in scope....but why do we have dead code here??

AloeareV: Seems to no longer be dead.

packages/zaino-state/src/chain_index.rs
```diff
@@ -175,30 +175,36 @@ fn branch_len_to_active_chain(
...
 /// return `IndexedBlock` directly — see issue #1096.
-pub(crate) enum ChainBlock {
-    Finalized(IndexedBlock),
-    NonFinalized(ProvisionalBlock),
+pub enum ChainBlock {
+    /// We have indexed all blocks from this block down to the genesis block.
+    /// We know the chain's entire cumulative chainwork as of this block.
+    Reified(IndexedBlock),
+    /// We have not indexed some blocks below this block.
+    /// We know the sum of chainwork from this block down to the finalized tip,
```
zancas, line 183: We will sum finalization_ceiling on demand to calculate provisional chain_work.

AloeareV (approved):
This is ready to go, except that without the reification feature this would be a regression: we'd lose the ability to report chainwork entirely. I am building reification on top of this, it will land as part of that PR.

PR #1151, opened 2026-05-27
Define provisional block type for nfs type safety

zancas:
Next step towards #1096 .

A full migration to instantiate on run NFS snapshots involves extensive changes.  This PR compiles and passes lints and tests including 3 tests specific to invariants known to be necessary before and after this redesign.

### DIFF
--- a/packages/zaino-state/src/chain_index/non_finalised_state.rs
+++ b/packages/zaino-state/src/chain_index/non_finalised_state.rs
@@ -1,7 +1,12 @@
 use super::{finalised_state::ZainoDB, source::BlockchainSource, NON_FINALIZED_DEPTH};
+use crate::chain_index::types::db::{
+    legacy::{BlockData, CompactTxData},
+    CommitmentTreeData,
+};
 use crate::{
     chain_index::types::{
-        self, BlockHash, BlockIndex, BlockMetadata, BlockWithMetadata, Height, TreeRootData,
+        self, BlockContext, BlockHash, BlockIndex, BlockMetadata, BlockWithMetadata, Height,
+        TreeRootData,
     },
     error::FinalisedStateError,
     ChainWork, IndexedBlock,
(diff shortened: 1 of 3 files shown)


packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -94,6 +99,142 @@ pub(crate) struct NonfinalizedBlockCacheSnapshot {
...
+/// misattributed.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub(crate) struct ProvisionalBlock {
+    /// Height + hash of this block.
+    index: BlockIndex,
+    /// Parent block hash as *claimed by this block's header*.
+    ///
+    /// UNTRUSTED while the snapshot is provisional: the linkage it asserts is
+    /// not validated against the finalized chain until the seam connects, so
+    /// it must not be used for trusted chain-walks (e.g. fork-point recursion)
```
zancas, line 167: This might be too strong.

AloeareV: Agreed. The validator is inside of zaino's trust boundary. We should be able to find fork points just as well using provisional blocks as indexed blocks.

packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -94,6 +99,142 @@ pub(crate) struct NonfinalizedBlockCacheSnapshot {
...
+    /// during the provisional stage.
+    parent_hash: BlockHash,
+    /// Cumulative work relative to the seam, header-derived. Never absolute —
+    /// the type enforces that (see [`ProvisionalCumulativeWork`]).
+    provisional_cumulative_work: ProvisionalCumulativeWork,
```
AloeareV, line 172: With the exception of this one field, all the rest of the provisional block is the same as an indexed block. Can we represent this somehow?

AloeareV (changes requested):
This is great, but I think it can be DRYed out further relatively simply.

packages/zaino-state/src/chain_index/non_finalised_state.rs
AloeareV: Looking into this further, the `IndexedBlock`'s `Chainwork` being inside of the `BlockContext` makes this a bit more complex to disentangle. I don't think it's worth blocking on refactoring this at the moment.

AloeareV (approved):
I'm not merging this on its own, as it adds unused code (and therefore would leave dev in violation of our linting rules). This can be considered approved for the purpose of merging the PRs that build atop this, and this code will land as a result of one of those PRs being merged.

PR #1152, opened 2026-05-27
Wire in provisional blocks

zancas:
Addresses #1096

This PR depends on #1151 so it should only be reviewed if that PR has passed review.

It wires the conservative ProvisionalBlocks that are defined in the PR into the NFS snapshot.

This provides type level safety against mistaking accumulated work for provisional work.

### DIFF
--- a/packages/zaino-state/src/chain_index.rs
+++ b/packages/zaino-state/src/chain_index.rs
@@ -175,30 +175,36 @@ fn branch_len_to_active_chain(
 /// Interim shape: once the Availability/Resolved work lands, a resolved NFS
 /// block promotes to `IndexedBlock` (via `into_indexed`), and these methods
 /// return `IndexedBlock` directly — see issue #1096.
-pub(crate) enum ChainBlock {
-    Finalized(IndexedBlock),
-    NonFinalized(ProvisionalBlock),
+pub enum ChainBlock {
+    /// We have indexed all blocks from this block down to the genesis block.
+    /// We know the chain's entire cumulative chainwork as of this block.
+    Reified(IndexedBlock),
+    /// We have not indexed some blocks below this block.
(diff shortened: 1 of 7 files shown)


packages/zaino-state/src/chain_index.rs
```diff
@@ -330,23 +367,28 @@ pub trait ChainIndex {
     ///
     /// **NOTE: This Method is currently not "passthrough aware", cumulative
     /// chain work must be made optional to enable this.**
+    // Interim return type while the NFS holds provisional blocks: a finalized
+    // hit is an `IndexedBlock`, a non-finalized hit a `ProvisionalBlock`. Once
+    // the FS is fully synced (Resolved), these resolve to `IndexedBlock` (#1096).
+    #[allow(private_interfaces)]
```
AloeareV: We can't allow a private interface here, `cargo check` is right to warn about this. This type being non-public makes it unusable by a consumer, which makes this method unusable to a consumer.

ChainBlock needs to become public. Also, this needs to be changelogged.

packages/zaino-state/src/chain_index.rs
```diff
@@ -330,23 +367,28 @@ pub trait ChainIndex {
...
-    ) -> impl std::future::Future<Output = Result<Option<IndexedBlock>, Self::Error>>;
+    ) -> impl std::future::Future<Output = Result<Option<ChainBlock>, Self::Error>>;
 
     /// Returns Some(IndexedBlock) for the given block height.in the best chain.
     ///
     /// Returns None if the specified block height is above the best chain tip.
     ///
     /// **NOTE: This Method is currently not "passthrough aware", cumulative
     /// chain work must be made optional to enable this.**
+    #[allow(private_interfaces)]
```
AloeareV: As above

PR #1157, opened 2026-05-27
Return indexed blocks after fs nfs convergence

### LINKED ISSUE #1096: collapse lazy NonFinalizedState instantiation

zancas:
Builds on: #1150
The New NFS architecture returns IndexedBlocks.
Fixes: #1096

### DIFF
--- a/packages/zaino-state/src/chain_index.rs
+++ b/packages/zaino-state/src/chain_index.rs
@@ -70,16 +70,28 @@ mod tests;
 /// zingolabs/zaino#1130.
 pub(crate) const NON_FINALIZED_DEPTH: u32 = zebra_state::MAX_BLOCK_REORG_HEIGHT + 1;
 
-/// Lower bound on zaino's finalized-DB tip, derived from the current
-/// best-known chain tip.
+/// The ceiling of the finalized (immutable) chain for a given chain tip:
+/// `NON_FINALIZED_DEPTH` below `best_tip`. A block is finalized — and therefore
+/// reorg-safe — exactly when its height is at or below this value.
 ///
-/// After a chain-shortening reorg this floor can move backwards while
-/// the on-disk `finalized_height` does not — finalized blocks are
-/// never evicted. Callers comparing this floor against
-/// `finalized_height` should account for the asymmetry
-/// (see zingolabs/zaino#1128).
-pub(crate) fn finalized_height_floor(chain_tip: u32) -> crate::Height {
-    crate::Height(chain_tip.saturating_sub(NON_FINALIZED_DEPTH))
+/// Reorg-safety and finality are the same notion: a reorg is bounded to the
+/// non-finalized window and can never evict a finalized block, so the ceiling
+/// monotonically increases and is unaffected by reorgs.
+///
+/// This is the single boundary the passthrough path needs: a height at or below
+/// it is reorg-safe to fetch from the backing validator by height (see
+/// [`is_finalized`]). It also drives the finalized DB's sync target.
+pub(crate) fn finalization_ceiling(best_tip: u32) -> crate::Height {
+    crate::Height(best_tip.saturating_sub(NON_FINALIZED_DEPTH))
+}
+
+/// Whether `height` is finalized given the chain `best_tip` — at or below the
+/// [`finalization_ceiling`], in the immutable range, and therefore reorg-safe
+/// to fetch from the backing validator by height. Heights above it are in the
+/// (reorg-mutable) non-finalized window and must be served only from the NFS's
+/// own validated view.
+pub(crate) fn is_finalized(best_tip: types::Height, height: types::Height) -> bool {
+    height <= finalization_ceiling(best_tip.0)
 }
 
 /// Builds a zcashd-compatible `getchaintips` response from the local non-finalized snapshot.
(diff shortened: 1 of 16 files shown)


packages/zaino-state/src/chain_index.rs
```diff
@@ -71,16 +70,28 @@ mod tests;
...
-pub(crate) fn finalized_height_floor(chain_tip: u32) -> crate::Height {
-    crate::Height(chain_tip.saturating_sub(NON_FINALIZED_DEPTH))
+/// Reorg-safety and finality are the same notion: a reorg is bounded to the
+/// non-finalized window and can never evict a finalized block, so the ceiling
+/// monotonically increases and is unaffected by reorgs.
+///
+/// This is the single boundary the passthrough path needs: a height at or below
+/// it is reorg-safe to fetch from the backing validator by height (see
+/// [`is_finalized`]). It also drives the finalized DB's sync target.
+pub(crate) fn finalization_ceiling(best_tip: u32) -> crate::Height {
```
AloeareV, line 84: This is not actually a ceiling. This is a lower bound for the ceiling. A ceiling-floor, if you will. 

In the vast majority of cases, we should be asking the nfs what it thinks the ceiling is, based on the highest block it is aware of (which may be higher than the best tip.) This sort of wooly calculation I think is only useful in one specific case: determining what height to seed initialization of the provisional nfs at. In all other cases, we should be asking the nfs. This function as written probably shouldn't exist at all.

PR #1160, opened 2026-05-28
stop adding nonbest blocks to heights_to_hashes map of best blocks

AloeareV:
note: commit message erroneously says `hash_or_height` map instead of `heights_to_hashes` map


idky137 (approved):
utACK

Issue #801, opened 2026-01-19
Add "transparent outpoint to spending tx" index and gRPC service


idky137:
### Task description

This endpoint has been requested, specifically "Given a provided outpoint (or set of outpoints), which, if any, txids have spent this / these outpoint(s)".

This can be divided into several sub-tasks:
 - [x] 1: Add "Outpoint to spending txid" index to finalised state (Complete). https://github.com/zingolabs/zaino/blob/c16bebca6a8f45750417ccbac12bb1d9bcca6330/zaino-state/src/chain_index/finalised_state/reader.rs#L359 
 - [ ] 2: Add "Outpoint to spending txid" index to non-finalised state.
 - [ ] 3: Expose new index in ChainIndex API.
 - [ ] 4: Update proto files, adding new compact formats required and new gRPC service definition.
 - [ ] 5: Write ZIP draft.
 - [ ] 6: Implement new gRPC endpoint.

NOTES: Where possible, the transparent address indexing implementation in the non-finalised state should follow that in the finalised state. Although not explicitly part of this task, the decision was made to move all transparent address indexing from zebra to zaino, this has now been implemented in full in the finalised state so the non-finalised implementation should follow this to enable serving the full set of transparent address indexes required.

### Context

- This is the second task in #799

### Acceptance criteria

Implement new RPC endpoint and show correct functionality in tests.

PR #1167, opened 2026-06-01
Add ChainIndex::get_outpoint_spenders

idky137:
Adds a method to the chain index that, given a batch of transparent outpoints, returns whether (and where) each was spent on the best chain. Lets clients check spend status of their own outpoints.

- Completes task 2 of https://github.com/zingolabs/zaino/issues/801

---
- Shape
```
  fn get_outpoint_spenders(
      &self,
      snapshot: &Self::Snapshot,
      outpoints: Vec<types::Outpoint>,
      scope: types::ChainScope,
  ) -> impl Future<Output = Result<Vec<Option<types::TransactionHash>>, Self::Error>>;
```
  - Output is index-aligned with the input: result[i] is the spender of outpoints[i].
  - Some(txid) = the tx that spent that outpoint; None = unspent or unknown.

  - ChainScope: finalised vs. full chain

  New enum gates how far the search reaches:

  - Finalised — finalised DB only. Reorg-stable: never reports a spend that lives solely in a non-finalised block, at the cost of not seeing recent spends.
  - FullChain — scans the non-finalised best chain first, then falls back to the finalised DB for anything still unresolved. Reports the latest known spend but may include spends a reorg could roll back.

  The non-finalised pass deliberately walks only blocks reachable via heights_to_hashes (the blocks map also holds reorged-away blocks, which must not count). The finalised lookup is batched into a single DB call, and
  identical TxLocations are deduped so a block spending several queried outpoints is only resolved to a txid once.

  Mempool is currently excluded

  Neither scope consults the mempool, only mined blocks (non-finalised best chain + finalised DB) are searched. So a spend that exists only as an unmined mempool tx will report None. This keeps results tied to actual
  chain state; mempool-aware lookups can be layered on later if needed but wallets are expected to track the mempool state separately.

  Tests

  - mockchain (get_outpoint_spenders, get_outpoint_spenders_empty_and_single): derives ground truth from the indexed test-vector chain — every transparent spend resolves to its spending txid, created-but-unspent
  outpoints return None, plus empty/single-element edge cases.
  - integration (get_outpoint_spenders_{zebrad,zcashd}, #[ignore]d as slow): end-to-end against a real regtest chain, building a spent+finalised, a spent+non-finalised, and an unspent outpoint, then asserting FullChain
  sees both spends while Finalised sees only the buried one. This is the only place the finalised TxLocation → txid resolution runs end-to-end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,9 @@ and this library adheres to Rust's notion of
   `ephemeral_finalised_state` config option (default `false`) that runs Zaino
   without a persistent finalised-state database, serving finalised reads from
   the backing validator via an ephemeral passthrough.
+- `ChainIndex::get_outpoint_spenders` — resolves, for each transparent
+  outpoint, the txid that spent it on the best chain (or `None` if unspent),
+  with a `ChainScope` selecting finalised-only or full-chain search.
 ### Changed
 - Finalised-state sync and the v1.1.0 -> v1.2.0 migration are substantially
   faster on large/mainnet caches. The txout-set accumulator is built in bulk at
--- a/live-tests/e2e/tests/devtool.rs
+++ b/live-tests/e2e/tests/devtool.rs
@@ -1970,7 +1970,9 @@ async fn sole_recipient_outpoint(
     use zaino_state::ChainIndex as _;
 
     let utxos = indexer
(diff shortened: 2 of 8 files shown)


idky137:
all tests pass locally

zancas:
Will return to this after release.

zancas (approved):
I have non-blocking followups for test upgrades.

zancas:
Local tests pass.

PR #1168, opened 2026-06-01
Tooling/CI: setup shellcheck and extract bash from Makefile.toml

dannywillems:
It is cleaner and more maintainable to have all the bash code into files. We also can run `shellcheck` independently. In addition to that, developers aren't forced to use `cargo-make` is they don't want to.

Helped by Claude Code.


AloeareV (approved):
Looks good!

PR #1169, opened 2026-06-01
packages: run makers fmt

dannywillems:
Files were not formatted correctly.


fluidvanadium (approved):
This is a better place to run rustfmt than where it was previously checked in CI.

AloeareV (approved):
Thanks! We had rustfmt and clippy in CI in the past, not sure what happened. Adding rustfmt back is a good improvement.

PR #1171, opened 2026-06-01
Setup hadolint

dannywillems:
Fixes:


AloeareV:
@nachog00 could you take a look at this one? I don't work with the dockerfiles often enough to have much in the way of opinions on how they're linted.

nachog00 (approved):

PR #1173, opened 2026-06-03
Bound GetAddressUtxos result materialization

### LINKED ISSUE #974: get_address_utxos materializes full backend result before applying max_entries cap
both backend implementations of `get_address_utxos` and `get_address_utxos_stream` in `zaino-state` resolve the full backend UTXO set before applying the caller-supplied `max_entries` or `start_height` filter.

**fetch backend** (`packages/zaino-state/src/backends/fetch.rs`): `z_get_address_utxos` is invoked with the address list alone. neither `max_entries` nor `start_height` is forwarded to the upstream RPC. the entire reply is decoded into memory before the loop enforces the cap.

**state backend** (`packages/zaino-state/src/backends/state.rs`): `ReadRequest::UtxosByAddresses(address_set)` resolves the full set before `.take(max_entries)` runs. the unary wrapper adds a second in-process accumulation via `try_collect::<Vec<_>>()`.

both paths are reachable through the gRPC methods `GetAddressUtxos` and `GetAddressUtxosStream` in `packages/zaino-serve/src/rpc/grpc/service.rs`. memory footprint grows proportional to the real UTXO set, not to `max_entries`.

reproduced at commit `20cc6d855a97a5e8d12990ed0587363b6a345c94`.

**suggested mitigations (zaino-local, no upstream changes required):**
- cap address list size in the gRPC handler before reaching the backend
- reject requests with unrestricted `max_entries` at the handler level
- remove the unary `try_collect::<Vec<_>>()` layer in the state backend
- document that `max_entries` limits response size, not backend work

the upstream interfaces (`ReadRequest::UtxosByAddresses` in zebra, `getaddressutxos` in zcashd) do not currently accept entry-cap fields, so a complete pushdown fix would require coordinated interface changes.

related public history: PR #98 (UTXO RPC implementation), PR #100 + `ef5ba469` (timeout hardening around `get_address_utxos`), issue #712, issue #715.

Zk-nd3r:
## Summary

Fixes #974 by applying `start_height` and `max_entries` inside the current `chain_index` path before Zaino converts the full result.

This replaces the obsolete `backends` implementation removed by #1361.

## Changes

- Added `AddressUtxosRequest` with addresses, inclusive `start_height`, and optional global `max_entries`.
- Added `ChainIndex::get_address_utxos_bounded(...)` while preserving the existing unbounded method.
- Routed unary and streaming handlers through the same bounded path.
- Applied bounds before `GetAddressUtxos` conversion in State, Fetch, and mockchain sources.
- Preserved the 1,000-address guard before any source call.
- Added tests for inclusive height filtering, global caps, `max_entries = 0`, unary and stream parity, conversion counts, and early address-limit rejection.
- Set the shared temporary test LMDB map to 1 GiB instead of inheriting the 384 GiB production default.

The backing State query and Fetch RPC response are still materialized upstream. The Fetch RPC has no height or limit parameters. This patch bounds Zaino-side conversion and returned materialization.

## Validation

Base: `zingolabs:dev` at `d98984d10ce802bc7cb553dae091f6581272eec0`

- `cargo fmt --all -- --check`
- `git diff HEAD^ HEAD --check`
- `cargo +stable-x86_64-pc-windows-msvc test -p zaino-state --lib --no-default-features --no-run`
- `chain_index::tests::mockchain_tests::get_address_utxos`: passed
- `chain_index::tests::mockchain_tests::address_utxo_handlers_apply_identical_bounds_before_source_conversion`: passed

Current public CI:

- Seven distinct generic checks pass: Hadolint, cargo publish dry-run, Shellcheck, Rustfmt, repo guards, cargo-deny, and Docker-tag computation.
- `Trigger integration tests` fails before product tests because fork PR context has no GitHub App `client-id`.
- `Build test artifacts`, `No-zcashd build`, and `Tests` are skipped behind that dispatch gate.

The patch is ready for maintainer review. Production/canonical jobs still need maintainer-context execution; the trigger failure is not treated as a code-test failure.

### DIFF
--- a/packages/zaino-state/src/chain_index.rs
+++ b/packages/zaino-state/src/chain_index.rs
@@ -33,7 +33,7 @@ use arc_swap::ArcSwapOption;
 use futures::{FutureExt, Stream};
 use hex::FromHex as _;
 use non_finalised_state::NonfinalizedBlockCacheSnapshot;
-use source::{BlockchainSource, ValidatorConnector};
+use source::{AddressUtxosRequest, BlockchainSource, ValidatorConnector};
 use tokio_stream::StreamExt;
 use tokio_util::sync::CancellationToken;
 use tracing::{info, instrument};
@@ -512,6 +512,17 @@ pub trait ChainIndex {
     fn get_address_utxos(
         &self,
         address_strings: GetAddressBalanceRequest,
+    ) -> impl std::future::Future<Output = Result<Vec<GetAddressUtxos>, Self::Error>>;
+
+    /// Returns a bounded subset of unspent transparent outputs for the given addresses.
+    ///
+    /// `start_height` is inclusive. `max_entries = None` leaves the result unrestricted.
+    /// These bounds apply to Zaino's converted result, not the backing validator's query work.
+    fn get_address_utxos_bounded(
+        &self,
+        address_strings: GetAddressBalanceRequest,
+        start_height: u64,
+        max_entries: Option<u32>,
     ) -> impl std::future::Future<Output = Result<Vec<GetAddressUtxos>, Self::Error>>;
 
     /// For each outpoint, returns the txid of the transaction that spent it on the best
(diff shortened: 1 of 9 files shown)


Zk-nd3r:
@idky137 @zancas — this is now ported to the current chain_index path and ready for review. Local build-only validation and the two focused bounds tests pass. The public product jobs did not dispatch because the fork PR integration trigger lacks its GitHub App client-id; the PR body now records that boundary. Could you review it or route it to the current owner?

zancas:
Thank you for this work. We are closing this PR, and the reasons follow.

The PR no longer applies to `dev`. It is one commit on a base from 2026-07-16,
and `dev` has more than 500 commits since then. Commit a4fda8f1d deleted
`chain_index/source/validator_connector.rs`, which holds most of this change.
The `getaddressutxos` path now runs through the `OneShotGetAddressUtxos` port
in `zaino-source`, with one adapter for each transport.

We examined a port of the change to that structure. The bounds save the most
work in the read-state adapter. That adapter copies each row that Zebra returns
and clones each lock script. We plan to remove the read-state transport. See
#1495, which is deferred until the RPC optimizations land in Zebra. A port
would therefore change a public trait for a code path that we intend to delete.

On the RPC path, `zaino-rpc` already limits each validator reply to 32 MiB
while it reads the body. That limit bounds the memory that this PR set out to
protect. One problem remains. A reply above the limit fails the full query,
even when the client asks for few entries. Zebra's `getaddressutxos` accepts no
limit, so the fix belongs in Zebra.

Your analysis identified a real problem. `dev` still applies `start_height` and
`max_entries` only after the full result exists.

PR #1176, opened 2026-06-04
BUG FIX:  get_block_deltas (state path)

idky137:
This PR fixes get_block_deltas on the State backend so transparent spend inputs are included in block deltas instead of being dropped when their value/address is not present on Zebra’s stateless transaction object.

The fix resolves transparent prevouts through ReadStateService, reports missing or out-of-range prevouts as typed errors, keeps coinbase/nonstandard-input handling unchanged, and removes adjacent expect/unwrap panics from the same response path.

### DIFF
--- a/live-tests/e2e/tests/devtool.rs
+++ b/live-tests/e2e/tests/devtool.rs
@@ -1828,7 +1828,7 @@ async fn get_block_deltas_resolves_transparent_spend() {
         Some(zebra_chain::parameters::NetworkKind::Regtest),
     )
     .await;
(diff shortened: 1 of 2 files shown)


idky137:
all tests pass locally

zancas:
Possible in this release.

zancas (approved):
Tests pass, filed non-blocking followups as an issue:

https://github.com/zingolabs/zaino/issues/1349

I didn't attempt to handle reorg'd blocks here, because the correct behavior is not-yet-well-defined.

PR #1177, opened 2026-06-05
bound get_address_utxos address fan-out (#974)

Zk-nd3r:
partial fix for #974.

#974 has two parts: the post-fetch cap (already in dev) and the fact that both backends materialize the full backend utxo set before applying max_entries/start_height. a complete pushdown fix needs upstream interface changes (zebra UtxosByAddresses and zcashd getaddressutxos take no entry cap), so this does the zaino-local part only.

what this does:
- caps the address list size at all four utxo entry points (state + fetch, unary + stream) before the backend call, via a shared validate_utxo_address_count with tests. bounds the one input the service controls locally, the address fan-out, so a caller cannot force an unbounded number of backend address lookups in a single request.
- documents that max_entries bounds response size, not backend work.

UTXO_MAX_ADDRESSES is 1000, set well above realistic wallet usage. left as a const with a TODO; happy to make it config-driven if you prefer.

deliberately not here:
- backend pushdown of max_entries/start_height: needs upstream coordination, tracked by the rest of #974.
- rejecting unrestricted max_entries: that changes the documented 0 = unrestricted contract, so i left it out. tell me if you want it gated in public service mode.
- try_collect removal: already gone from dev.

questions: prefer a config field over the const for the address cap? and do you want unrestricted max_entries rejected in public mode, or is the existing clamp-after-fetch enough?

### DIFF
--- a/packages/zaino-state/src/backends.rs
+++ b/packages/zaino-state/src/backends.rs
@@ -15,6 +15,32 @@ fn latest_network_upgrade(
     })
 }
 
+/// Maximum number of addresses a single `get_address_utxos` / `get_address_utxos_stream`
+/// request may carry.
+///
+/// Both backends resolve the full backend UTXO set before applying `max_entries` /
+/// `start_height` (issue #974). A complete pushdown fix needs upstream interface changes
+/// the caller-supplied entry cap cannot reach today, so until then this bounds the one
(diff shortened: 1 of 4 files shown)


idky137 (approved):
Thanks! Looks good!

PR #1178, opened 2026-06-05
add check for reorgs to blocks that aren't higher than the current best tip

AloeareV:
Note: This is a quick fix, I've added a post-normal-sync check, where we handle reorgs if the chaintip hashes don't match. This couple perhaps fit more elegantly in the sync loop itself.


packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -512,6 +514,38 @@ impl<Source: BlockchainSource> NonFinalizedState<Source> {
         Ok(indexed_block)
     }
 
+    async fn check_for_nonhigher_reorgs(
+        &self,
+        working_snapshot: &mut NonfinalizedBlockCacheSnapshot,
+        // Callers should provide None. Used for self-recursion case only
+        height_to_recurse_to: Option<Height>,
+    ) -> Result<(), SyncError> {
```
idky137, line 554: Should this also have the recursion limit as in #1179 ?

AloeareV: > Should this also have the recursion limit as in #1179 ?

It should! Sorry for the delay in adding it. Will get to this soon!

zancas (approved):

PR #1179, opened 2026-06-05
add bound to reorg loop

AloeareV:
Note: It was suggested that replacing the recursive call with an accumulator would be more efficient, due to needing to recursively box during recursive async functions. 

This is true in the abstract. As this function isn't a hot path, and will very rarely recurse a double-digit number of times, I suspect that this have unnoticeable benefit in terms of how fast zaino actually functions.


idky137 (dismissed):
utACK

idky137 (approved):

PR #1180, opened 2026-06-06
deps: bump librustzcash/zebra to orchard-0.14 compatible line

dianaravenborne:
All pre-0.14 `orchard` releases were yanked from crates.io; zaino transitively pulls them in via `zcash_primitives 0.27` and `zebra-chain 7.0`. Bump the workspace dependency pins forward and pick up a few related upstream churn points:

| crate | from | to |
| --- | --- | --- |
| zcash_address | 0.11 | 0.12 |
| zcash_keys | 0.13 | 0.14 |
| zcash_protocol | 0.8 | 0.9 |
| zcash_primitives | 0.27 | 0.28 |
| zcash_transparent | 0.7 | 0.8 |
| zcash_client_backend | 0.22 | 0.23 |
| zebra-chain | 7.0 | 9.0 |
| zebra-state | 6.0 | 8.0 |
| zebra-rpc | 7.0 | 9.0 |

**Source-side changes:**

- `zebra-chain 9.0` added an `Nu6_2` variant to `NetworkUpgrade` and an `nu6_2` field on `ConfiguredActivationHeights`. Zaino's `ActivationHeights` doesn't model `Nu6_2` yet, so the value is dropped at the boundary with a `TODO` for proper wire-through.
- Two `u32::is_multiple_of` call sites (stable since Rust 1.87) are replaced with plain modulo so the code still compiles on Zallet's MSRV (1.85).
- A deprecated public `.fetcher()` accessor is added on `FetchServiceSubscriber` so downstream consumers (Zallet) can keep using the legacy `JsonRpSeeConnector` surface while migrating off of it.

---
Requested by @nullcopy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1463,9 +1463,9 @@ dependencies = [
 
 [[package]]
 name = "halo2_gadgets"
-version = "0.4.0"
+version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45824ce0dd12e91ec0c68ebae2a7ed8ae19b70946624c849add59f1d1a62a143"
+checksum = "fb2a697cad929f706b7987fe804ad57d43622cd37463ba7e4d662a926fdcfea3"
 dependencies = [
  "arrayvec",
  "bitvec",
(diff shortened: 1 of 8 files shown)


dianaravenborne:
Closing in favor of #1181.

PR #1181, opened 2026-06-06
Add NU6.2 activation height support

czarcas7ic:
Adds NU6.2 support to Zebra RPC parsing and preserves the NU6.2 activation height in zaino-common network config. This lets custom network activation heights flow through into Zaino's Zebra network setup.

Validation: cargo test -p zaino-common

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ license = "Apache-2.0"
 
 # Zingo
 
-zingo_common_components =  "0.3.0"
+zingo_common_components = { git = "https://github.com/valargroup/zingo-common.git", rev = "3d58a4399c00ca10779fcef5ecf8ed46f913cb84" }
 
 # Librustzcash
 incrementalmerkletree = "0.8"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,17 +32,17 @@ zingo_common_components =  "0.3.0"
 
 # Librustzcash
 incrementalmerkletree = "0.8"
-zcash_address = "0.11"
-zcash_keys = "0.13"
-zcash_protocol = "0.8"
-zcash_primitives = "0.27"
-zcash_transparent = "0.7"
-zcash_client_backend = "0.22"
+zcash_address = "0.12"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -28,21 +28,21 @@ license = "Apache-2.0"
 
 # Zingo
 
-zingo_common_components =  "0.3.0"
+zingo_common_components = { git = "https://github.com/valargroup/zingo-common.git", rev = "3d58a4399c00ca10779fcef5ecf8ed46f913cb84" }
```
czarcas7ic: Update with https://github.com/zingolabs/zingo-common/pull/78 once tagged

czarcas7ic:
Do you have any recommendations on testing this on mainnet without having to reindex the entire chain? Previously passing `DatabaseSize::Gb(0)` let us skip this but I think now it is required? Am trying to better understand, but thought leaving a question here would possibly save me some time.

packages/zaino-fetch/src/jsonrpsee/response/address_deltas.rs
```diff
@@ -278,7 +278,7 @@ mod tests {
 
     fn sample_delta_with_block_index(i: u32, bi: Option<u32>) -> AddressDelta {
         AddressDelta {
-            satoshis: if i.is_multiple_of(2) { 1_000 } else { -500 },
```
czarcas7ic, line 281: Changed this to keep rust 1.85 compatible, which is what upstream Zallet is on

Cargo.toml
zancas: We should be able to point at this now:  https://crates.io/crates/zingo_common_components/0.3.1

czarcas7ic: Done here https://github.com/zingolabs/zaino/pull/1181/commits/1e9e339952888ee0222621263c83c6145601fac0

zancas (approved):
Our integration-tests are already broken by the orchard yank, we're getting them back online ASAP independently.   This will be tested (by those tests) post-facto.

PR #1199, opened 2026-06-08
  Remove zingolib from *default* zaino-testutils

zancas:
This PR factors integration tests into wallet and walletless workspaces.

Invariants:

* all tests that run on dev run on this feature
* new tools allow running integration-tests that do NOT compile/link or otherwise depend on zingolib

To see the full integration test suite:

**makers integration-test**

To see all zingolib independent tests:

**makers container-test && makers walletless-integration-test**

Note that ALL changes are in **integration-test** except for modifications to config files, and test running tools (Makefile.toml).


zancas:
There's a build cost since the walletless and wallet trees build completely separately.

nachog00 (approved):
I don't like the idea of expanding the set of workspaces. One ws for "integration testing" is the cleanest approach.
I think this approach tackles the symptom and not the cause of the problem: the testing code depending on zingolib is a deviation from the approach we took for other testing dependencies. the wallet functionality should be provided by local_net and integration testing code shouldn't depend on specific implementation detail libraries.

I'm inclined to approve this pr as is, as it adds valuable refactors, but i see the added workspace as a disadvantage, as it adds a bunch of extra kruft and a mental model which we shouldn't have to begging with, just like we don't have a "validator enabled tests" workspace.

I think work should be scheduled to apply the correct long term fix.

zancas:
I agree with your assessment.   The two-workspace split is a hack to allow a subset of tests to be runnable even if the zingolib dep breaks some of them.

Once zingolib is entirely factored out the workspace will unify again.

So..  yes.  It's a temporary fix.

PR #1200, opened 2026-06-09
Optional finalised state DB, background sync, and background DB migration

### LINKED ISSUE #249: Issues encountered while integrating Zaino into Zallet
This is an ongoing list of issues I have encountered while integrating Zaino into Zallet. I'll update it either as I find more things, or if any of these are moved out to standalone issues.

# Missing functionality

- A full block index.
  - `zaino-state` provides a `CompactBlock` cache, which is insufficient for full node wallet syncing as `CompactBlock` omits large parts of the full block (currently Zallet makes multiple round trips through Zaino to the backing full node to fetch the data it needs).
  - The equivalent type in `zcashd` is `CBlockIndex` (and the main chain logic built around it); there's probably already a similar type in Zebra that Zaino should be able to access and replicate.
- Chain fork point logic (most efficiently implemented with direct access to a block index).
  - I currently implement this in Zallet with individual `JsonRpcConnector::get_block` calls.
- Background cache / index building.
  - AFAICT setting `FetchServiceConfig.no_db` to `false` causes `IndexerService::spawn` to block on completely filling its cache. This makes sense for a Zaino-as-lightwalletd-binary use case (where you can wait until a new node is fully spun up before exposing it to users), but is not practical for bundling Zaino as a library within a wallet app (because it blocks anything that requires chain data such as RPCs, not just the things that require the particular cache or index).
  - `IndexerService::spawn` should have an option for filling caches etc. in the background and/or lazily, by computing and storing missing cache entries on-demand, while optionally filling out yet-unrequested cache entries in the background.
    - Some indexes might not be possible to produce without full data access, and may require callers to block on indexing completion (a la `zcashd -reindex`). 

# API issues

- AFAICT Zaino's block cache does not have an API to get the hash of the chain tip block atomically.
  - Instead you need to call `BlockCacheSubscriber::get_chain_height` followed by `BlockCacheSubscriber::get_compact_block`, and the latter may fail to return a block for the given height during reorgs.
- The `hash_or_height: String` type in several APIs should instead be expressed as an enum, to simplify the public API.
- The Zaino libraries like `zaino-fetch` would be easier to use if they had typed APIs. For example, with `JsonRpcConnector::get_subtrees_by_index`:
  - Use `zcash_protocol::ShieldedProtocol` for the pool argument.
  - Parse the returned hex roots into at least byte arrays.
  - Use `zcash_protocol` types as appropriate.
- `zaino_fetch::jsonrpc::response::GetTreestateResponse` represents `height` as an `i32`; it should use `Option<u32>` instead.
  - Bitcoin Core (and thus `zcashd`) used `int` for height in various places, but only so `-1` could mean "no height". `z_gettreestate` will never return a height of `-1` for blocks currently in the chain, but (at least for `zcashd`) _might_ do so for blocks that have never been connected to any chain (I have not checked this). The `-1` case should be mapped to `None`, and any other negative value is an error.
- `zaino_fetch::jsonrpc::response::GetTreestateResponse` uses `Treestate<String>` for its `sapling` and `orchard` fields, which is confusing. It should use `Treestate<Vec<u8>>` instead.
  - Zebra's `Treestate<Tree: AsRef<[u8]>>` uses a `serde(with)` annotation to parse the hex encoding output of `z_gettreestate` into a raw byte form. By storing the tree state as a `String`, it was no longer clear to me whether or not the hex decoding in Zebra actually occurs. If it did, that would almost certainly violate the requirement that `String` only contains valid UTF-8 bytes. If it didn't, that foists the decoding burden onto the downstream user, and given that the purpose of this API is to be consumed by wallet code rather than users, doing the decoding inside Zaino makes the most sense.
  - For testing purposes, I made a guess that due to internal magic within `serde`, `String` does cause hex decoding to be skipped. This appears to have been the case.
- Because `zaino_proto::proto` and `zcash_client_backend::proto` are different types, we have to inefficiently convert between them. We should figure out a place to de-duplicate them (though this won't be a blocker for Zallet as we will be using full blocks eventually).
- The stream returned by `MempoolSubscriber::get_mempool_stream` ends when the chain tip changes, but there's no indication or guarantee about the state of the chain tip when the stream _starts_. This creates the possibility of a race condition wherein the chain tip changes between a call to `JsonRpcConnector::get_block` (or `BlockCacheSubscriber::get_chain_height`) and the call to `MempoolSubscriber::get_mempool_stream`, causing the caller to miss the chain tip change and be sent a mempool stream that is invalid for their current state.
  - A possible solution would be to allow the caller to specify what their view of the chain tip is, so that `get_mempool_stream` can exit immediately if the mempool stream would be invalid for the caller.
- `MempoolSubscriber::get_mempool_stream` returns the `zebra_rpc::methods::GetRawTransaction` type directly, which forces Zallet to have the entire node-side JSON-RPC implementation as a dependency.
  - This is why I didn't use `zebra_rpc` for Zallet's JSON-RPC server implementation, and instead implemented the server separately. For Zallet's use case, factoring out the server logic may or may not help, but I don't know the full extent of how Zaino is depending on `zebra_rpc`, so IDK if it's actually possible to remove this dependency. Maybe more feature flags in `zebra_rpc` could help?
  - Additionally, AFAICT there's no way for the caller to control whether `get_mempool_stream` returns `GetRawTransaction::Raw` or `GetRawTransaction::Object`, so the only reliably-available data is the raw bytes of the transaction. It feels like Zaino should just return the raw transaction bytes here, or maybe have a generics-based API that lets the caller parse those bytes into their desired `Transaction` type.

# Documentation issues

- Either `JsonRpcConnector::get_treestate` or `zaino_fetch::jsonrpc::response::GetTreestateResponse` should document that the returned treestates are encoded with the v0 frontier encoding (`zcash_primitives::merkle_tree::read_frontier_v0`).
  - This may eventually change if zcashd and/or zebrad adds a flag to allow choosing the returned encoding format, but for now it's v0.
- `MempoolSubscriber::get_mempool_stream` doesn't document how the returned `task` should be managed, or whether it can be ignored.
- `MempoolSubscriber::get_mempool_stream` doesn't document the full stream item type, or what properties the stream has (for example, will txids ever be repeated?).
- `MempoolValue` is documented as being a copy of the txid, but appears to actually store the transaction. Its documentation says this was a possible future change; were the docs not updated in some earlier PR?

### LINKED ISSUE #861: ZainoDB "feature unavailable" error pass-through
## Problem

DbV2 will intentionally implement a reduced capability set and will return FeatureUnavailable for capabilities it does not provide. When that happens callers must still get data from the backing validator so services (Zallet, explorers) continue to work.

## Options

A — NodeBackedChainIndex passthrough: extend the existing NodeBackedChainIndex logic to forward requests to the validator when DB returns FeatureUnavailable.
Pro: minimal changes. Con: responsibility lives outside DB; duplicate logic needed for other callers.

B — ZainoDB passthrough: implement passthrough inside finalised-state using ValidatorConnector, so callers always ask the DB and the DB delegates when a feature is unavailable.
Pro: clean, single API surface; callers don’t need version-awareness. Con: small added complexity in finalised-state.

## Tasks (Option B)
- Wire ValidatorConnector into finalised-state/router.
- Implement a small wrapper that intercepts FeatureUnavailable, calls the validator (with a short configurable timeout), and returns the validator response.
- Ensure passthrough triggers only for FeatureUnavailable (not for other DB errors).
- Add unit/integration tests: success, timeout, validator error, and V1 (no passthrough).
- Add a config knob for timeout and enable/disable passthrough; emit a simple metric for passthrough events.

## Acceptance criteria (Option B)
- Callers receive validator data transparently when DB returns FeatureUnavailable.
- Passthrough is configurable, time-bounded, and only invoked for FeatureUnavailable.
- NodeBackedChainIndex can be simplified to rely on DB passthrough.

### LINKED ISSUE #1196: Unable to disable Zaino DB
### Security check

### Usage mode

Library integration (using zaino-* crates)

### Describe the bug

Previously in Zallet, we were able to disable Zaino's LMDB by setting `zaino_common::DatabaseSize(0)`. This bypass no longer seems to work. Now, Zaino ignores this and still creates the embedded db with a small size, which quickly throws `MDB_MAP_FULL` during scan.

This is currently preventing Zallet from running and consequently causing integration tests to fail.

Is it no longer possible to disable the embedded db?

### Steps to reproduce

Attempt to instantiate Zaino with `zaino_common::DatabaseSize(0)`

### Expected behavior

Should disable embedded db and run without caching

### Environment

Zaino 0a41125a712a8cc7de75e903a827a7ef0f08b4a0

### Configuration

```toml

```

### Logs

```shell

```

### Additional context

_No response_

idky137:
Makes the finalised state (FS) optional and turns FS sync + version migration into background, non-blocking operations so startup and serving are never blocked on persistence.

  - **Ephemeral mode** (`ephemeral_finalised_state` config, default `false`): runs without a persistent FS database, serving finalised reads from the backing validator via a stateless passthrough.
  - **Background sync/migration**: large syncs and migrations run in the background (small syncs run inline) while a stateless passthrough serves reads. Background failures retry and escalate to `CriticalError`.
  - **NFS memory bound** (`MAX_NFS_DEPTH`): caps non-finalised cache retention to a fixed depth below the tip, so it can't grow unbounded when `db_height` lags (background sync) or is pinned at `0` (ephemeral mode).

  ## Renames

  The FS facade and backing were renamed to reflect that the backing is no longer necessarily a database (internal, `pub(crate)`):

  - `ZainoDB` → `FinalisedState`
  - `db` module → `finalised_source`; `DbBackend` → `FinalisedSource` (variant `Stateless` → `Ephemeral`)
  - `StatelessFinalisedState` → `EphemeralFinalisedState`

  ## Other

  - DB version bumped to **v1.2.1** (metadata-only migration; no on-disk layout change).
  - Added `FinalisedState::wait_until_synced` (waits for background work to reach target, vs `wait_until_ready`'s serving-readiness).
  - Docs + changelogs updated.

  ## Notes for reviewers

  - During a large background sync/migration, passthrough-served blocks carry chainwork `0` — consistent for NFS relative fork-choice, but absolute chainwork is offset-low until the persistent DB catches up. (Acceptable
  until the in-flight NFS rework #1157 lands.)
 
---
- Closes #1196
- Closes #861
- Closes #249
- Required for #858


packages/zaino-state/src/chain_index/finalised_state/db/ephemeral.rs
```diff
@@ -0,0 +1,821 @@
+//! StatelessFinalisedState provides access to the finalised portion of the
```
zancas: What does "Stateless" mean in this context?

Does it indicate that the mode of operation does not use a stateful DB?

If yes, isn't that DB an index of chain-state that is useful to lightclient consumers?

idky137: Stateless in this context means it has no persistent / on-disk ZainoDB state; reads are passed through to the backing chain source instead. I agree the name could be clearer - maybe EphemeralFinalisedState?

And yes, the DB is an index of finalised chain data that is useful to lightclient consumers (and non-lightclient consumers). This mode is not saying the DB is useless, only that Zaino can operate (with limited functionality) without maintaining that index. 

There will always be drawbacks with this mode though:
1. Speed of data servicing will be much slower
2. Certain capability - spend detection, possibly taddr indexing if it moves from zebra, some RPCs (get_tx_out_set_info) - is not serviceable in this mode, this is a hard constraint due to the fact we have to build these indexes from genesis and store them somewhere.

Note for drawback 2: This was the reasoning for the "small footprint DB" planned - https://github.com/zingolabs/zaino/issues/858 (issue slightly stale, I will update once the optional DB work lands).

zancas: OK.   I am confused about **ephemeral** in this context.   In some cases (e.g. maybe miners) the intention is to never use the index/zainodb, so in that case the mode of operation is permanent.

docs/example_configs/zainod.toml
```diff
@@ -14,6 +14,11 @@ backend = 'fetch'
 zebra_db_path = '/home/cupress/.cache/zebra'
 network = 'Testnet'
 
+# Run the finalised state in ephemeral/stateless mode. When true, Zaino opens no
```
zancas, line 17: I am confused how a zainod operator should set this if they want instant responsiveness, and eventually full zainodb index functionality.

I that case I want to always have all the data I can get from zainod, as soon as **I**, the consumer, exist.

idky137: With this update zaino (ChainIndex and Zainod) always serves data immediately. This setting is related to whether you want zaino to build the database at all.

Setting "ephemeral" to "true" means zaino will not build the database, and will always serve data using the backing validator.

Leaving "ephemeral" as "false" means zaino will build the database, but now that is a non-blocking background operation, so service is still immediate.

idky137: Short answer, if you leave it as default (false), zaino will build the database but also give immediate responsiveness.

packages/zaino-state/src/chain_index/finalised_state/db/ephemeral.rs
idky137: Ephemeral here means it is not persistent / permanent:
- "ephemeral": no persistent / permanent FinalisedState, - only held in memory.
- non-"ephemeral": data is held permanently / in persistence - in the database.

nullcopy:
FYI I've confirmed this PR closes https://github.com/zingolabs/zaino/pull/1200 as well

zancas:
Superceded by #1250 which merges in the 0.4.0-backport plus dev, and resolves conflicts.

PR #1207, opened 2026-06-11
0 4 0 release bug fixes

idky137:
Fixes slow sync and migration observed in release test flows.

### DIFF
--- a/packages/zaino-state/src/chain_index/finalised_state/db/v1/block_core.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state/db/v1/block_core.rs
@@ -307,71 +307,26 @@ impl DbV1 {
         txid: &TransactionHash,
     ) -> Result<Option<TxLocation>, FinalisedStateError> {
         let ro = self.env.begin_ro_txn()?;
-        let mut cursor = ro.open_ro_cursor(self.txids)?;
 
-        let target: [u8; 32] = (*txid).into();
+        // Reverse-index point lookup: `txid_location` maps a txid directly to its
+        // `TxLocation`, replacing the former full scan of the height-keyed `txids` table.
+        let key: [u8; 32] = (*txid).into();
 
-        for (height_bytes, stored_bytes) in cursor.iter() {
-            if let Some(tx_index) =
-                Self::find_txid_position_in_stored_txid_list(&target, stored_bytes)
-            {
-                let height = Height::from_bytes(height_bytes)?;
-                return Ok(Some(TxLocation::new(height.0, tx_index as u16)));
+        match ro.get(self.txid_location, &key) {
+            Ok(stored_bytes) => {
+                let entry =
+                    StoredEntryFixed::<TxLocation>::from_bytes(stored_bytes).map_err(|e| {
+                        FinalisedStateError::Custom(format!("corrupt txid_location entry: {e}"))
+                    })?;
+                if !entry.verify(key) {
+                    return Err(FinalisedStateError::Custom(
+                        "txid_location entry checksum mismatch".to_string(),
+                    ));
+                }
+                Ok(Some(*entry.inner()))
             }
+            Err(lmdb::Error::NotFound) => Ok(None),
+            Err(e) => Err(FinalisedStateError::LmdbError(e)),
         }
-        Ok(None)
-    }
-
-    /// Efficiently scans a raw `StoredEntryVar<TxidList>` buffer to locate the index
-    /// of a given transaction ID without full deserialization.
-    ///
-    /// The format is:
-    /// - 1 byte: StoredEntryVar version
-    /// - CompactSize: length of the item
-    /// - 1 byte: TxidList version
-    /// - CompactSize: number of the item
-    /// - N x (1 byte + 32 bytes): tagged Hash items
-    /// - 32 bytes: checksum
-    ///
-    /// # Arguments
-    /// - `target_txid`: A `[u8; 32]` representing the transaction ID to match.
-    /// - `stored`: Raw LMDB byte slice from a `StoredEntryVar<TxidList>`.
-    ///
-    /// # Returns
-    /// - `Some(index)` if a matching txid is found
-    /// - `None` if the format is invalid or no match
-    #[inline]
-    fn find_txid_position_in_stored_txid_list(
-        target_txid: &[u8; 32],
-        stored: &[u8],
-    ) -> Option<usize> {
-        const CHECKSUM_LEN: usize = 32;
-
-        // Check is at least sotred version + compactsize + checksum
-        // else return none.
-        if stored.len() < TransactionHash::VERSION_TAG_LEN + 8 + CHECKSUM_LEN {
-            return None;
-        }
-
-        let mut cursor = &stored[TransactionHash::VERSION_TAG_LEN..];
-        let item_len = CompactSize::read(&mut cursor).ok()? as usize;
-        if cursor.len() < item_len + CHECKSUM_LEN {
-            return None;
-        }
-
-        let (_record_version, mut remaining) = cursor.split_first()?;
-        let vec_len = CompactSize::read(&mut remaining).ok()? as usize;
-
-        for idx in 0..vec_len {
-            // Each entry is 1-byte tag + 32-byte hash
-            let (_tag, rest) = remaining.split_first()?;
-            let hash_bytes: &[u8; 32] = rest.get(..32)?.try_into().ok()?;
-            if hash_bytes == target_txid {
-                return Some(idx);
-            }
-            remaining = &rest[32..];
-        }
-
-        None
     }
 }
--- a/packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
@@ -34,6 +34,9 @@ impl DbV1 {
     /// Writes a given (finalised) [`IndexedBlock`] to ZainoDB.
     ///
     /// NOTE: This method should never leave a block partially written to the database.
+    // `u32::is_multiple_of` is only stable from Rust 1.87; the `% 100 == 0` form below keeps the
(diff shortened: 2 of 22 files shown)


zancas:
This should not be PR'd against dev.
It should be PR's against 0.4.0

packages/zaino-state/src/chain_index/finalised_state/db/v1/block_core.rs
```diff
@@ -307,71 +307,26 @@ impl DbV1 {
...
-            // Each entry is 1-byte tag + 32-byte hash
-            let (_tag, rest) = remaining.split_first()?;
-            let hash_bytes: &[u8; 32] = rest.get(..32)?.try_into().ok()?;
-            if hash_bytes == target_txid {
-                return Some(idx);
+        // Reverse-index point lookup: `txid_location` maps a txid directly to its
+        // `TxLocation`, replacing the former full scan of the height-keyed `txids` table.
+        let key: [u8; 32] = (*txid).into();
+
+        match ro.get(self.txid_location, &key) {
```
zancas, line 315: Nice.

packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
```diff
@@ -311,6 +314,22 @@ impl DbV1 {
         let (txids, transparent): (Vec<TransactionHash>, Vec<Option<TransparentCompactTx>>) =
             transactions.into_iter().unzip();
 
+        // Reverse txid index entries (`txid -> TxLocation`). Built before `txids` is moved into
+        // the `TxidList` below, and sorted by txid so the random-keyed `txid_location` B-tree
+        // sees locally-ordered inserts.
+        let mut txid_location_entries: Vec<([u8; 32], TxLocation)> =
+            Vec::with_capacity(txids.len());
+        for (tx_index, txid) in txids.iter().enumerate() {
```
zancas, line 322: Why is an O(n) loop necessary here?

idky137: This is not looking through the complete chain, only the txids in the transaction. we must loop over all of the to index each.

zancas:
Can we make the PR against release 0.4.0 instead of dev?

packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
zancas: What do you think of this proposal:


Here's the single-pass proposal, written against PR head b62b0418. It folds the new second pass into the existing
  per-transaction loop and, as a bonus of the same structural change, eliminates the feature-gated O(n²) .find. I'm
  presenting it as a diff rather than editing files here, since this worktree is on instantiate_NFS_at_launch, not the
  PR branch.

  1. Replace the dup-guard HashSet with an index HashMap

```Rust
  // before (line 145)
  let mut txid_set: HashSet<TransactionHash> = HashSet::with_capacity(tx_len);

  // after
  // txid -> in-block index. `insert` returning `Some` is the duplicate-txid guard
  // (formerly a separate HashSet); the index makes in-block prevout lookups O(1).
  let mut txid_index: HashMap<TransactionHash, u16> = HashMap::with_capacity(tx_len);
  let mut txid_location_entries: Vec<([u8; 32], TxLocation)> = Vec::with_capacity(tx_len);
```
  2. Single pass: hoist the u16 conversion, fuse the dup check, emit the index entry

  The loop head loses its #[allow(clippy::unused_enumerate_index)] and the _tx_index underscore — the index is now genuinely used at the top:

```Rust
  for (tx_index, tx) in block.transactions().iter().enumerate() {
      let hash = tx.txid();

      // Bound the index first: the dup map, the reverse-index entry, and the
      // spent map all want the narrow u16 form.
      let tx_index = u16::try_from(tx_index).map_err(|_| FinalisedStateError::InvalidBlock {
          height: block_height.0,
          hash: block_hash,
          reason: format!("transaction index {tx_index} does not fit into u16"),
      })?;
      let tx_location = TxLocation::new(block_height.into(), tx_index);

      if txid_index.insert(*hash, tx_index).is_some() {
          return Err(FinalisedStateError::InvalidBlock {
              height: block_height.0,
              hash: block_hash,
              reason: format!("duplicate transaction hash in block: {hash:?}"),
          });
      }

      txid_location_entries.push(((*hash).into(), tx_location));

      // transparent / sapling / orchard pushes: unchanged
      // spent_map per-input loop: unchanged (uses `tx_location` already in scope)
      ...
  }
```

  The original mid-loop try_from block (lines 199–207) is deleted — tx_index/tx_location now exist from the top of the body. One behavioral footnote: on a block that is invalid two ways at once (duplicate txid and >65 535 transactions), the overflow error now wins over the dup error at the crossover transaction. Both paths reject the block, so nothing observable changes.

  3. Delete the second pass; keep only the sort

  Lines 317–330 (the flagged loop, its with_capacity, and the duplicated weaker-message try_from) go away entirely.
  After the main loop, only this remains:

```Rust
  // Sorted by txid so the random-keyed `txid_location` B-tree sees
  // locally-ordered inserts.
  txid_location_entries.sort_by_key(|entry| entry.0);
```

  The "built before txids is moved" ordering constraint in the old comment is now satisfied by construction, so that
  half of the comment disappears too.

  4. Free payoff: the addrhist in-block prevout lookup becomes O(1)

  In the transparent_address_history_experimental block, the linear scan (lines 246–252):

```Rust
  if txid_set.contains(&prev_tx_hash) {
      if let Some((tx_index, (_, Some(prev_transparent)))) = transactions
          .iter()
          .enumerate()
          .find(|(_, (h, _))| h == &prev_tx_hash)
```

  becomes:

```Rust
  if let Some(&prev_idx) = txid_index.get(&prev_tx_hash) {
      if let (_, Some(prev_transparent)) = &transactions[prev_idx as usize] {
```

  with prev_output_tx_location = TxLocation::new(block_height.0, prev_idx) inside. The else if let Ok(...) =
  tokio::task::block_in_place(...) / final else error arms are untouched, and the silent-skip semantics for an in-block prevout whose output index is missing are preserved exactly (the nested if lets just don't fire, same as the old .find chain). The indexing is in-bounds by construction: prev_idx was assigned when that transaction was pushed into transactions, and the current transaction is pushed before its inputs are processed, matching today's behavior for self-references.

  Net effect

  - One pass over the block instead of two; the 14-line flagged loop and its unreachable, weaker-error try_from are gone.
  - Same allocation count (the HashMap replaces the HashSet one-for-one; txid_location_entries existed in the PR already).
  - The only quadratic pattern in the file's write path becomes linear, without adding any structure that wasn't
  already needed for the dup check.

packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
```diff
@@ -364,6 +384,17 @@ impl DbV1 {
                 WriteFlags::NO_OVERWRITE,
             )?;
 
+            // Reverse txid index: `txid -> TxLocation`.
+            for (txid_bytes, tx_location) in &txid_location_entries {
+                let entry_bytes = StoredEntryFixed::new(txid_bytes, *tx_location).to_bytes()?;
```
zancas, line 389: ● With the single-pass and cursor changes assumed in, the remaining costs in write_block split cleanly into CPU (serialization does redundant work per entry) and I/O (one durable commit — two fsyncs — per block). The I/O side is where "slow sync and migration" actually lives; the serialization fixes are real but secondary. Everything below is grounded in entry.rs and the PR-head write_core.rs.

  Serialization

  1. Every value is currently serialized twice. StoredEntryFixed::new / StoredEntryVar::new (entry.rs:119-127, entry.rs:297-305) encode the item into a Vec to compute the BLAKE2b checksum, then discard those bytes, storing only {item, checksum}. Inside the write txn, .to_bytes() encodes the same item all over again. Restructure construction to produce the final wire bytes once:

```Rust
  pub(crate) fn encode<K: AsRef<[u8]>>(key: K, item: &T) -> io::Result<Vec<u8>> {
      // version tag + item body + checksum, built in one buffer
  }
```
  returning the ready-to-put bytes. Halves encode CPU and deletes one Vec per entry.

  2. The checksum input allocates a concatenation it never needed. Both new and verify do blake2b256(&[key.as_ref(),&body].concat()). Blake2bVar is streaming — hasher.update(key); hasher.update(body) does the same hash with zero extra allocation. Trivial, applies to every entry written and verified.

  3. verify() re-serializes the decoded item — but the encoded bytes are sitting in the DB value. The stored format is tag ‖ item_bytes ‖ checksum, and the checksum is defined over key ‖ item_bytes. So verification never needs to re-encode anything, let alone loop over candidate versions re-encoding each (entry.rs:314-330): hash the stored body slice directly during from_bytes, before decoding. This is the read-path twin of #1 and makes the post-commit validate_block_blocking read-back dramatically cheaper.

  4. Encode outside the write txn. The to_bytes() calls currently execute between begin_rw_txn and commit. With #1, the encoded buffers exist before spawn_blocking even starts, so the single-writer window shrinks to pure puts. For the fixed-size tables (txid_location, heights, commitment_tree_data, spent), lmdb 0.8 also exposes RwTransaction::reserve (MDB_RESERVE) — encode straight into the mapped page and skip the intermediate buffer + memcpy entirely.

  5. Stop cloning the block's transaction data. The main loop does tx.transparent().clone() / tx.sapling().clone() / tx.orchard().clone() per transaction, but write_block owns block: IndexedBlock. Destructure it and consume transactions by value; the clones of every shielded bundle in every block disappear.

zancas:
This is superceded by #1209

nachog00:
During a long deploy of this ref against zebra 5.1.0:

```
MDB_MAP_FULL — the LMDB database hit its configured size limit. It synced to height 3,142,607 (deep into the
  chain) and the DB filled up the 64GB default map size. The config has storage.database.size = 64 but the
  accumulator + reverse txid index makes the DB much larger than pre-#1112 versions.
```

PR #1208, opened 2026-06-11
Reify nfs when fs synced

AloeareV:
Builds atop #1150


zancas:
Thank you for this work, Hazel. I propose that we close this PR. `dev` now
reaches the same goal by a different route.

PR #1440 merged on 2026-08-19. Its commit 00902f5ae removed
`chain_index/non_finalised_state.rs` and
`ChainIndexSnapshot::StillSyncingFinalizedState`. The chain head now anchors
before `ChainIndex::new` returns, so no snapshot can be in the "still syncing"
state. The snapshot type is `Arc<MapBackedSnapshot>`. This is the result that
#1096 asked for.

This PR cannot be brought up to date. Its base is from 2026-07-01, and `dev`
has more than 700 commits since then. Twelve of its 21 files no longer exist
on `dev`. Those files include `backends/fetch.rs`, `backends/state.rs`, and
the non-finalised state module. #1150, which this PR builds on, closed without
a merge.

`dev` did not adopt `ProvisionalBlock` or the reify step. The design in #1550
expresses partial availability as serviceability. A read returns
`NotServiceable` until its index is built, and `ServiceableRange` reports the
covered heights. The question of chainwork relative to an anchor continues in
#1560.

PR #1209, opened 2026-06-11
0 4 0 release bug

idky137:


packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
```diff
@@ -311,6 +314,22 @@ impl DbV1 {
         let (txids, transparent): (Vec<TransactionHash>, Vec<Option<TransparentCompactTx>>) =
             transactions.into_iter().unzip();
 
+        // Reverse txid index entries (`txid -> TxLocation`). Built before `txids` is moved into
+        // the `TxidList` below, and sorted by txid so the random-keyed `txid_location` B-tree
+        // sees locally-ordered inserts.
+        let mut txid_location_entries: Vec<([u8; 32], TxLocation)> =
+            Vec::with_capacity(txids.len());
+        for (tx_index, txid) in txids.iter().enumerate() {
```
zancas, line 322: See this comment on the original PR:

https://github.com/zingolabs/zaino/pull/1207/changes#r3398927655

packages/zaino-state/src/chain_index/finalised_state/db/v1/write_core.rs
```diff
@@ -364,6 +384,17 @@ impl DbV1 {
                 WriteFlags::NO_OVERWRITE,
             )?;
 
+            // Reverse txid index: `txid -> TxLocation`.
+            for (txid_bytes, tx_location) in &txid_location_entries {
+                let entry_bytes = StoredEntryFixed::new(txid_bytes, *tx_location).to_bytes()?;
```
zancas, line 389: See this comment on the original PR: https://github.com/zingolabs/zaino/pull/1207/changes#r3399020933

zancas (changes requested):
Let's consider the AI proposed further optimizations.

idky137:
Looking at these, they are mostly small efficiency improvements and not related to code validity. 

Additionally we must be careful with this migration, it holds extra steps that may look redundant, but are important to ensure that users that started syncing Zaino with the nu6.2 alpha_1 build, helping with early testing, do not have corrupt cache files on migration completion.

As I am away for the next 4 days, and blocking on these could block release by up to a week, can we open an issue with these comments in?

packages/zaino-state/src/chain_index/finalised_state/db/v1.rs
```diff
@@ -227,6 +227,12 @@ pub(crate) struct DbV1 {
     /// Used to check spent status of given outpoints, retuning spending tx.
     spent: Database,
 
+    /// Reverse txid index: `TransactionHash` -> `StoredEntryFixed<TxLocation>`
+    ///
+    /// Maps a transaction id to its on-chain `TxLocation`, giving O(log n) previous-output
+    /// resolution instead of a full scan of the height-keyed `txids` table.
+    txid_location: Database,
+
```
zancas, line 235: Here're are more Fable provided optimization suggestions.

Ideally we can grab the most high impact one(s) for this PR, and reference this comment in a new issue for the rest:

  LMDB commit

  The env opens with only NO_TLS | NO_READAHEAD (db/v1.rs:331), so each per-block commit is fully durable: data fdatasync plus meta-page sync — two fsyncs per block, which is the hard floor on sync throughput regardless of how fast the CPU side gets.

  6. NO_META_SYNC is nearly free. It halves the fsync count, and its documented failure mode is exactly one: a crash may undo the last committed transaction, with integrity intact. For a rebuildable index whose sync loop already  resumes from validated_tip, losing the final block to a crash is indistinguishable from having stopped one block earlier.

  7. Batch K blocks per commit during sync/migration — the dominant win. Extract the put-sequence into a
  write_block_in_txn(&self, txn: &mut RwTransaction, block) helper; write_block stays begin → helper → commit for tip-following, and a write_blocks(batch) path loops the helper before a single commit. The "never leave a block  partially written" invariant is per-transaction, so it survives unchanged — a crash atomically drops the whole in-flight batch. This divides the fsync cost by K and also amortizes the per-block spawn_blocking dispatch, the Self{ ... } handle reconstruction (write_core.rs:340-361), and the validation read-back (validate once per batch).

  8. NO_SYNC for the initial-sync/migration phase only. Without WRITE_MAP, on a write-order-preserving filesystem, MDB_NOSYNC loses only durability — integrity holds; worst case is re-syncing from the validator, which is the recovery path anyway. Call env.sync(true) every N blocks and at completion/shutdown. One wrinkle: lmdb 0.8 doesn't wrap mdb_env_set_flags, so flipping back to durable mode mid-process means either a raw lmdb_sys::mdb_env_set_flags call or reopening the env when sync completes. If that plumbing is unappealing, #6 + #7 together get most of the benefit with none of the mode-switching.

  9. APPEND for the height-keyed tables. Height::encode_v1 writes big-endian precisely so keys sort numerically (legacy.rs:473-476), and sync writes heights in ascending order — so headers, txids, transparent, sapling, orchard, and commitment_tree_data qualify for WriteFlags::APPEND on the sync path. That skips the root-to-leaf search per put and, more valuably, packs leaf pages to ~full instead of the ~half-full fill that sequential inserts-with-splits produce: smaller DB, fewer dirty pages, fewer bytes per fsync. It must stay off any rewrite/reorg path (APPEND errors if the key isn't strictly greater than the table's last), so gate it the same way as the batch path.

  10. Sort the spent_map writes. The loop at write_core.rs:427 iterates a HashMap in random order into the outpoint-keyed table. Collect-and-sort before the put loop — the identical locality argument that justified sorting txid_location_entries, and the same cursor treatment applies.

  Expected impact, roughly in order: batching (#7) and/or NO_SYNC-during-bulk (#8) are multiplicative on sync throughput because fsync is the bottleneck; NO_META_SYNC (#6) is a free 2× on the same axis; APPEND (#9) compounds by shrinking what each fsync flushes; the serialization items (#1–5) cut per-block CPU meaningfully but only matter once the fsyncs stop dominating.

idky137: I actually almost added the NO_SYNC option in this PR, but decided against it as it would add more code churn in a hotfix that does not need it. 

I agree that we should implement this but it should probably be added as a separate feature on dev and fully tested, as it has a large surface it could effect.

zancas (approved):

zancas (changes requested):
UtAck   let's ensure that the CI fails aren't showing a bug in zaino.

zancas (approved):

zancas:
Local tests pass!

PR #1212, opened 2026-06-12
Leverage zebrad's new mine-to-orchard, for test

zancas:
Leverage zebrad 5.1.0 's mine-to-orchard capability to make regtest-driven tests more efficient.

All tests that run on dev, run on this feature.
Point Estimate Delta in test run times:

* walletless = 187s - 60s = 127 seconds faster
* wallet       =  950s - 458 = 492 seconds faster

Significant DRY refactoring nets:  - 491 LOC 

The 42f1369e18a72e451db25711e785ec21b5a0cb5c commit in this series assumes that per-block checks during chain generation are unnecessary for the subset of tests it affects.  That assumption should be carefully considered.

I also note that some tests use a hard wired 100 block generation represent the finalization depth.. this is not a good assumption.


zancas:
Remember friends...  you cannot exploit the vuln in the loc that's been deleted from the repo.

idky137 (approved):
Looks good!

PR #1214, opened 2026-06-12
Add write-path performance benchmark tests, and test sync optimizations

zancas:
Four #[ignore]d benchmark harnesses under chain_index/tests/benchmarks.rs covering the costs identified in the PR #1207 review (threads r3398927655 and r3399020933): end-to-end write_block ingest of the vector chain, and StoredEntry fixed encode / decode+verify / var encode micro-benchmarks. Run with:

  **cargo nextest run -p zaino-state --run-ignored ignored-only --no-capture benchmarks**


zancas:
# Write-path benchmark baseline

Baseline numbers for the `zaino-state` finalised-state write path, captured
before applying the optimizations proposed in the PR
[#1207](https://github.com/zingolabs/zaino/pull/1207) review threads
([r3398927655](https://github.com/zingolabs/zaino/pull/1207#discussion_r3398927655),
[r3399020933](https://github.com/zingolabs/zaino/pull/1207#discussion_r3399020933)).

## Environment

| | |
|---|---|
| Commit | `3e23f97` (branch `optimize_sync`) |
| Date | 2026-06-11 |
| CPU | 12th Gen Intel Core i7-1280P |
| Kernel | Linux 7.0.11-arch1-1 |
| rustc | 1.95.0 |
| Profile | nextest `test` profile (optimized + debuginfo) |

## How to reproduce

```sh
# Full suite (tempdirs land on /tmp — tmpfs on this machine):
cargo nextest run -p zaino-state --run-ignored ignored-only --no-capture benchmarks

# Disk-backed ingest (point TMPDIR at a real filesystem so the per-block
# durable LMDB commit pays its actual fsync cost):
mkdir -p target/bench-tmp
TMPDIR=$PWD/target/bench-tmp cargo nextest run -p zaino-state \
  --run-ignored ignored-only --no-capture write_block_chain_ingest
```

Benchmark sources: `packages/zaino-state/src/chain_index/tests/benchmarks.rs`.

## End-to-end `write_block` ingest

Regtest vector chain: **201 blocks, 830 transactions**; fresh `ZainoDB` per
run, 3 runs, median reported.

| Tempdir backing | Median total | Per block | Throughput |
|---|---:|---:|---:|
| tmpfs (`/tmp`) | 13.98 ms | 69.6 µs | 14,373 blocks/s |
| ext4 (`TMPDIR=target/bench-tmp`) | 167.07 ms | 831 µs | 1,203 blocks/s |

## Serialization micro-benchmarks

CPU-only; independent of disk backing. 5 runs, median reported.

| Benchmark | Work per run | Per item | Throughput |
|---|---|---:|---:|
| `StoredEntryFixed<TxLocation>` encode | 100,000 entries | 254 ns | 3.94 M entries/s |
| `StoredEntryFixed<TxLocation>` decode+verify | 100,000 entries | 246 ns | 4.06 M entries/s |
| `StoredEntryVar<TxidList>` encode (2,000 txids/list) | 50 lists | 92.4 µs | 10,819 lists/s |

## Interpretation

- The ext4 ingest is **12× slower** than tmpfs for identical work: on real
  disk, ~760 µs of the 831 µs per block is the durable LMDB commit (two
  fsyncs) plus the disk-touching post-commit validation read-back. This
  matches the r3399020933 analysis that the I/O side is where slow sync
  lives.
- The CPU-side costs targeted by the single-pass loop refactor
  (r3398927655) and the `entry.rs` serialization fixes (encode-once,
  streaming checksum input, verify-without-re-encode) all live inside the
  ~70 µs/block tmpfs slice. Gauge those changes against:
  - the **micro-benchmarks** for the `entry.rs` fixes (expect roughly 2× on
    encode, more on verify), and
  - the **tmpfs ingest** number for `write_block`-internal changes
    (single-pass loop, encoding outside the write txn, dropping per-tx
    bundle clones).
- Moving the **ext4 ingest** number materially would require amortizing
  commits across blocks (batched writes) or relaxing durability — a design
  decision beyond the current proposals.

zancas:
# Results after optimizations

Follow-up to the [baseline numbers](https://github.com/zingolabs/zaino/pull/1214#issuecomment-4687655704) above. Same machine, harness, and vector chain (201 blocks / 830 transactions); both changes are implemented on `optimize_sync` locally (`e30c5556`, `dcbf57c5`) and not yet part of this PR's pushed head.

## `e30c5556` — single-pass `write_block` loop ([r3398927655](https://github.com/zingolabs/zaino/pull/1207#discussion_r3398927655))

Folds the reverse-`txid_location` build into the main per-transaction loop and removes the feature-gated O(n²) in-block prevout scan. Performance is **neutral within run-to-run noise** on this chain (tmpfs medians 69–87 µs/block across repeat invocations vs the 69.6 µs baseline): at ~4 transactions per vector block the removed work sits below the noise floor. The win is structural and scales with per-block transaction count.

Noise characterization from repeat runs: tmpfs ingest medians cluster at 13.9–14.9 ms; ext4 is bimodal across invocations (160–170 ms typical, occasional ~440 ms runs), so single ext4 runs should not be compared in isolation.

## `dcbf57c5` — byte-budgeted batched writes

`WriteBatcher` + `DbV1::write_blocks`: contiguous blocks share one durable LMDB commit, flushing on a 128 MiB byte budget or when a block depends on uncommitted batch state (it spends an output created by — or a sibling output of a transaction spent from by — the pending batch). Reproduce with the `write_block_chain_ingest_batched` benchmark (runs a 64 KiB and the default 128 MiB budget).

| Backing | Per-block ingest | Batched (64 KiB budget) | Speedup |
|---|---:|---:|---:|
| ext4 | 830 µs/block (1,205 blocks/s) | 468 µs/block (2,139 blocks/s) | **1.8×** |
| tmpfs | 121 µs/block | 134 µs/block | ~0.9× (no fsync to save) |

Key finding: on this vector chain the **dependency rule, not the byte budget, is the amortization ceiling**. The regtest wallets spend freshly created outputs almost every other block, so the chain splits into ~2-block batches (103 batches at a 64 KiB budget; still 98 at 128 MiB — the whole chain is only ~1 MB, so at the default budget every split is dependency-driven), and the 1.8× comes from roughly halving commit count. Chains with sparse cross-block transparent spends (e.g. early mainnet) form much larger batches and amortize proportionally more of the ~760 µs/block commit cost.

Follow-up that would lift the ceiling: make the batch build-phase reads pending-aware (consult the open batch before the DB in `resolve_spent_outpoints_for_set_info`, the double-spend pre-check, and `apply_prior_block_transitions`), so the byte budget becomes the only flush trigger.

zancas:
# Results: pending-aware batch reads (dependency ceiling lifted)

Follow-up to the [batched-writes results](https://github.com/zingolabs/zaino/pull/1214#issuecomment-4687963399) above. Implemented locally on `optimize_sync` as `51fb3f02` (same machine, harness, and 201-block / 830-tx vector chain).

A `PendingBatchState` overlay — the accumulator after the latest pending block, every pending transaction with its location and transparent data, and every pending spend — is threaded through the batch build phase and consulted before the committed tables at each read that can touch uncommitted state (prevout resolution, the already-spent pre-check, `apply_prior_block_transitions`, and the feature-gated address-history input path). `WriteBatcher` drops its dependency tracking entirely: **the byte budget is the only flush trigger**, capping batches at ~128 MiB of estimated write volume.

Batch counts: 64 KiB budget 103 → **18** batches; 128 MiB budget 98 → **1** batch (this fixture's whole chain is ~1 MB, so the cap doesn't bind here; on mainnet batches are 128 MiB chunks).

Single invocation, medians (this invocation's per-block ext4 run landed in the slow bimodal mode at 1.78 ms/block; the baseline's typical mode is 830 µs/block):

| Backing | Per-block | Batched 64 KiB | Batched 128 MiB |
|---|---:|---:|---:|
| ext4 | 1.78 ms/block (563 blocks/s) | 469 µs/block (2,130 blocks/s) | **137.7 µs/block (7,264 blocks/s)** |
| tmpfs | 103 µs/block | 112 µs/block | **46.9 µs/block (21,315 blocks/s)** |

Speedups: ext4 single-batch is **12.9×** this invocation's per-block run and **6.0×** the baseline's typical mode. tmpfs single-batch is ~2.2× — one transaction setup and one commit save CPU even with fsyncs nearly free.

Correctness: the batched-vs-per-block equivalence test ingests this chain — whose blocks spend each other's fresh outputs almost every other block, now *inside* batches — and reproduces identical tip, `txid_location` mappings, and txout-set accumulator state. The already-spent pre-check now also catches double spends across batch blocks. 148/148 default-feature tests pass, plus the finalised-state suite under `transparent_address_history_experimental`.

idky137 (changes requested):
This is not a clear  "hot fix", this must be merged into dev and fully tested.

packages/zaino-state/src/chain_index/finalised_state/db.rs
```diff
@@ -244,6 +244,24 @@ impl DbBackend {
         Ok(Self::V1(DbV1::spawn(cfg).await?))
     }
 
+    /// Writes a contiguous, dependency-free batch of blocks with a single durable commit
+    /// where the backend supports it (see `DbV1::write_blocks` for the batch contract).
+    /// The legacy v0 backend has no batched path and falls back to per-block writes.
+    pub(crate) async fn write_blocks(
```
idky137, line 251: This does not use the capability trait system set out. it is at odds with the current finalised state framework. If we want batch block write it should be added to the WRITE_CORE trait.

packages/zaino-state/src/chain_index/finalised_state/router.rs
```diff
@@ -339,6 +339,21 @@ impl DbCore for Router {
     }
 }
 
+impl Router {
+    /// Writes a contiguous, dependency-free batch of blocks with one durable commit via
+    /// the backend currently serving `WRITE_CORE` (see `DbV1::write_blocks` for the batch
+    /// contract). Inherent rather than part of [`DbWrite`] because only backends with a
+    /// batched path benefit; others fall back to per-block writes internally.
+    pub(crate) async fn write_blocks(
```
idky137, line 348: similar to above.

packages/zaino-state/src/chain_index/tests/finalised_state/v1.rs
```diff
@@ -307,6 +313,286 @@ async fn try_write_invalid_block() {
     dbg!(zaino_db.db_height().await.unwrap());
 }
 
+/// Spawns a v1 `DbBackend` over a fresh tempdir, for tests that need direct
+/// backend access (e.g. the batched write path and accumulator reads).
+async fn spawn_fresh_v1_backend() -> (TempDir, DbBackend) {
```
idky137, line 407: we have this functionality already. is there a reason we need another spawn method? we should try to use the current methods where possible.

packages/zaino-state/src/chain_index/tests/benchmarks.rs
```diff
@@ -0,0 +1,295 @@
...
+/// production sync-time profile.
+///
+/// multi_thread required: `DbV1::write_block` calls
+/// `tokio::task::block_in_place`, which panics on a current-thread runtime.
+#[tokio::test(flavor = "multi_thread")]
+#[ignore = "benchmark: run with `cargo nextest run -p zaino-state --run-ignored ignored-only --no-capture benchmarks`"]
+async fn write_block_chain_ingest() {
+    const RUNS: usize = 3;
+
+    let test_vector_data = load_test_vectors().expect("test vectors load");
```
idky137, line 81: this data is not a good representation of sync. Even testnet does not show the problems we get on mainnet. we must run larger tests to actually see the bottlenecks.

idky137 (changes requested):
This PR does not integrate with the current FS trait system. 

It is also at odds with #1200, and does not clearly fix any obvious problems. We should look at applying this work on top of #1200, so we do not delay getting instant responsiveness out to the community.

packages/zaino-state/src/chain_index/tests/benchmarks.rs
zancas: The unit-test level benchmarks are NOT in any way intended to replace long running or other more definitive testing.   Rather they augment.

The existence of integration testing (deployment testing) doesn't eliminate the value of unit testing.

zancas: Improving the realism of test-vectors is a valuable idea.

Obviously the long run deployment data is definitive.

packages/zaino-state/src/chain_index/tests/finalised_state/v1.rs
zancas: I will look into this.

zancas:
@idky137 thanks for the review!

idky137 (changes requested):
This pr now has multiple changes that touch a big part of the zaino-state codebase.

We separately agreed that updating block validation was a big change that should not be rushed into a release as it has a very large surface for introducing bugs and needs to be carefully reviewed.

This pr needs to be broken into the separate fixes it implements.

There are already several problems with this pr that needs fixing, these new commits add a very large bug surface and should be reviewed separately.

AloeareV:
A lot of conflicts, and this likely wants to go on dev rather than as a hotfix. Leaving open as a breadcrumb, this should be closed as soon as a PR into dev supersedes it.

zancas:
Not integrated.

PR #1215, opened 2026-06-12
Rc 0 4 0 slow sync

idky137:
[fixed 1.1.0->1.2.0 migration, moved db to f_sync](https://github.com/zingolabs/zaino/commit/6a2090399a0c03a4ed1140ab5191e10d6c444ae6)

### DIFF
--- a/packages/zaino-state/src/chain_index/finalised_state/migrations.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state/migrations.rs
@@ -187,7 +187,7 @@ use crate::{
     chain_index::{
         finalised_state::{
             capability::{CapabilityRequest, DbMetadata},
-            db::v1::{DB_VERSION_V1, TX_OUT_SET_INFO_ACCUMULATOR_KEY},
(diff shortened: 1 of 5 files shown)


zancas:
If I understand correctly this turns off fsync and moves the overhead of file I/O to the interval dictated by the 1_000 block batch size?

Wouldn't it likely increase performance more if that window were specified in bytes, probably picked as a function of a standard 512MiB page size?

packages/zaino-state/src/chain_index/finalised_state/migrations.rs
```diff
@@ -575,6 +576,12 @@ impl<T: BlockchainSource> Migration<T> for Migration0To1 {
 
         info!("promoting v1 database to primary.");
 
+        // The migrated v1 data is about to become primary and v0 is wiped below. Under `NO_SYNC`
+        // the shadow's tail blocks and its `Complete` status may not be on disk yet; force them
+        // durable now so a crash during or after promotion can never lose migrated blocks that
+        // exist only in v1 (v0 is removed and cannot serve as a fallback).
```
zancas, line 582: Is there a performance trade-off for this?

How robust to in-sync crashes do we need to be?

PR #1216, opened 2026-06-12
Add Prometheus /metrics endpoint

nachog00:
## Summary

- Adds a Prometheus metrics scrape endpoint (`/metrics`) to zainod, disabled by default
- Enable via config: `metrics_endpoint = "0.0.0.0:9998"`
- Uses the `metrics` + `metrics-exporter-prometheus` crates (same stack as Zebra)

### Metrics exposed

| Metric | Description |
|--------|-------------|
| `zaino.sync.finalized_height` | Current finalized block height being synced |
| `zaino.sync.target_height` | Target finalized block height for current sync iteration |
| `zaino.chain.tip_height` | Latest chain tip height reported by the validator |

### Feature gating

The exporter is behind a non-default `prometheus` feature in zainod. The `metrics` facade crate stays as a regular dep in zaino-state — its macros compile to an atomic pointer check and silently no-op without a recorder installed, so there's zero runtime cost when the feature is off.

`prometheus` is bundled into `no_tls_use_unencrypted_traffic` rather than exposed as a standalone feature. Zaino is a privacy-focused indexer, so observability capabilities should not ship in default builds. The `no_tls` flag already targets operators deploying behind trusted infrastructure (service meshes, internal networks) — the same users who need metrics scraping. Bundling avoids expanding the CI image matrix while keeping observability explicitly opt-in.

The `metrics_endpoint` config field is always deserialized regardless of feature state to avoid `deny_unknown_fields` parse failures when a config file is shared across builds. Without the feature, the field is silently ignored.

### Context

The devops cluster already has kube-prometheus-stack, ServiceMonitors, and a Zaino sync Grafana dashboard ready to consume these metrics. This lands the zaino side so we can start scraping sync progress from deployed instances.

Rebased from the original `feature/prometheus-metrics` branch (nachog00) onto current dev.

## Test plan

- [x] Deploy on devops cluster with `metrics_endpoint = "0.0.0.0:9998"`
- [x] Verify `/metrics` returns Prometheus text format with all three gauges
- [x] Confirm ServiceMonitor scrapes and Grafana dashboard populates
- [x] Verify metrics endpoint stays disabled when config field is omitted
- [x] Verify default build (no feature) compiles and ignores the config field
- [x] Verify `cargo check --package zainod` and `cargo check --package zainod --features prometheus` both pass


zancas (changes requested):
This looks great but can you please gate out all instrumentation and metrics code, and make the new observeability-only dependencies be optional as a function of that gate?

Also please make that feature non-default, only a consumer who explicitly asks for observeability should get any of the code or dependencies.

packages/zainod/Cargo.toml
```diff
@@ -18,7 +18,10 @@ path = "src/lib.rs"
 
 [features]
 # Removes network restrictions.
-no_tls_use_unencrypted_traffic = ["zaino-serve/no_tls_use_unencrypted_traffic"]
+no_tls_use_unencrypted_traffic = ["zaino-serve/no_tls_use_unencrypted_traffic", "prometheus"]
```
zancas: Let's make it that the user has to explicitly ask for prometheus instead of picking it up incidentally with no_tls_use_unencrypted_traffic.

Cargo.toml
```diff
@@ -95,6 +95,10 @@ async-stream = "0.3"
 async-trait = "0.1"
 futures = "0.3.30"
 
+# Metrics
+metrics = "0.24"
```
zancas, line 99: Should this not also be "default-features = false"?

packages/zaino-state/Cargo.toml
```diff
@@ -45,6 +45,9 @@ zebra-rpc = { workspace = true }
 # Tracing
 tracing = { workspace = true }
 
+# Metrics
```
zancas, line 51: optional = true?

packages/zaino-state/src/chain_index.rs
```diff
@@ -872,6 +872,7 @@ impl<Source: BlockchainSource> NodeBackedChainIndex<Source> {
                                 "node returned no best block height",
                             ))
                         })?;
+                    metrics::gauge!("zaino.chain.tip_height").set(chain_height.0 as f64);
```
zancas, line 876: How is this accessed in a binary that doesn't compile metrics?

zancas (changes requested):

Cargo.toml
nachog00: I think this only pins the ws version and each individual crate's manifest can determine the actual dependency.

packages/zainod/src/config.rs
```diff
@@ -76,6 +79,11 @@ pub struct ZainodConfig {
     pub zebra_db_path: PathBuf,
     /// Network to connect to (Mainnet, Testnet, or Regtest).
     pub network: Network,
+    /// Prometheus metrics endpoint listen address.
+    ///
+    /// Set to enable the `/metrics` scrape endpoint. Disabled when `None`.
+    /// Requires the `prometheus` feature; ignored without it.
+    pub metrics_endpoint: Option<SocketAddr>,
```
zancas, line 86: Is this properly gated?

packages/zainod/src/metrics.rs
```diff
@@ -0,0 +1,47 @@
+//! Prometheus metrics endpoint for Zaino.
```
zancas, line 1: Is this whole file gated out.

zancas (approved):

PR #1219, opened 2026-06-12
Checkpoint durability on written bytes instead of block count

zancas:
Replace SYNC_CHECKPOINT_INTERVAL (1,000 blocks) with SYNC_CHECKPOINT_BYTE_WINDOW (128 MiB of written key + value bytes). A byte window keeps the crash-loss tail, the per-checkpoint sync stall, and dirty-page accumulation uniform across block sizes: ~64 max-size mainnet blocks per checkpoint instead of up to ~2 GB per 1,000 such blocks, and no over-frequent fsyncs across the near-empty early chain.

The write path counts exact bytes: a put_counted helper wraps every put in write_block's transaction so the count can never drift from the puts it describes, and the spawn_blocking task returns the total. DbV1 carries the window position in a shared bytes_since_checkpoint atomic; the checkpoint fires when the window fills, plus unconditionally at genesis so a brand-new cache has a durable root. The v1.2.0 migration Stage A / Stage B backfill loops count their entry bytes the same way with loop-local accumulators. (Per-height progress-key and accumulator rewrites are a few dozen bytes and not counted.)

Fixes:


zancas:
Commit 240e40b needs review for a potentially important behavior change here:

  What changed mechanically: the heights put (keyed by block hash) used to execute second, between headers and txids.
  Folding the six height-keyed puts into a loop moved heights to after all six. So the execution order of puts inside
  the transaction is different.

  Why that's unobservable:

  1. Atomic visibility. LMDB is MVCC: a read transaction sees the database as of some committed transaction. Every put
  inside one write transaction becomes visible in a single atomic step at txn.commit(). No reader — same process or
  another process sharing the env — can ever observe a state where some of these puts have landed and others haven't,
  so no reader can detect their internal order.
  2. Crash behavior. An uncommitted transaction leaves no trace: copy-on-write pages aren't referenced by a meta page
  until commit, so a crash mid-transaction yields "nothing happened" regardless of how many puts had executed or in
  what order. (This holds under NO_SYNC too — the meta-page flip is what publishes the transaction.)
  3. No read-after-write inside the window. Nothing between the first and last of these puts reads what an earlier put
  wrote. The only within-transaction read in write_block is the address-history spent-marking, which runs after all of
  these in both orderings.
  4. The byte counter sums the same set of key+value lengths; addition commutes.

  The single theoretically distinguishable artifact is error selection when multiple puts would fail in the same
  transaction. All seven use NO_OVERWRITE, so in a multi-process race where another writer landed the same block
  between our pre-check and our write, several puts could each be poised to return KeyExist — and the first one in
  execution order is the error that surfaces. Before, that would have been headers then heights; now heights reports
  last among the seven. But every such case aborts the whole transaction identically (nothing written), and
  write_block's KeyExist recovery handler doesn't branch on which table reported it — it re-reads the header at the
  height and validates. So the worst case is a different table named in a transient race's log line, never a different
  database state or control-flow outcome.

AloeareV:
The described benefits seem good. I don't understand the target branch of `rc_0_4_0_slow_sync`. Deferring.

zancas:
Probably obsolete.

PR #1221, opened 2026-06-12
Rc 0 4 0 slow sync fixed data window

zancas:
This has the #1215 optimizations plus a fixed to data width window for fs access.

For visibility.


zancas:
Again failure to integrate.

PR #1238, opened 2026-06-16
rc-0.4.0: Speed up finalised-state sync & v1.1→v1.2 migration (sandblast-height stall fix)

idky137:
Fixes: Speed up finalised-state sync & v1.1→v1.2 migration (sandblast-height stall fix)

This PR takes lessons learned in #1214 and tries to implement them within the current finalised state framework, ensuring full crash resistance and resumability (sync and migration), compatibility with #1200/#1218, ensuring that feature is not held up on sync optimisations, and compatibility with the future "small footprint db". Changes have also been kept more directed, to ease review into release, and to give us more time to discuss the larger architectural changed proposed in #1214 without blocking release.

---

  Problem

  Finalised-state sync — and the v1.1→v1.2 migration — crawls to a stall around sandblast height. The per-block work is dominated by random I/O that faults once the DB exceeds RAM:
  - the txout-set accumulator did an unbounded fan-out of random spent reads per block,
  - every written block was re-read and re-validated,
  - and the random-key index writes (spent, txid_location) faulted a scattered B-tree leaf per insert.

  This PR removes the per-block random-read amplification, takes validation off the write hot path, and converts the random-key index writes into sorted batched writes — while preserving crash-safety and resumability
  throughout.

  Changes (oldest → newest commit)

  3773b8c1 — updated 1.1.0→1.2.0 upgrade
  Adds a transparent_db() accessor so the migration’s spent-index backfill reads each block’s transparent data directly from the raw table, bypassing the validated-read path (resolve_validated_hash_or_height →
  validate_block_blocking). Fixes: the migration was re-validating the whole chain (merkle recompute + checksum) inside the backfill loop — pure redundant CPU on already-on-disk, trusted data.

  6a209039 — fixed 1.1.0→1.2.0 migration, moved db to f_sync
  Opens the LMDB env with MDB_NOSYNC and forces durability at explicit env.sync(true) checkpoints (SYNC_CHECKPOINT_INTERVAL = 1000) in the write path and migration; fixes the migration’s finalisation ordering (record the
  version gate durably before deleting progress keys). Fixes: per-commit fsync was the dominant cost now that the write path does many random-key inserts; this batches write-back while bounding crash rollback to a known
  tail. (This is the foundation the rest of the PR relies on.)

  413d1ed9 — txout-set accumulator bulk update
  Adds a sequential bulk accumulator builder (build_tx_out_set_accumulator_blocking / rebuild_tx_out_set_accumulator) that recomputes the accumulator from the finalised transparent+spent tables via sequential scans
  (XOR-multiset cancellation; in-memory per-tx unspent counting), plus a freshness watermark. Fixes: the per-block accumulator’s apply_prior_block_transitions did an unbounded fan-out of random spent reads — the primary
  stall driver. Verified equivalent to the incremental result across shard counts.

  06648b36 — batch block write (write_blocks_to_height)
  Moves the tip→height ingestion loop into the backend behind a new DbWrite::write_blocks_to_height(height, source); the v1 backend defers accumulator maintenance across the run and rebuilds it once at the tip; v0 loops
  write_block; sync_to_height delegates. DRYs write_block into write_block_with_options. Fixes: removes per-block accumulator random-reads during catch-up sync; the rebuild runs once at the tip (and only when blocks were
  actually written).

  2fb30f60 — 3-stage migration
  Splits the v1.1→v1.2 migration into Stage A (txid_location), Stage B (spent only), Stage C (bulk accumulator rebuild via the new builder). Fixes: removes the migration’s per-block accumulator fan-out. Stage C always
  recomputes from scratch, so a partially-run original (2-stage) migration is recomputed correctly, never corrupted; resumable via per-stage progress watermarks.

  58dd461d — remove validation from write path
  Replaces the post-commit validate_block_blocking re-read with two cheap in-memory checks (parent-hash continuity + merkle root) and advances validated_tip directly; full validation is now startup-only. Also fixes
  initial_block_scan to validate in ascending height order (was hash order → cache thrash + non-monotonic tip). Fixes: the per-block re-read re-faulted spent for every input; reads now hit the validated fast path and
  never re-validate after startup. (validated_tip is intentionally not persisted — an on-disk marker isn’t tamper-evident; startup remains the integrity gate.)

  df58450c — sorted batched index writes
  Sync (write_blocks_to_height) and migration Stage B now buffer work up to a configurable byte budget and write each batch in one transaction with the random-keyed spent/txid_location entries inserted in sorted key
  order. Fixes: the random-key index writes faulted a scattered B-tree leaf per insert once the DB exceeded RAM; sorted insertion turns that into a sequential sweep. New config storage.database.sync_write_batch_bytes
  (default 4 GiB). The address-history experimental feature keeps the per-block path (its prev-output resolution can’t see earlier-in-batch uncommitted blocks).
  
  88723ce — bound bulk-sync write batch by count/time; report in-flight progress
  Caps each bulk-sync write batch on the first of {byte budget, SYNC_WRITE_BATCH_MAX_BLOCKS = 100_000, SYNC_WRITE_BATCH_MAX_INTERVAL = 60s} instead of bytes alone, and logs the in-flight height (throttled to 10 s) plus
  the per-batch commit from inside write_blocks_to_height; removes the old sync_to_height reporter that polled the committed db tip. Fixes: on a fresh sync the early-chain blocks are tiny, so the byte-only budget
  buffered an enormous number of blocks before the first commit — delaying durability, inflating RAM, widening the crash-loss window, and making the reporter sit at height 0 (committed tip) so the sync looked hung. The
  count/time caps make the first commit prompt and bound memory; the new logging reports real progress (in-flight + committed). Crash-safety is unchanged: still one atomic txn + fsync per batch, resume from the committed
  tip.
  
  18ef025 — short-sync accumulator update path

  Replaces the unconditional from-genesis accumulator rebuild at the end of write_blocks_to_height with a watermark-driven dispatch: on a steady-state catch-up the accumulator is advanced by applying only the delta for
  the just-written range (DbV1::update_tx_out_set_accumulator_for_range), reading the stored accumulator + _tx_out_set_accumulator_built_height watermark, incorporating the range's created/spent outputs, and rewriting
  both in one txn + fsync. The full rebuild_tx_out_set_accumulator is kept only for the first build (no watermark) or an unusually large gap (tip − watermark > ACCUMULATOR_INCREMENTAL_MAX_GAP = 1_000, e.g. a sync
  interrupted far behind the on-disk tip).

  Fixes: the bulk rebuild added in 413d1ed re-scans the entire spent + transparent tables (a fixed chain-length cost, ~45 min on mainnet once the DB exceeds RAM). Because it ran at the end of every write_blocks_to_height
  call — and the sync loop calls that on every newly-finalised block — each new block triggered a fresh full-chain scan, so the node never reached the tip and stayed stuck "Syncing" (finalised tip pinned ~one rebuild
  behind the validator). The incremental path is O(range): for a steady-state gap of a few blocks it reads only those blocks plus a bounded number of point lookups, independent of chain length.

  Crash-safety & resumability

  Every batch commits atomically (block data + its sorted indexes, or migration spent entries + progress watermark) and fsyncs, so headers/the watermark never run ahead of the indexes. A crash rolls back to the last
  fsynced batch; sync resumes from the on-disk tip, the migration from its watermark. Re-done work is idempotent (NO_OVERWRITE + verify-match).

  Testing

  - New batched_sync_is_batch_size_independent: syncing the same chain as one huge batch vs one-block-per-batch yields an identical db tip, validated tip, and accumulator.
  - New bulk_tx_out_set_accumulator_builder_matches_incremental and write_path_advances_validated_tip.
  - Full zaino-state chain_index suite green (incl. migration crash-resume, accumulator/spent/validated-tip via the batch path).
  - proptest_blockgen::make_chain is #[ignore]d: it drives the indexer with partial_chain_strategy blocks (invalid merkle roots) that the finalised state now correctly rejects; MockchainSource tests cover the finalised
  state. Re-enable when the optional-db PR lands.

  Config

  - storage.database.sync_write_batch_bytes (default 4 GiB) — bulk-sync/migration write-batch budget; raise on large-RAM hosts (trades RAM + dirty pages against the page cache the sorted sweep needs).

  Known follow-ups (not in this PR)

  - Apply the sorted-batch treatment to migration Stage A (txid_location) — same random-write cliff as Stage B; and consider merging Stage A+B into one pass (Stage B no longer depends on txid_location).
  - Fully batch the address-history feature (resolve prev-outputs from the in-memory batch buffer).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this library adheres to Rust's notion of
   validator.
 
 ### Added
+- `storage.database.sync_write_batch_bytes` config (default 4 GiB) tunes the
+  finalised-state bulk-sync / migration write-batch size.
 - `zainod` gains an `allow_unencrypted_public_json_rpc_bind` build feature that
   lifts the new private-only JSON-RPC bind restriction for trusted
   private-network deployments (logs a `WARN` on startup when enabled).
@@ -24,6 +26,13 @@ and this library adheres to Rust's notion of
   `FinalisedTxOutSetInfoAccumulator` with the non-finalised state to produce
   the full `GetTxOutSetInfoResponse`.
 ### Changed
+- Finalised-state sync and the v1.1.0 -> v1.2.0 migration are substantially
+  faster on large/mainnet caches. The txout-set accumulator is built in bulk at
+  the tip instead of per block (removing an unbounded fan-out of random reads),
+  block validation is off the write path, and the random-keyed `spent` /
+  `txid_location` indexes are written in sorted batches — together removing the
+  random-fault stall around sandblast height. See the `zaino-state` changelog for
+  details; tune the write-batch size with `storage.database.sync_write_batch_bytes`.
 - The `zainod` JSON-RPC server now refuses to bind to public or unspecified
   (`0.0.0.0` / `::`) addresses by default; `check_config` enforces the same
   private/loopback rule already applied to gRPC. The unencrypted JSON-RPC
--- a/packages/zaino-common/CHANGELOG.md
+++ b/packages/zaino-common/CHANGELOG.md
@@ -8,6 +8,11 @@ and this library adheres to Rust's notion of
 ## [Unreleased]
 
 ### Added
+- `StorageConfig::database.sync_write_batch_bytes` (default 4 GiB) — byte budget
+  for the finalised-state bulk-sync / migration write batch. Larger batches
+  insert the random-keyed `spent` / `txid_location` indexes in bigger sorted
+  sweeps (fewer random B-tree faults once the DB exceeds RAM), at the cost of
+  more RAM; raise it on large-RAM hosts.
 ### Changed
 ### Deprecated
 ### Removed
--- a/packages/zaino-common/src/config/storage.rs
+++ b/packages/zaino-common/src/config/storage.rs
@@ -68,6 +68,23 @@ pub struct DatabaseConfig {
     /// Database size limit. Defaults to 128 GB.
     #[serde(default)]
     pub size: DatabaseSize,
+    /// Approximate in-memory byte budget for the finalised-state bulk-sync write batch.
+    ///
+    /// Bulk sync buffers fetched blocks up to this many bytes, then writes the whole batch in one
+    /// LMDB transaction with the random-keyed `spent` / `txid_location` entries inserted in **sorted**
+    /// key order. Sorting turns the random B-tree leaf faults (which dominate once the DB exceeds
+    /// RAM) into a sequential sweep; larger batches mean fewer sweeps.
+    ///
+    /// NOTE: peak RAM is roughly this budget (buffered blocks) plus the transaction's dirty pages,
+    /// and it competes with the OS page cache the sorted sweep relies on — larger is not always
+    /// better. Defaults to 4 GiB; raise it on large-RAM hosts.
+    #[serde(default = "default_sync_write_batch_bytes")]
+    pub sync_write_batch_bytes: u64,
+}
+
+/// Default [`DatabaseConfig::sync_write_batch_bytes`]: 4 GiB.
+fn default_sync_write_batch_bytes() -> u64 {
+    4 * 1024 * 1024 * 1024
 }
 
 impl Default for DatabaseConfig {
@@ -75,6 +92,7 @@ impl Default for DatabaseConfig {
         Self {
             path: resolve_path_with_xdg_cache_defaults("zaino"),
             size: DatabaseSize::default(),
+            sync_write_batch_bytes: default_sync_write_batch_bytes(),
         }
     }
 }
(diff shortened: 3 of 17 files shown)


zancas:
After ironing out bottlenecks due to DB read, dynamic batch-width, and inefficient LMDB config options, fetch/networking became the unambiguous chokepoint.

@nuttycom pointed out that the StateService zebrad_db backed data source:

(a)  is required by Zallet to access as a library
(b)  expands the source throughput for fetch, such that it is no longer binding.

Upon moving to the StateService, I discovered that the chokepoint on sync in the sandblast  that bound the 1214 state-service sync was [back pressure](https://github.com/zingolabs/zaino/pull/1214/changes/79eca190fc3aed47727f80bbfbf2e1e94f900fdc) from zebra streaming the full set of blocks for certain requests, coupled with a graceless shutdown, and insensitive detection on the zaino side.

We'll look into PRing against zebrad after testing the fix.

zancas (commented):
I'm filing follow up issues for non-blocking improvements.

zancas (approved):
Local unit and integration tests pass!

zancas (approved):
I am merging based on manual test of send and receive with zkool.

PR #1239, opened 2026-06-16
Add liveness and readiness HTTP endpoints

emersonian:
For container orchestrators like Kubernetes to know if Zaino is alive and synced, ready for end-user traffic.

- New opt-in HTTP server on a dedicated port (health.listen_addr, disabled by default).
- Readiness is decided by proximity to the chain tip (blocks-behind plus tip-age), like Zebra. This avoids ready-state flapping when the chain isn't perfectly at the tip.
- Endpoint names (/livez, /readyz) follow the Kubernetes API-server convention; configuration style ([health] section) mirrors Zebra.

Reference zcash-stack for examples of how we use this in production for Zec.rocks:

- Zebra: https://github.com/zecrocks/zcash-stack/blob/main/charts/zcash-stack/templates/zebra-statefulset.yaml#L128 (livenessProbe, readinessProbe)
- Lightwalletd: https://github.com/zecrocks/zcash-stack/blob/main/charts/zcash-stack/templates/lightwalletd-statefulset.yaml#L121

Related:
- https://github.com/zcash/wallet/issues/366


zancas:
Moving this to draft until we address https://github.com/zingolabs/zaino/issues/1311

zancas:
Thank you for this work, Emerson. I propose that we close this PR. The feature
continues in the runtime design of #1550.

This PR cannot be brought up to date. Its base is from 2026-06-12, and `dev`
has more than 1,000 commits since then. The PR changes `backends/fetch.rs` and
`backends/state.rs`, and `dev` no longer has those files. One
`NodeBackedIndexerService` replaced `FetchService` and `StateService`. The
direct read-state path that `StateService` used is deprecated, and we will
delete it.

#1550 contains a `HealthServer` in `zaino-runtime`. It serves `/livez`,
`/readyz`, and `/startupz` as a supervised component. It answers from the
statuses of the components.

Your readiness rule is the part that #1550 did not have. Draft PR #1598, on
top of #1550, adds `max_blocks_behind` to `ReadinessCriteria`, with your
default of two blocks. The bound on the age of the tip does not move yet. A
note on #1550 asks for your `[health]` configuration section, for the start of
the server from `zainod`, and for a decision on the age of the tip. The
zcash-stack probe examples stay our reference.

PR #1241, opened 2026-06-17
perf(zaino-state): Concurrent block fetches

emersonian:
This PR pipelines the finalized bulk-sync fetch path: blocks are fetched concurrently and committed strictly in height order. It switches the validator RPC client to HTTP/2 (prior-knowledge h2c) so a single connection multiplexes all in-flight requests, removing the connection-per-request churn the new concurrency would otherwise put on Zebra, and retries transient connection errors (with exponential backoff in the sync loop).

Background: write_blocks_to_height was serial and fetch-latency-bound: each block awaited two sequential RPCs (get_block + get_commitment_tree_roots) with zero overlap, so at the sandblast band we observed throughput collapse to ~1 blk/s while both Zaino and Zebra sat practically CPU-idle.

This update is an attempt to help Zaino sync past the sandblasted blocks faster.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3430,6 +3430,7 @@ dependencies = [
  "cookie",
  "cookie_store",
  "futures-core",
+ "h2",
  "http",
  "http-body",
  "http-body-util",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,6 +71,7 @@ url = "2.5"
 reqwest = { version = "0.12", default-features = false, features = [
     "cookies",
     "rustls-tls",
+    "http2",
 ] }
 tower = { version = "0.4", features = ["buffer", "util"] }
 tonic = { version = "0.14" }
--- a/docs/example_configs/zainod.toml
+++ b/docs/example_configs/zainod.toml
@@ -33,3 +33,21 @@ shard_power = 4
 [storage.database]
 path = '/home/cupress/.cache/zaino'
 size = 128
+
+# Approximate in-memory byte budget for the finalised-state bulk-sync write batch.
+# Bulk sync buffers fetched blocks up to this many bytes, then writes the whole batch in one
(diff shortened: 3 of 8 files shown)


zancas:
@emersonian I am considering redirecting this to dev, opinions.

emersonian:
It sped things up overall in my testing and I think is similar to our [sync speedup in our lightwalletd fork](https://github.com/zecrocks/lightwalletd/commit/b6e0c4f88728a64054e611d8e444f639fdd0ea63), but it did not help much through the sandblasting period.

emersonian:
If anything, HTTP/2 should definitely be merged

AloeareV:
> @emersonian I am considering redirecting this to dev, opinions.

I am inclined to agree this should be moved to dev.

zancas:
Thank you for this work, Emerson. We are closing this PR. Two of its three
changes are on `dev` now, and #1599 carries the third change.

This PR targets `rc_0_4_0_slow_sync_plus_updates`, which is no longer a release
branch in use. It cannot move to `dev`. `dev` has more than 1,100 commits since
the base of this PR, and `zaino-fetch/src/jsonrpsee/connector.rs` no longer
exists there.

- Concurrent fetches with commits in height order are on `dev`. See the
  `buffered` stream in
  `zaino-chain-store-zainodb/src/store/finalised_source/v1/write_core.rs`.
- The retry of connection errors with exponential backoff is on `dev`. See
  `zaino-source/src/validator_client.rs`.
- HTTP/2 is not on `dev`. You wrote that it should merge, and we agree. Draft
  PR #1599 adds it to `zaino-rpc` as an option, with your adaptive window
  setting.

#1599 keeps HTTP/1.1 as the default, because a client with prior knowledge
fails against an endpoint that speaks only HTTP/1.1.

PR #1242, opened 2026-06-17
Add transaction and shielded-action sync throughput metrics

nachog00:
## Summary

Sync speed has so far only been observable in **blocks/sec**, which understates indexing work: early-chain blocks are near-empty while post-NU5 blocks can carry thousands of note commitments each. This adds workload-based throughput counters so sync speed can be read in transactions/sec and shielded-actions/sec.

Builds on the `/metrics` endpoint from #1216.

## Metrics added

Three monotonic counters, emitted from the `sync_to_height` loop behind the existing `prometheus` feature (no-op when the feature is off):

| Metric (scraped name) | Meaning |
|---|---|
| `zaino_sync_transactions_total` | Total transactions indexed during sync |
| `zaino_sync_sapling_outputs_total` | Total Sapling outputs indexed during sync |
| `zaino_sync_orchard_actions_total` | Total Orchard actions indexed during sync |

Example PromQL:

```promql
rate(zaino_sync_transactions_total[5m])                                            # tx/sec
rate(zaino_sync_sapling_outputs_total[5m]) + rate(zaino_sync_orchard_actions_total[5m])  # shielded actions/sec
```

## Notes

- **No observer-effect cost on the hot path.** Counts are read from data already materialized in the `IndexedBlock` by the preceding `IndexedBlock::try_from` step — just a walk over in-memory vectors and a few integer adds. No I/O, no parsing, no allocation.
- Sapling outputs and Orchard actions are kept as **separate** counters (sum in PromQL for total shielded throughput) rather than conflated under one ambiguous "actions" name.
- Descriptions registered alongside the existing gauges in `zainod`'s `describe_metrics`.
- Verified `cargo check` clean both with `--features prometheus` and in the default (feature-off) build.

A follow-up Grafana dashboard update will add tx/sec and actions/sec panels.

### DIFF
--- a/packages/zaino-state/src/chain_index/finalised_state.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state.rs
@@ -654,6 +654,8 @@ impl ZainoDB {
                 };
                 parent_chainwork = chain_block.context.chainwork;
 
+                record_block_throughput(&chain_block);
+
                 self.write_block(chain_block).await?;
             }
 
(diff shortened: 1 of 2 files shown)


zancas (approved):
Awesome!

PR #1250, opened 2026-06-18
Finalised state passthrough conflicts resolved

zancas:
Merges in dev+backports from 0.4.0 and fixes all merge conflicts.   The original #1200 is still in place and can be reopened if this proposed fix is not workeable.


idky137 (approved):
Thanks for dealing with all the conflicts! Looks good!

PR #1252, opened 2026-06-18
Zcashd support

zancas:
This PR changes the CONTRACT for:

**makers container-test**

and

**makers integration-test** they no longer run zcashd by default!

To run tests INCLUDING zcashd tests use:

**makers zcashd_test**


nachog00:
Why adding a new interface? couldn't we keep the "run integration tests" command unified and have it accept an optional flag like `--with-zcashd` or something like that?

zancas:
Because developers hate adding a single flag, or changing a single invocation.  This makes it just work without human intervention.. the way it always has.

And we're NOW in the situation where testing against zcashd is not the default flow.. it's the special case (driven in part by the 2x runtime overhead).

nachog00:
If the `--with-zcashd` flag defaults to false, existing calls to `makers integration-tests` would still work just like before. Moreover, future improvements/changes to the integration tests run logic wouldn't need to be propagated to the new integration tests running command.

zancas:
Fine..   please push a commit with the proposed change.  I like it.  It's DRYer.

nachog00:
Done. Now we need a new reviewer.

zancas (commented):
UtAck..  nice upgrade @nachog00 !

AloeareV (approved):

PR #1256, opened 2026-06-18
Update to latest zebrad library releases.

nuttycom:
Zebra just made new releases of the zebra-chain, zebra-state, and zebra-rpc crates; we need to adapt to those new releases. This PR is in draft because there are version conflicts related to how zingolib_testutils also brings in zebra crates; please somebody take it over and make the needed fixes?

PR #1259, opened 2026-06-18
excise zingolib, replace with devtool

### LINKED ISSUE #249: Issues encountered while integrating Zaino into Zallet

zancas:
Unblocks: #1256 

The dependency on zingolib caused dependency clashes in upstream and forced an inefficient upgrade cascade in developers' hot-development-loops.

This PR completely removes zingolib, and zingolib-testutils.

The functionality that was provided is now replaced with zcash-devtool.

This port does NOT provide 100% test parity.  See the table below for outstanding tests which I am actively porting.

This test set should be marked as a known issue, until I finish the port.  I don't believe it should block upgrade, or merge of this PR.

To confirm excision:

```
cd integration-tests && cargo tree -i zingolib_testutils; cargo tree -i zingolib
```

The somewhat awkward comprehensive test invocation
```
makers container-test && makers integration-test --with-zcashd
```

## All 9 `devtool-incompatible`-gated instances

5 distinct tests across 9 gated instances.

| # | Test | Reason / method of exclusion | Location |
|---:|---|---|---|
| 1 | `address_deltas` | `devtool-incompatible` gate — devtool shield coinbase-maturity off-by-one | `integration-tests/wallet-tests/tests/devtool.rs:2142` (`mod zebrad`) |
| 2 | `send_to_transparent_finalization` | `devtool-incompatible` gate — heavy: ~99 halo2 proofs; needs round-3 P2 transparent filler | `integration-tests/wallet-tests/tests/devtool.rs:1977` (`mod fetch_service`, `FetchService`) |
| 3 | `send_to_transparent_finalization` | `devtool-incompatible` gate — heavy: ~99 halo2 proofs; needs round-3 P2 transparent filler | `integration-tests/wallet-tests/tests/devtool.rs:2202` (`mod state_service`, `StateService`) |
| 4 | `monitor_unverified_mempool` | `devtool-incompatible` gate — devtool `WalletBalance` lacks `unconfirmed_*` / `confirmed_*` fields; balance asserts commented out | `integration-tests/wallet-tests/tests/devtool.rs:2008` (`mod fetch_service`, `FetchService`) |
| 5 | `monitor_unverified_mempool` | `devtool-incompatible` gate — devtool `WalletBalance` lacks `unconfirmed_*` / `confirmed_*` fields; balance asserts commented out | `integration-tests/wallet-tests/tests/devtool.rs:2223` (`mod state_service`, `StateService`) |
| 6 | `create_200_block_regtest_chain_vectors` | `devtool-incompatible` gate — vector builder funds via transparent-coinbase shielding (round-2 P1); same shield gap as `address_deltas` | `integration-tests/wallet-tests/tests/test_vectors.rs:41` |
| 7 | `send_to_transparent_finalization` | `devtool-incompatible` gate + `zcashd_support` crate gate — heavy: ~99 halo2 proofs; needs round-3 P2 | `integration-tests/wallet-tests/tests/devtool_zcashd.rs:542` (`mod wallet_to_validator`) |
| 8 | `send_to_all` | `devtool-incompatible` gate + `zcashd_support` crate gate — heavy: ~100 halo2 proofs; needs round-3 P2 | `integration-tests/wallet-tests/tests/devtool_zcashd.rs:590` (`mod wallet_to_validator`) |
| 9 | `monitor_unverified_mempool` | `devtool-incompatible` gate + `zcashd_support` crate gate — `WalletBalance` lacks `unconfirmed_*` / `confirmed_*` fields | `integration-tests/wallet-tests/tests/devtool_zcashd.rs:631` (`mod wallet_to_validator`) |

### Distinct tests

| Test | Gated instances | Files |
|---|---:|---|
| `address_deltas` | 1 | `devtool.rs` |
| `send_to_transparent_finalization` | 3 | `devtool.rs`, `devtool_zcashd.rs` |
| `monitor_unverified_mempool` | 3 | `devtool.rs`, `devtool_zcashd.rs` |
| `create_200_block_regtest_chain_vectors` | 1 | `test_vectors.rs` |
| `send_to_all` | 1 | `devtool_zcashd.rs` |

### Feature-gate notes

The last three instances in `devtool_zcashd.rs` carry the `devtool-incompatible` gate and sit inside the `#![cfg(feature = "zcashd_support")]` binary. They need `--with-zcashd` to compile and `--features devtool-incompatible` to un-ignore.

The first six instances need only `--features devtool-incompatible`.


integration-tests/wallet-tests/src/devtool.rs
```diff
@@ -0,0 +1,241 @@
...
+}
+
+/// Convert a devtool txid — the hex string `send`/`shield` return, which
+/// devtool prints in display (reversed) order via `TxId`'s `Display` — into
+/// the internal-order 32 bytes that zaino's `TxFilter` and compact-tx
+/// comparisons use (the order zingolib's `TxId::as_ref()` yields). Any test
+/// that then queries zaino with the result validates the order: a wrong one
+/// simply fails to match the indexed transaction.
+pub fn txid_internal_bytes(devtool_txid_hex: &str) -> Vec<u8> {
+    let mut bytes = hex::decode(devtool_txid_hex.trim()).expect("devtool txid is valid hex");
```
pacu, line 68: I bet there's a better way to do this that's already implemented

integration-tests/wallet-tests/src/devtool.rs
```diff
@@ -0,0 +1,241 @@
...
+        self.sync_recipient().await;
+    }
+}
+
+impl Pool {
+    /// The spendable balance received in this pool, in zatoshis — the
+    /// devtool-backend counterpart of [`Pool::received_balance`]. Spendable
+    /// equals received once the funding transaction is mined and the wallet
+    /// synced, which is the state every asserting test establishes first.
+    pub fn spendable_balance(self, balance: &WalletBalance) -> u64 {
```
pacu, line 234: Why is this not Zatoshi  and leave conversion to primitive value left for the caller?

nachog00 (approved):
I would prefer this work to not tie everything so much to the specific devtool implementation detail, but to a general client. But this work is valuable in it of itself.

LGTM!

zancas:
Upon investigation I reproduced the CI failure in a zaino-free environment:

https://github.com/zcash/integration-tests/issues/136

I hope to circle back and fix...

pacu (changes requested):
Requesting reviewing a small check. The hand-crafted TxId reversing is a recurring solution we encounter that when patched, brings more problems that it actually solves. Please double check that and seek to reuse proven solutions either in Zaino itself or librustzcassh

Issue #1261, opened 2026-06-18
Sync progress lost on restart in 0.5.0-rc.1 (background sync persistence regression)


nachog00:
## Summary

In `0.5.0-rc.1`, zaino loses nearly all sync progress when the process restarts. After syncing millions of blocks, a restart causes it to resume from ~101k (the golden snapshot baseline) instead of from where it left off. This is a regression from `rc/0.4.0` / PR #1238, where restarts correctly resume near the last committed height.

Related: #1260 (OOM during accumulator rebuild) — these two bugs compound: the OOM crashes the process, and this bug ensures no progress is retained across those crashes.

## Environment

- **Zaino**: `0.5.0-rc.1` (regression), vs PR #1238 / `rc_0_4_0_slow_sync_plus_updates` (works correctly)
- **Zebra**: 5.1.0
- **Network**: mainnet (chain tip ~3,382,500)
- **Golden snapshot baseline**: ~101,162 blocks

### 0.5.0-rc.1 (broken)
After each OOM restart (45 so far), zaino re-syncs from height ~101k:
```
zainod v0.4.0
write_blocks_to_height: syncing finalised blocks 101162..=3382533 on Mainnet
write_blocks_to_height: syncing height 103361 / 3382528 on Mainnet
write_blocks_to_height: syncing height 102154 / 3382528 on Mainnet
```
Every restart shows the same `101162..=3382533` range — no progress is retained.

### PR #1238 (works correctly)
After each OOM restart, zaino resumes from near chain tip:
```
write_blocks_to_height: syncing finalised blocks 3382524..=3382530 on Mainnet
write_blocks_to_height: committed batch to height 3382530 (7 blocks)
```
Only the last few blocks need re-syncing — LMDB state is properly persisted.

## Root cause analysis

The regression was introduced by PRs #1249 and #1250, which added a background sync architecture with ephemeral passthrough (`EphemeralMode`/`EphemeralReference` in `router.rs`).

When `sync_to_height` encounters a range larger than `LONG_RUNNING_SYNC_THRESHOLD`, it spawns the sync as a background task and serves reads from a validator passthrough. In PR #1238 (without this architecture), sync writes happen inline/synchronously, so each LMDB batch commit is durable and survives crashes.

With the background sync approach, when the process is killed (e.g., OOM), the spawned task's uncommitted work is lost. On restart, the persistent DB only has data up to ~101k (the snapshot baseline), because the background task hadn't committed its batches to durable storage.

[...]

Issue #1260, opened 2026-06-18
OOM during txout-set accumulator rebuild at chain tip (>16 GiB)


nachog00:
## Summary

Zaino consistently OOM-crashes during the txout-set accumulator rebuild when it reaches chain tip on mainnet. This affects both `rc_0_4_0_slow_sync_plus_updates` (PR #1238) and `0.5.0-rc.1`.

## Environment

- **Zaino versions**: PR #1238 (`1f06894`) and `0.5.0-rc.1`
- **Zebra**: 5.1.0
- **Network**: mainnet (chain tip ~3,382,500)
- **Memory limit**: 16 GiB (k8s pod limit)
- **LMDB map size**: 96 GiB
- **Cargo features**: `no_tls_with_prometheus` (0.5.0-rc.1), `no_tls_use_unencrypted_traffic` (PR 1238)

## Observed behavior

Both versions successfully sync all finalized blocks from height 0 to chain tip (~3.38M). Immediately after committing the final batch, the process begins "rebuilding txout-set accumulator" and memory usage exceeds 16 GiB, triggering OOMKill (exit code 137).

### PR 1238 logs (before crash)
```
write_blocks_to_height: committed batch to height 3382530 (7 blocks)
write_blocks_to_height: rebuilding txout-set accumulator to height 3382530
```
→ OOMKilled

### 0.5.0-rc.1 logs (before crash)
Same pattern — reaches tip, starts accumulator rebuild, dies.

### Crash loop

- PR 1238: **28 restarts** — but resumes near tip each time (LMDB state persisted), so only re-syncs last few blocks before hitting the accumulator rebuild again
- 0.5.0-rc.1: **45 restarts** — compounded by #1261, re-syncs from ~101k every time

## Expected behavior

The txout-set accumulator rebuild at chain tip should complete within 16 GiB. If it genuinely needs more memory, the required amount should be documented.

## Reproduction

```bash
# Deploy from golden mainnet snapshot (zebra synced to tip)
# with 16Gi memory limit on zaino pod
argo submit --from workflowtemplate/deploy-ephemeral -n argo \
  -p namespace=oom-repro \
  -p ref=0.5.0-rc.1 \
  -p cargo-features=no_tls_with_prometheus \
  -p zebra-tag=5.1.0
```

Wait for sync to reach chain tip — OOM occurs during accumulator rebuild.

idky137:
@nachog00 are we still getting this error or can we close this issue?

PR #1263, opened 2026-06-19
RC-0.5.0 bug fixes

idky137:
Contains fixes for bugs/issues found in release testing of the v0.5.0 release candidate, detailed by commit / issue:
---
**Sync progress lost on restart - #1261**:

- 82aa6d03 — Restrict the finalised-state sync to a single running task

The ChainIndex worker calls sync_to_height on every poll (~500 ms), and for long ranges that method spawns the sync as a detached task and returns immediately. With no in-flight guard, each poll re-spawned another background sync, piling up concurrent write_blocks_to_height runs that contended on the single LMDB writer and multiplied memory (concurrent block buffering plus repeated full accumulator rebuilds) — driving the process to OOM before any batch committed, so on restart the persistent tip was still at the old snapshot baseline and sync started over. The fix adds a single-flight check at the top of sync_to_height using the existing Router::has_background_ops() tracker (incremented synchronously before the spawn, held for the task's whole lifetime): if a sync is already in progress, the poll is a no-op. The worker being the sole sequential caller makes the check race-free.

- 7e78a6f9 — Seed the non-finalised state no deeper than chain_height − NON_FINALIZED_DEPTH

While the finalised DB was still catching up, seeding the non-finalised state (NFS) called get_chain_block_by_height(tip − NON_FINALIZED_DEPTH), which returned Ok(None) and was .expect("todo") unwrapped down to genesis. sync() then re-anchored to the lagging finalised tip and walked every block up to the chain tip one at a time, never converging. The fix adds resolve_anchor_block, which fetches the seed block reader-first (finalised DB, or the validator via the ReadOnly passthrough during catch-up) with a direct-from-validator fallback, so the anchor never drops to genesis. Both the init and re-anchor paths now floor the anchor at max(finalised_tip, chain_height − NON_FINALIZED_DEPTH), keeping the NFS to the reorg window (~101 blocks). The two .expect("todo") calls were replaced with proper error propagation.

  ---
**OOM on TxOutSetInfo accumulator build - #1260**:

- 8f4c2348 — Add sharding to the txout-set accumulator build for low-memory systems

At chain tip the from-genesis accumulator rebuild was pinned to ACCUMULATOR_BUILD_SHARDS = 1, loading the entire spent table into one in-memory HashSet — which exceeds 16 GiB at mainnet tip and OOM-crashed (exit 137). The fix enables the builder's existing sequential, zero-random-read sharding: the spent set is partitioned by the creating-txid's first byte into shards contiguous key ranges, each range is accumulated independently, and the results are recombined. Because the accumulator is an XOR multiset plus additive counters, partitioning is order-independent and result-identical, and the accumulator struct is fixed-size — so only the per-shard spent set scales with shard count, and capping it caps RSS. (Batching via the incremental path was rejected: its O(range) random lookups reintroduce the sandblast-era sync slowdown.)

- f56ee9a8 — Separate the block-batch budget from the accumulator-build budget

A single sync_write_batch_size config sized both the bulk-sync block buffer and the accumulator spent set, so they couldn't be tuned independently — and because DatabaseConfig was #[serde(default)] with no deny_unknown_fields, a config still using the old sync_write_batch_bytes key was silently ignored and fell back to the large default, defeating the budget on exactly the hosts that needed it. The fix adds a dedicated accumulator_rebuild_memory_size config (the rebuild uses this, never the block-batch budget) and adds #[serde(deny_unknown_fields)] so any stale or unknown key now fails loud at startup instead of silently reverting to defaults. Both budgets default to 8 GiB and sync_checkpoint_interval defaults to 120 s, with saturating byte-count conversion. The example config and changelogs are updated; new tests confirm the old key is rejected and the new keys parse.

- baf9a6a6 — Make accumulator sharding a hard memory cap

Shard count was derived from an estimate of spent-set bytes, so if the estimate was low (or one first-byte bucket was unusually dense) a shard could still exceed the budget and OOM — the cap wasn't actually guaranteed. The fix replaces the estimate with a strict hard cap: each shard loads spent entries via a range-seek and aborts the moment it hits max_spent_entries (returning None), and the work-list then bisects that range and retries, so no shard ever exceeds the budget regardless of key distribution. If a single first-byte bucket can't be split below the cap it fails with an actionable error instead of OOMing. A new combine method on the accumulator (with overflow-checked counter sums) plus unit tests prove combine is order-independent and correctly recombines partitioned outputs.

- 26c20744 — Diagnostics, durability docs, and tests for the spent scan

On a torn or corrupt spent page the startup scan aborted with a cryptic "unsupported Zaino version tag 253" and no context, several scans used LMDB cursor iterators that can panic or silently truncate on edge cases, and doc comments wrongly claimed the NO_SYNC env "never corrupts" on a crash (false on NFS / evicted containers where write ordering isn't preserved). The fix rewrites initial_spent_scan (and the estimator and per-shard loads) as explicit, panic-safe cursor walks that propagate every LMDB error, and on a decode failure reports the key (hex), value length, and value head plus a "wipe and re-index" instruction — turning "tag 253" into a definitive torn-page diagnosis. It also adds phase logging around the full-table scans so a native abort in stderr is attributable, corrects the misleading durability comments, and adds unit and end-to-end tests (including one that injects a corrupt spent value and asserts the diagnostic error).

  ---

### DIFF
--- a/packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs
@@ -177,7 +177,7 @@ pub(crate) const ACCUMULATOR_INCREMENTAL_MAX_GAP: u32 = 1_000;
 /// independent of the shard count.
 ///
 /// The shard count is chosen at rebuild time so the per-shard spent set fits the configured
-/// [`zaino_common::DatabaseConfig::sync_write_batch_size`] budget (see
+/// [`zaino_common::DatabaseConfig::accumulator_rebuild_memory_size`] budget (see
 /// [`DbV1::rebuild_tx_out_set_accumulator`]): a single optimal pass on hosts with enough RAM for
 /// the full spent set, scaling up on memory-constrained deployments. Sharding partitions on the
 /// creating-txid's first byte (256 distinct values), so the count is capped here.
--- a/packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs
@@ -184,9 +184,14 @@ pub(crate) const ACCUMULATOR_INCREMENTAL_MAX_GAP: u32 = 1_000;
 pub(crate) const ACCUMULATOR_BUILD_MAX_SHARDS: u16 = 256;
 
 /// Conservative per-entry RAM estimate for the rebuild's in-memory spent set, used only to size the
-/// shard count. Each entry is a boxed `spent` key (~37 bytes) plus `HashSet` bucket overhead; an
-/// over-estimate just adds shards (less memory), never under-bounds.
-pub(crate) const SPENT_SET_ENTRY_BYTES_ESTIMATE: u64 = 128;
+/// shard count.
+///
+/// Each entry is a 37-byte `spent` key heap-allocated as a `Box<[u8]>` (rounded up by the
+/// allocator), a 16-byte fat pointer stored in the `HashSet` table, plus hashbrown control bytes and
+/// load-factor slack — realistically ~120 bytes, but allocator behaviour varies. Set deliberately
+/// *above* that so the chosen shard count over-provisions: the per-shard set then stays within the
+/// budget, and over-counting only adds shards (less memory per shard), it never under-bounds.
+pub(crate) const SPENT_SET_ENTRY_BYTES_ESTIMATE: u64 = 256;
 
 /// Number of committed block writes / migration heights between explicit
 /// `env.sync(true)` durability checkpoints.
--- a/packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs
@@ -196,15 +196,54 @@ pub(crate) const SPENT_SET_ENTRY_BYTES_ESTIMATE: u64 = 256;
 /// Number of committed block writes / migration heights between explicit
 /// `env.sync(true)` durability checkpoints.
 ///
+/// This governs the durability-sync cadence of the **per-block steady-state append**
+/// ([`DbWrite::write_block`]) and the **migration backfill scan**: both commit frequently but, under
+/// `MDB_NOSYNC`, force an `env.sync(true)` only every `SYNC_CHECKPOINT_INTERVAL` committed
+/// writes/heights. (The separate **bulk catch-up** path batches differently — by the
+/// `sync_write_batch_size` byte budget, a block-count cap, and the wall-clock
+/// `DatabaseConfig::sync_checkpoint_interval` — and fsyncs once per committed batch, so it does not
+/// use this constant.)
+///
 /// The LMDB environment is opened with `MDB_NOSYNC` (see [`DbV1::spawn`]), so an individual
-/// `txn.commit()` is *not* flushed to disk. Because the environment does not use `WRITE_MAP`,
-/// LMDB still guarantees ACI on crash — only durability (D) is lost: a crash rolls the database
-/// back to the last on-disk-consistent transaction, it never corrupts it (copy-on-write + dual
-/// meta pages always leave a recoverable committed snapshot). Forcing a sync every
-/// `SYNC_CHECKPOINT_INTERVAL` writes bounds how much committed-but-unflushed tail a crash can
-/// discard. The tail is always safe to re-do: clean sync resumes from the on-disk tip and
-/// re-fetches the missing blocks, and migrations resume idempotently from their progress keys.
+/// `txn.commit()` is *not* flushed to disk. On a **write-order-preserving local filesystem** this
+/// only costs durability (D): LMDB's copy-on-write + dual meta pages mean a crash rolls back to the
+/// last fully-flushed transaction without corrupting the database, and the checkpoint cadence bounds
+/// how much committed-but-unflushed tail a crash can discard. The tail is always safe to re-do:
+/// clean sync resumes from the on-disk tip and re-fetches the missing blocks, and migrations resume
+/// idempotently from their progress keys.
+///
+/// CAVEAT: that integrity guarantee relies on the filesystem preserving write order. On networked
+/// storage (NFS), overlay filesystems, or a container/pod hard-eviction that drops the unflushed
+/// page cache, write order is *not* guaranteed and a crash under `MDB_NOSYNC` **can** leave torn
+/// pages — surfacing later as an LMDB cursor assertion or a checksum/decode failure on the affected
+/// table. The recovery is to wipe the finalised-state DB and re-index (its tables are all
+/// re-derivable from the validator). A shorter checkpoint interval shrinks, but does not eliminate,
+/// this window.
 pub(crate) const SYNC_CHECKPOINT_INTERVAL: u32 = 1000;
+
+/// Formats a one-line corruption report for a malformed `spent` entry: its position in the table,
+/// the key (hex), the value length, and the value's leading bytes (hex), plus the re-index hint.
+///
+/// The leading bytes are the diagnostic that distinguishes failure modes: a value that should begin
+/// with the `StoredEntryFixed` version tag but starts with arbitrary bytes (e.g. `0xfd`) is a torn
+/// write, whereas a recognisable-but-old structure would indicate a format problem.
+fn spent_corruption_detail(
+    entry_index: u64,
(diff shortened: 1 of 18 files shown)


packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs
```diff
@@ -636,26 +693,62 @@ impl DbV1 {
...
+                    StoredEntryFixed::<TxLocation>::from_bytes(val_bytes).map_err(|error| {
+                        FinalisedStateError::Custom(spent_corruption_detail(
+                            entry_index,
+                            key_bytes,
+                            val_bytes,
+                            &error.to_string(),
+                        ))
+                    })?;
 
                 if !entry.verify(key_bytes) {
```
nachog00, line 739: ### Log `error!` at corruption detection site

The diagnostic from `spent_corruption_detail` is packed into `Err(FinalisedStateError::Custom(...))` and only surfaces when the caller logs it. If the process dies between detection and that caller log (OOM, signal, etc.), the diagnostic is lost — which is exactly the scenario this code is designed for.

Suggestion — emit a structured `error!` immediately, in both the `from_bytes` and `verify` paths:

```rust
let entry =
    StoredEntryFixed::<TxLocation>::from_bytes(val_bytes).map_err(|error| {
        let detail = spent_corruption_detail(entry_index, key_bytes, val_bytes, &error.to_string());
        error!(
            entry_index,
            key = %hex::encode(key_bytes),
            value_len = val_bytes.len(),
            reason = %error,
            "corrupt spent entry detected — wipe database and re-index"
        );
        FinalisedStateError::Custom(detail)
    })?;
```

Same for the checksum mismatch path below. Structured fields make each value independently queryable in Loki (`{reason="checksum mismatch"}` or `entry_index > 1000`).

packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1/transparent_address_history.rs
```diff
@@ -1745,184 +1774,305 @@ impl DbV1 {
...
+                    .map_err(|error| FinalisedStateError::Custom(error.to_string()))?,
+                None => {
+                    if hi - lo <= 1 {
+                        // A single creating-txid first-byte value cannot be split further. Fail with
+                        // an actionable error rather than OOM-ing: the spent outpoints sharing this
+                        // first byte do not fit the configured budget.
+                        return Err(FinalisedStateError::Custom(format!(
+                            "txout-set accumulator: spent shard for creating-txid first-byte {lo} \
+                             exceeds the per-shard budget ({max_spent_entries} outpoints) and cannot \
+                             be split further; raise accumulator_rebuild_memory_size"
```
nachog00, line 1900: ### Budget-exceeded error: `error!` before returning

This is an actionable operator intervention ("raise accumulator_rebuild_memory_size"), but it is only returned as `Err`, never explicitly logged. Same risk as the corruption path: the `?` chain may not reach a logger before the process exits.

```rust
if hi - lo <= 1 {
    error!(
        first_byte = lo,
        max_spent_entries,
        "txout-set accumulator shard cannot be split further — raise accumulator_rebuild_memory_size"
    );
    return Err(FinalisedStateError::Custom(format!(...)));
}
```

packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1/transparent_address_history.rs
```diff
@@ -1745,184 +1774,305 @@ impl DbV1 {
...
+                        )));
                     }
-                    spent_set.insert(Box::from(key_bytes));
+                    let mid = lo + (hi - lo) / 2;
+                    pending.push((lo, mid));
+                    pending.push((mid, hi));
```
nachog00, line 1905: ### Silent shard bisections — add `debug!` for visibility

When a shard exceeds `max_spent_entries` and gets bisected, it is completely silent. If the memory estimate is off and the builder does 50 bisections, that is a significant slowdown with zero visibility.

```rust
None => {
    if hi - lo <= 1 { /* ... error path above ... */ }
    let mid = lo + (hi - lo) / 2;
    debug!(
        lo,
        hi,
        mid,
        max_spent_entries,
        "accumulator shard exceeded budget, bisecting"
    );
    pending.push((lo, mid));
    pending.push((mid, hi));
}
```

packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1/transparent_address_history.rs
```diff
@@ -1715,9 +1715,38 @@ impl DbV1 {
...
+            db_tip.0
+        );
+        // `shards` is the *initial* partition (a good first guess from the memory estimate);
+        // `max_spent_entries` is the *hard* per-shard cap the builder enforces during the spent
+        // load, bisecting any shard that would exceed it. So the estimate only affects how many
+        // passes we make, never whether we stay within budget.
+        let shards = self.accumulator_build_shards(budget)?;
+        let max_spent_entries = (budget / SPENT_SET_ENTRY_BYTES_ESTIMATE).max(1);
+        info!(
+            "txout-set accumulator rebuild to height {}: {shards} initial shard(s), \
```
nachog00, line 1742: ### Use structured tracing fields on new `info!` logs

The new `info!` calls interpolate values into format strings, making them unqueryable in structured logging backends (Loki/Grafana). Since these are new lines:

```rust
// Instead of string interpolation:
info!(
    "txout-set accumulator rebuild to height {}: {shards} initial shard(s), ...",
    db_tip.0
);

// Prefer structured fields:
info!(
    height = db_tip.0,
    initial_shards = shards,
    max_spent_per_shard = max_spent_entries,
    budget_bytes = budget,
    "txout-set accumulator rebuild: shards computed"
);
```

The pre-sizing breadcrumb (L1730) could also be `debug!` since the post-sizing log carries a superset of its information.

Same applies to the integrity check completion in v1.rs:
```rust
// Instead of:
info!("finalised-state spent table integrity check passed ({entry_index} entries)");
// Prefer:
info!(entries = entry_index, "finalised-state spent table integrity check passed");
```

nachog00:
### LMDB assertion crash reproduced on `26c20744`

Both ephemeral deployments (`preview-1263-26c20744-2g` and `preview-1263-26c20744-8g`) crashed during initial sync with the same LMDB assertion failure seen on the previous commit (`8f4c234`):

```
/usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lmdb-sys-0.8.0/lmdb/libraries/liblmdb:5800:
Assertion 'IS_BRANCH(mc->mc_pg[mc->mc_top])' failed in mdb_cursor_sibling()
```

Exit code 139 (SIGSEGV) on both variants (batch size 2 and 8). Both pods restarted and are currently running.

Grafana panel at the time of crash: https://grafana.vaquita-altair.ts.net/goto/ffq10zsbnfvggb?orgId=1

This confirms the new commits (separated batch/accumulator sizes, accumulator sharding, better error logging) did not resolve the underlying LMDB torn-page corruption. The C-level assertion fires inside `mdb_cursor_sibling()` before any Rust error handling can intercept it — no structured error log is emitted, only the stderr assertion message visible via `kubectl logs --previous`.

idky137:
> ### LMDB assertion crash reproduced on `26c20744`
> Both ephemeral deployments (`preview-1263-26c20744-2g` and `preview-1263-26c20744-8g`) crashed during initial sync with the same LMDB assertion failure seen on the previous commit (`8f4c234`):
> 
> ```
> /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lmdb-sys-0.8.0/lmdb/libraries/liblmdb:5800:
> Assertion 'IS_BRANCH(mc->mc_pg[mc->mc_top])' failed in mdb_cursor_sibling()
> ```
> 
> Exit code 139 (SIGSEGV) on both variants (batch size 2 and 8). Both pods restarted and are currently running.
> 
> Grafana panel at the time of crash: https://grafana.vaquita-altair.ts.net/goto/ffq10zsbnfvggb?orgId=1
> 
> This confirms the new commits (separated batch/accumulator sizes, accumulator sharding, better error logging) did not resolve the underlying LMDB torn-page corruption. The C-level assertion fires inside `mdb_cursor_sibling()` before any Rust error handling can intercept it — no structured error log is emitted, only the stderr assertion message visible via `kubectl logs --previous`.

Hmm, when you say "ephemeral deployment" do you mean that the "ephemeral_finalised_state" bool in the config was set to true?

Also were these sync runs from scratch or from a cache file from a previously crashed run, I think I have found a bug in the crash resistance (credit also due to @zancas) https://github.com/zingolabs/zaino/issues/1299, I wonder if this could be connected?

PR #1265, opened 2026-06-19
Migrate Integration-Tests to ztest harness

elicbarbieri:
# Migrate integration tests to the ztest harness

## Still Outstanding
- [x] Full Audit that all test setup and verified assertions/invariants are still intact
- [x] Some minor ztest harness issues to cleanup


nachog00:
@elicbarbieri this has diverged quite a bit... i would love to land this into dev. could you look into the conflicts?

zancas:
@nachog00 I notice the conflicts are resolved...   but @elicbarbieri it seems like CI is unhappy?

elicbarbieri:
Yeah this will break all the tests, that is expected.  Im finalizing another big push that addresses some review findings from your pass, and claude

nachog00:
**108 files, +14,965/−11,169, two headline-only commits, a one-line body.** The migration itself is solid — the packaging is what's expensive: review means reading the raw diff end-to-end, and after merge nothing here can be bisected. Specifics, with asks:

**1. The PR body does not describe the PR.** Beyond the test migration this diff changes: the workspace topology (`live-tests/` leaves the root workspace, gains its own committed lockfile), the CI surface (partition matrix is now hand-maintained, its validator is deleted, fork PRs get compile-only signal), and served production behavior (see 3). None of this is in the body. Ask: expand the body to list every category of change and its consequence.

**2. Twenty-two force-pushes, arriving as two headline-only commits.** The branch spans 2.5 months and was force-pushed 22 times, including six times in the two days after review comments landed — each push replaced the history those comments anchored to, and breaks "changes since last review" for every reviewer. Rebasing onto dev to avoid merge noise is fine and preserves commit granularity; collapsing the sequence to two commits is a separate choice, and it deletes the reviewer's ability to follow the work and the team's ability to bisect it. Ask: rebase freely, but keep logical units as separate commits with bodies until merge; flatten at merge time, not before review.

**3. Production behavior changes ride inside a test-infra PR.** Three served-output changes are bundled here: single-block `GetBlock`/`GetBlockNullifiers` pool filtering, validator error-message propagation, and `getaddressdeltas` registration. Each looks correct. Their only trace is CHANGELOG entries. Ask: either split them into a small precursor PR, or name them in the body and state which migrated assertions require them.

**4. An architectural decision with no record.** The re-split reverses ADR-0002. The reversal is justified — ztest tests link no production crate (except `zaino-proto`), and the new lockfile is committed, so 0002's drift-flake failure mode cannot recur — but that justification exists nowhere in the repo after this PR, because 0002 is deleted rather than superseded. ADR-0003's two-crate split and taxonomy still govern the new layout and are deleted too. Ask: add one short ADR ("live tests leave the root workspace for the ztest harness", supersedes 0002–0004) and restore the old records with superseded status. Policy argument: [#1490 comment](https://github.com/zingolabs/zaino/pull/1490#issuecomment-5543148919).

**5. The trust model is absent from the PR.** This PR makes zaino the first repo whose PR code executes on the production cluster. The model — federated OIDC, `tag:ztest-runner`, operator proxy, group `ztest-ci` → `ztest-remote` — was worked out across devops#9, machines#16, and the [design note on devops#9](https://github.com/zingolabs/devops/pull/9#issuecomment-5516873401); none of it is derivable from this diff. Ask: add a "Trust model" section to the body covering (a) the auth chain and what each hop can reach; (b) who can trigger a cluster run and under what approval (currently: any same-repo push, no approval — the `ztest-cluster` environment has no protection rules); (c) why the protected-environment gate from the converged design was dropped; (d) the blast radius (`ztest-remote` is resource-scoped but namespace-wide, on the production cluster) and what contains a hostile or runaway payload.

For scale: the migrated tests themselves are a clear improvement over what they replace — inline topology, independent oracles, honest QoS tiers. The asks are about making the work reviewable and its decisions recoverable, not about the work.

zancas (approved):
UtAck!

PR #1266, opened 2026-06-19
Chore/0.5.0 update zebra

pacu:
replaces #1256


daira (approved):
utACK

AloeareV:
We've updated our zebra versions in latest dev. Do we need to backport this into rc-0.5.0 without incorporating the other changes from dev? I am inclined to let these version updates land when we next add latest dev into an RC.

zancas:
Superceded, dev will become the new rc.

Issue #1264, opened 2026-06-19
Abstract devtool coupling behind a general client interface


zancas:
From review feedback on #1259 ("excise zingolib, replace with devtool", review by @nachog00):

> I would prefer this work to not tie everything so much to the specific devtool implementation detail, but to a general client.

The zingolib→devtool replacement landed coupled to the concrete devtool implementation. Follow-up: introduce a general client abstraction so test/consumer code depends on the interface, not on devtool specifics, making the backend swappable.

Ref: https://github.com/zingolabs/zaino/pull/1259#pullrequestreview-4535077387

zancas:
We are closing this issue. It tracks the coupling between the tests of Zaino
and the devtool client, from the review of #1259. That coupling is gone.

#1265 (commit 9fd5bb9b3, merged 2026-09-07) moved the live suite to the `ztest`
harness. It deleted `live-tests/e2e/src/devtool.rs` and
`live-tests/e2e/tests/devtool.rs`. No file in `live-tests/` refers to the
devtool client now. `zcash_local_net` appears in one comment, which records
that `ztest` replaced it.

#1272 began the abstraction that this issue asks for, and it closes for the
same reason. If a general wallet client is still wanted, it belongs in
https://github.com/zingolabs/infrastructure.

Issue #1262, opened 2026-06-19
Follow-Up to MVP devtool migration, necessary for full coverage and coherent upstream tracking


zancas:
The zingolib-excision branch (#1259) ports the wallet tests to the
zcash-devtool wallet client and gates 9 test instances (5 distinct tests)
behind the non-default `devtool-incompatible` feature, so they are
`#[ignore]`'d in the default run. Each is blocked on a devtool/`zcash_local_net`
capability or a perf optimization. This issue tracks un-gating them.

Run the gated set with `--features devtool-incompatible` (the three
`devtool_zcashd.rs` rows additionally need `--with-zcashd`).

| # | Test | Reason / blocker | Location |
|---|------|------------------|----------|
| 1 | `address_deltas` | devtool `shield` coinbase-maturity off-by-one: `shield` drains all transparent funds, always including the just-mined `tip-99` coinbase; devtool computes coinbase maturity vs target height (tip+1), zebra's mempool enforces vs current tip → immature by one block | `integration-tests/wallet-tests/tests/devtool.rs:2142` (mod `zebrad`) |
| 2 | `send_to_transparent_finalization` [FetchService] | heavy: ~99 halo2 proofs (99-block orchard advance); needs round-3 P2 per-call mining-pool override for transparent filler | `integration-tests/wallet-tests/tests/devtool.rs:1977` |
| 3 | `send_to_transparent_finalization` [StateService] | heavy: ~99 halo2 proofs; needs round-3 P2 transparent filler | `integration-tests/wallet-tests/tests/devtool.rs:2202` |
| 4 | `monitor_unverified_mempool` [FetchService] | devtool `WalletBalance` lacks `unconfirmed_*`/`confirmed_*` fields; balance asserts commented out | `integration-tests/wallet-tests/tests/devtool.rs:2008` |
| 5 | `monitor_unverified_mempool` [StateService] | devtool `WalletBalance` lacks `unconfirmed_*`/`confirmed_*` fields; balance asserts commented out | `integration-tests/wallet-tests/tests/devtool.rs:2223` |
| 6 | `create_200_block_regtest_chain_vectors` | vector builder funds via transparent-coinbase shielding (round-2 P1); same shield gap as #1 | `integration-tests/wallet-tests/tests/test_vectors.rs:41` |
[...]

PR #1272, opened 2026-06-20
Abstract devtool coupling behind a general client interface

Timi16:
- Introduce TestWallet trait, WalletPoolBalances, and TestClients<W> generic struct so wallet-test code depends on a
  backend-agnostic interface rather than devtool specifics (#1264).
  - Replace hand-rolled txid_internal_bytes with TxId-based txid_from_devtool — uses the proven ecosystem type from
  zcash_primitives instead of raw byte manipulation.
  - Return Zatoshis from balance accessors instead of raw u64, enforcing type-safe amount handling throughout test assertions.

  Addresses nachog00's review feedback on PR #1259 and pacu's review comments.

  Refs #1264, #1262

  Test plan

  - cargo check --tests passes in integration-tests/wallet-tests/
  - cargo clippy --tests passes with no new warnings
  - Existing devtool integration tests (devtool.rs, devtool_zcashd.rs, test_vectors.rs) compile and run unchanged against a local
  regtest node
  - Verify that TestWallet trait is the only interface test bodies depend on (no direct DevtoolClients method calls remain)


integration-tests/wallet-tests/Cargo.toml
```diff
@@ -60,6 +60,7 @@ zebra-state = "9.0.0"
 # Runtime / misc
 tokio = { version = "1.38", features = ["full"] }
 zcash_primitives = "0.28"
+zcash_protocol = "0.9"
```
zancas: We don't want to introduce dependencies on more fundamental libraries from the zcash_ecosystem, especially ones that are more fundamental (upstream) than zebra.

zancas:
If you can produce a variant of this that does not add new dependencies that would be more interesting.

Timi16:
> If you can produce a variant of this that does not add new dependencies that would be more interesting.

Okat alright would look into it thank you

zancas:
Moving to draft until it removes new dependencies.

Timi16:
> Moving to draft until it removes new dependencies.

Made changes and ready to review

zancas (commented):
Why isn't this PR'd against https://github.com/zingolabs/infrastructure ?

Timi16:
The changes in this PR live entirely in `integration-tests/wallet-tests/`. It adds `TestWallet`, `WalletPoolBalances`, `TestClients<W>`, and the `TestWallet` impl for the ZcashDevtool adapter — all inside Zaino’s own test harness. Nothing in `zcash_local_net` or the infrastructure was modified.

The abstraction wraps `zcash_local_net::client::Client` so that Zaino’s test bodies depend on a backend-agnostic trait (`TestWallet`) rather than directly calling ZcashDevtool methods. This decoupling is on the Zaino side: it lets us swap or add wallet backends (e.g. a future `zingolib` or mock backend) without touching the infrastructure.

The `TestWallet` impl for `ZcashDevtool` lives here because it adapts an external type (`ZcashDevtool`) to a Zaino-local trait (`TestWallet`), which is exactly where Rust’s orphan rules require it to live.

integration-tests/wallet-tests/src/lib.rs
```diff
@@ -37,6 +44,132 @@ impl Pool {
...
+            Pool::Transparent => self.transparent,
+        }
+    }
+}
+
+/// The operations a test wallet backend must support.
+///
+/// Each method panics on failure — test wallets are not expected to
+/// recover from errors, so the trait surface stays minimal.
+pub trait TestWallet {
```
zancas: Couldn't this be defined in zcash_local_net?

Timi16: The `Client` trait in `zcash_local_net` already defines the same set of operations (`address`, `balance`, `send`, `shield`, `sync`, `rescan`) with `Result` return types, and `WalletBalance` already carries the per-pool spendable fields. `TestWallet` is essentially a near-duplicate that replaces `Result` with panics.

That is a fair point. The trait could live in `zcash_local_net` as a panicking convenience layer alongside `Client`. Alternatively, the tests here could just use `Client` directly and call `.expect()` at each call site (or make the test functions return `Result` and use `?`).

Either approach would remove the need for `TestWallet` as a separate trait in this repo.

I think the best option is to drop `TestWallet` entirely and make the test helpers (`TestClients<W>`) generic over `Client` from `zcash_local_net` instead, with each method calling `.expect(...)` inline. This avoids duplicating the trait surface in either repository.

zancas: Yes..  let's avoid duplication, and leverage the Client trait as much as possible.

Timi16:
Let me get back to work

zancas:
OK marking draft for now.

Timi16:
> OK marking draft for now.

Okay

Timi16:
> OK marking draft for now.

Why did you make this a draft please would love to know

zancas:
Thank you for this work, and I apologize for the late answer. On 2026-07-02 you
asked why this PR became a draft. There were two reasons. The first version
added new dependencies. The second reason was the location of the change. A
general wallet client belongs in https://github.com/zingolabs/infrastructure
and not in the test code of Zaino.

We are closing this PR, for two reasons.

The PR has no content now. Commit 48cd6abf4, "Merge dev: drop TestWallet, use
Client trait directly", removed the `TestWallet` trait. The diff against `dev`
contains one file, `tools/.DS_Store`. That file is macOS metadata, and it must
not merge.

The code that this PR changed is gone. #1298 renamed `integration-tests` to
`live-tests`. #1265 then moved the live suite to the `ztest` harness. It
deleted `live-tests/e2e/src/devtool.rs` and `live-tests/e2e/tests/devtool.rs`.
No test in Zaino calls the devtool client now, so no coupling remains to
remove.

Work on a general client interface is welcome in the infrastructure
repository.

PR #1274, opened 2026-06-21
feat: gate gettxoutsetinfo + txout-set accumulator behind a non-default feature (functional)

zancas:
## Summary

Gates the `gettxoutsetinfo` RPC and its sole dependency — the finalised **txout-set
accumulator** (schema table #9, `tx_out_set_info_accumulator`) — behind a **non-default
`gettxoutsetinfo` Cargo feature**. The default build neither stores nor builds the accumulator
(whose from-genesis build OOM-killed a 16 GiB host, #1260) and answers the RPC with
`FinalisedStateError::FeatureUnavailable`; `--features gettxoutsetinfo` restores it.

This is the **functional half** of the work — production + test code only. The design/ADR/
architecture docs, the CHANGELOG prose, and the CLAUDE.md tooling rules land in a **follow-on
docs PR** (kept separate to ease review).

## What's here

- **Remove the legacy v0 finalised-state schema** — the gate is built on the v1-only tree.
- **The gate, at the capability-dispatch seam.** The cost path (the accumulator field, its table
  creation, the builders, write-path maintenance, and migration Stage C) is hard-`#[cfg]`'d out;
  method signatures stay stable; the dispatch body returns `FeatureUnavailable` when the feature
  is off, and everything above the seam (NFS fold, backends, indexer trait, JSON-RPC handler) is
  untouched and just propagates the typed error. The schema-hash text keeps the table *described*,
  so feature-on/off databases stay mutually openable (no rebuild/cutover).
- **Consolidation.** The entire accumulator capability — its impl, the `FinalisedSource` dispatch,
  the write-path glue, the migration Stage C step, and every accumulator test + oracle — lives in
  one gated module, `finalised_source/v1/tx_out_set_accumulator.rs`. No accumulator-specific code
  remains inline anywhere else; the remaining `#[cfg(feature = "gettxoutsetinfo")]` points are the
  irreducible floor (mod gate, field, constructions, the two dispatch arms, and single gated
  calls into the module).
- **CI matrix** building + testing both feature states (default off, and `--features gettxoutsetinfo`).
- **Breaking version bumps** (semver: removing always-on functionality from the default set):
  `zaino-state` → **0.4.0**, `zaino-serve` → **0.4.0**, `zainod` → **0.5.0**. (The CHANGELOG prose
  documenting the break + consumer-migration guidance is in the follow-on docs PR.)

## Verification

- `cargo check` and `cargo clippy` clean in **both** feature states (default and `gettxoutsetinfo`).

## Base note

Cut from `0.5.0-rc.1`. `rc/0.5.0` has since advanced to `0.5.0-rc.2`, which made **no code
changes** — only manifest/version/CHANGELOG bumps. A small rebase onto the current `rc/0.5.0`
resolves the manifest conflicts (`zainod` version `0.4.1` vs `0.5.0`, `zaino-proto` `0.1.3`,
`Cargo.lock`); there are no code conflicts.


idky137 (changes requested):
Im a bit confused as to why we are jumping to this solution when we do not yet know the issue/bug we need to fix.

It is still unknown if this OOM is caused by a bug in sync or in configuration, feature gating this code and deleting the tests that test it does not help us track down and fix the bug we have.

There is also a lot of unrelated changes mixed into this PR (V0 deletion, drying of unrelated code), I think these are good changes to make but should be in separate PRs into dev, rather than mixed into a hotfix PR into an RC.

If we want to add this feature we need to add new migration tests to ensure if zaino has been synced with / without the feature, and then is rebuilt in a different configuration and re-launched with the same cache file, does not crash or build / serve corrupt data.

zancas:
Please explain what these migration tests look like, so we can upgrade our codebase:

_If we want to add this feature we need to add new migration tests to ensure if zaino has been synced with / without the feature, and then is rebuilt in a different configuration and re-launched with the same cache file, does not crash or build / serve corrupt data._

zancas:
I am porting the fixes for known bugs here.

For the spawn background-syncs race, I am preferring @emersonian 's solution based on this analysis:

|                                            | idky #1263 (82aa6d03)                                  | emersonian #1275                                                                                    |
| ------------------------------------------ | ------------------------------------------------------ | --------------------------------------------------------------------------------------------------- |
| Guard site                                 | top of `sync_to_height` (line ~688) — guards all paths | inside `if sync_is_long_running` (line ~708) — guards only the detached-spawn path                  |
| Inline near-tip sync while a backfill runs | suppressed (coarse)                                    | proceeds (surgical)                                                                                 |
| TOCTOU                                     | non-atomic check-then-act, unmentioned                 | non-atomic, explicitly documented + proposes the fix (`AtomicBool`/`compare_exchange` or `TryLock`) |
| Writeup                                    | one inline comment                                     | full SIGSEGV/LMDB-NOTLS root cause + before/after (6 restarts→0, RSS 15 GiB→1.3 GiB)                |


For the other two, I intend to be port them #1263.

zancas:
This ports all of the root-caused and patched bug-fixes @idky137 found and implemented on #1263.

idky137 (changes requested):
This pr was already too large and confused for a hotfix pr. We should not be adding more code to this pr, we need to work on extracting the bits of wode we want from it.

idky137:
> This ports all of the root-caused and patched bug-fixes @idky137 found and implemented on #1263.

#1263 implements these fixes in isolation, without all the extra unrelated changes. That pr should be merged, and then this can be reviewed in isolation.

idky137 (changes requested):
This should not land. It removes production features we have committed to supporting for no reason.

nachog00:
@zancas is this superseeded/stale?

PR #1285, opened 2026-06-22
changed feature to test only

idky137:
Makes the new txoutsetinfo accumulator feature flag a test-only feature, simplifying db version safety.

**Reasoning**:
A subtractive test-only feature is preferred here because it keeps the production database format canonical and unconditional.

If the new behaviour is added as a normal/default production feature, then database compatibility becomes dependent on both the database version and the binary’s feature set. That means migrations must account for:

    previous database version
    target database version
    previous binary feature configuration
    current binary feature configuration

This creates unnecessary state-space growth and makes version safety harder to reason about.

By contrast, a subtractive test-only feature lets production builds always compile the real production path, while tests can explicitly disable that path to construct legacy fixtures, exercise migration behaviour or test limited builds. The feature does not define a supported production configuration, so it does not need to be tracked in the database metadata or migration matrix.

### DIFF
--- a/integration-tests/walletless-tests/Cargo.toml
+++ b/integration-tests/walletless-tests/Cargo.toml
@@ -11,10 +11,11 @@ license.workspace = true
 version.workspace = true
 
 [features]
-# gettxoutsetinfo RPC + its finalised txout-set accumulator (non-default in zaino-state).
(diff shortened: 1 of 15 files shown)


zancas:
This enables us to toggle off a feature for test, special distribution, etc.

Looks good.  Thanks for this.

PR #1291, opened 2026-06-23
Patch reorg depth unit tests

### LINKED ISSUE #1288: test vectors insufficient to cover new finalization threshold
### Security check

### Observation

Our mockchain tests use a hard-coded test vector. This vector isn't large enough to cover the new finalization threshold, now that zebra has updated it to 1000. 

Currently, this causes one test failure: `race_pre_mine_finalized_height_block_is_evicted_when_source_advances_mid_iter`. This test asserts that blocks below the finalization threshold will be deleted once that threshold is reached...but that threshold is now never reached. we need to either add 1000-ish blocks to our test vector, or adapt this test to use proptest_mockchain (which means implementing the adding-blocks-on-a-hook mechanism that mockchain_source uses).

### Location in codebase

zaino-state tests

### Type of concern

Other

### Impact

_No response_

### Suggested direction

_No response_

zancas:
Fixes:  #1288

This does not fix issues in the integration tests.

We'll keep a back-compatibility patch for tests that are impractical or inaccurate with a 1_000 block reorg bound, and begin adding tests that specifically enforce invariants that are scale dependent.

See https://github.com/zingolabs/zaino/issues/1292 for planned work.

### DIFF
--- a/integration-tests/Cargo.toml
+++ b/integration-tests/Cargo.toml
@@ -43,9 +43,9 @@ zcash_protocol = { version = "0.9", features = ["local-consensus"] }
 zcash_client_backend = "0.23"
 
 # Zebra
-zebra-chain = "9.0.0"
-zebra-state = "8.0.0"
-zebra-rpc = "9.0.0"
+zebra-chain = "10.0.0"
+zebra-state = "9.0.0"
+zebra-rpc = "10.0.0"
 
 # Runtime
 tokio = { version = "1.38", features = ["full"] }
--- a/packages/zaino-state/src/chain_index/tests/non_finalised_state.rs
+++ b/packages/zaino-state/src/chain_index/tests/non_finalised_state.rs
@@ -229,8 +229,6 @@ async fn shutdown_terminates_sync_loop_cleanly() {
 /// land in iter N+1 (which would compute the correct post-mine finalized
 /// height and trim properly).
 #[tokio::test(flavor = "multi_thread")]
-#[ignore = "Fails due to finalized height never reached. \
-See https://github.com/zingolabs/zaino/issues/1288"]
 async fn race_pre_mine_finalized_height_block_is_evicted_when_source_advances_mid_iter() {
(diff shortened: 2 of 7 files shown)


AloeareV (approved):
The new finalized depth pin for tests is good by me!

integration-tests/Cargo.toml
```diff
@@ -43,9 +43,9 @@ zcash_protocol = { version = "0.9", features = ["local-consensus"] }
 zcash_client_backend = "0.23"
 
 # Zebra
-zebra-chain = "9.0.0"
-zebra-state = "8.0.0"
-zebra-rpc = "9.0.0"
+zebra-chain = "10.0.0"
+zebra-state = "9.0.0"
+zebra-rpc = "10.0.0"
```
nuttycom, line 48: Nit: these are a little bit overspecialized, they could just be "10", "9", "10"

packages/zaino-state/src/chain_index/tests/non_finalised_state.rs
```diff
@@ -234,7 +234,7 @@ async fn race_pre_mine_finalized_height_block_is_evicted_when_source_advances_mi
         load_test_vectors_and_sync_chain_index(MockchainMode::Active).await;
 
     let initial_active = mockchain.active_height();
-    let pre_mine_finalized_height = finalized_height_floor(initial_active);
+    let pre_mine_finalized_height = dbg!(finalized_height_floor(initial_active));
```
nuttycom, line 237: dbg! left in code.

integration-tests/Cargo.toml
zancas: Will update, thanks.

packages/zaino-state/src/chain_index/tests/non_finalised_state.rs
zancas: Ack.

PR #1295, opened 2026-06-24
Update zebra-* dependencies to ZcashFoundation/zebra@9a27f886a5bfb143f65d1712e912cef252426800

nuttycom:
This updates to the in-flight ZcashFoundation/zebra@9a27f886a5bfb143f65d1712e912cef252426800 that is intended to be included in the next round of zebrad releases. It provides functionality to allow use of Zebra's internal logic for finding the fork point relative to Zebra's current "best chain" tip given a block locator, which Zallet depends upon. Zebra is able to find the fork point much more efficiently than an external client is able to because they have an indexed in-memory tree of all of the current chain branches that isn't otherwise exposed (except as chain tips) to consumers.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,3 @@
+[env]
+# Workaround for GCC 15+: zebra's librocksdb-sys vendored C++ relies on transitive <cstdint>.
+CXXFLAGS = { value = "-include cstdint", force = false, relative = false }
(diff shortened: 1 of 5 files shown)


zancas (commented):
UtAck

zancas:
I kicked a local test run against this.

zancas (approved):
Tested locally.

zancas:
@AloeareV please check this against your reify-nfs work, when you get a chance.

PR #1300, opened 2026-06-25
set ephemeral_finalised_state to true in zainod example config

zancas:
Reduces the functionality of zaino, so we can focus on what's strictly required for Z3 / ironwood.


zancas:
Deferred.

PR #1302, opened 2026-06-26
Dependabot bumps

zancas:
Bumps dependencies where dependabot has recommended to do so.

All tests run and pass locally.


AloeareV (approved):

AloeareV (approved):
shellcheck lint fail is not introduced by this PR, therefore non-blocking

PR #1303, opened 2026-06-26
Encapsulate tx out set info

zancas:
This branch refactors gettxoutsetinfo into it's own mod.

PR #1307, opened 2026-06-26
build: port build_deterministic.sh to rust-script, fixing Shellcheck CI

zancas:
## What

The `bootstrapped_deterministic_build` merge added `utils/build_deterministic.sh`, which the separately-merged Shellcheck gate (`lint-shell.sh` over `git ls-files '*.sh'`) then rejected on `dev`: one **SC2034** (unused `DIR`) and three **SC2086** (unquoted expansions). This surfaced as the failing Shellcheck job on PR #1303.

Rather than patch the shell, this ports it to a single-file **rust-script** per the repo's Rust-first rule for committed tooling. The Shellcheck job goes green by construction — the `.sh` no longer exists in the tracked set (33 → 32).

## Improvements the port buys

- **Result-based error propagation** (no `unwrap`; explicit "not a git repository?" failure) in place of `set -eu`'s implicit abort.
- **Per-command `.env(...)`** instead of exported globals, so the script stays `#![forbid(unsafe_code)]`-clean (`env::set_var` is `unsafe` under edition 2024).
- The shared `docker build` flags kept **DRY** in one `deterministic_build()` fn.

## Caller impact

Nothing in the repo references the script (no Makefile task, docs, or CI), so the only change is the extension: `./utils/build_deterministic.rs [args]`.

## Verification

- `./tools/scripts/lint-shell.sh` → green (32 scripts pass).
- `rust-script --test utils/build_deterministic.rs` → compiles clean, zero warnings.
- Not runtime-tested against a live docker daemon.


zancas:
Already merged with #1304

Issue #1328, opened 2026-06-29
Finalised spend-index pilot: parallel-buildable `transparent_outpoint → spending_txid` (refines #1326)


zancas:
## Summary

A **refined, per-index proposal** developed from first principles, linked to the umbrella conversation in #1326. Where #1326 surveys the whole finalised-state schema, this issue scopes the **first concrete deliverable**: a **proof of concept** that a single served finalised index — `transparent_outpoint → spending_txid` — can be built **entirely in parallel** from the zebra block stream, in isolation from every other index.

This is the *simplest useful case*: it lets us land the zaino design change that enables isolated/parallel per-index builds (step 0) and debug/optimise the sync procedure against it before generalising.

Decisions are recorded in **`docs/adr/0002-finalised-spend-index-parallel-build.md`**; glossary terms in `packages/zaino-state/src/chain_index/CONTEXT.md`. Implementation lands in a **separate, discrete PR** after the ADR is committed.

## The index

> Given a transparent outpoint, return the txid of the transaction that **consumed** it, or `None` if it is unspent.

```
transparent_outpoint  ->  spending_txid
```

- **Key:** the 36-byte outpoint (`prev_txid[32] ‖ LE(u32) prev_index`). We key by the outpoint because that is what the query hands us — keying direction *is* the design.
- **Value:** the bare 32-byte spending txid. Not a `TxLocation` (which would reintroduce a dependency on `txids`/`txid_location` and break isolation).
- **Absence ⇒ unspent.**

Terminology: an outpoint already names its **creating** transaction (its first field). The transaction that *contains* an outpoint — as the prevout field of one of its **inputs** — is its **spending** transaction. This index maps to the spending txid, never the creating txid.

## Why build it (rather than proxy zebra)

zebra already answers this via `ReadRequest::SpendingTransactionId(Spend::OutPoint(..))`, but only behind `#[cfg(feature = "indexer")]`. Building it in zaino gives: (a) **independence** from a non-default zebra feature, (b) **serving scale**, and (c) a **pilot** for the parallel-sync architecture. Critically, **zebra's `indexer` service is slated to be deprecated and removed once this is stable in zaino** — so zaino becomes the sole provider and this index becomes *required*, not merely offered.

zebra's `SpendingTransactionId` is therefore a **temporary correctness oracle**: diff zaino's index against it over a mainnet range. **Golden vectors must be captured before zebra removes the feature.**

## Step (0): the design change — the read-free extractor

[...]

Issue #1326, opened 2026-06-29
Assess whether the persistent finalised-state's shape should be updated to offer more efficient build and service times (Conversation)


idky137:
# Zaino Finalised State Extended Services
This Issue documents the current and planned (agreed in principle) indexes that zaino does (or will) serve (exclusive from zebrad), and discusses / explores possible shapes for storing this data.

### 1.1 CompactBlock Service
**Indexes**:
- `BlockFilter -> CompactBlock`
- `BlockRange -> Stream<CompactBlock>`

*Potentially possible to serve directly from the backing validator, performance dependant.

**Action**: benchmark validator-backed vs internally-backed CompactBlock service over a large mainnet block range, including high concurrency benchmarks.

| Design | Stored Indexes | Build | Serve | Build Complexity | Serve Complexity |
|--------|----------------|-------|-------|------------------|------------------|
| **Current** | `headers: Height -> BlockHeaderData`<br>`txids: Height -> Vec<TxId>`<br>`transparent: Height -> Vec<Option<TransparentCompactTx>>`<br>`sapling: Height -> Vec<Option<SaplingCompactTx>>`<br>`orchard: Height -> Vec<Option<OrchardCompactTx>>`<br>`commitment_tree_data: Height -> CommitmentTreeData`<br>`heights: BlockHash -> Height` | Store per-height component tables. CompactBlock is reconstructed at serve time from header, txids, optional pool tables, and commitment tree data. | Fetch header, txids, requested pool tables, and commitment tree data; then construct `CompactBlock`. Range streaming walks headers and keeps the other tables cursor-aligned by height. | `O(txs_in_block + compact_elements_in_block)` writes across multiple component tables. | Single block: `O(number_of_component_tables + txs_in_block + compact_elements_in_requested_pools)`.<br>Range: `O(blocks_in_range × requested_component_tables + compact_elements_in_range)`. |
[...]

PR #1329, opened 2026-06-29
ADR proposing an outpoint --> spending_txid single index

zancas:
Describes a POC single outpoint --> spending transaction index, similar in spirit to "Proposal 1" from #1326 and extended in #1328.


zancas:
Superceded.

PR #1332, opened 2026-06-29
Dry spent outpoint

zancas:
DRYs spent_outpoint collection.

All local test pass.

This PR unifies some ~8 disparate copies of the same logic into a single helper, and adds a unit-test to protect against regression across the helper.

With the ADDITION of a test, and documentation, this PR still nets ~ -11 LOC.


idky137 (commented):
utACK

idky137 (approved):

PR #1338, opened 2026-06-30
Finalize 0.4.2: strip -rc.1 release-candidate suffixes

nachog00:
PR #1306 ("Release 0.4.2") merged `rc/0.4.2` into `stable` and the `0.4.2` tag was stamped on the result — but it never stripped the release-candidate suffixes. The tagged tree still self-identifies as `0.4.2-rc.1`, so it cannot be published to crates.io and no `0.4.2` Docker image / GitHub Release exist.

This drops `-rc.1` across the workspace so the crate versions match the tag:

| Crate | Before | After |
|---|---|---|
| `zainod` | `0.4.2-rc.1` | `0.4.2` |
| `zaino-fetch` | `0.2.1-rc.1` | `0.2.1` |
| `zaino-state` | `0.3.1-rc.1` | `0.3.1` |
| `zaino-serve` | `0.3.1-rc.1` | `0.3.1` |

Root `[workspace.dependencies]` pins updated to match and `Cargo.lock` regenerated (`cargo update --workspace`, 4 packages, no other churn). `zaino-common` (0.2.0) and `zaino-proto` (0.1.3) were already final and are untouched.

### Follow-up after merge
- Move the `0.4.2` tag onto the merge commit (`git tag -d 0.4.2 && git push origin :refs/tags/0.4.2`, then retag). The current tag never produced a Docker image, crates.io release, or GitHub Release, so nothing references it.
- Publish in dependency order — `zaino-fetch 0.2.1`, `zaino-state 0.3.1`, `zaino-serve 0.3.1` are new crates.io releases (registry currently has `.2.0`/`.3.0`/`.3.0`), then `zainod 0.4.2`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5554,7 +5554,7 @@ dependencies = [
 
 [[package]]
 name = "zaino-fetch"
-version = "0.2.1-rc.1"
+version = "0.2.1"
 dependencies = [
  "base64",
  "byteorder",
@@ -5596,7 +5596,7 @@ dependencies = [
 
 [[package]]
 name = "zaino-serve"
-version = "0.3.1-rc.1"
+version = "0.3.1"
 dependencies = [
  "futures",
  "jsonrpsee",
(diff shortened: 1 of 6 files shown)


zancas (approved):
Thanks for fixing this!

PR #1313, opened 2026-06-27
types: add CompactDifficulty and ChainWork domain primitives

nachog00:
## Summary

- Introduce `CompactDifficulty` (wraps zebra's type, validates nBits at construction) and `ChainWork` (wraps `NonZeroU128`, strictly positive, fallible arithmetic) as domain primitives under `types/primitives/`.
- Replace `BlockData.bits: u32` with `CompactDifficulty` and delete hand-rolled `compact_to_target_u256`, `target()`, `work()`, `difficulty()`, `u256_to_f64` — all conversion now delegates to zebra's validated pipeline.
- Remove old `ChainWork([u8; 32])` from `legacy.rs`; replace zero-chainwork sentinels with `Option<ChainWork>` (`None` = no parent / genesis).
- Add `PersistentChainWork` and `PersistentCompactDifficulty` DB boundary types in `db/block.rs`.
- Fix genesis chainwork double-counting bug in `non_finalised_state` init.

Closes #1158

## Context

`BlockData.bits` was a raw `u32` with no validation — any value flowed through a buggy hand-rolled `compact_to_target_u256` (missing sign-bit and overflow checks) into `ChainWork` accumulation. The old `ChainWork([u8; 32])` accepted zero values and used unchecked U256 arithmetic.

The new types enforce invariants at construction time (parse-don't-validate): `CompactDifficulty` rejects invalid nBits encodings via zebra's consensus-tested validation, and `ChainWork` is structurally nonzero via `NonZeroU128`.

PR #1346, opened 2026-07-01
Fix the #1313 chainwork on-disk byte-order regression (big-endian)

zancas:
#1313 (5e4dae4a) reworked ChainWork into PersistentChainWork but serialized it little-endian, while the established v1 on-disk format -- set by the original ChainWork([u8;32]) via U256::to_big_endian -- is big-endian. Reading real v1 data (the v1_test_db fixture: regtest cumulative chainwork 17 = big-endian [00..00,11]) as little-endian pushed the value into the high bytes and rejected it as 'chainwork exceeds u128 range', breaking load_db_backend_from_file and decode of every existing v1 DB. Fix: PersistentChainWork now reads/writes big-endian; no migration or fixture regeneration needed -- the fixture was correct all along.

Archaeology + regression proof: recovers the actual deleted encoder from history (git show 5e4dae4a^:.../types/db/legacy.rs) into a #[cfg(test)] reference module, and adds a cross-encoder equivalence test asserting the current encoder reproduces the original big-endian bytes byte-for-byte (and decodes them) across the whole nonzero u128 domain -- every power of two, every byte position, the extremes, plus proptest. This is the test #1313 lacked: red against its little-endian encoder, green against the corrected one. The likely root of the mistake is a documentation bug in the original (struct doc said big-endian but FixedEncodedLen said '32 bytes, LE' and the writer is write_fixed_le), highlighted in the recovered reference. Also corrects the canonical golden (expected_v2_bytes), which #1313 minted little-endian -- a golden enshrining the very bug it should have caught.

### DIFF
--- a/packages/zaino-state/src/chain_index/tests/types.rs
+++ b/packages/zaino-state/src/chain_index/tests/types.rs
@@ -53,11 +53,15 @@ pub(crate) fn expected_v2_bytes() -> Vec<u8> {
     // BlockHash (parent_hash): V1 tag + 32-byte body.
     out.push(version::V1);
     out.extend_from_slice(&[0x02; 32]);
-    // ChainWork: V1 tag + 32-byte LE (value = 0x42, stored as u128 LE in lower 16 bytes).
+    // ChainWork: V1 tag + 32-byte big-endian (value = 0x42, in the low-order 16
+    // bytes). Corrected from little-endian: the established v1 on-disk format is
+    // big-endian (the original `ChainWork([u8;32])` via `U256::to_big_endian`,
+    // and the `v1_test_db` fixture) — #1313 wrongly minted this golden LE, which
+    // is exactly a golden enshrining the bug it should have caught.
     out.push(version::V1);
     {
         let mut cw_bytes = [0u8; 32];
-        cw_bytes[..16].copy_from_slice(&0x42u128.to_le_bytes());
+        cw_bytes[16..].copy_from_slice(&0x42u128.to_be_bytes());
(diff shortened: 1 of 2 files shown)


AloeareV (approved):
This is a fantastic fix. I don't want to block this PR on a single comment that's still mostly intelligible. Will open a followup! 

Edit: I see a lot of instances of 'golden' being used as a noun here, which I find extremely confusing, but it seems to be intentional language and not a grammatical error.

packages/zaino-state/src/chain_index/tests/types.rs
```diff
@@ -53,11 +53,15 @@ pub(crate) fn expected_v2_bytes() -> Vec<u8> {
     // BlockHash (parent_hash): V1 tag + 32-byte body.
     out.push(version::V1);
     out.extend_from_slice(&[0x02; 32]);
-    // ChainWork: V1 tag + 32-byte LE (value = 0x42, stored as u128 LE in lower 16 bytes).
+    // ChainWork: V1 tag + 32-byte big-endian (value = 0x42, in the low-order 16
+    // bytes). Corrected from little-endian: the established v1 on-disk format is
+    // big-endian (the original `ChainWork([u8;32])` via `U256::to_big_endian`,
+    // and the `v1_test_db` fixture) — #1313 wrongly minted this golden LE, which
+    // is exactly a golden enshrining the bug it should have caught.
```
AloeareV, line 60: This comment doesn't scan to me. 

Edit: I see a lot of instances of 'golden' being used as a noun here, which I find extremely confusing, but it seems to be intentional language and not a grammatical error.

PR #1354, opened 2026-07-01
Fix nfs depth bug

zancas:
Multiple instances of "magic" values meant to represent the NFS depth drifted when the upstream value bumped from 100 to 1_000.  This PR is meant to remedy those bugs, and eliminate that "magic" value drift for at least some consts.

This PR proposes a single place for const declaration in zaino-common, and places several consts there.


idky137 (approved):
utACK

PR #1359, opened 2026-07-03
Add publish dry-run CI check

nachog00:
## Summary

- Adds a `cargo publish --dry-run` workflow that runs on every PR and push
- **Advisory (warn-only)** on regular PRs — step fails but doesn't block merge
- **Hard-fails on `rc/*` and `stable`** — catches unpublishable `[patch.crates-io]` overrides before release
- Publishes crates in dependency order: zaino-proto, zaino-common, zaino-fetch, zaino-state, zaino-serve, zainod

This would have caught the zebra fork patch blocker that prevented publishing zaino-state/zaino-serve/zainod for the 0.5.0 release.

### DIFF
--- a/.github/workflows/publish-dry-run.yml
+++ b/.github/workflows/publish-dry-run.yml
@@ -1,0 +1,80 @@
+name: Publish dry-run
+
+on:
+  workflow_call:
+  workflow_dispatch:
+  push:
+    branches:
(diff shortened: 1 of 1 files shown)


zancas (approved):
Nice.   Pro-move.

Issue #1368, opened 2026-07-05
First NU6.3 block's coinbase is served as Orchard, not Ironwood (zebrad routing off-by-one, or a zaino pool-swap?)


zancas:
## Situation

The new e2e wire-tier era tests (`live-tests/e2e/tests/compact_block_wire.rs`, ee9ed33c on `feat/ironwood_nu6_3`) fail on their first container run, both at **exactly the first ironwood-era height**:

```
ironwood_only_wire_serving_zebrad
  served orchard actions mismatch at height 2 (orchard 1, ironwood 0)

orchard_to_ironwood_transition_wire_serving_zebrad
  served orchard actions mismatch at height 6 (orchard 1, ironwood 0)
```

Image: `zingodevops/zaino-ci:RUST_1.96.0-ZCASH_6.20.0-ZEBRA_6.0.0-rc.0-DEVTOOL_252114bbfa69-CONTAINER_4bbade12d61d2f`. Fixture: regtest, `MinerPool::Orchard` (orchard-receiver miner), NU6.3 active at height 2 (ironwood-only test) or height 6 (transition test).

Expected from NU6.3 consensus: the orchard-receiver coinbase must have an **empty Orchard component**, with the reward routed to **Ironwood** actions. Observed at the first post-activation height: one Orchard action, zero Ironwood.

What the failures already localize:

- **Wire fidelity passed** (asserted before era composition): what the tonic client received equals what the in-process subscriber produced, so zainod's gRPC layer is faithful — the mislabel (if zaino-side) is upstream, or the block itself is what it says.
- **The transition test's Orchard-era half passed** (heights 2–5 served as Orchard coinbases), so the fixture mechanics and orchard serving are sound. The failure is confined to the ironwood boundary.
- Both tests abort at the first failing height, so heights *after* the boundary (3+ / 7+) are unobserved in this run.

## Hypotheses

**A — zebrad-side (fixture premise false for this binary).** `zebrad 6.0.0-rc.0`'s coinbase builder routes the reward into Orchard at the first NU6.3 height. Two sub-variants:
  - *A1, off-by-one*: the miner evaluates the network upgrade at the **tip** height instead of the **next-block** height, so only the first post-activation block misroutes and heights 3+/7+ are V6/Ironwood. Circumstantially supported: earlier devtool runs on NU6.3-active chains produced "Unsupported tx version 6" errors in released zainod, proving V6 blocks do appear on these chains.
[...]

PR #278, opened 2025-04-14
Update Cargo.lock

nachog00:

PR #1375, opened 2026-07-07
feat!: enforce the Validator as the sole source of activation-height truth

### LINKED ISSUE #1076: Regtest fixture has zero NU6.1 coverage — depends on zcash_local_net lockbox disbursements
## Today

`zaino-common::ZEBRAD_DEFAULT_ACTIVATION_HEIGHTS` sets `nu6_1` at height 1000. This is mirrored in `zcash_local_net::validator::regtest_test_activation_heights` (the alignment is documented in that helper's doc-comment as required to avoid `InvalidData("Block commitment could not be computed")` from zaino's chain-index sync).

Effect: zaino's regtest integration tests never reach the NU6.1 activation block. NU6.1 is mainnet-active; the codepaths it touches (lockbox accounting, ZIP-271 disbursements at activation, NU6.1-specific block-commitment scheme) get zero regression coverage.

## Why height 1000

zebrad's `subsidy_is_valid` (`zebra-consensus/src/block/check.rs`) rejects the NU6.1 activation block when `Network::lockbox_disbursements(height).is_empty()`. On regtest the default is empty, so the activation block can never be mined. The 1000 sentinel sidesteps the failure by ensuring tests never reach it.

This is being addressed upstream in zingolabs/infrastructure: a `LockboxDisbursement` value type plus `ZebradConfig.lockbox_disbursements` field that serializes into Zebra's regtest TOML at `[[network.testnet_parameters.lockbox_disbursements]]`. See zingolabs/infrastructure#243.

## What zaino needs to do once that lands

1. **Update `zaino-common::ZEBRAD_DEFAULT_ACTIVATION_HEIGHTS`** to put NU6.1 at the new low height (likely 2 — same block as NU6 — but the exact value is settled by the upstream proof-of-life test).
2. **Wire lockbox disbursements into zaino's integration test fixtures.** Every test that launches a zebrad must inject a non-empty disbursement list (a single dummy entry suffices). Likely a one-line addition wherever `ZebradConfig` is built for tests.
3. **Verify chain-index sync** with NU6.1 active. The doc-comment in `zcash_local_net::validator::regtest_test_activation_heights` warns that drift between zaino-common and the validator's heights causes `InvalidData("Block commitment could not be computed")` — i.e. zaino expects a specific block-commitment scheme per height. With NU6.1 at height 2, the chain-index sync code paths that branch on activation are exercised for the first time and must be confirmed correct.

## Acceptance criteria

- [ ] `ZEBRAD_DEFAULT_ACTIVATION_HEIGHTS` and `regtest_test_activation_heights` agree on a low NU6.1 height.
- [ ] zaino's integration tests configure `lockbox_disbursements` whenever they launch zebrad.
- [ ] At least one zaino integration test verifies behavior across the NU6.1 activation boundary (a sync that crosses the activation block, or a query that returns NU6.1-era data).
- [ ] No `InvalidData("Block commitment could not be computed")` regressions surface.

## Out of scope

- The infrastructure-side plumbing itself (LockboxDisbursement type, ZebradConfig field, TOML serialization). Tracked at zingolabs/infrastructure#243.

## Cross-reference

- Upstream enabler: zingolabs/infrastructure#243.
- Zebra source: `zebra-consensus/src/block/check.rs::subsidy_is_valid` (the gating check).
- ZIP-271: https://zips.z.cash/zip-0271

zancas:
The Validator's configured activation heights become the only source of truth for every component, enforced at compile time on both sides of the zcash_local_net seam. This closes #1076 and retires the failure class behind #1368 (the hand-maintained three-repository height mirror and its `InvalidData("Block commitment could not be computed")` drift crashes).

On the indexer side, `zaino_common::Network` is now a payload-free kind, so a configuration value cannot carry heights and a pre-handshake height read does not typecheck. Both backends construct the runtime `zebra_chain::parameters::Network` in exactly one place — the spawn handshake — using zebra's compiled parameters for the public networks and adopting the validator's reported schedule from `getblockchaininfo.upgrades` for regtest. An upgrade absent from the validator's map is never-activated, nothing is backfilled from defaults, and a failed handshake aborts startup naming the validator endpoint.

On the wallet side, the zcash_local_net pin moves to `78511f7` (infrastructure ADR 0003, infrastructure PR #278), where the devtool client's regtest network is only constructible by querying the running validator. The harness no longer mirrors fixture heights into zainod, so every custom-heights launch doubles as a regression test of the adoption path. The inbound design spec is committed as `zainod-heights-from-validator-spec.md`.

Regression coverage: `clientless/tests/validator_heights.rs` proves a kind-only-configured zainod syncs a NU6.3-at-6 schedule its configuration never saw and pins the `upgrades` map a live zebrad reports; three unit tests beside `activation_heights_from_upgrades` pin the mapping itself, including the never-activated treatment of absent upgrades.

Breaking for library consumers: `zaino_common::Network` loses its heights payload, and `to_zebra_network` and the `From` conversions into zebra's network type are deleted.

A follow-up PR stacked on this branch carries the Ironwood activation demo suite that this work unblocks.


zancas:
All local tests pass!

zancas:
Review comments addressed, local tests pass.

nachog00 (approved):

PR #1379, opened 2026-07-07
Fix: ironwood db migration

### LINKED ISSUE #1374: Persistent finalised state validation starts before migration.
BUG: Persistent finalised state validation starts before migration, causing failure due to incorrect schema:
```
  13:08:29.975 ERROR zaino_state::chain_index::finalised_state::finalised_source::v1: initial validation failed, e: LMDB database error: MDB_NOTFOUND: No matching key/data pair found, desc: "block scan"
    at packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1.rs:639

...

  13:08:30.086 ERROR zainodlib: Zaino failed to start, e: NodeBackedIndexerService error: Tonic status error: code: 'Internal error', message: "ChainIndex initial sync failed, check full log for details."
    at packages/zainod/src/lib.rs:62

Error: NodeBackedIndexerService error: Tonic status error: code: 'Internal error', message: "ChainIndex initial sync failed, check full log for details."
```

FIX: make sure all migrations complete before validation starts.

idky137:
- Closes https://github.com/zingolabs/zaino/issues/1374

Fixes: Fixes zainod startup failure on pre-v1.3.0 caches (Ironwood migration)

Opening an old testnet cache failed with MDB_NOTFOUND: "block scan" → ChainIndex initial sync failed. Two independent defects, one per commit.

### 5e4d45e7 — make persistent db validation wait on migration

The finalised-state background validator was started in DbV1::spawn, concurrently with the migration that FinalisedState::spawn kicks off afterward. Its initial block scan read the freshly-created-but-empty commitment_tree_data_1_3_0 table before the migration populated it → MDB_NOTFOUND → permanent CriticalError.

- Split "open/heal DB" from "start validator": DbV1::spawn now opens + heals only; the old spawn_handler becomes start_validator(&self), exposed via FinalisedSource::start_validator.
- FinalisedState::spawn starts the validator only after migrate() returns Ok (or immediately when no migration is pending), so validation always runs against a fully-migrated, current-schema DB.
- Test-only fixtures never validate a pre-migration DB: spawn_v1_0_0 marks the empty fixture Ready directly; spawn_with_target_version marks Ready after its migration without validating — no version-comparison, robust across future schema bumps.

###  b1551012 — fixed ironwood migration

The v1.2.1 → v1.3.0 migration hard-refused when the chain tip was at/above NU6.3 activation, because it could only rebuild commitment rows from legacy on-disk data. It now backfills the missing ironwood data from the backing validator.

- Dropped the NU6.3 guard; the migration now uses its source.
- Per height: below NU6.3 → in-place rebuild of the commitment row from the legacy fixed-length table (ironwood None); at/above NU6.3 → refetch via build_indexed_block_from_source and write both the commitment row (with ironwood root/size) and the sparse ironwood row (new shared build_block_ironwood_entry helper + ironwood_db() accessor).
- Preserves resumability, checkpoint fsync, idempotent (NO_OVERWRITE + verify-match) writes, and version-bump finalisation.
- Adds migrations/v1_2_to_v1_3.rs regression test: builds a v1.2.1 cache, reopens through the production spawn path, and asserts it migrates to the current schema and reaches Ready (i.e. migration completed before validation).

Caveats: the at/above-activation ironwood backfill path is verified by review + compilation but not integration-tested — MockchainSource serves no ironwood roots and the vectors carry no ironwood actions (see the test-file note). For an on-disk cache already past NU6.3, the operator's validator source must be able to serve that height range.

### DIFF
--- a/packages/zaino-state/src/chain_index/finalised_state.rs
+++ b/packages/zaino-state/src/chain_index/finalised_state.rs
@@ -543,12 +543,25 @@ impl<T: BlockchainSource> FinalisedState<T> {
                         source: migration_source,
                     };
 
-                    if let Err(error) = migration_manager.migrate().await {
-                        tracing::error!("FinalisedState migration failed: {error}");
+                    match migration_manager.migrate().await {
+                        Ok(()) => {
+                            // Start the background validator only now that every migration has
+                            // finished: its initial scan reads tables a migration populates (e.g.
+                            // `commitment_tree_data_1_3_0`), so starting it earlier would race the
+                            // migration and fail on a not-yet-written row.
+                            migration_router.primary_backend().start_validator();
+                        }
+                        Err(error) => {
+                            tracing::error!("FinalisedState migration failed: {error}");
 
-                        migration_router.store_primary_status(StatusType::CriticalError);
+                            migration_router.store_primary_status(StatusType::CriticalError);
+                        }
                     }
                 });
+            } else {
+                // No migration to run, so the on-disk tables the validator scans are already at the
+                // current schema: start it immediately.
+                router.primary_backend().start_validator();
             }
 
             Ok(Self { db: router, cfg })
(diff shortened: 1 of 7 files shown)


zancas:
Review surfaced six small polish items (error remediation text, allocation and DRY nits, visibility, test-flavor comment). None blocks this PR; they are tracked as non-blocking follow-up in #1382.

zancas (approved):

PR #1391, opened 2026-07-09
Dry fetch and proto

zancas:
This reduces duplication and prefers more function expression where reasonable.


zancas:
It may that the right pattern is to make IndexedBlock etc. actually depend on ZainoVersionedSerde...

nachog00 (approved):

PR #1394, opened 2026-07-13
bump deps

zancas:
Bumps dependencies.

Also removes unused dependencies, and makes test summary more informative including "packages" summary, and per-suite times.

### DIFF
--- a/.env.testing-artifacts
+++ b/.env.testing-artifacts
@@ -5,7 +5,7 @@
 # zcashd Git tag (https://github.com/zcash/zcash/releases)
 ZCASH_VERSION=6.20.0
 
-ZEBRA_VERSION=6.0.0-rc.0
+ZEBRA_VERSION=6.0.0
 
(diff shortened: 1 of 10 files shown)


nachog00:
On the timing addition in live-summary: I'd avoid parsing nextest's human-readable summary line for wall-clock durations — that `[ 510.718s]` format is a display detail, not a stable contract, and it could change between nextest versions.

Since `live-summary.rs` already owns the process lifecycle (calling `run_partition()` sequentially), the simplest approach is to just time it directly:

```rust
let start = Instant::now();
let (rc, log) = run_partition("live-clientless", with_zcashd)?;
let elapsed = start.elapsed();
```

No parsing, no external format dependency, and it captures build time too (which the nextest stamp doesn't include). The `Summary` struct gets a `Duration` field instead of parsing millis from ANSI output, and the `duration_millis` parser + its tests go away.

nachog00 (approved):

PR #1395, opened 2026-07-13
feat!: remove zcashd support entirely

zancas:
zcashd support was gated behind the opt-in `zcashd_support` feature (docs/adr/0001, 0005) as a staging post on the deprecation path. This PR takes the final step, recorded in docs/adr/0008: the feature and everything it gated are deleted — the zcashd-shaped `getpeerinfo` parsing in zaino-fetch, the zcashd validator launchers and `ValidatorKind::Zcashd` in zaino-testutils, and the zcashd-backed live suites (`json_server`, `devtool_zcashd`, and the zcashd arms of the shared suites). Zebra (or another Zaino) is the only supported backing validator.

The test-environment image no longer downloads or builds zcashd: the `ZCASH_VERSION` pin, the zcashd Containerfile stages, and the zcashd-conditional image targets are gone, collapsing the build matrix to final-prebuilt and final-zebrad-source, and the image tag no longer embeds a zcashd version. The `--with-zcashd` flag, the `zcashd_test` task, and the check-no-zcashd guard family are deleted; the guard is moot now that the no-zcashd build is the only build.

Zaino still serves the zcashd-compatible RPC surface to its own clients. That wire contract — field names, error codes, response shapes, the `zcashd_build`/`zcashd_subversion` protobuf fields — is Zaino's product and is unaffected by which validator backs the index, so the comments citing zcashd RPC semantics for the serving surface stay.

This PR is stacked on #1394 and should merge after it: the branch contains #1394's seven commits, so until that PR merges the diff here shows them too. Only the final commit belongs to this PR.

Note for reviewers: `cargo clippy --all-features -D warnings` is red on the base branch already (E0425 on `network`/`pool_lists` in the feature-gated `write_core.rs` code); this PR neither causes nor fixes that.


elicbarbieri (dismissed):
Remove the ADRs explaining why we dropped zebra, and delete all old mentions of it.  No need to confuse agents.  Its gone, we will never talk about it again

elicbarbieri (approved):
LGTM

PR #1414, opened 2026-07-17
docs: propose the Zaino driving port design

zancas:
## What this proposes

This pull request presents the design of Zaino's **driving port** — the contract through which consumers drive Zaino — as a code-free proposal, so the design can be discussed and judged on its own rather than inside an implementation diff. It adds one directory, `docs/driving-port/`, holding two kinds of document:

- **`CONTEXT.md`** — the glossary that fixes the port's language: thirteen terms (Driving port, Snapshot, Conformance kit, Block locator, Reported upgrades, Broadcast, Tip event, Fork point, Treestate, Transaction status, Spend status, Mempool view, Transient failure), each with the meanings we committed to and the near-synonyms we deliberately avoid.
- **Three ADRs** (architecture decision records) recording the load-bearing decisions, each with its rejected alternatives.

The port spans snapshot-pinned chain reads, an explicit chain tip-change subscription, a mempool stream that stands apart from chain state, and transaction broadcast. Zallet is its first driver; zainod-for-lightclients is its expected second.

## The three decisions

1. **ADR 0001 — Decouple the mempool from chain state.** Zallet today learns of a new block through a side effect: Zaino's mempool stream ends when the tip changes. The port retires that idiom. It provides an explicit tip-change subscription, and its mempool stream survives tip changes, tagging every delivery with the tip it was validated against; drivers compose the two by resubscribing on tip events.
2. **ADR 0002 — Raw consensus-serialized bytes at the boundary.** Payloads (blocks, transactions, treestate frontiers) cross the port as bytes; only identifiers and locators are typed, drawn from `zaino-primitives`. Consumers already parse consensus bytes into their own types, and typing payloads with `zcash_primitives` would put Zaino's engines in version-lockstep with the wallet stack.
3. **ADR 0003 — Snapshot pinning is unconditional.** Every read through a snapshot observes the chain as of the pinned tip, and that data stays readable while any clone of the snapshot lives — across reorgs, without exception. The rejected alternative (a `Lapsed` answer engines may give when they have forgotten a pinned view) would export the engine's storage policy into every driver's sync loop.

## How the design was pressure-tested

A complete first-draft implementation was built alongside these documents and then put through a structured review: eight independent finder passes (three correctness angles, three cleanup angles, altitude, and conventions) followed by adversarial verification of all 35 deduplicated candidates. The draft — a contract crate with a scriptable in-memory mock and a conformance kit that is the executable half of this contract — lives on the `docs/driving-port-design` branch and is deliberately **not** part of this pull request.

All ten ranked findings of that review are implemented in the draft. The ones that fed back into the design itself:

- The conformance kit now enforces ADR 0003 across **all thirteen** snapshot capabilities, using harness-supplied handles (an address, an outpoint, a mined txid) for the surfaces the kit cannot fabricate — closing the gap where an adapter answering address queries from the live UTXO set would have passed the kit while serving torn reads.
- Every liveness promise in the kit is held to a deadline that panics naming the broken promise, so a nonconforming adapter fails loudly instead of hanging an adapter developer's CI.
- The driver-side mempool view is contractually a **superset** trued up by resubscription: the stream signals arrivals only, and mempool presence is a hint, never authoritative (glossary: *Mempool view*).
- Payload newtypes share their bytes (cloning is O(1)), and the unspent-outpoints capability takes a height range, because both are contract shapes every real engine inherits.

## Open questions for this discussion

The review surfaced five design questions we deliberately left unresolved for this proposal:

1. **Crate name.** The draft crate is `zallet-driving-port`, but the glossary defines the port as the contract through which consumers drive *Zaino* — a port belongs to the hexagon, not to its first adapter. Rename to `zaino-driving-port`?
2. **Vocabulary hoisting.** The port's `BlockId`, `Outpoint`, and treestate types structurally duplicate existing workspace types (`zaino-state`'s `BlockIndex`, two existing outpoint shapes, and `zaino-primitives`' own `Treestate`/`TreeBytes`). Should the shared vocabulary be hoisted into `zaino-primitives`, which both sides already depend on?
3. **Backend error structure.** `BackendError` carries only a retry classification and a message; adapters must stringify structured transport errors. Adding a `source` chain preserves structure but costs the `PartialEq`/`Eq` derives.
4. **Empty domain-error enums.** Twelve capabilities declare empty per-capability error enums. The review's verifier defended them (real semver headroom, grep-able names) against consolidation into a shared uninhabited type; we lean toward keeping them, but it deserves a decision.
5. **Mock helper visibility.** The mock's scripted-chain accessors are `pub` with no in-repo external caller — plausibly the right testing API for out-of-tree adapter tests, but formally wider than anything requires.

## Relationship to code

No code changes in this pull request. The reference draft (crate, mock, conformance kit, 33 passing tests) is on `docs/driving-port-design` for readers who want to see the contract executed; it will be proposed separately once this design settles.

### DIFF
--- a/docs/driving-port/0001-decouple-mempool-from-chain-state.md
+++ b/docs/driving-port/0001-decouple-mempool-from-chain-state.md
@@ -1,0 +1,28 @@
+# Decouple the mempool from chain state in Zaino's driving port
+
+Zallet today learns of a new block through a side effect: Zaino's mempool
+stream ends when the chain tip changes, and the sync loop treats that closure
+as its only new-block signal (zcash/zallet,
+`zallet-core/src/components/sync.rs:136`). We decided that the driving port
+Zallet will consume from Zaino must not carry this coupling. The port serves
+the mempool through an abstraction that stands apart from chain state, because
+the protocol itself imposes no structural relation between them: a transaction
+commits to no block hash. Consensus binds a transaction only to chain
+*state* — shielded anchors, nullifier non-membership, transparent prevouts,
+expiry height, and the consensus branch id — so a mempool view is coherent
+only relative to the tip it was validated against, and the port expresses
+exactly that.
+
+The port therefore provides an explicit chain tip-change subscription, and its
+mempool capability is an independent stream whose view is tagged with the tip
+it was validated against. Zallet composes the two: on a tip event it resyncs
+and resubscribes to the mempool. This costs Zallet a sync-loop rewrite during
+adoption.
+
+## Considered Options
+
+We rejected canonizing the closure idiom (keeping "stream ends on tip change"
+as the contract), because it hides a chain-state dependency in the mempool
+abstraction's fine print and forces every future engine to implement the
+entanglement. We rejected tip polling by Zallet, because it adds latency and
+busy-work to every implementation's hot path.
--- a/docs/driving-port/CONTEXT.md
+++ b/docs/driving-port/CONTEXT.md
@@ -1,0 +1,106 @@
+# The Zallet driving port
+
+This context covers the contract through which Zallet (and later other
+consumers) drives Zaino. The language below was settled in a design review
+held in the Zallet repository, where the consumer side of this contract lives.
+
+## Language
+
+**Mempool**:
+The set of transactions awaiting mining. It is chain-validity-dependent but
+chain-structure-independent: consensus binds its transactions to chain state
+(anchors, nullifiers, prevouts, expiry, branch id), yet its representation and
+service abstraction stand apart from chain state. A mempool view is coherent
+only relative to the chain tip it was validated against.
+
+**Driving port**:
+The contract through which consumers drive Zaino. It spans chain reads,
+mempool streaming, and transaction broadcast. Zallet is its first driver, and
+zainod-for-lightclients is its expected second; zingolib is not a consumer.
+_Avoid_: primary port, wallet API, indexer interface
+
+**Snapshot**:
+A pinned view of the best chain, taken through the port. Every read through
+a snapshot observes the chain as of the tip it was pinned to, and that data
+stays readable while any clone of the snapshot lives — across reorgs. The
+guarantee is unconditional: an engine must retain the pinned view for as
+long as any clone lives, and an engine that cannot is not an implementation
+of the port. The pinned tip is a property of the snapshot, not a query
+against the engine.
+_Avoid_: view (Zallet's consumer-side name for the same idea)
+
+**Conformance kit**:
+The executable half of the port contract: a suite of invariant cases that
+every implementation of the port must pass, shipped with the port itself.
+An engine adapter passing the kit is what qualifies it as an implementation.
+
+**Block locator**:
+A descending-by-height sample of blocks a driver believes are on the chain,
+offered to the port to locate the fork point between the driver's view and
+the best chain. A block's identity is its hash; its height is position, so
+presence is always judged by hash.
+
+**Reported upgrades**:
+The network-upgrade schedule as the validator reports it, ascending by
+activation height, each upgrade active or pending relative to the current
+tip. The port passes the validator's schedule through — activation heights
+come from the validator, never from constants of the port's own — and
+drivers feed it into their node-compatibility checks.
+
+**Broadcast**:
+Submission of a transaction to the network through the port — there is no
+side channel. Acceptance returns the txid and means the engine admitted the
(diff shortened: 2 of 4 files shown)


docs/driving-port/0001-decouple-mempool-from-chain-state.md
```diff
@@ -0,0 +1,28 @@
+# Decouple the mempool from chain state in Zaino's driving port
+
+Zallet today learns of a new block through a side effect: Zaino's mempool
```
idky137, line 3: This is not a side effect, it is an API that was specifically designed and agreed upon in both zallet API meetings and the lightclient working group calls. This is misleading.

docs/driving-port/0001-decouple-mempool-from-chain-state.md
```diff
@@ -0,0 +1,28 @@
...
+stream ends when the chain tip changes, and the sync loop treats that closure
+as its only new-block signal (zcash/zallet,
+`zallet-core/src/components/sync.rs:136`). We decided that the driving port
+Zallet will consume from Zaino must not carry this coupling. The port serves
```
idky137, line 7: This is confusing to me, the mempool requires this functionality in some form, even if not for zallet then for the light-client interface. #1416 separates this functionality from the core mempool service so that users that dont want it are not forced to use it. 

If we are separating the mempool from the ChainIndex / ChainView as discussed then why would we tie zallet into any specific configuration? It should be up to the zallet devs to decide whether they want just the MempoolService or the CoherentMempoolService.

docs/driving-port/0001-decouple-mempool-from-chain-state.md
```diff
@@ -0,0 +1,28 @@
...
+the mempool through an abstraction that stands apart from chain state, because
+the protocol itself imposes no structural relation between them: a transaction
+commits to no block hash. Consensus binds a transaction only to chain
+*state* — shielded anchors, nullifier non-membership, transparent prevouts,
+expiry height, and the consensus branch id — so a mempool view is coherent
```
idky137, line 12: I think there is a bit of confusion about the issue / bug that lead to the greater tip tracking we have, it is not that a transaction must be tied to a particular block hash, the issue is that the mempool must be correct for the chain being served.

The first problem is, if a reorg occurs and the mempool and non-finalised state are out of sync, transactions could be missing completely (not shown in mempool or chain), or shown twice (both in mempool and chain), the fix that #1416 adds for the coherentmempool (freeze until the two re-agree) fixes this race condition.

The second is that clients that use the `get_mempool_stream` method **as intended** see a separate race condition: non-finalised state serves the chain_tip to clients, they then call get_mempool_stream with the given chain tip, and if out of sync with the mempool, return none, also fixed in #1416.

docs/driving-port/0001-decouple-mempool-from-chain-state.md
```diff
@@ -0,0 +1,28 @@
...
+only relative to the tip it was validated against, and the port expresses
+exactly that.
+
+The port therefore provides an explicit chain tip-change subscription, and its
+mempool capability is an independent stream whose view is tagged with the tip
+it was validated against. Zallet composes the two: on a tip event it resyncs
+and resubscribes to the mempool. This costs Zallet a sync-loop rewrite during
```
idky137, line 19: See above, this is not just a Zallet API, it is also a part of the lightclient interface.

docs/driving-port/0001-decouple-mempool-from-chain-state.md
```diff
@@ -0,0 +1,28 @@
...
+adoption.
+
+## Considered Options
+
+We rejected canonizing the closure idiom (keeping "stream ends on tip change"
```
idky137, line 24: No we have not. we must keep this behaviour, just separate from core mempool functionality.

docs/driving-port/CONTEXT.md
```diff
@@ -0,0 +1,106 @@
+# The Zallet driving port
```
idky137, line 1: Are we sure we want a single bespoke driving port for zallet? what are the benefits / disadvantages vs providing generic, functionality based driving ports? 

Eg. providing MempoolService, ChainView, BlockChainSource, etc and letting Zallet use what they want how they want, rather than tying them into a set configuration.

idky137 (changes requested):
I think there are some incorrect statements in `docs/driving-port/0001-decouple-mempool-from-chain-state.md` that need to be addressed, I think we should also consider not tying zallet into a set configuration and providing a modular system for them to use.

PR #1416, opened 2026-07-20
Feat/mempool rework

idky137:
## Mempool subsystem rework

Rebuilds Zaino's mempool from the ground up and lifts it out of `zaino-state` into two dedicated crates behind hexagonal ports/adapters. The previous mempool lived inside `zaino-state` on a general-purpose `Broadcast` (shared `DashMap` + watch channel) and carried a cluster of confirmed defects (audit **AP-05**) plus a mempool-height indexing bug; a follow-up audit (**AP-12**) raised seven residual findings. This PR closes all of them. See **ADR-0007** for the design rationale.

---

  ### Architecture

- **Two new crates, hexagonal layering.** `zaino-mempool` holds the ports (traits) and foundational types — depends only on `zebra-chain` (no `zaino-state`, no `zaino-proto`). `zaino-mempool-rpc` holds the runtime: the polling core service, the coherence service, and the read handles. `zaino-state` supplies the adapters.
- **Tip-agnostic core + tip-aware coherence.** The core is an always-live, never- frozen mirror of the validator's mempool that serves `getrawmempool` / `getmempoolinfo` / `GetMempoolTx` and a `MempoolUpdate` change feed. An optional coherence layer (feature `tip_aware_mempool`) adds freeze/thaw dual-tip coherence for `get_raw_transaction` / `get_transaction_status` / the raw-tx stream, so the *live* reads never block on tip transitions. Plus a validator-only mode.
- **Single-source rule.** The mempool's validator tip comes from the *same* fetcher that serves the mempool data, never from `ReadStateService`. The core tags every set with that tip (`source_tip`); coherence is then a **pure function of (core set + source_tip, NS)** — it never re-fetches and needs no before/after guards, resolving freeze/thaw by comparing `source_tip` to Zaino's non-finalized tip. A tag-stability guard discards any poll the tip moved across, so the tag and the data are always a single-source pair.
- **Bounded, lock-free, copy-free.** Serving is `ArcSwap` (no lock on the read path); transactions fan out as shared `bytes::Bytes` (refcount bump, no per- subscriber clone); the change feed is bounded with an explicit `Lagged` signal; inputs and memory are bounded by a ZIP-401 cost model.
- **It is a read model, not a validator.** No validation, gossip, fee policy, dependency resolution, or eviction — a bounded, coherent, local view.

---

### Vulnerabilities & bugs fixed

**Coherence correctness — the bug the whole tip-aware layer exists to fix:** (**Zallet sync loop bug**)
- **Mempool ↔ chain-tip incoherence race.** Zaino reads the mempool over JSON-RPC — there is no atomic "mempool + its tip". The old mempool had no notion of the tip its data was valid for, so tip-coherent reads (`get_raw_transaction`, `get_transaction_status`, the raw-tx stream) could serve a set fetched against a validator tip that differed from the caller's chain snapshot — reporting a tx as in-mempool while the caller's chain already had it *mined* (same tx in both best chain and mempool), or the inverse across a block boundary. Coherence now gates those reads on `source_tip == NS` and freezes otherwise.

**Original cluster (AP-05):** (**Private vuln / bug disclosures, tracked internally**)
- **Unbounded mempool exclude filter** — a DoS: a client could drive unbounded work with a large/short exclude list. Now bounded (count + min suffix length), rejecting empty/near-empty suffixes with `InvalidArgument`.
- **O(N²) polling / full-map rescans** on every mempool update. Now O(N) diff-based.
- **Unbounded & duplicated buffers**, **per-subscriber byte clones** — replaced by a bounded feed and shared `Bytes`.
- **`GetMempoolStream` snapshot→open tip race** — a chain-tip change between taking the snapshot and opening the stream served a stale-height stream. Now guarded (retryable error).
- **Mempool-height indexing bug** — unconfirmed txs reported the chain-tip height instead of the wire "unconfirmed" sentinel.

**Phase 1 (merge blockers + audit fixes):** (**Internal audit**)
- **Lagging-consumer silent stream end** — a slow `GetMempoolStream` consumer ended exactly like a normal tip-change close, so a client took a *partial* mempool for the whole one. Now a retryable error.
- **Unmodelled source error silently dropped a transaction** from the set. Fixed.
- **Tip-coherent reads blacked out for a poll interval after every block** (indefinitely under sync lag) — the coherence layer is now signalled on NS advance.
- **Capacity-refused txs were re-fetched every poll**, and the heavy verbose listing could time out on a busy validator — both left the mempool degraded exactly when the chain was busiest.
- **`GetTransaction` returned "not found"** for a tx in the validator's mempool but not yet tip-coherent — a wallet could be told its own transaction didn't exist. Now retryable `Unavailable`.

**Phase 2 (AP-12):** (**Internal rework audit**)
- **Fetch not bounded by the cost cap (Medium)** — additions were materialised in full, then refused; against a non-conforming validator the code meant to stop the blow-up performed it first. Now truncated *before* the fetch, with an early-out at the bound.
- **Grindable admission order** — at capacity, admission followed reversed-txid bytes (sender-controlled), letting an attacker displace honest txs from Zaino's view. Now keyed on `(entry_time, entry_height, per-process-salted hash(txid))`.
- **One refusal froze every tip-coherent read chain-wide** — completeness now splits into *short* (accurate but missing known txids) vs *possibly-wrong*; only the latter freezes. Negative lookups stay precise per-txid via `unadmitted`.
- **Metadata coalescing extended the post-block freeze** — deferred polls now still publish removals + tip re-tag, so `metadata_min_interval` no longer stalls thaw.
- **Low-water underflow** (retry threshold `0` for small bounds); redundant pre-fetch tip read removed (C4); startup rejection of a sub-floor cost bound.
    
---

### Key design decisions

- **`unadmitted` set on the snapshot, not set-level gating.** Gating negative lookups on "is the set short?" would make *every absent* txid `Unavailable` — the common case. Precise per-txid gating is the point.
- **A short set is served, never withheld.** Freezing doesn't make missing txs appear; it hides the ones we have. Rejected serving a full txid listing over a short entry set (internal inconsistency) — the served set stays consistent, with `completeness` + `unadmitted` carrying the honest signal.
- **Stream shortness is log+metric, not in-band** — the lightwallet wire has no field for it and `MempoolStreamError` is terminal.
- **`entry_height` is never derived locally** — it's a protocol field the validator owns; deriving it would be exact only under an assumption that can silently fail. The whole-mempool verbose walk is retained to keep it authoritative.
- **Boundary conversions as named methods** (`to_wire`/`try_from_wire`, `from_business`/`into_business`), enforced by a CI lint.

---

### Wire-compatibility (deliberate, client-visible)

- Mempool `RawTransaction.height` is `0` (unconfirmed sentinel), not the tip height.
- `GetMempoolTx` excludes a tx only on a *unique* shortened-txid match; suffixes `< 4` bytes are rejected (`InvalidArgument`).
- Mempool-but-not-coherent lookups return `Unavailable`, not `Ok(None)`.

---

### Observability

New metrics: `zaino.mempool.short_set_streams_total`, `zaino.mempool.coherence_frozen_seconds` (escalates a freeze outlasting normal thaw).

---

### Docs

- ADR-0007 (subsystem separation)
- ADR-0008 (documentation workbook)
- per-crate `usage.md` guides
- implementation `audit.md`
- mempool lifecycle doc.

---

### Future work (recorded, deferred)

- **Zebra `getmempoolentry`-by-txid** would remove the whole-mempool verbose walk that dominates per-poll cost.
- Compact-tx cache at the RPC boundary (C2); a completeness-carrying library stream for an embedder that must *refuse* a short set - both deferred until a need is raised.

---

### Testing

Full mock-driven matrix in the mempool crates (core + coherence suites, **both** feature modes), `zaino-state` mockchain integration for the ChainIndex-routed reads, and a small regtest live set. Green across `makers lint` (fmt, clippy incl. `--all-features`, rustdoc, cargo-deny), `check-no-zcashd`, `lint-boundary-conversions`,
  and **`makers test live` — 117/117**.


idky137:
Self review: `zaino-mempool-rpc` should be renamed `zaino-mempool-service`, as it is not necessarily rpc specific (driven ports were added which could be implemented with non rpc based fetch services)

idky137:
superceded by #1439

PR #1429, opened 2026-07-28
 fix(chain-index): wait for the validator's chain instead of failing at start-up

### LINKED ISSUE #1006: zaino assumes genesis block is confirmed
### Usage mode

Both / Not sure

### Describe the bug

Ugh. This is a Zaino bug. Zaino assumes the
   genesis block is committed. We have a 10 second delay to work around this but there is of course a race condition.

### Steps to reproduce

I'm researching currently.

### Expected behavior

Not sure.

### Environment

Crosslink workshop

### Configuration

```toml

```

### Logs

```shell

```

### Additional context

_No response_

steven3002:
Fixes #1006. Zaino's start-up raced the validator's: if the validator had not yet committed its genesis block, `Mempool::spawn` failed immediately and the daemon never started. It now waits for the chain the same way it already waits for the mempool.

## Root cause

`Mempool::spawn` already anticipated a validator that isn't up, it polls `get_mempool_txids()` in a loop. The very next call had no such tolerance:

```rust
// Wait for mempool in validator to come online.
loop {
    match fetcher.get_mempool_txids().await {
        Ok(_) => break,
        Err(_) => { info!("Waiting for Validator mempool to come online"); sleep(3s).await; }
    }
}

let best_block_hash = match fetcher.get_best_block_hash().await {
    Ok(Some(hash)) => hash.into(),
    Ok(None) => return Err(MempoolError::Critical(...)),   // ← no retry
    Err(_)    => return Err(MempoolError::Critical(...)),
};
```

`get_best_block_hash()` returns `Ok(None)` in exactly one situation: **no block is committed yet.** So a validator that was up and answering mempool queries, but had not yet committed genesis, produced a hard `Critical` and killed start-up. That is the issue's "zaino assumes the genesis block is committed", precisely, and why the reported workaround was an external 10-second delay, which can only narrow the window, never close it.

A second, later failure point compounds it. The chain-index sync loop opens each iteration with `get_best_block_height()`, and a validator with no chain reaches it as the same `SyncError::ErrorFromSource` a *broken* validator produces:

```rust
let chain_height = source.get_best_block_height().await
    .map_err(source_error)?                                      // ← validator broken
    .ok_or_else(|| source_error(std::io::Error::other(
        "node returned no best block height")))?;                // ← validator has no chain yet
```

The loop cannot distinguish the two, so the steady-state budget, 10 consecutive failures, ~40 s, designed for a validator that *was* working and then broke, applied to start-up. On expiry the worker does not merely report `CriticalError`; it `return`s, so it is gone and a validator arriving later can never revive it.

## The fix

**`mempool.rs`**, fold the chain check into the existing wait loop, with a distinct log line per condition. This is the primary fix.

**`chain_index.rs`**, add `startup_max_consecutive_failures` (80 ≈ 10 min), applied until the loop's first successful iteration. The steady-state budget of 10 is unchanged, so a validator that breaks mid-run still escalates in ~40 s. Defence in depth: with the mempool fix in place the sync loop should not see a chainless validator, but the two failure modes are independent and the loop should not be the thing that assumes otherwise.

## Verification

Reproduced first, then fixed. The reproduction test passes against the *unfixed* tree, showing the failure is terminal:

```
WARN  sync loop iteration failed, retrying  consecutive_failures=1   max=10  backoff=25ms
...
ERROR sync loop failed, giving up           consecutive_failures=10
```

— after which the source recovered and mined 5 blocks, and the tip never moved.

Tests added:

| Test | Proves |
|---|---|
| `waits_out_validator_that_is_unavailable_at_startup` | Regression test: construction waits instead of dying, then syncs once the validator arrives |
| `critical_error_is_terminal_and_source_recovery_does_not_revive_sync` | The original reproduction, the worker is *gone*, not merely unhealthy |
| `startup_budget_outlasts_steady_state_budget` | The property the fix rests on; if it ever fails, the fix has silently become a no-op |

Also adds a `build_index_with_source` test helper. None of the existing helpers could express "validator unavailable at construction", they build the source internally and block until synced, so the scenario was previously untestable.

**366/366 tests pass** (`cargo nextest run --no-default-features`), `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean.

### DIFF
--- a/packages/zaino-state/src/chain_index.rs
+++ b/packages/zaino-state/src/chain_index.rs
@@ -868,6 +868,22 @@ pub(crate) struct SyncTimings {
     pub(crate) initial_backoff: Duration,
     pub(crate) max_backoff: Duration,
     pub(crate) max_consecutive_failures: u32,
+    /// Failure budget applied *before the loop has ever completed an
+    /// iteration*, i.e. while waiting for the validator to become usable at
+    /// start-up.
+    ///
+    /// A validator that has not yet committed its genesis block reports no best
+    /// block height, which reaches the loop as the same
+    /// `SyncError::ErrorFromSource` a broken validator produces. The two are
+    /// indistinguishable here, so the steady-state budget — designed for a
+    /// validator that *was* working and then failed — used to give zaino only
+    /// ~40 s to wait for a node that was still starting up, after which the
+    /// worker returned and could never recover (issue #1006).
+    ///
+    /// Start-up therefore gets its own, much larger budget. It stays bounded so
+    /// a misconfigured validator address still fails loudly instead of hanging
+    /// the daemon forever.
+    pub(crate) startup_max_consecutive_failures: u32,
 }
 
 impl Default for SyncTimings {
@@ -877,6 +893,9 @@ impl Default for SyncTimings {
             initial_backoff: Duration::from_millis(250),
             max_backoff: Duration::from_secs(8),
             max_consecutive_failures: 10,
+            // ~10 minutes: 7.75 s of ramp (250 ms doubling to the 8 s cap)
+            // followed by 74 × 8 s.
+            startup_max_consecutive_failures: 80,
         }
     }
 }
@@ -890,6 +909,10 @@ impl SyncTimings {
             initial_backoff: Duration::from_millis(25),
             max_backoff: Duration::from_millis(800),
             max_consecutive_failures: 10,
+            // Only ~2× the steady-state budget, not the production 8×: start-up
+            // tests must observe the loop *outliving* the steady-state window
+            // without waiting a full production-scale grace period.
+            startup_max_consecutive_failures: 20,
         }
     }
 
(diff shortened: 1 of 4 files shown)


zancas:
Thank you for this work. Your analysis of the two failure points was correct.
We are closing this PR, because `dev` replaced both code paths after you
opened it.

1. `Mempool::spawn` and its call to `get_best_block_hash()` are gone. #1439
   replaced that mempool with the `zaino-mempool-service` crate.
   `zaino-state/src/chain_index/mempool.rs` now contains two adapters only, so
   the primary fix has no code to change.
2. Start-up moved to the chain head in #1440. `anchor_with_retry` in
   `zaino-chain-head-service/src/service.rs` tries again when the validator
   cannot supply a tip. The default limit is ten failures. The delay doubles
   from 500 ms to a maximum of 30 s. Zaino therefore waits about two minutes
   for the validator. The test `spawn_retries_a_transient_failure` covers the
   retry.

One of your points remains open. `anchor_with_retry` uses the same limit as the
steady-state loop. A validator that needs more than two minutes to commit its
genesis block still stops the start-up. Your separate, longer limit for
start-up is the correct idea. The runtime design in #1550 owns the start-up
sequence, so a note there records your proposal.

PR #1402, opened 2026-07-13
feat: modular sync engine — 10 new crates replacing zaino-state/fetch monolith

nachog00:
## Overview

Replaces the monolithic `zaino-state` + `zaino-fetch` architecture with 10 isolated crates following hexagonal architecture. The sync engine, source adapters, persistence layer, and index definitions are fully decoupled — each crate has a single responsibility and minimal dependencies.

## Crate Architecture

See [docs/sync-engine/crate-architecture.md](docs/sync-engine/crate-architecture.md) for the full Mermaid diagram with port-adapter relationships and dependency flow invariants.

```mermaid
graph TB
    subgraph domain_core ["Domain Core"]
        primitives["zaino-primitives<br/><i>Domain types</i>"]
        sync["zaino-sync<br/><i>DAG sync engine</i>"]
    end

    subgraph ports ["Ports (Trait Crates)"]
        source["zaino-source<br/><i>Driven port: validator queries</i>"]
        persistence["zaino-persistence<br/><i>Driven port: storage</i>"]
    end

    subgraph adapters ["Adapters"]
        zebra_rpc["zaino-source-zebra-rpc<br/><i>Zebra JSON-RPC adapter</i>"]
        zebra_rs["zaino-source-zebra-readstate<br/><i>Zebra ReadState adapter</i>"]
        lmdb["zaino-backend-lmdb<br/><i>LMDB storage adapter</i>"]
    end

    subgraph orchestration ["Orchestration"]
        indexes["zaino-indexes<br/><i>Zcash index definitions + sets</i>"]
    end

    subgraph infra ["Shared Infrastructure"]
        rpc["zaino-rpc<br/><i>JSON-RPC 2.0 client</i>"]
        convert["zaino-convert-zebra<br/><i>zebra-chain to domain conversion</i>"]
    end

    source -->|"depends on"| primitives
    persistence -->|"depends on"| primitives
    sync -->|"depends on"| persistence
    sync -->|"depends on"| primitives
    indexes -->|"depends on"| sync
    indexes -->|"depends on"| primitives

    zebra_rpc -.->|"implements<br/>GetBlock, GetChainTip,<br/>GetTreestate"| source
    zebra_rpc -->|"depends on"| rpc
    zebra_rpc -->|"depends on"| convert
    zebra_rs -.->|"implements<br/>GetBlock, GetChainTip"| source
    zebra_rs -->|"depends on"| convert
[...]

PR #1432, opened 2026-07-31
Isolate zaino-primitives, zaino-rpc and zaino-source from #1402 and ready for production

idky137:
# Isolate zaino-primitives, zaino-rpc and zaino-source and ready for production

## What this does, and why

Takes the `zaino-primitives` / `zaino-source` foundation drafted in **#1402** and lands it as a **rewire** rather than an addition: the old code is removed, `zaino-fetch` is deleted, and every consumer moves onto the new stack in the same PR.

`BlockchainSource` was Zaino's abstraction over the backing validator. Its purpose was right; its boundary was wrong — a 34-method trait whose signatures returned `zebra_chain`, `zebra_rpc` and `zaino_fetch` types. Three consequences:

- **Errors were unclassifiable.** `BlockchainSourceError` had two variants, both "Unrecoverable". The distinction the code most needed — *"the validator answered, and the answer is: no such block"* vs *"the validator could not be reached"* — could not be expressed. `zaino-serve` recovered zcashd error codes by downcast-walking for a `zaino-fetch` type; the sync loop hand-rolled a retry ladder that just counted consecutive failures.
- **Semantics lived in comments.** `get_block(HashOrHeight)` behaved differently per variant (the hash path fell back to JSON-RPC for side-chain blocks; the height path did not). `get_best_block_hash` / `get_best_block_height` were separate calls, so callers could observe a torn pair.
- **Capability was conflated with preference.** A 3,145-line enum matched in every method to answer two unrelated questions: *can* this transport serve the query, and *should* it when both can.

Landing this first means the mempool rework (#1416) rebases onto it, NFS/ChainHead isolation builds atop it, and the remaining sync work rebases last.

## What's in it

Eight new crates, in dependency order. Each carries a `usage.md`.

```
zaino-primitives   (thiserror only)         domain types, no serde
      ↑
zaino-source       (+ tokio)                36 single-question ports, QueryError, Resilient
      ↑                    ↑
zaino-rpc                  zaino-convert-zebra
(transport only)           (zebra-chain → domain)
      ↑                    ↑
zaino-source-zebra-rpc     zaino-source-zebra-readstate
                  ↑        ↑
              zaino-source-zebra  (ZebraValidator composite)
                       ↑
              zaino-state / zaino-serve
```

Plus `zaino-address` (a leaf isolating the `zcash_keys` / `sapling-crypto` stack), and `zaino-serve/src/rpc/jsonrpc/wire/` — which now owns the served JSON schema.

Three ideas do the work:

1. **`QueryError<E>` separates an answer from a failure.** `Domain(E)` is returned immediately; `Fetch(FetchError)` is retried by `Resilient` according to a machine-readable `FailureMode`. This is what lets a zcashd legacy code survive from the validator to the served response.
2. **Capability is structural.** The read-state adapter does not implement the mempool ports, because a state service has no mempool — so mis-routing is a compile error, not a wrong answer.
3. **Preference is a routing table.** `Fetch`-only and `State` become configurations of one type, not enum variants.

`BlockchainSource` survives as documented **temporary scaffolding** with a "do not extend" note. Its single implementation, `ValidatorSource<V>`, is generic over the ports, so production and the test mocks reach ChainIndex through the same conversion code — which turns `mockchain_tests.rs` and `proptest_blockgen.rs` into coverage of that conversion rather than of a parallel implementation of it.

Two ADRs record the reasoning: **ADR-0008** (the port model) and **ADR-0009** (the served JSON schema).

## Problems found, and what was done about them

Eight real defects surfaced. Six were pre-existing and latent; two I introduced and caught.

**1. A missing block looked like an unreachable node.** The RPC adapter constructed **zero** `QueryError::Domain` values — every port already defined the right variant, the adapter simply never produced one. So "no block at that height", which the sync loop asks on *every iteration*, arrived as an unrecoverable transport fault and exhausted the retry ladder against a healthy validator. Caught by the live suite: a test that passed on `dev` in 0.9s failed after 40s of retries.
→ Added classification helpers, one per class of method rather than one per method. The `-8`-as-absent reading is safe **only** where every parameter is rendered from a domain type (a `Height` is always a decimal `u32`, a hash always 64 hex chars); the address-keyed methods take `-5` only, because there `-8` means a malformed envelope — a Zaino bug — and must stay a failure. This distinction is written down, not just implemented.

**2. zcashd error-code recovery was silently inert.** `zaino-serve` downcast-walked for `zaino_fetch`'s `RpcError`, which the new stack never constructs, so every code reached clients as a generic internal error. A downcast that never matches simply falls through — which is why it survived.
→ Retargeted at `FailureMode::RpcError(i64)` and a new typed `LegacyRpcError` for Zaino's own rejections, with tests on the walker itself.

**3. `getblockdeltas` was broken against zebrad.** The read-state derivation was omitted with the rationale that it avoided "a second copy of logic the validator already has, for no capability gain". That was **false**: zebrad does not implement `getblockdeltas` at all.
→ Ported the derivation; the composite routes to it *first*. The module doc now records why the original reasoning was wrong.

**4. `getblockchaininfo` was broken against zebra 6.0** — it serialises the deferred-development-fund pool as `lockbox` where zcashd calls it `deferred`.
→ Both spellings accepted, so the served answer no longer depends on which validator is behind the adapter.

**5. Network upgrade names disagreed between transports** — `Nu5` (RPC, from the validator's reply) vs `NU5` (read-state, from `Debug`). Zebra's `Display` is its `Debug`, and neither is the wire name.
→ Take it from serde.

**6. 13 tests had never run.** `zaino-source`'s `mock` module was gated on a `testing` feature that **nothing in the workspace enables**, so the module never compiled and its tests silently never ran — invisible to `cargo nextest`, because a module that isn't compiled reports no failures. It was also hiding a compile error only `--all-features` could reach, which is why `makers lint` could not pass.
→ Gated `any(test, feature = "testing")`. `zaino-source` went 11 → 24 tests.

**7. `getspentinfo` was modelled on the wrong neighbour.** It reached the port as `Result<Option<SpentInfo>, _>`, sharing a helper with `gettxout` whose doc claimed both "return JSON null for absent". True of `gettxout`; false of this. zcashd has no null answer for `getspentinfo` — absence is `-5 Unable to get spent info`, and the code is what clients match on. Modelled as `None`, that code had to be reinvented downstream, and was reinvented wrongly (`-8`). The predecessor was no better: it consumed `-5` into a typed error, destroying the `RpcError` the recovery walks for.
→ `Result<SpentInfo, _>` with `NotSpent` (`-5`) and `Unsupported` (`-32601`). The second matters because **zebrad does not implement the method**: conflating it with `NotSpent` would tell every client that every output is unspent.

**8. An unreachable `Err`.** `clamp_deltas_range_to_tip` returned a `Result` whose error could not be provoked — a bound only reaches `Height::try_from` after being clamped below `tip`, which is already a valid `Height`.
→ Made total. A `Result` nobody can provoke is a test that can never be written.

Also: the mempool stream parsed each transaction **twice** (deserialising the same bytes into a `zebra_chain` transaction and then again into `zaino-fetch`'s `FullTransaction` purely to reach `to_compact`) — fixed, closing an in-code TODO and removing an `.unwrap()`; two dead `TryFrom` impls in `types/db/legacy.rs`, unreachable but invisible to the dead-code lint because trait impls are; and `zaino-state`'s `zcashd_support` feature, which gated nothing.

## Decisions worth flagging for review

- **`z_getblock` and `getrawtransaction` keep zebra's presentation shapes.** They are built from a block's own bytes plus a few chain facts, and modelling them in the domain would give one fact two sources. This means `zebra-rpc` is *not* fully removed from `zaino-state`'s public API — the PR delivers less than "no zebra types in the port" on those two methods, deliberately.
- **Wire conversions are named `from_domain`, not `to_wire`.** CLAUDE.md specifies `to_wire`/`try_from_wire` as inherent methods on the *business* type; that shape isn't available, because the business types live in `zaino-primitives`, whose only dependency is `thiserror`. Giving them a `to_wire` means giving that crate serde — the exact coupling ADR-0008 exists to prevent. ADR-0009 records the exception explicitly rather than quietly breaking the rule.
- **Zaino still does not answer `getspentinfo` from its own index.** It never has — it has been a passthrough since it was added. That is now a *problem* rather than a preference, because zebrad will never implement it: as things stand the method dies with zcashd. Zaino already holds most of the answer (`FinalisedStateReader::get_outpoint_spender`, DB v1.2+, gives `height` directly and `txid` with one more read); what's missing is the input index. Documented as a TODO on the trait, not implemented — it would be new indexer capability, not this rewire.
- **`nonfinalized_listener` is removed** rather than ported. Nothing consumes it, but it is a behaviour change inside a foundation PR.
- **`PolledChainTip`** is built and tested but unwired, deferred with the `SubscribeChainTip` consumer move.

## Size

Whole-repo line counts, `dev` → `HEAD`:

| | dev | HEAD | Δ |
|---|---:|---:|---:|
| Production Rust | 50,507 | 54,769 | **+4,262** |
| Test Rust | 24,839 | 29,256 | **+4,417** |
| Docs (`.md`) | 4,815 | 6,235 | **+1,420** |

Diff churn: **248 files, +21,457 / −11,212**. New package code is +15,875; `zaino-fetch`'s deletion and the `ValidatorConnector` removal account for most of the ~10,300 lines dropped from existing packages.

**This got much bigger than planned, and I want to be straight about it.** The original plan targeted a **net-negative** PR on the grounds that deleting `zaino-fetch` (7,586 lines) would outweigh the new stack. It did not: the net is **+10,099**. 

**I do not think this is comfortably reviewable as a single unit, and I think split it but will leave that choice to the reviewr.** The natural cut lines, in decreasing order of how much they'd help:

If reviewers would rather take it whole, the commits are ordered so each compiles and is green, and reading them in sequence is much easier than reading the diff.

## Verification

- **514 unit tests** pass on default-members (dev: ~486 across the workspace). `zaino-serve` went 2 → 119; the wire modules are the only place served field names are asserted.
- **Full live suite passes** (`clientless` + `e2e`), compared against a `dev` baseline in a separate worktree. All four live-suite failures listed above were found this way and fixed.
- `mockchain_tests.rs` and `proptest_blockgen.rs` pass and now exercise the production conversion layer.
- `cargo fmt --check` clean; `cargo clippy --workspace --all-targets` at the same 5 warnings as `dev`, all in `live-tests/`.
- Test-inventory reconciliation (`cargo nextest list` diffed against `dev`): of 110 dev-only test paths, 72 are renames, 33 have equivalent-or-better replacements, 3 test cases the type system now makes unrepresentable, and **1** was a genuine gap — closed.

## Known gaps

- **`--all-features` does not compile**, on `write_core.rs` behind `transparent_address_history_experimental`. Byte-identical on `dev`; verified, not introduced here. It is the only thing keeping `makers lint` red.
- **Nothing in the live suites exercises `getspentinfo`.** Its five unit tests are its only coverage.
- **Spend deltas remain unreported** where zebra-rpc hardcodes `Input::NonCoinbase { address: None, value_zat: None }`. Reproduced exactly; recovering them is a behaviour change plus N extra reads per transaction.
- **The live-test crates fail `--features zcashd_support`** on an unrelated `zcash_local_net` import. Pre-existing and verified as such, but it means the zcashd live path may not build at all right now — worth someone looking at separately.


idky137:
Closing a duplicate of https://github.com/zingolabs/zaino/pull/1434

PR #1434, opened 2026-08-03
Feat: Isolate zaino-source, zaino-rpc and zaino-primitives

idky137:
# Isolate zaino-primitives, zaino-rpc and zaino-source, and ready for production

## What this does, and why

Takes the `zaino-primitives` / `zaino-source` foundation drafted in **#1402** and lands it as a **rewire** rather than an addition: `zaino-fetch` is deleted and every consumer moves onto the new stack in the same PR. No dead code is introduced at any point.

`BlockchainSource` was Zaino's abstraction over the backing validator. Its purpose was right; its boundary was wrong — a 34-method trait whose signatures returned `zebra_chain`, `zebra_rpc` and `zaino_fetch` types. Three consequences:

- **Errors were unclassifiable.** `BlockchainSourceError` had two variants, both "Unrecoverable". The distinction the code most needed — *"the validator answered: no such block"* vs *"the validator could not be reached"* — could not be expressed. `zaino-serve` recovered zcashd error codes by downcast-walking for a `zaino-fetch` type; the sync loop hand-rolled a retry ladder that just counted consecutive failures.
- **Semantics lived in comments.** `get_block(HashOrHeight)` behaved differently per variant. `get_best_block_hash` / `get_best_block_height` were separate calls, so callers could observe a torn pair.
- **Capability was conflated with preference.** A 3,145-line enum matched in every method to answer two unrelated questions: *can* this transport serve the query, and *should* it when both can.

Eight new crates replace it, in dependency order. Each carries a `usage.md`.

```
zaino-primitives   (thiserror only)      domain types, no serde
      ↑
zaino-source       (+ tokio)             36 single-question ports, QueryError, Resilient
      ↑                  ↑
zaino-rpc                zaino-convert-zebra
(transport only)         (zebra-chain → domain)
      ↑                  ↑
zaino-source-zebra-rpc   zaino-source-zebra-readstate
                ↑        ↑
            zaino-source-zebra  (ZebraValidator composite)
                     ↑
            zaino-state / zaino-serve
```

Plus `zaino-address` (a leaf isolating the `zcash_keys` / `sapling-crypto` stack) and `zaino-serve/src/rpc/jsonrpc/wire/`, which now owns the served JSON schema.

Three ideas do the work:

1. **`QueryError<E>` separates an answer from a failure.** `Domain(E)` returns immediately; `Fetch(FetchError)` is retried by `Resilient` according to a machine-readable `FailureMode`. This is what lets a zcashd legacy code survive from validator to served response.
2. **Capability is structural.** The read-state adapter does not implement the mempool ports, because a state service has no mempool — mis-routing is a compile error, not a wrong answer.
3. **Preference is a routing table.** `Fetch`-only and `State` become configurations of one type, not enum variants.

`BlockchainSource` survives as documented **temporary scaffolding** with a "do not extend" note. Its single implementation, `ValidatorSource<V>`, is generic over the ports, so production and the test mocks reach `ChainIndex` through the same conversion code — which turns `mockchain_tests.rs` and `proptest_blockgen.rs` into coverage of that conversion rather than of a parallel implementation of it.

Two ADRs record the reasoning: **ADR-0008** (the port model) and **ADR-0009** (the served JSON schema).

Landing this first means the mempool rework (#1416) rebases onto it, NFS/ChainHead isolation builds atop it, and the remaining sync work rebases last.

## Issues found in the #1402 implementation, and what changed

Eight real defects surfaced while productionising the draft. Six were pre-existing and latent; two were introduced in the draft and caught here.

| # | Defect | Fix |
|---|---|---|
| 1 | **A missing block looked like an unreachable node.** The RPC adapter constructed **zero** `QueryError::Domain` values — the ports defined the variant, the adapter never produced one. "No block at that height", asked on *every* sync iteration, exhausted the retry ladder against a healthy validator. | Classification helpers, one per *class* of method. The `-8`-as-absent reading is safe only where every parameter is rendered from a domain type; address-keyed methods take `-5` only, because there `-8` means a malformed envelope — a Zaino bug — and must stay a failure. Written down, not just implemented. |
| 2 | **zcashd error-code recovery was silently inert.** `zaino-serve` downcast-walked for `zaino_fetch`'s `RpcError`, which the new stack never constructs. A downcast that never matches falls through, which is why it survived. | Retargeted at `FailureMode::RpcError(i64)` plus a typed `LegacyRpcError` for Zaino's own rejections, with tests on the walker itself. |
| 3 | **`getblockdeltas` was broken against zebrad.** The read-state derivation was omitted on the rationale that it avoided duplicating logic "the validator already has". That was false — zebrad does not implement the method at all. | Ported the derivation; the composite routes to it first. The module doc records why the original reasoning was wrong. |
| 4 | **`getblockchaininfo` was broken against zebra 6.0** — it serialises the deferred-development-fund pool as `lockbox` where zcashd calls it `deferred`. | Both spellings accepted, so the served answer no longer depends on which validator is behind the adapter. |
| 5 | **Network upgrade names disagreed between transports** — `Nu5` (RPC) vs `NU5` (read-state, from `Debug`). Zebra's `Display` *is* its `Debug`, and neither is the wire name. | Take it from serde. |
| 6 | **13 tests had never run.** `zaino-source`'s `mock` module was gated on a `testing` feature nothing enables, so it never compiled and its tests never ran — invisible to nextest, because an uncompiled module reports no failures. It was also hiding a compile error only `--all-features` reaches. | Gated `any(test, feature = "testing")`. `zaino-source` went 11 → 24 tests. |
| 7 | **`getspentinfo` was modelled on the wrong neighbour** — `Result<Option<SpentInfo>, _>`, sharing a helper with `gettxout`. zcashd has no null answer here; absence is `-5`, and the code is what clients match on. Modelled as `None`, it had to be reinvented downstream, and was reinvented wrongly (`-8`). | `Result<SpentInfo, _>` with `NotSpent` (`-5`) and `Unsupported` (`-32601`). The second matters: **zebrad does not implement the method**, and conflating it with `NotSpent` would tell every client that every output is unspent. |
| 8 | **An unreachable `Err`.** `clamp_deltas_range_to_tip` returned a `Result` whose error could not be provoked. | Made total. A `Result` nobody can provoke is a test that can never be written. |

Also fixed: the mempool stream parsed each transaction **twice** (deserialising the same bytes into a `zebra_chain` transaction and again into `zaino-fetch`'s `FullTransaction` purely to reach `to_compact`) — closes an in-code TODO and removes an `.unwrap()`; two dead `TryFrom` impls in `types/db/legacy.rs`, unreachable but invisible to the dead-code lint because trait impls are; and `zaino-state`'s `zcashd_support` feature, which gated nothing.

**Decisions worth flagging:**

- **`z_getblock` and `getrawtransaction` keep zebra's presentation shapes.** They are built from a block's own bytes plus a few chain facts; modelling them in the domain would give one fact two sources. So `zebra-rpc` is *not* fully removed from `zaino-state`'s public API — this delivers less than "no zebra types in the port" on those two methods, deliberately.
- **Wire conversions are named `from_domain`, not `to_wire`.** CLAUDE.md specifies `to_wire`/`try_from_wire` on the *business* type; that shape isn't available, because the business types live in `zaino-primitives`, whose only dependency is `thiserror`. Giving them `to_wire` means giving that crate serde — the exact coupling ADR-0008 exists to prevent. ADR-0009 records the exception rather than quietly breaking the rule.
- **`nonfinalized_listener` is removed** rather than ported. Nothing consumes it, but it is a behaviour change inside a foundation PR.
- **`PolledChainTip` is built but unwired, and has no tests of its own.** `ZebraValidator::with_tip_polling` is the opt-in wiring point; nothing calls it yet. Deferred with the `SubscribeChainTip` consumer move — flagging it because it is the one piece of new production code in this PR that no test exercises.

## Review guide: commit by commit

> **This history is built to be read one commit at a time.** Each commit contains only final-shape code — no commit shows a stage that a later commit corrects — and the branch was rebuilt from the finished tree rather than rebased, so no fixups are folded in anywhere. **If you would rather review this as two or three separate PRs, it splits cleanly at the commit boundaries** (see the note after the table).
>
> Commits 1–9 and 13 are individually green on `cargo check --workspace --all-targets`, `clippy`, `fmt` and `nextest`. Commits 10 and 11 build (`cargo check` / `cargo build`) but do not test — the `#[cfg(test)]` mockchain harness is still old-shaped until commit 12. That was a deliberate trade to keep the ~3.4k-line production rewire separate from ~3.3k lines of test-harness churn.

Every line of every commit is classified by where it actually lives, not by filename: an inline `#[cfg(test)] mod tests` block counts as test code even inside `src/`, and `///` / `//!` doc comments count as documentation even inside a production module. Test code also covers test-only modules (`mock.rs`, `mockchain_source.rs`), `tests/` directories, `examples/`, and everything under `live-tests/`.

| # | Commit | Files | Production | Tests | Docs | Lockfile |
|---|---|---:|---|---|---|---|
| 1 | `feat(primitives): add zaino-primitives` | 48 | 47f +1,313 | 7f +250 | 45f +895 | +7 |
| 2 | `feat(address): add zaino-address` | 7 | 6f +152 | 2f +280 | 4f +167 | +13 |
| 3 | `feat(source): add zaino-source` | 42 | 40f +1,135 | 2f +481 | 39f +597 | +9 |
| 4 | `feat(rpc): add zaino-rpc` | 9 | 8f +383 | 3f +157 | 6f +110 | +14 |
| 5 | `feat(convert): add zaino-convert-zebra` | 4 | 3f +195 | 1f +14 | 1f +35 | +9 |
| 6 | `feat(source): add zaino-source-zebra-rpc` | 17 | 7f +1,721 | 12f +805 | 8f +414 | +22 |
| 7 | `feat(source): add zaino-source-zebra-readstate` | 5 | 4f +1,259 | 1f +28 | 2f +191 | +17 |
| 8 | `feat(source): add zaino-source-zebra` | 6 | 5f +519 | 1f +33 | 3f +82 | +11 |
| 9 | `feat(serve): add the served JSON-RPC wire module` | 22 | 21f +1,766 | 18f +2,515 | 18f +668 | +6 |
| 10 | `refactor: rewire zaino-state and zaino-serve` | 25 | 22f +2,057 −3,566 | 9f +661 −515 | 13f +643 −557 | +9 −3 |
| 11 | `refactor: delete zaino-fetch` | 29 | 24f −4,544 | 15f −2,132 | 23f −1,123 | −51 |
| 12 | `test: rewire the mockchain harness and the live suite` | 41 | — | 38f +3,059 −1,659 | 24f +255 −38 | +14 |
| 13 | `docs: changelogs, usage guides and ADRs` | 30 | — | — | 30f +1,510 −48 | — |
| | **Sum of commits** | **285** | **+10,500 −8,110** | **+8,283 −4,306** | **+5,567 −1,766** | **+131 −54** |

Two notes on reading the table. The per-bucket file counts overlap — a single `.rs` file normally contributes production, test *and* doc lines, so the three bucket counts sum to more than the **Files** column. And the sum of commits (+24,481 / −14,236) exceeds the squashed diff against `dev` (**248 files, +21,457 / −11,212**), because in the squashed view git pairs some deleted `zaino-fetch` files with their replacements as renames and counts only the delta.

**Where the review effort actually is.** Only **10,500 lines of production code are added** across the whole PR, against +8,283 test and +5,567 documentation. Two commits dominate the logic: **6** (`zaino-source-zebra-rpc`, +1,721 prod) and **10** (the rewire, +2,057 / −3,566 prod). Commit **9** looks like the second-largest at 4,949 lines, but only 1,766 of those are logic — over half is inline golden tests.

**1 — `zaino-primitives`.** The domain vocabulary: heights, hashes, blocks, transactions, and the response shapes the JSON-RPC surface is defined in terms of. Zero dependencies on `zebra-chain` or any other zaino crate. *Read:* `src/types/rpc/` (the served shapes), `src/types/block.rs`, `src/types/transaction.rs`.

**2 — `zaino-address`.** A leaf isolating the librustzcash address stack behind one crate. Carries Sapling diversifier/pk_d extraction including zcashd's big-endian byte order, with the test vector pinning it. *Read:* `src/classify.rs`, `src/sapling.rs`.

**3 — `zaino-source`.** The driven-port traits — one per question — plus the source-agnostic machinery built on them: `Resilient` (retry/backoff), `PolledChainTip`, the chain-tip subscription, and the in-memory mock. **This is the conceptual centre of the PR.** *Read:* `src/lib.rs` (the port set), `src/error.rs` (`QueryError` / `FailureMode`), `src/resilient.rs`.

**4 — `zaino-rpc`.** JSON-RPC 2.0 transport only: envelope, HTTP client with cookie/basic auth, connection probe, retry classifier. Knows nothing about Zcash methods. *Read:* `src/retry.rs`, `src/client.rs`.

**5 — `zaino-convert-zebra`.** The single place `zebra_chain` types become domain types. Small and mechanical.

**6 — `zaino-source-zebra-rpc`.** The adapter answering every port over JSON-RPC. Contains the error classification from defect #1 and the `getspentinfo` handling from #7. *Read:* `src/adapter.rs` (classification helpers), `src/parse.rs`. `tests/block_parity.rs` runs the parser against captured mainnet blocks spanning Sapling → NU5+. Also moves raw-transaction validation into `zaino-common::consensus` as a typed `RawTransactionError`.

**7 — `zaino-source-zebra-readstate`.** The adapter backed by zebra's in-process `ReadStateService`. Implements only the ports zebra's state can answer and **does not implement the rest** — that absence is the design. Contains the `getblockdeltas` derivation from defect #3. *Read:* `src/adapter.rs`, especially which traits are absent.

**8 — `zaino-source-zebra`.** The composite. Capability already decides what each adapter *can* answer; this decides which to *ask*. *Read:* `src/routing.rs`, `src/fallback.rs` (a domain miss is retried against the other source, a success never is).

**9 — `zaino-serve` wire module.** The served JSON schema: zcashd's exact field names, hex encodings and byte orders. All new files, no deletions — and **2,515 of its 4,955 lines are inline `#[cfg(test)]` golden tests against 1,766 lines of logic**, so the reviewable surface is a good deal smaller than the line count suggests. Landed ahead of the rewire so it reads on its own; nothing calls it until commit 10, so `validate_address_from_domain` warns as unused in this one commit. *Read:* `wire.rs` (the direction/naming rationale), then any one module — they are all the same shape.

**10 — the rewire.** `ValidatorConnector` (one 3,000-line enum) becomes `ValidatorSource<V>`, generic over the ports. `ChainIndex` and the indexer traits answer in domain types; 21 JSON-RPC methods move onto the domain and `service.rs` renders them through commit 9's wire module. **This is the commit to spend review time on.** *Read:* `chain_index/validator_source.rs` (+1718), `indexer.rs` (+119 −527), `indexer/node_backed_indexer.rs` (+383 −210, includes the mempool double-parse fix), `chain_index/source_ports.rs`, `source_caps.rs`, `chain_index/network_adoption.rs`, `zaino-serve/src/rpc/jsonrpc/service.rs`.

**11 — delete `zaino-fetch`.** Pure deletion; nothing to review beyond confirming nothing was kept. Drops six workspace dependencies it was the sole consumer of (`nu-ansi-term`, `url`, `jsonrpsee-types`, `base64`, `byteorder`, `derive_more`).

**12 — test harness and live suite.** The mockchain now implements the `zaino-source` ports directly, so unit tests exercise `ChainIndex` through the same traits a real adapter satisfies. The live suite gains a validator oracle and a legacy parser — these tests compare Zaino's domain answers against a raw JSON-RPC oracle, and the legacy parser keeps the pre-rewrite decoding available as an *independent* check rather than reusing the code under test. *Read:* `mockchain_source.rs`, `live-tests/zaino-testutils/src/validator_oracle.rs`.

**13 — docs.** Per-crate `CHANGELOG` / `usage.md`, ADR-0008 and ADR-0009.

### If you would rather this were split

It splits at commit boundaries with no rework:

- **1–9 / 10–13.** The first nine commits are purely additive — eight new crates plus the wire module — and land as one PR that touches nothing existing. The cost is that the new crates are unused until the second PR, which is the one thing this history was built to avoid; say the word if you'd prefer it anyway.
- **1–12 / 13.** Docs as a trailing PR. Trivial, low value.
- **10–12 cannot be split further.** `zaino-state`'s new `ChainIndex` API returns domain types, and the old-shaped `zaino-serve` cannot compile against it. Separating them means shipping an intermediate API that a later commit rewrites.

## Testing

- **514 unit tests** pass on default-members (`dev`: ~486 across the workspace). Per binary: `zaino-state` 229, `zaino-serve` 119 (from 2 — the wire modules are the only place served field names are asserted), `zainod` 30, `zaino-primitives` 28, `zaino-source` 24, `zaino-common` 24, `zaino-source-zebra-rpc` 19, `zaino-rpc` 17, `zaino-address` 8, `zaino-source-zebra` 3, `zaino-source-zebra-readstate` 2.
- **Full live suite passes** (`clientless` + `e2e`), compared against a `dev` baseline in a separate worktree. Defects #1, #3, #4 and #5 above were all found this way.
- `mockchain_tests.rs` and `proptest_blockgen.rs` pass and now exercise the production conversion layer rather than a parallel implementation.
- `cargo fmt --all --check` clean. `makers lint-boundary-conversions` passes. `cargo clippy --workspace --all-targets` reports 5 warnings (1 doc-list indentation, 1 too-many-arguments, 3 clone-on-`Copy`); the same 5 are present at the base of this branch, and three further warnings that existed there are removed by this PR.
- **Test-inventory reconciliation** (`cargo nextest list` diffed against `dev`): of 110 dev-only test paths, 72 are renames, 33 have equivalent-or-better replacements, 3 test cases the type system now makes unrepresentable, and **1** was a genuine gap — closed.

## Known gaps and next steps

- **`--all-features` does not compile**, on `write_core.rs` behind `transparent_address_history_experimental`. Verified byte-identical on `dev` — pre-existing, not introduced here. It is the only thing keeping `makers lint` red, and worth a separate fix.
- **The live-test crates fail `--features zcashd_support`** on an unrelated `zcash_local_net` import. Pre-existing and verified as such, but it means the zcashd live path may not build at all right now — worth someone looking at separately.
- **Zaino still does not answer `getspentinfo` from its own index.** It never has — a passthrough since it was added. That is now a *problem* rather than a preference, because zebrad will never implement it: as things stand the method dies with zcashd. Zaino already holds most of the answer (`FinalisedStateReader::get_outpoint_spender`, DB v1.2+, gives `height` directly and `txid` with one more read); what's missing is the input index. Documented as a TODO on the trait, not implemented — that would be new indexer capability, not this rewire.
- **Nothing in the live suites exercises `getspentinfo`.** Its five unit tests are its only coverage.
- **Spend deltas remain unreported** where zebra-rpc hardcodes `Input::NonCoinbase { address: None, value_zat: None }`. Reproduced exactly; recovering them is a behaviour change plus N extra reads per transaction.
- **Follow-on work this unblocks:** the mempool rework (#1416) rebases onto these ports; NFS/ChainHead isolation builds atop them; the remaining sync work rebases last. `PolledChainTip` gets wired with the `SubscribeChainTip` consumer move.


packages/zaino-primitives/src/types/rpc/block_deltas.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub outputs: Vec<OutputDelta>,
+}
+
+/// One transparent input, spending a previous output.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct InputDelta {
+    /// Address the spent output paid to.
+    pub address: TransparentAddress,
+    /// Value moved. Always negative — this is value leaving the address.
+    pub satoshis: SignedZatoshis,
```
elicbarbieri, line 71: zatoshis or satoshis?

packages/zaino-primitives/src/types/rpc/block_deltas.rs
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/// One transparent output, paying an address.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OutputDelta {
+    /// Address paid.
+    pub address: TransparentAddress,
+    /// Value moved. Unsigned because an output can only add value — the wire
+    /// form is a non-negative amount, and the type says so.
+    pub satoshis: Zatoshis,
```
elicbarbieri, line 87: Zatoshis?

packages/zaino-primitives/src/types/transaction.rs
```diff
@@ -0,0 +1,107 @@
...
+use super::{
+    EncryptedCiphertext, EphemeralKey, NoteCommitment, Nullifier, OutputIndex, Script,
+    SignedZatoshis, TransactionHash, TxIndex, Zatoshis,
+};
+
+/// A transaction within a block.
+#[derive(Debug, Clone)]
+pub struct Transaction {
+    /// Transaction id.
+    pub txid: TransactionHash,
```
elicbarbieri: Add clarity/separate between TransactionHash and TransactionID?

In NU 1-4, txid == txhash (sha256 of serialized tx)
From NU 5 onward, txid != txhash

Probably just rename TransactionHash to TransactionID and add a few lines of docs and point to: https://github.com/zcash/zips/blob/main/zips/zip-0244.rst

packages/zaino-primitives/src/types/zatoshis.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Create a zatoshi amount, validating against the protocol maximum.
+    pub fn new(amount: u64) -> Result<Self, ZatoshisOverflow> {
+        if amount > MAX_ZATOSHIS {
+            Err(ZatoshisOverflow { got: amount })
+        } else {
+            Ok(Self(amount))
+        }
+    }
+
+    /// Checked addition.
```
elicbarbieri: This is where documentation is needed.  Should an overflow return none or throw that error?

elicbarbieri:
Overall I think this looks really good.  Im exhaused by the comments, and part of me wants to change the linting/cargo-fmt rules to remove comments for all struct members.  Such a waste of codebase length to have:

```rust

struct BlockHeader {
    // block hash
    pub hash: BlockHash,
    
    // block number
    pub number: u64
}
```

elicbarbieri:
I did reproduce passing tests on my machine via makers test all

```
     Summary [ 439.873s] 63 tests run: 63 passed (7 slow), 7 skipped
[cargo-make][3] INFO - 
[cargo-make][3] INFO - Build File: Makefile.toml
[cargo-make][3] INFO - Task: notify
[cargo-make][3] INFO - Profile: development
[cargo-make][3] INFO - Running Task: notify
[cargo-make][3] INFO - Build Done in 0.62 seconds.
[cargo-make][2] INFO - Build Done in 517.08 seconds.
[cargo-make][1] INFO - Build Done in 518.18 seconds.

====================== test summary ==========================
  packages:           514 run,  514 passed,    0 failed,    3 skipped,     18.448s
  clientless:          54 run,   54 passed,    0 failed,   14 skipped,    235.711s
  e2e:                 63 run,   63 passed,    0 failed,    7 skipped,    518.190s
  TOTAL:              631 run,  631 passed,    0 failed,   24 skipped,    772.349s
==============================================================
[cargo-make] INFO - Build Done in 773.04 seconds.
```

packages/zaino-primitives/src/types/rpc/block_deltas.rs
idky137: Confusingly the field in the RPC response type is satoshis - https://zcash.github.io/rpc/getblockdeltas.html (due to zcash originally being a fork of bitcoin and this never being updated).

If this was an internal type I would definitely agree it should be zatoshis but due to it being an RPC type I'm less sure what is best. Im happy to update to zatoshis if we want though.

packages/zaino-primitives/src/types/rpc/block_deltas.rs
idky137: Similar to above, I'm happy to change this if we want though.

packages/zaino-primitives/src/types/transaction.rs
idky137: Updated in [d8f2cff](https://github.com/zingolabs/zaino/pull/1434/commits/d8f2cff6f6c342243361aea30b5280259942791a)

packages/zaino-primitives/src/types/zatoshis.rs
idky137: Updated in [a59bd5c](https://github.com/zingolabs/zaino/pull/1434/commits/a59bd5cb96c7fa68b0dd4cefd2801b25d9aca466)

idky137:
> Overall I think this looks really good. Im exhaused by the comments, and part of me wants to change the linting/cargo-fmt rules to remove comments for all struct members. Such a waste of codebase length to have:
> 
> ```rust
> struct BlockHeader {
>     // block hash
>     pub hash: BlockHash,
>     
>     // block number
>     pub number: u64
> }
> ```

Ye I'm not sure what the best solution is here, w had an issue of there being lots of undocumented code so enforcing the lint was a good way of reminding everyone to add documentation, its true that it does not always lead to the best / most useful documentation though. I wonder if there is a middle ground that would give devs a bit more freedom whilst still requiring code to be documented?

zancas:
I'm in favor of updating the lints to remove redundant docs.

zancas:
@idky137 can you look into this CI fail:  https://github.com/zingolabs/zaino/actions/runs/31091647352/job/92583812634?pr=1434  ?

zancas:
Here's CoPilot's fix:

```Rust
async fn call_parsed_or_handle<T, E, F>(
    &self,
    method: &str,
    params: Vec<serde_json::Value>,
    parse: impl FnOnce(&serde_json::Value) -> Result<T, parse::ParseError>,
    handle_error: F,
) -> Result<T, QueryError<E>>
where
    E: std::fmt::Debug + std::fmt::Display,
    F: FnOnce(zaino_rpc::RpcError) -> QueryError<E>,
{
    let value = self
        .rpc
        .call(method, params)
        .await
        .map_err(handle_error)?;
    parse(&value).map_err(|e| QueryError::Fetch(from_parse(e)))
}

async fn call_parsed_or_absent<T, E>(
    &self,
    method: &str,
    params: Vec<serde_json::Value>,
    parse: impl FnOnce(&serde_json::Value) -> Result<T, parse::ParseError>,
    absent: impl FnOnce() -> E,
) -> Result<T, QueryError<E>>
where
    E: std::fmt::Debug + std::fmt::Display,
{
    self.call_parsed_or_handle(method, params, parse, |error| {
        absent_or_fetch(error, absent)
    })
    .await
}

async fn call_parsed_or_invalid_address<T, E>(
    &self,
    method: &str,
    params: Vec<serde_json::Value>,
    parse: impl FnOnce(&serde_json::Value) -> Result<T, parse::ParseError>,
    invalid: impl FnOnce(String) -> E,
) -> Result<T, QueryError<E>>
where
    E: std::fmt::Debug + std::fmt::Display,
{
    self.call_parsed_or_handle(method, params, parse, |error| {
        invalid_address_or_fetch(error, invalid)
    })
    .await
}
```

idky137:
> @idky137 can you look into this CI fail: https://github.com/zingolabs/zaino/actions/runs/31091647352/job/92583812634?pr=1434 ?

Fixed in [0a3c83c](https://github.com/zingolabs/zaino/pull/1434/commits/0a3c83c908c1b91bd39b893b1a8771b4c4ba6c84)

nachog00 (approved):

Issue #1164, opened 2026-05-31
Add multi-arch Docker images, particularly ARM64 alongside the default AMD64.


nmarley:
### Security check

- [x] This issue is **not** a security vulnerability.

### Usage mode

zainod (running the daemon)

### Problem or use case

I'm trying to pull the zainod image on my Apple Silicon Macbook but it's not able to because of the architecture. Also, ARM64 is becoming more and more mainstream, especially with AWS' Graviton instances and being less expensive to run and generally a bit faster. I think this is a reasonable request in 2026. I have a PoC with these building and am happy to open a PR shortly to add this to this repo.

### Proposed solution

The GitHub actions runners already have native arm64 in addition to the amd64 ones, so I plan to open a PR w/a build matrix with both amd64 and arm64 runners to build via `docker buildx` and then a manifest job to use `docker buildx` to merge the two architectures into a multi-arch manifest.

### Alternatives considered

QEMU emulation is a possibility but it's generally inefficient. Additionally, this should be a simple add with lots of benefit. Low downside, high upside and I've done the work to add this already.

### Area

Other

### Additional context

This relations to the packaging and distribution via Docker images.

subashmokshya:
@nmarley's outline here is exactly the right shape — a native-runner build matrix plus a manifest merge job, no QEMU. Since the issue has been open a couple of months without the PR landing, I went ahead and implemented it in #1435:

- amd64 and arm64 are each built on their native GitHub runner (`ubuntu-latest` / `ubuntu-24.04-arm`), for both image flavors (default and `-no-tls`), pushed by digest, and merged into multi-arch manifests tagged by the existing metadata rules.
- Dry-runs (`workflow_dispatch` with `push=false`) now work without DockerHub credentials — login, digest export and manifest assembly are skipped — so forks and feature branches can verify builds.
- Registry build caches are scoped per flavor+arch (`buildcache-<flavor>-<arch>`) to avoid cross-architecture cache pollution.

Green verification run on my fork covering all four legs: https://github.com/subashmokshya/zaino/actions/runs/30882013050

@nmarley if your PoC covers ground this doesn't, happy to defer or combine.

Issue #866, opened 2026-02-25
Explore Docker image size optimizations


nachog00:
The current release image (`0.2.0-rc.1`) is 120MB, which is already reasonable for a Rust binary on a slim Debian base.

There are some further optimizations that could be explored if image size becomes a concern in the future:

- Stripping debug symbols from the binary
- Enabling LTO (Link Time Optimization) in the release profile
- Using size-optimized compiler settings
- Switching to an Alpine base image with musl libc

These are not urgent and would require benchmarking to understand the trade-offs between size, build time, and runtime performance.

PR #1435, opened 2026-08-04
ci: publish multi-arch (amd64+arm64) Docker images via native runners

subashmokshya:
### Summary

Publishes multi-arch (linux/amd64 + linux/arm64) Docker images for zaino/zainod, closing #1164, using the approach outlined in that issue: native runners per architecture (no QEMU), digest pushes, and a manifest-merge job.

### Changes

- `docker` job becomes a 4-leg matrix: {amd64, arm64} × {default, no-tls}, with arm64 on GitHub's free `ubuntu-24.04-arm` runners.
- Each leg pushes by digest only; a new `merge` job assembles per-arch digests into one multi-arch manifest per flavor and applies the existing metadata-action tags (both `zaino` and `zainod` names).
- Dry-run mode (`workflow_dispatch`, `push=false`) no longer requires DockerHub secrets: login/push/merge are conditionally skipped while the image is still fully built (`type=cacheonly`). Previously a dry-run attempted DockerHub login and registry cache writes, which fails on forks and feature branches without secrets.
- Registry build caches scoped per flavor+arch (`buildcache-<flavor>-<arch>`) to prevent cache pollution between architectures.
- `create-release` unchanged, now depends on `merge`.

### Verification

Green `workflow_dispatch push=false` run on my fork — all four build legs on a clean cache, including both arm64 flavors:
https://github.com/subashmokshya/zaino/actions/runs/30882013050

### Note on the second commit

The branch includes the Dockerfile from #1430 as a clearly-labeled, verification-only commit: without its stale-pin fix (`protobuf-compiler=3.21.12-3` → `3.21.12-3+deb12u1`), no image builds from a clean checkout on any architecture (this is what failed my first verification run). I'll drop that commit and rebase as soon as #1430 lands — or fold the one-line pin bump in here if maintainers prefer.

### Follow-ups I'm happy to do

- `linux/arm/v7` and/or ghcr.io publishing
- image size work per #866

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,13 +14,116 @@ on:
 
 jobs:
   docker:
-    name: Docker images (build+push)
-    runs-on: ubuntu-latest
+    name: Build ${{ matrix.flavor }} (${{ matrix.platform }})
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      # One failing leg should not cancel the others: per-arch results are
+      # independently useful when debugging a platform-specific failure.
+      fail-fast: false
+      matrix:
+        include:
+          # Native runners per architecture — no QEMU emulation. The arm64
+          # lane uses GitHub's free arm64 runners for public repos.
+          - platform: linux/amd64
+            runner: ubuntu-latest
+            arch: amd64
+            flavor: default
+            no-tls: 'false'
+          - platform: linux/arm64
+            runner: ubuntu-24.04-arm
+            arch: arm64
+            flavor: default
+            no-tls: 'false'
+          - platform: linux/amd64
+            runner: ubuntu-latest
+            arch: amd64
+            flavor: no-tls
+            no-tls: 'true'
+          - platform: linux/arm64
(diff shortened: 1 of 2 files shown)


zancas:
Thank you for this work. We are closing this PR in favour of #1528, which
adds the same images and fits the present release pipeline.

Both PRs build each architecture on a native runner and push by digest. This PR
writes the matrix and the manifest merge by hand. #1528 uses the reusable
workflow `docker/github-builder`, which does the same work in fewer lines.
#1528 also handles the `zainod-X.Y.Z` tags that the release pipeline now
pushes. This PR started before that pipeline existed.

Two more points apply. `dev` has more than 500 commits since the base of this
PR, and `release.yaml` conflicts. The second commit copies the Dockerfile from
#1430 for verification, and that commit must not merge.

One part of this PR has value for #1528. Your dry-run builds all four images
with no DockerHub secrets, so a fork can verify the workflow. Your green run
proves that the arm64 images build. A review comment on #1528 asks for the same
dry-run.

PR #1436, opened 2026-08-04
Add Stress testing to Zaino Integration Tests

elicbarbieri:
Based on the work from #1378


live-tests/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+[workspace.dependencies]
+# Workspace member.
+zaino-testutils = { path = "zaino-testutils" }
+
+# ztest — the k8s-based test harness that owns the validator / indexer / wallet
+# lifecycle and hands tests typed RPC handles. Lives in a sibling repo
+# (`ZingoLabs/ztest`); requires a sibling checkout for path resolution. The
+# tests drive Zaino out-of-process (in a pod) through this, so the live-test
+# crates link no zaino/zebra production code — hence no rocksdb / libstdc++.
+ztest = { path = "../../ztest" }
```
idky137, line 27: Would it be possible to update this to a git ref rather than a path so that a manual checkout is not required? Otherwise if there is a reason this is not possible could we add a script that pulls and builds the required repo?

I think keeping test set up as simple as possible for newcomers and external contributors would help in having everyone run all tests before submitting code.

PR #1439, opened 2026-08-10
Feat - Mempool Rework

idky137:
# Mempool subsystem: replace the `Broadcast` mempool with `zaino-mempool`

> **Base branch:** this PR is built on top of `chore/isolate_zaino_consensus_and_status`
> and should be merged after it. Review the diff against that branch, not `dev`.

## What this does

Zaino's mempool lived inside `zaino-state` on a general-purpose `Broadcast`
(a shared `DashMap` + a watch channel) and carried a cluster of confirmed
defects: an unbounded exclude filter, O(N²) polling, unbounded and duplicated
buffers, per-subscriber byte clones, full-map rescans, a mempool-stream tip
race, and a mempool-height indexing bug.

It is now its own subsystem — a bounded, coherent, local read model of the
validator's mempool, explicitly *not* a validator mempool. Split hexagonally
across two new crates:

- **`zaino-mempool`** — domain types and ports. Depends on `zaino-primitives`
  and `zaino-source` only; no `zaino-state`, no `zaino-proto`, no node library.
- **`zaino-mempool-service`** — the runtime: the tip-agnostic polling core, the
  tip-aware coherence layer (feature-gated on `tip_aware_mempool`), and the
  read handles.

The validator side reuses `zaino-source` (ADR 0008) rather than defining a
bespoke port: `MempoolPorts` is a consumer-defined bound over
`GetMempoolTxids`, `GetMempoolMetadata`, `GetRawMempoolTransaction`,
`GetMempoolSourceTip` and `SubscribeBlocks`.

### The tip-agnostic / tip-aware split

Reads that *list* the mempool (`getrawmempool`, `getmempoolinfo`) stay live
across a tip transition. Reads that *place* a transaction relative to a chain
tip (`getrawtransaction`'s mempool arm, transaction status, `GetMempoolStream`)
go through the coherence layer and return `Unavailable` rather than an answer
computed against a stale tip — an unmined transaction's consensus branch id is
derived from the height it would be mined at, and a caller cannot tell a wrong
branch id from a right one.

The sync loop pushes an epoch signal on each non-finalized publication so the
coherence layer thaws on the block rather than on its next poll tick, and warns
(plus a Prometheus gauge) if a freeze outlives `COHERENCE_FREEZE_ESCALATION`.

### Also in here

- `RawTransaction.data` is served as `Bytes`, so mempool transaction bytes reach
  the wire without a per-subscriber copy.
- `zaino-rpc` bounds response bodies (`MAX_RESPONSE_BYTES`) and adds per-request
  timeouts (`HEAVY_METHOD_TIMEOUT` for the methods the validator answers by
  walking a whole data structure).
- Mempool bounds (`max_cost_bytes`) are operator-configurable via `zainod`
  config, live-updatable across service clones.

See `docs/adr/0010-mempool-subsystem-separation.md` for the full rationale.

## Code changes by commit

Counted per commit at `-U0`, so a modified line is one addition **and** one
deletion (git's own convention — there is no separate "changed" count).
Classification: path first (`docs/**`, `*.md` → documentation;
`live-tests/**`, `**/tests/**`, `**/tests.rs` → test), then per line inside a
`.rs` file (inside a `#[cfg(test)]` block → test; `///`, `//!`, `#[doc]` →
documentation; otherwise production).

| Commit | Files | Production | Test | Documentation |
|---|---:|---:|---:|---:|
| `e83da12` fix(rpc): bound response bodies and add per-request timeouts | 4 | +55 / −6 | +27 / −0 | +38 / −0 |
| `c9ad484` fix(ci): drop the stale live-tests workspace from the proto-heavy check | 1 | +9 / −4 | +0 / −0 | +0 / −0 |
| `6bddfc1` feat(source): add the mempool sourcing ports | 11 | +237 / −6 | +79 / −0 | +178 / −1 |
| `3f38a83` feat(mempool): add zaino-mempool | 16 | +538 / −0 | +64 / −0 | +1149 / −0 |
| `f5d8ee9` feat(mempool-service): add the tip-agnostic service and read handle | 14 | +979 / −1 | +1057 / −0 | +512 / −10 |
| `6d84dfc` feat(mempool-service): add the tip-aware coherence layer | 5 | +518 / −0 | +548 / −0 | +65 / −1 |
| `d1342a7` feat(proto): serve RawTransaction.data as Bytes | 9 | +23 / −11 | +0 / −0 | +8 / −1 |
| `e56dedf` refactor(state): replace the Broadcast mempool with the mempool subsystem | 34 | +622 / −1061 | +287 / −626 | +203 / −80 |
| `247ee5f` feat(zainod): make the mempool bounds operator-configurable | 3 | +51 / −3 | +86 / −0 | +30 / −0 |
| `56b833d` test: mempool coverage across mockchain and the live suite | 4 | +0 / −0 | +143 / −123 | +0 / −0 |
| `ecf7b99` docs: ADR, changelogs and usage guides for the mempool subsystem | 6 | +0 / −0 | +0 / −0 | +232 / −2 |
| **Total** (85 distinct files) | **85** | **+3032 / −1092** | **+2291 / −749** | **+2415 / −95** |

Roughly: 39% production, 29% test, 32% documentation by added lines. The one
net-negative commit is `e56dedf`, which deletes the old `Broadcast` mempool
(`packages/zaino-state/src/broadcast.rs`, `chain_index/mempool.rs`) and its
tests as it swaps in the new subsystem.

### Where the changes land

- **New crates:** `packages/zaino-mempool/`, `packages/zaino-mempool-service/`
- **Ports:** `packages/zaino-source/src/get_mempool_{metadata,source_tip}.rs`,
  `get_raw_mempool_transaction.rs`, plus the `zaino-source-zebra-rpc` adapter
- **Integration:** `packages/zaino-state/src/chain_index.rs`,
  `chain_index/mempool_ports.rs` (removed: `broadcast.rs`,
  `chain_index/mempool.rs`)
- **Config / ops:** `packages/zainod/src/config.rs`, `src/metrics.rs`
- **Docs:** `docs/adr/0010-mempool-subsystem-separation.md`,
  `packages/zaino-mempool/docs/{audit,mempool_lifecycle}.md`, the `usage.md`
  guides, and per-crate CHANGELOGs


packages/zaino-mempool/src/snapshot.rs
```diff
@@ -0,0 +1,171 @@
...
+    pub mempool_generation: u64,
+
+    /// Monotonic event sequence (increments on each published snapshot).
+    pub event_sequence: u64,
+
+    /// Entries indexed by txid.
+    pub by_txid: Arc<HashMap<TransactionId, Arc<MempoolEntry>>>,
+
+    /// Txids sorted by [`reversed_txid_key`], for shortened-txid suffix lookup.
+    pub txids_sorted: Arc<[TransactionId]>,
```
nachog00: **Blocking.** This ordering (`reversed_txid_key`) is the precondition for the shortened-txid suffix binary search, but `MempoolSnapshot` exposes all fields `pub` with no checked constructor (only `empty_not_ready`). If a builder sorts by the wrong key or skips a re-sort, the search silently misses matches — no panic, no test failure.

Make the fields private and add a constructor that owns the sort and derives `by_txid`/counts/`cost_bytes`, so an inconsistent snapshot can't be built. Fold the documented "unadmitted empty iff Complete" invariant in there too.

Confirmed reproducible: a test that builds a snapshot with a present txid but an unordered `txids_sorted` and calls `get_filtered_entries` fails — the exclude filter leaks the txid it was asked to hide.

packages/zaino-mempool/src/config.rs
```diff
@@ -0,0 +1,152 @@
...
+    /// undelivered updates a subscriber may fall behind before it is told to
+    /// resync (`MempoolUpdate::Lagged`). State-losslessness does not depend on it —
+    /// a lagged consumer recovers the full set from `current()` — so it can be kept
+    /// safely bounded. Since buffered updates carry no snapshots (only entries and
+    /// small facts), memory is `~capacity` small slots regardless of subscriber
+    /// count. The default trades a generous window for that bounded cost.
+    pub event_buffer_len: usize,
+
+    /// How often the update loop polls the source when no wake signal arrives.
+    pub poll_interval: Duration,
```
nachog00: **Blocking (parse-don't-validate).** `poll_interval`/`metadata_min_interval` are plain `Duration`, fed from the `[mempool]` TOML via `from_millis` with no floor: `poll_interval_ms = 0` becomes `Duration::ZERO`, which panics `tokio::time::interval` at spawn (`` `period` must be non-zero ``). `event_buffer_len = 0` panics `broadcast::channel`. Only `max_cost_bytes` is validated.

Model these as `NonZeroU64`/`NonZeroUsize` and give `MempoolConfig` a checked constructor instead of `pub` fields, so `0` is rejected at deserialize rather than reaching the runtime.

Confirmed reproducible: `load_config` with `poll_interval_ms = 0` returns `Ok` today.

packages/zaino-mempool-service/src/service.rs
```diff
@@ -0,0 +1,807 @@
...
+        let mut added_meta: Vec<MempoolTxMeta> = Vec::new();
+
+        if !added_txids.is_empty() {
+            // How many additions can possibly fit, assuming every one costs the
+            // ZIP-401 floor. An entry can only cost *more*, so this over-counts
+            // and the exact check in `publish_snapshot` still decides.
+            let max_cost_bytes = self.config.max_cost_bytes();
+            let headroom = max_cost_bytes.saturating_sub(current.cost_bytes);
+            let max_admissible =
+                (headroom / zaino_mempool::config::MEMPOOL_TRANSACTION_COST_THRESHOLD) as usize;
```
nachog00: `as` numeric cast (forbidden by the no-`as` rule) — and a real truncation edge: `(u64 / u64) as usize` truncates on a 32-bit target when `max_cost_bytes` is large, corrupting `max_admissible`. Use `usize::try_from(...)`.

Same conversion for the other new-code casts: service.rs:567, subscriber.rs:133, entry.rs:79, parse.rs:179/184/1109, client.rs:193. A diff-scoped lint would keep new `as` casts out.

nachog00 (commented):
## Mempool rework — review

Reviewed against the stacked base `chore/isolate_zaino_consensus_and_status` (not `dev`): 11 commits, 85 files. Strong PR — clean domain/service split behind `zaino-source` ports, each of the 7 named defects has a targeted test, and the tip-race / capacity-bound / no-per-subscriber-copy fixes hold up. Docs are thorough (ADR-0010, per-crate CHANGELOGs + usage guides). Four blocking items, all fixable without design change.

### Blocking

1. **`MempoolSnapshot` invariants aren't enforced by the type** (`snapshot.rs`) — all fields `pub`, only `empty_not_ready()`. The suffix binary search needs `txids_sorted` ordered by `reversed_txid_key`; if a builder drifts from that the search silently misses matches (no panic, no test failure). Private fields + a checked constructor that owns the sort and accounting. *(inline)*
2. **Config accepts illegal values** (`zaino-mempool/config.rs`, `zainod/config.rs`) — `poll_interval_ms = 0` → `Duration::ZERO`, which panics `tokio::time::interval` at spawn; `event_buffer_len = 0` panics `broadcast::channel`. Only `max_cost_bytes` is validated. Parse-don't-validate: `NonZeroU64`/`NonZeroUsize` + a checked constructor so `0` is rejected at deserialize. *(inline)*
3. **No tracing in the two new crates** — neither `zaino-mempool` nor `zaino-mempool-service` depends on `tracing` or emits any. Two brand-new crates with zero instrumentation is immediate tech debt. Minimum bar before merge: add `tracing`, `#[instrument]` the poll/reconcile loops, and edge-log state transitions (completeness changes, freeze/thaw, task start/stop) with structured fields. Related: **stop swallowing source errors** — the poll loop discards validator errors into `IncompleteSourceError` unlogged, so the mempool degrades silently. *(inline)*
4. **`as` numeric casts in new code** violate the no-`as` rule; `service.rs:352` is also a 32-bit truncation edge. Convert to `try_from` (service.rs:352/567, subscriber.rs:133, entry.rs:79, parse.rs:179/184/1109, client.rs:193). A diff-scoped lint would keep new ones out. *(inline)*

> Blocking items 1 and 2 are **confirmed reproducible** with local failing unit tests that drive the real code paths — #1 through `get_filtered_entries` (the exclude filter silently leaks a present txid when `txids_sorted` is unordered), #2 through `load_config` (accepts `poll_interval_ms = 0`). They'll ride along on the fix branch.

### Notes (non-blocking)

- **Background-task lifecycle** — the poll tasks stop only via the `CancellationToken`; dropping the service `Arc` alone doesn't (the task holds its own `Arc<Self>`). Correct as composed (`NodeBackedChainIndex::Drop` cancels the tree), but it means `zaino-mempool-service` isn't self-sufficient standalone — a consumer that just drops it leaks the task. Either document the shutdown contract on `spawn()`, or have the service cancel its own child token on `Drop` (keeping the parent token for hierarchical shutdown). Minor: teardown calls `mempool.close()` but leaves coherence to token-cancel only.
- **`GetRawMempoolTransaction` returns `Vec<u8>`** — consistent with the deliberate `Raw*` byte-exactness carve-out, not new debt. Worth one line in ADR-0008 registering the exception (or a thin `RawTxBytes` newtype) if the rule is read strictly.
- **Pre-existing, not this PR:** the duplicate `zaino.chain.tip_height` emission in `chain_index.rs` (a hardcoded literal alongside the `CHAIN_TIP_HEIGHT` constant) comes from the base branch (`0c2709a8`); flag for that PR / a follow-up.

packages/zaino-mempool-service/src/service.rs
```diff
@@ -0,0 +1,807 @@
...
+
+    async fn tick(&self, state: &mut PollState) {
+        // Tag: the validator tip this poll's fetch window opens at.
+        //
+        // Read fresh every poll, never carried over from the last one: this read
+        // is also how a tip *change* is detected over an otherwise unchanged
+        // mempool, which the coherence layer depends on to thaw.
+        let tip_before = match self.source.get_mempool_source_tip().await {
+            Ok(tip) => BlockRef::from_tip(tip),
+            Err(_) => return self.source_error(state),
```
nachog00: Source errors are discarded here (`Err(_) => …`) and collapse into `IncompleteSourceError` with nothing logged — the mempool degrades silently. Same at the txids arm below (306), metadata (375), and the raw-fetch error (590, where a built `MempoolError` is dropped).

Don't swallow: log the error with structured fields where it's discarded, and `warn!` on the transition into the degraded state. Part of the crate-level tracing gap (both new crates currently have no `tracing` dependency at all).

packages/zaino-mempool-service/src/coherence.rs
```diff
@@ -0,0 +1,564 @@
...
+            mode: MempoolMode::Frozen {
+                valid_for: prev.valid_for,
+                reason,
+            },
+            valid_for: prev.valid_for,
+            observed_tips: observed,
+            event_sequence: next_sequence,
+        });
+        self.coherent.store(snapshot);
+        let _ = self.events.send(Arc::new(MempoolEvent::Frozen {
```
nachog00: The `FreezeReason` is emitted only as a broadcast event, never traced. An operator who sees the upstream freeze-escalation warning has no log of *why* coherence froze. Emit the freeze/thaw transition with the reason as structured fields (`debug!` on freeze, `info!` on thaw).

packages/zaino-mempool/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+pub use config::MempoolConfig;
+pub use entry::MempoolEntry;
+pub use error::MempoolError;
+pub use ports::{BlockRef, Mempool, MempoolPorts};
+pub use snapshot::{reversed_txid_key, MempoolCompleteness, MempoolSnapshot};
+pub use update::MempoolUpdate;
+
+/// Re-exported from [`zaino_source`] so adapters and consumers of this crate can
+/// name the metadata type without also depending on the port crate.
+pub use zaino_source::MempoolTxMeta;
```
nachog00: Non-blocking, but this re-export is dead and its rationale doesn't hold — suggest removing it.

- **Unused:** `zaino_mempool::MempoolTxMeta` is referenced nowhere but this line. Every consumer names `zaino_source::MempoolTxMeta` directly (`zaino-state` mempool_ports.rs:88 / validator_source.rs:449 / mockchain_source.rs:638, `zaino-mempool-service` service.rs:42).
- **Not load-bearing:** `MempoolTxMeta` doesn't appear in this crate's public API (no port/fn signature surfaces it), so nothing here forces consumers to name it.
- **Rationale is counterfactual:** the doc says it lets consumers avoid depending on the port crate, but every consumer already depends on `zaino-source` directly.

More generally: a `pub use` of a foreign type isn't an abstraction boundary — it's the same type, so it adds no encapsulation, creates two paths to one name, and couples this crate's public API to `zaino-source`'s type churn for no gain. Re-export only types your own public API surfaces. If the goal is for `zaino-mempool` to own its metadata vocabulary independently, define a domain type here + a named boundary conversion from `zaino_source::MempoolTxMeta` (the `try_from_wire`/`from_business` idiom) rather than a re-export.

packages/zaino-mempool/src/ports.rs
```diff
@@ -0,0 +1,234 @@
...
+use crate::update::MempoolUpdate;
+
+/// A minimal `(hash, height)` reference to a block.
+///
+/// A named pair rather than a tuple because the mempool compares these for
+/// equality in the one place that decides whether a transaction set is still
+/// coherent with the chain; `tip.hash != other.hash` reads correctly where
+/// `tip.0 != other.0` invites the wrong field.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct BlockRef {
```
nachog00: This duplicates an existing primitive. `zaino-primitives` already has `BlockRef` (`types/rpc/address_deltas.rs:64`) with the identical shape — `{ hash: BlockHash, height: Height }`, same field order, same types (`derive(Debug, Clone, Copy, PartialEq, Eq)`). This one differs only by an extra `Hash` derive and the `from_tip` constructor.

"A block named by hash+height" isn't mempool domain — it's a chain-wide primitive; the mempool just uses it as the source-tip tag. And `zaino-mempool` already depends on `zaino-primitives` (it imports `BlockHash`/`Height` from there), so reusing it costs no new dependency.

Suggest: promote a single `BlockRef` into `zaino-primitives` as a top-level type (the existing copy is itself mis-homed under `rpc/address_deltas` for a general concept), give it the superset derives (`+ Hash`) and a `from_tip` / `From<(BlockHash, Height)>`, and have both this bound and `address_deltas` use it. One canonical type instead of two that will drift.

packages/zaino-mempool/src/ports.rs
```diff
@@ -0,0 +1,234 @@
...
+/// tags each published set with
+/// [`get_mempool_source_tip`](zaino_source::GetMempoolSourceTip::get_mempool_source_tip)
+/// so the coherence layer can judge the set's coherence without re-fetching it,
+/// and that comparison is only sound for a single-source pair. `ZebraValidator`
+/// upholds this by routing all four to JSON-RPC; see
+/// [`GetMempoolSourceTip`](zaino_source::GetMempoolSourceTip)'s documentation.
+///
+/// `Clone` is required because the core clones the source to fan out bounded,
+/// concurrent raw-transaction fetches, so implementations must be cheap to clone.
+pub trait MempoolPorts:
```
nachog00: Naming nit: `MempoolPorts` names the architectural role ("a bag of ports") rather than the capability the combined bound represents. A trait bound reads best as a capability — *what a type that satisfies it can do* — not as its place in the hexagon.

What this actually says is "a single transport that can source a validator's mempool coherently" (the five reads, the single-source rule, plus block signals). Something like `MempoolSource` (or `ValidatorMempoolAccess`) states that capability directly and reads naturally at every bound: `impl<S: MempoolSource>` / `where S: MempoolSource`.

Trade-off: there's a sibling `ChainIndexSourcePorts` in `zaino-state` using the same `...Ports` convention, so renaming diverges from it. Worth it — ideally both move to capability names, and this new one is the place to start.

packages/zaino-mempool/src/snapshot.rs
```diff
@@ -0,0 +1,171 @@
...
+///   [`IncompletePendingMetadata`](Self::IncompletePendingMetadata)). Serving it
+///   is more useful than withholding it; the missing txids are named in
+///   [`MempoolSnapshot::unadmitted`].
+/// - **Possibly wrong** — the set may not reflect the source at all
+///   ([`IncompleteSourceError`](Self::IncompleteSourceError),
+///   [`NotReady`](Self::NotReady)). Only these justify a freeze.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum MempoolCompleteness {
+    /// No set has been built yet.
+    NotReady,
```
nachog00: Non-blocking modeling note: `NotReady` conflates readiness (a lifecycle state) with completeness (the fidelity of an existing set). The other four variants describe *a set that exists*; `NotReady` describes *the absence of one*.

Three tells it's on the wrong axis:
- The live classifier never emits it: `completeness_for` only returns `Complete`/`IncompleteCapacityLimited`/`IncompletePendingMetadata`, and errors set `IncompleteSourceError`. `NotReady` is set only in `empty_not_ready()` (the pre-first-poll sentinel) — the state machine transitions out of it, never into it.
- Readiness is already tracked elsewhere — twice: `StatusType` (`Spawning`/`Syncing`/`Ready`) and `MempoolMode::NotReady` in the coherence view. So "no set yet" is encoded three times.
- It muddies the enum's own invariants: `may_be_wrong()` lumps `NotReady` with `IncompleteSourceError` (different reasons, same outcome), and the documented "`unadmitted` empty iff `Complete`" is really "iff `Complete` or `NotReady`".

Not a bug (both `may_be_wrong` states correctly freeze). Suggestion: let `MempoolCompleteness` describe only an existing set (drop `NotReady`) and represent "not ready" on the axis that already owns it — the status/mode — or make the core set `Option` so a consumer matching completeness never has to handle a lifecycle case, and "empty iff Complete" holds again.

packages/zaino-mempool/src/tip.rs
```diff
@@ -0,0 +1,178 @@
...
+
+use zaino_primitives::types::TransactionId;
+
+use crate::entry::MempoolEntry;
+use crate::ports::{BlockRef, NonFinalizedEpoch};
+use crate::snapshot::MempoolSnapshot;
+
+/// The tip of the source that supplies mempool data ("V").
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ValidatorTip {
```
nachog00: Non-blocking: `ValidatorTip` is a single named field over `BlockRef`, and nothing ever touches it but `.best_tip` (`agree()` compares `validator.best_tip.hash`). The named field + its doc line is pure ceremony when the struct *is* a `BlockRef` — and it reads oddly next to `NonFinalizedEpoch`, which legitimately has a second field (`generation`).

Home is right, though: it's the "V" operand of the V-vs-NS coherence diff, so it's coherence-domain and belongs here under `tip_aware` — it only means anything the moment it's compared to the NS tip.

On shape, two better options:
- Newtype: `pub struct ValidatorTip(pub BlockRef);` with `impl Deref<Target = BlockRef>` (or a `hash()`), so call sites stay `validator.hash` rather than `validator.0.hash`. Keeps the role tag, drops the field + doc.
- Or drop the wrapper and use `BlockRef` directly (`validator: Option<BlockRef>`): the NS side is already a distinct type, so there's no sibling `BlockRef` to confuse it with, and `ObservedTips.validator` already names the role.

Either is better than the single-named-field struct.

packages/zaino-source-zebra-rpc/src/adapter.rs
```diff
@@ -315,12 +321,33 @@ impl ZebraRpcAdapter {
...
-        self.call_parsed_classified(method, params, parse, |error| {
+        self.call_parsed_classified(method, params, None, parse, |error| {
             QueryError::Fetch(error.into())
         })
         .await
     }
 
+    /// [`call_parsed`](Self::call_parsed) with a per-request timeout, for the
+    /// methods the validator answers by walking a whole data structure.
+    async fn call_parsed_slow<T, E>(
```
nachog00: Non-blocking design note: timeout here is encoded as a method-name axis (`call_parsed` vs `call_parsed_slow`), but the adapter already has a second name-axis for error classification (`call_parsed_or_absent`, `call_parsed_or_invalid_address`, `call_parsed_optional`). Those two multiply: a heavy method that also needs `or_absent` would force a `call_parsed_or_absent_slow`. That cross-product is the incoherence.

The general form already exists — `call_parsed_classified(method, params, Some(timeout), parse, err_mapper)` — and `call_parsed_slow` is just it with the timeout preset. Cleaner to treat timeout as a *parameter* each method supplies (default `None` = client timeout), composing freely with any error classification, and drop `_slow`.

Fair caveat: Rust has no default args, so a bare `timeout` param would add `None` to ~20 terse call sites — presumably why the split exists. Keeping the terse `call_parsed` for the common case while routing heavy calls through `call_parsed_classified` with the method's own value avoids both the noise and the variant explosion.

Related (see the summary): `HEAVY_METHOD_TIMEOUT` is 30s, identical to the default `request_timeout` (30s), so `call_parsed_slow` buys zero headroom today. A per-method value (e.g. the whole-mempool `getrawmempool verbose` walk gets its own, larger timeout) is exactly the "each method informs its own timeout" model and would make the mechanism actually do something.

packages/zaino-source/src/get_mempool_source_tip.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use std::future::Future;
+
+use zaino_primitives::types::{BlockHash, Height};
+
+use super::QueryError;
+
+/// Domain error for [`GetMempoolSourceTip`].
+#[derive(Debug, thiserror::Error, Clone, PartialEq, Eq)]
+pub enum GetMempoolSourceTipError {
```
nachog00: Please address the correctness of these mempool error types — three domain variants added here are never produced in production:

- `GetMempoolSourceTipError::NotReady` — `get_mempool_source_tip` uses `call_parsed` (over `getblockchaininfo`), whose classifier maps every failure to `QueryError::Fetch`. Never yields `Domain`. (Only the test mock constructs `NotReady`.)
- `GetMempoolTxidsError::Unavailable` — `get_mempool_txids` uses `call_parsed`. Same.
- `GetMempoolMetadataError::Unavailable` — `get_mempool_metadata` uses `call_parsed_slow`, whose classifier is also `|e| Fetch(e.into())`. Same.

So in production a syncing / mempool-disabled Zebra collapses to `Fetch` -> the consumer's `Err(_) => source_error` -> `IncompleteSourceError`. The variants carry nothing a consumer reads.

Two things to resolve here:

1. **Is `NotReady` (tip) the same condition as `Unavailable` (listing)?** They read as two names for overlapping ideas: `NotReady` names node-global syncing but sits on a method that reads *chain state* (`getblockchaininfo`), which can't observe mempool-subsystem readiness at all; `Unavailable` names mempool-subsystem availability on the methods that actually read the mempool. Please reconcile — pick one concept, put it on the methods that can observe it.
2. **If any of these are real domain failure modes, route them through a call path that actually produces them** — a real `classify` closure, as `get_raw_mempool_transaction` does via `call_parsed_or_absent` for `NotFound` — instead of `call_parsed`/`call_parsed_slow`, which swallow everything to `Fetch`. If they are *not* real domain failures, drop them and let the method be transport-only (type it `QueryError<Infallible>`).

The broader pattern behind this (dead domain variants, the swallowing no-domain presets, and the classification×timeout preset cross-product) is tracked in #1442, which proposes the `Infallible` typing + classifier-as-argument direction (and a capability-trait-dispatch alternative). This comment is just the in-PR correctness ask; the general refactor lives there.

packages/zaino-state/src/chain_index/mempool.rs
```diff
@@ -0,0 +1,182 @@
+//! What the mempool subsystem needs from `zaino-state`, and how it gets it.
```
nachog00, line 3: Naming: this module is named `mempool_ports` but it defines **adapters**, not ports — the header itself calls them "the two adapters below" / "these adapters." The ports (traits) live in `zaino-mempool` (`NfsEpochObserver`, the source-port bounds) and `zaino-source`. It's also inconsistent within the crate: `source_ports.rs` holds an actual consumer bound (`ChainIndexSourcePorts`), so the `_ports` suffix means opposite things in the two files. Suggest `mempool_adapters.rs`.

Placement is right, to be clear: `NfsEpochAdapter` wraps a concrete `NonFinalizedState`, so it must live in zaino-state, and the dependency already points inward (adapters -> mempool ports; the mempool crates never name a zaino-state type) — so DIP is satisfied without a separate crate. Only the name is off.

packages/zaino-state/src/chain_index/mempool.rs
```diff
@@ -0,0 +1,182 @@
...
+    block_wake: tokio::sync::watch::Receiver<()>,
+}
+
+impl<S> MempoolSourceAdapter<S> {
+    pub(crate) fn new(source: S, block_wake: tokio::sync::watch::Receiver<()>) -> Self {
+        Self { source, block_wake }
+    }
+}
+
+impl<S: BlockchainSource> zaino_source::GetMempoolTxids for MempoolSourceAdapter<S> {
```
nachog00, line 79: `MempoolSourceAdapter` is bounded on `BlockchainSource`, but it only forwards the four zaino-source mempool ports and adds `SubscribeBlocks` — it never calls a `BlockchainSource` method. `BlockchainSource` is the legacy wire-typed driven-port god-trait being decomposed (its own doc calls it "temporary scaffolding"), and the mempool subsystem has otherwise already migrated off it. Binding this adapter to it re-couples the mempool wiring to the scaffolding it was meant to leave behind.

Narrow the bound to the ports it actually forwards — `GetMempoolTxids + GetMempoolMetadata + GetRawMempoolTransaction + GetMempoolSourceTip` (i.e. `MempoolPorts` minus `SubscribeBlocks`). Once it's off `BlockchainSource`, its only coupling to zaino-state is gone, so it could optionally move to `zaino-mempool-service` as a generic "add a block-wake to a pull source" adapter — a reuse call, not required (one consumer today, and DIP is already satisfied here).

nachog00:
## A design-pass note tying several threads together

Beyond the four blocking items in the review above (snapshot invariant, config validation, tracing, `as`-casts), a second set of comments share one root cause — flagging it here so they read as a coherent pass, not scattered nits.

**Theme: the crate is mid-migration off the wire-typed `BlockchainSource` god-trait toward per-port consumption over `zaino-source`, and naming / bounds / error-modeling haven't caught up.** The mempool subsystem is the first fully-migrated consumer (it reads `zaino-source` directly and never sees `BlockchainSource`, per `source.rs`'s own note) — which is exactly why the seams surface here.

**Ports vs. adapters — the vocabulary slipped**
- [`MempoolPorts`](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771119383) names an arch role, not the capability it represents (`MempoolSource`?).
- [`mempool_ports.rs`](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771577851) holds *adapters*, not ports (its own header says so) → `mempool_adapters.rs`; note `source_ports.rs` in the same crate really does hold a port, so the suffix means two things.
- [`MempoolSourceAdapter`](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771577903) is still bounded on the legacy `BlockchainSource` though it only forwards `zaino-source` mempool ports — re-coupling to the scaffolding the mempool otherwise left behind.

**Domain modeling — put the type where the concept lives**
- [`BlockRef`](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771119314) re-defines a type `zaino-primitives` already has (identical fields).
- [`ValidatorTip`](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771197112) is a single-field wrapper over `BlockRef` — ceremony without safety.
- [`MempoolCompleteness`](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771165952) conflates readiness (`NotReady`) with set-fidelity; readiness already lives in `StatusType`/`MempoolMode`.
- the [`MempoolTxMeta` re-export](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771102542) is unused and its rationale is counterfactual.

**Error model — declared but unwired (tracked in #1442)**
- [`NotReady`/`Unavailable`](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771427409) domain variants are never produced — `call_parsed`/`call_parsed_slow` swallow everything to `Fetch`; only `NotFound` is wired. Decide the contract, or type the no-domain methods `Infallible`.
- [timeout](https://github.com/zingolabs/zaino/pull/1439#discussion_r3771245547) is encoded as a preset axis (`call_parsed_slow`) that cross-products with the error-classification presets → make it a parameter.

None of these are correctness-breaking on their own — they're the residue of an in-flight decomposition, cheap to clean up, and cleaning them keeps the migration legible. #1442 captures the general error-classification direction.

packages/zaino-mempool-service/src/service.rs
idky137: We don't have a "no-`as`" rule. (not in any ADR, or repo documentation - there are currently ~277 as casts in packages/* today).

In this case:
- The divisor is MEMPOOL_TRANSACTION_COST_THRESHOLD = 10,000, so to lose bits on a 32-bit target you'd need headroom / 10_000 ≥ 2^32  - max_cost_bytes around 43 TB. The ZIP-401 default is 80 MB and our own default is 128 MiB.
- Crucially, if it did truncate it fails in the safe direction. as is modulo 2^32, so the result is always ≤ the true quotient - max_admissible can only come out too small, never too large. The cost cap can't be breached by this cast, and publish_snapshot does the exact cost_bytes + cost > max_cost_bytes check and is authoritative regardless; this value is only a pre-fetch estimate.

On the other sites: service.rs:567 and subscriber.rs:133 are usize → u64, lossless on every platform Rust supports, and entry.rs:79 is inside #[cfg(test)].

packages/zaino-state/src/chain_index/mempool.rs
idky137: This is a fair observation. I agree the staging isn't obvious enough from the current diff, and I'm going to improve the BlockchainSource docs to make that clearer.

I would like to keep the broader bound in this PR, though. The decomposition is intentionally staged: mempool first, then non-finalised state, then finalised state, followed by the final ChainHead cutover. During that process BlockchainSource is acting as the temporary composition point. As each subsystem moves out, its dependency moves from wire-typed methods on BlockchainSource to capability supertraits, so the trait's own method surface progressively shrinks toward something that can be removed mechanically at the end.

You're right that MempoolSourceAdapter itself only uses the four mempool ports, so narrowing its bound would make that local dependency more precise. The reason I'm not doing that here is that, in the current wiring, the adapter is constructed from the same Source already constrained by NodeBackedChainIndex<Source: BlockchainSource>. Narrowing the adapter alone therefore doesn't change the concrete source boundary used here, while pushing the narrower capability bounds through the surrounding generics adds churn that the staged approach is intended to avoid.

So I think the architectural observation is valid; I was just choosing to defer that additional narrowing until the surrounding boundary moves with it. I'll make the migration plan explicit in the docs, including the temporary duplication between BlockchainSource and ChainIndexSourcePorts and the expectation that they converge as the remaining subsystems are isolated.

packages/zaino-mempool/src/ports.rs
idky137: Agreed, and taken with one change.

It's actually a bit worse: the two types are already used side by side inside zaino-state. validator_source.rs:282-285 returns zaino_primitives::types::rpc::BlockRef while non_finalised_state.rs:146 builds zaino_mempool::BlockRef.

BlockRef is now a top-level zaino_primitives::types::BlockRef, with the superset derives (+ Hash), from_tip, and a From<(BlockHash, Height)>.

One deliberate difference from the suggestion: I dropped the types::rpc re-export entirely rather than keeping it for compatibility, and removed zaino-mempool's re-export too. Leaving either would give two paths to one type, which is the same problem in a different shape, and it's the reasoning from your MempoolTxMeta comment, which I've also taken. address_deltas and the three zaino-serve wire sites now name the canonical path.

(Added in [0b4238b](https://github.com/zingolabs/zaino/pull/1439/commits/0b4238baf688c5defe89cea0d03c81afdb6b5fb4))

packages/zaino-state/src/chain_index/mempool.rs
idky137: Updated in [5031f62](https://github.com/zingolabs/zaino/pull/1439/commits/5031f62dec66d7814635de9001bec209940e0cbf).

packages/zaino-source-zebra-rpc/src/adapter.rs
idky137: Good catch! I think it's actually worse than inert, because the constant contradicted its own documentation. The doc promises getrawmempool verbose room on a busy chain and explains that failing it "leaves a consumer's mempool marked incomplete precisely when the chain is busiest"; at 30s against a 30s default, that failure still happened. Good catch.

It's now 120s, with the constraint written on the constant and a unit test asserting it exceeds the default request_timeout, so it can't silently go inert again.

On the preset shape: agreed, and I've gone with the approach from your own caveat paragraph rather than the first proposal. call_parsed_slow is deleted and get_mempool_metadata passes Some(HEAVY_METHOD_TIMEOUT) to call_parsed_classified directly. That's "each method informs its own timeout" without adding None to ~26 terse call sites, and a future heavy method that also needs or_absent goes through the same door, so the cross-product you flagged can't form.

I've left call_parsed, _or_absent and _or_invalid_address as they are. They're presets on the classification axis, which genuinely has three values; it was the timeout axis that only ever had one occupant.

One thing I looked at and deliberately didn't change: call_parsed_optional bypasses call_parsed_classified and so structurally can't take a timeout, a real instance of the same wart. But it can't be folded in. Its classification turns gettxout's not-found code into Ok(None), and classify: FnOnce(RpcError) -> QueryError<E> can only produce errors. Folding it would mean generalising classify to return Result<T, QueryError<E>>, pushing Err(...) wrapping onto all four existing classifiers to serve one caller, for one line of shared code. Happy to revisit if you see a cleaner shape, but this looked like a net loss to me.

(Updated in [da7ffad](https://github.com/zingolabs/zaino/pull/1439/commits/da7ffad58641b0f2c3e585138a8282cac1eed37f))

packages/zaino-source/src/get_mempool_source_tip.rs
idky137: Sorry for leaving this in! And thanks for catching it!

GetMempoolSourceTipError is deleted. Not because the JSON-RPC adapter happens not to produce NotReady, but because the port's single-source rule guarantees no adapter ever can. ZebraValidator::get_mempool_source_tip routes to self.rpc and is deliberately not fast_or_slow!, unlike its GetChainTip neighbour, because the tip must come from whichever transport serves the mempool. ZebraReadStateAdapter doesn't implement the port at all. So the only transport that can answer is JSON-RPC, and getblockchaininfo either returns a tip or fails at the transport level. get_mempool_source_tip is now QueryError<Infallible>, which is your suggestion and is the honest type.

Unavailable stays on both listing ports, and is now actually produced. You were right that it was dead; the fix is wiring rather than deletion, because it names a condition that is observable on its transport. A validator answering -32601 on getrawmempool is telling you it has no mempool, a statement about the node, not the request, so retrying can't help. New mempool_unavailable_or_fetch classifier following spent_info_rejection's shape, with tests.

On reconciling the two names: they were two names for different things, and neither doc said so. Now stated on both. The tip port has no domain error; the listing ports have exactly one, meaning "this validator does not expose a mempool", sitting on the methods that can observe it — which is your point 1.

One thing I'd push back on generalising: GetChainTipError::NotReady should not get the same treatment, even though the JSON-RPC adapter doesn't produce it either. zaino-source is a multi-adapter port crate, and ZebraReadStateAdapter genuinely produces it (adapter.rs:242, ReadResponse::Tip(None)). A variant one transport can't see but another can is correct port design. The rule I've written into usage.md is narrower: a variant earns its place by being producible by some adapter, not by being plausible.

(Updated in [a500b99](https://github.com/zingolabs/zaino/pull/1439/commits/a500b99c3a66993cb3f25a3fc7d97dcca29ffdcf))

packages/zaino-mempool/src/lib.rs
idky137: removed in [8e3d742](https://github.com/zingolabs/zaino/pull/1439/commits/8e3d74241b7d8d2a33bab76ef03b39630cf84118)

packages/zaino-mempool/src/ports.rs
idky137: Updated to MempoolSource.

On the "renaming diverges from ChainIndexSourcePorts" concern: it doesn't, because the destination is already established on the branches that follow this one. feat/isolate_nfs has ChainHeadBlockSource (zaino-chain-head/src/ports.rs:63) and feat/isolate_finalised_state_plus_ports adds ChainStoreSource (zaino-chain-store/src/ports.rs:39), both consumer-defined bounds in a domain crate's ports.rs, structurally identical to this one, both named for the capability. That branch also introduces source_caps.rs with FinalisedSourceCaps, whose module doc spells out the convention.

So MempoolPorts was the outlier, not the standard. zaino-state is transitional wiring, it gets dissolved once the mempool, non-finalised state and finalised state are all extracted and ChainHead replaces ChainIndex, so ChainIndexSourcePorts is a name with an expiry date rather than a convention to preserve.

Separately, rewriting that doc turned up a real error worth flagging, since it was wrong in the same way in two places: both the rustdoc and usage.md said the single-source rule covers all five ports and that ZebraValidator "routes all five to JSON-RPC". It doesn't, SubscribeBlocks is supplied by zaino-state's sync loop, which is the whole reason MempoolSourceAdapter exists. usage.md then contradicted itself two paragraphs down. Now stated as the four validator reads, with SubscribeBlocks documented as the exception, which is itself an argument for the capability name, since the bound isn't one transport's worth of reads.
  
(Updated in [59094ee](https://github.com/zingolabs/zaino/pull/1439/commits/59094ee0a482006a68ae722e9ea954efd53243c2))

packages/zaino-mempool/src/snapshot.rs
idky137: Taken in full, your three tells were all accurate, and it turned out to be cheaper than expected.

NotReady is gone. MempoolCompleteness now describes only the fidelity of a set that exists, and "is there a set yet" is MempoolSnapshot::is_ready().

The thing that made this safe to do without an Option: readiness was already encoded structurally. A snapshot is tagged with the validator tip it was read at, so having no tag is having no set. is_ready() is source_tip.is_some(), derived from an existing field, no new state, no wrapper, no allocation. And the invariant you flagged now holds exactly: unadmitted is empty iff Complete, with no "or NotReady" caveat.

Tracing the freeze path also showed the pre-first-poll safety never actually rested on the variant. An unready snapshot has no source_tip, so observe_tips yields no validator tip, agree() returns None, and reconcile freezes, independently of may_be_wrong(). I've added an explicit !core.is_ready() guard anyway, with a comment saying it's belt-and-braces: the safety of the empty case shouldn't depend on following the tip path through three other functions, and a future change to observe_tips shouldn't be able to silently remove it.

One behaviour change worth calling out, since it's visible in logs: the startup freeze now reports ValidatorTipUnavailable rather than CoreIncomplete. That's more accurate, the empty startup set isn't incomplete, there's simply no tip to place it against. Both existing CoreIncomplete assertions are source-error cases and unaffected. New test pins it.

CoherentSnapshot::empty_not_ready() keeps its name deliberately: coherence is the axis readiness belongs on, and that view really is MempoolMode::NotReady. Renaming it would undo the point.

(Updated in [59ec2cc](https://github.com/zingolabs/zaino/pull/1439/commits/59ec2cc05e2dac65af5c8d8f6eab1783a1d3bb37))

packages/zaino-mempool/src/tip.rs
idky137: Updated in in [e3f2c27](https://github.com/zingolabs/zaino/pull/1439/commits/e3f2c27bc0e415fd75a43c97b96ffcf42bc1f5bd)

packages/zaino-mempool/src/snapshot.rs
idky137: Updated: fields are private, construction is sealed behind empty() / from_entries() / retag(), and reads go through accessors.

One correction to the framing, though, since it matters for how the PR reads: the invariant held at all three construction sites already. MempoolSnapshot was built in exactly three places, all in service.rs; the first sorts eleven lines before constructing, and the other two Arc::clone the already-sorted array forward. Your repro constructs a corrupt snapshot by hand it demonstrates the type permits an invalid state, not that any code creates one. So this is hardening, not a latent bug fix. The mechanism you describe is real and worth closing regardless, which is why I've closed it.

`from_entries` now owns the sort and derives entries_in_order, tx_count, raw_bytes and cost_bytes from the entry map, so a publisher can't get them wrong. I went further than the ordering in one respect: the old code carried the byte totals forward incrementally, with a comment about avoiding "three needless O(N) passes". They're now summed from the set. The sort is already O(n log n) over the same entries, so the extra O(n) is noise, and it makes a drifted running total unrepresentable rather than merely unlikely, which is the same argument you're making about the ordering.

`retag` covers the no-delta path and encodes a second rule that was previously only a comment: the set and mempool_generation are held while the sequence advances, because bumping the generation on unchanged contents makes the coherence layer redo its work on every tip re-stamp.

(Updated in [7f35b8b](https://github.com/zingolabs/zaino/pull/1439/commits/7f35b8ba9e9c1ea78db1d4a0add5869cd9d1645b))

packages/zaino-mempool/src/config.rs
idky137: Taken, the poll-interval crash is now fixed, both at the type and at the config boundary.

MempoolConfig's fields are private now. poll_interval_ms is a NonZeroU64 (read back as a Duration), event_buffer_len and max_concurrent_raw_fetches are NonZeroUsize, and zainod's check_config rejects poll_interval_ms = 0 with a named error next to the existing ZIP-401 floor check, so an operator gets a configuration error rather than an abort inside tokio::time::interval. Worth noting the panic reaches two call sites, not one: the coherence loop builds an interval from the same value.

Two corrections to the comment, both of which changed what I did:

`event_buffer_len` was never operator-reachable. It isn't a field of MempoolSettings, the TOML mirror exposes only max_cost_bytes, poll_interval_ms, metadata_min_interval_ms and max_exclude_count. So it could only be set from Rust. Still worth NonZeroUsize (a zero there does panic broadcast::channel), but it was not a config vulnerability.

`metadata_min_interval` = 0 is legal and I've deliberately left it that way. It feeds last.elapsed() >= self.config.metadata_min_interval(), so zero means "no floor beyond the poll cadence", a meaningful setting, and the same thing as setting it equal to the poll interval. Making it NonZero would have removed a valid configuration in the name of consistency. There's a test pinning that asymmetry so it doesn't get "fixed" later. max_exclude_count is the same story: zero disables client-supplied exclusion.

The audit also turned up something adjacent: max_concurrent_raw_fetches was read as .max(1), which silently rewrote an operator's zero instead of rejecting it. It's NonZeroUsize now and the guard is gone.

(Fixed in [3671f21](https://github.com/zingolabs/zaino/pull/1439/commits/3671f21a1088baf2c516da63119466fbe70da8e7))

packages/zaino-mempool-service/src/service.rs
idky137: Converting service.rs:352 anyway, for one reason that wasn't in my original reply.

`usize::try_from(...).unwrap_or(usize::MAX)`

The saturation direction is the point: max_admissible is a deliberate over-count, it assumes every addition costs the ZIP-401 floor, and publish_snapshot does the exact check, so saturating high keeps the error in the direction the estimate is already wrong in. Truncation goes the other way, and the boundary case is nastier than "under-admits": a quotient of exactly 2³² wraps to zero, which takes the refuse-everything-without-fetching branch while the set is nowhere near its bound. Unreachable at 43 TB on a 64-bit target, but it's a cheap line and the comment now records the reasoning.

The rest of my earlier reply stands. There's no as rule in the repo, entry.rs:79 is test code, and service.rs:567 / subscriber.rs:133 are usize → u64, which is lossless on every platform Rust supports, those are the only as casts left in either mempool crate outside test fixtures. Still happy to open an issue for a repo-wide policy and lint, which would need to account for the ~277 pre-existing sites.
  
(Added in [f9b771c](https://github.com/zingolabs/zaino/pull/1439/commits/f9b771cda6044de58487d3bda7c65150c3626727))

packages/zaino-mempool-service/src/service.rs
idky137: Taken in full, thanks for catching this.

All four Err(_) arms now carry the error. The transition into degradation logs warn with the cause, the underlying error and the tx_count still being served; recovery logs info, so a warning is always closed rather than leaving an operator to infer recovery from silence. While still degraded it drops to debug. That split is deliberate: at a sub-second poll cadence, logging every failed tick would turn a downed validator into thousands of identical warnings, so the edges carry the signal and the level knob answers "is it still failing, and why".

The :590 case you flagged, where a built MempoolError was dropped, now logs at the point it's constructed, inside the buffer_unordered fan-out where the per-transaction context still exists, rather than at the collection site where only "something failed" survives.

Threading the error through turned up something adjacent: publish_source_error had two callers with genuinely different causes. Besides the failed reads, the MAX_CONSECUTIVE_DISCARDS backstop lands there too, and in that case the validator is answering fine, the tip just won't hold still long enough to tag a set against. Reporting it as a source read failure would send an operator looking for a broken connection, so it has its own line and a cause = "tip_unstable" field.

Both loops carry a long-lived #[instrument] span (mempool_poll_loop, mempool_coherence_loop) with start/stop at debug, which is what ZAINOLOG_FORMAT=tree groups on. I deliberately didn't add per-tick spans, same reasoning as the per-tick logs.

One deviation from your ask: tracing is a dependency of zaino-mempool-service only. zaino-mempool has no runtime behaviour to instrument — it's types, ports and pure functions; its one validation path is a debug_assert, and its config knobs are now NonZero, so illegal values are unrepresentable rather than rejected-at-runtime. Adding the dependency there would leave it unused. I've documented that as a positive statement in the CHANGELOG ("nothing to log", not "nothing logged") so it reads as a decision rather than an omission. Happy to add it if you'd rather have the uniformity though.

`usage.md` now has an "Observing it" section documenting the levels and what each one means

(Updated in [b823481](https://github.com/zingolabs/zaino/pull/1439/commits/b823481b99173ed91d221024536089964f914818))

packages/zaino-mempool-service/src/coherence.rs
idky137: Taken, both transitions now trace with the reason as a structured field.

Your framing was the useful part: the reason existed only on the broadcast event, so an operator who saw the upstream freeze-escalation warning had no record of why, and FreezeReason has eight variants that answer precisely that question.

One deviation: both edges are at debug, not debug freeze / info thaw. At the default RUST_LOG=zaino=info that split shows the recovery and hides the degradation, you'd see "coherence thawed" with no preceding line explaining what happened. Since a freeze is routine (every block causes one, so roughly every 75s on mainnet), neither edge belongs at info; putting a line per block in a default-level log for a healthy node is noise. Keeping both at debug leaves the existing COHERENCE_FREEZE_ESCALATION warn and the zaino.mempool.coherence_frozen_seconds gauge as the abnormal-case signal, which is where it belongs.

The code was already shaped for this, freeze() has the transition-in check that starts the freeze clock, and the dedup guard for "already frozen, same reason, same tips", so the logs sit on real edges rather than firing per reconcile. I added a third, distinct line for a freeze whose cause changes while still frozen (a tip moving again, or tips diverging after one went unavailable): that publishes a new snapshot and is worth seeing, but it isn't the same event as entering a freeze. It's placed after the dedup guard so an unchanged freeze stays quiet.

The thaw edge keys off prev.mode rather than the freeze clock, since publish_live clears the clock on every live publication and the clock alone would have logged on every reconcile.

(Updated in [7e1dcb2](https://github.com/zingolabs/zaino/pull/1439/commits/7e1dcb21d8e5c69a2e0294af0ce9953efc489b1c))

packages/zaino-state/src/chain_index/mempool.rs
idky137: Documented in 0c793d8

nachog00 (approved):

PR #1440, opened 2026-08-10
Feat - Isolate nfs (readying for im based nfs rework + codebase modularisation)

idky137:
# Isolate the non-finalised chain head, and make it synchronise itself

## What this does

Takes the non-finalised state out of `zaino-state` and lands it as a subsystem that
runs itself: two new crates, one writer task, `zaino-state` demoted from driver to
consumer. `non_finalised_state.rs` (1,038 lines) and its 290-line test module are
deleted in the same PR that adds the replacement — no dead code, no window where
both exist.

The old module held the right things (recent block graph, reorg handling) and was
reached the wrong way:

- **It could not advance without the finalised state** — it read `db_height()` for its
  anchor *and* trim floors, which is pinned at `0` in ephemeral mode, so every read
  took the fallback arm. The dependency lived in the types, never in the logic.
- **It could not advance without being told to** — no task of its own; `ChainIndex`'s
  sync worker called `sync(...)`, so head freshness was set by database write throughput
  and any holder could drive it.
- **Publication was not atomic** — mid-catch-up `update()` calls stored into the shared
  cell directly, so readers could observe a half-filled window or half-applied reorg.
- **Its boundary was `IndexedBlock`** — a persistence row shape used as the currency of
  an in-memory graph with no database.

Replacement follows the `<domain>` / `<domain>-service` split used elsewhere:

```
zaino-chain-head          ports + vocabulary + ChainHeadSnapshot (trait). No runtime.
      ↑
zaino-chain-head-service  ChainHeadService (one writer task), MapBackedSnapshot,
                          ChainHeadSubscriber
      ↑
zaino-state               reads published snapshots; its worker drives only the
                          finalised half
```

Three load-bearing ideas:

1. **Advancing is not an operation.** No `sync` / `advance` / `reconcile` on any port at
   any visibility. The writer task installs a candidate snapshot with one store; nothing
   else reaches the published cell. Test-only stepping is `cfg`-gated out of production.
2. **Reads live on the snapshot; the handle only produces snapshots.**
   `ChainHeadBlockService` has `current()` and `subscribe_updates()`, nothing else — five
   questions to one snapshot describe one instant. Matches the `ChainView` /
   `ChainViewSnapshot` split in the Chain View Layering Spec.
3. **Capability is a bound over existing ports.** `ChainHeadBlockSource` is blanket-implemented
   over five `zaino-source` ports, so any adapter that can already answer earns it.
   `ZebraValidator` satisfying it is asserted by a compile-time test (`ports.rs`).

`ChainHeadSnapshot` is a trait, so replacing `MapBackedSnapshot` with persistent structures
is a one-crate change. `zaino-state` keeps one documented piece of scaffolding —
`chain_index/chain_head.rs` converts back to `IndexedBlock` for the existing query paths;
its chainwork is **anchor-relative**, so its output is served and never persisted. It
disappears when `ChainIndex` becomes the chain view layer. Reasoning recorded in **ADR-0011**.

## The chain head is now the mempool's notion of "which chain"

The mempool subsystem's coherence layer freezes and thaws against a non-finalised-state
epoch (ADR-0010). With the head extracted, that epoch is the chain head's, and this PR
wires the two together.

- **The epoch is one type, in `zaino-primitives`.** `ChainHeadEpoch` and the mempool's
  `NonFinalizedEpoch` were field-identical — same two fields, same order, both over
  `BlockRef`, both keyed to tip changes rather than publications, differing only in a
  `Hash` derive. Both are replaced by `zaino_primitives::types::ChainStateEpoch`. Both
  crates already depended on `zaino-primitives`, so this costs no new dependency and
  removes the reason the two could ever drift. **There is no translation function**: the
  coherence check compares one type rather than converting between two that started
  identical. Neither crate re-exports it — a `pub use` of a foreign type is not an
  abstraction boundary, it just gives one type two paths.
- **The epoch is readable from a snapshot, not just from the handle.**
  `ChainHeadSnapshot::epoch()` is new, and `MapBackedSnapshot` carries the generation of
  the publication that produced it, stamped before the snapshot is stored. This is the
  property the coherence check rests on: it asks whether a transaction set matches the
  chain *the caller is being served*, and the caller holds a captured view — reading the
  handle's epoch would compare against whatever has been published since. A test pins it
  (`a_snapshot_keeps_the_epoch_it_was_published_under`).
- **The snapshot owns the generation rule.** The field is private and written only by
  `stamp_generation`: a publication whose tip is unchanged inherits the previous
  generation, one whose tip moved advances past the highest yet published. Taking the
  *highest published* rather than the previous snapshot's is what keeps it monotonic
  across a re-anchor, where the graph is rebuilt from a single block and starts at zero.
  Covered by `the_generation_advances_across_a_re_anchor`.
- **The sync loop no longer relays a publication signal.** It does not drive the head and
  so does not know when the head publishes; the head's own epoch feed is both more
  accurate and one fewer thing to keep in step.

**Worth a reviewer's eye:** the mempool's `NfsEpochObserver::subscribe_epoch_changes`
returns `watch::Receiver<()>` and the chain head publishes `watch::Receiver<ChainStateEpoch>`.
`watch::Receiver` cannot be mapped, so `ChainHeadEpochAdapter::spawn` runs a small relay
task to bridge the two. The alternative is returning `None` (the port allows it), which
reintroduces a freeze-per-block lasting a full poll tick — the wake exists precisely to
avoid that. Making the port generic over its item type would remove the task; that felt
like a larger change to the mempool's public surface than this PR should make.

## Defects found while moving, and how they're fixed

| # | Defect | Fix |
|---|--------|-----|
| 1 | **Window could overshoot its own seam (#1126)** — the extension loop kept extending past the committed `chain_height` while the validator produced blocks, and trimming used the pre-mine floor. | Structural: `tick` reads the tip **once**, `next_graph` is bounded by it, a mid-iteration advance is deferred. Race reproducer and its `get_block_hook` mock scaffolding deleted with the mechanism. |
| 2 | **A tie in accumulated work flipped the tip nondeterministically** — `max_by_key(chainwork)` fed unconditionally into `handle_reorg`, so equal-work siblings resolved by map iteration order. | `if heaviest.work > tip_work` — strictly greater; on a tie the validator's answer stands. |
| 3 | **Every steady-state poll cost a block fetch** — the non-higher reorg check re-read the tip height each iteration. | Early return when the source tip equals the held tip; a hash commits to its parent, so an identical tip means an identical chain. |
| 4 | **Underflow at genesis** — the non-higher reorg check recursed on `target_height - 1` with unchecked `Sub<u32>`. | Explicit termination plus saturating `height_below` / `next_height`. |
| 5 | **Dead code that looked like a feature** — the block-carrying `nonfinalized_listener` path (every source returned `Ok(None)`), plus `NonFinalisedStateError`, `UpdateError` and friends. | Not carried into the new subsystem; the error types are merged and renamed into `ChainHeadAdvanceError`. Flagged as a behaviour change: push-channel branch discovery never worked. (The vestigial port remains on the legacy `BlockchainSource` trait, which dissolves separately.) |

Also: trimming leaves a **10-block margin** (`RETENTION_MARGIN`) below the configured depth
so it never cuts at the seam. Reorg metric names (`zaino.sync.reorg_total` / `_depth`) are
deliberately unchanged — only their constants moved crates — so existing dashboards keep working.

**Decisions worth flagging**

- **`getchaintips` is not in the driven port, on purpose.** The original never enumerated
  tips; it only knew branches it lived through, and still does. ADR-0011 records that
  **`zebra-rpc` does not implement `getchaintips` at all**, so real competing-tip discovery
  needs a ReadState-backed answer first.
- **The `getchaintips` "still syncing" fallback is removed** — the head anchors before
  `ChainIndex::new` returns, so the state it existed for cannot occur.
- **Startup is fallible where it was not.** No persistent state, no second source: it anchors
  or fails, and `ChainIndex::new` fails with it (transients still absorbed by the retry ladder).
  In exchange `current()` is total for the rest of the process's life.
- **The status fold no longer latches.** `ChainIndex::status` / `combined_status` account for
  the chain head, and component statuses are read live on each call rather than written back
  into the index's own cell — where the first transient failure in any component used to pin
  the index to `RecoverableError` for the life of the process.
- **`ChainHeadFreezeEvents` is new capability, not a move** — blocks below the seam are
  broadcast whole so a store can ingest without re-fetching. Best-effort by design
  (`Lagged(n)`, discontinuous floor after a re-anchor); subscribing is what starts collection.
- **`ChainHeadWork` is distinct from `ChainWork`** — accumulated from the anchor, not genesis,
  so it orders branches but is not the validator's absolute number.
- **ADR numbered 0011, not 0010** — the mempool subsystem separation took 0010.

## Applying the #1439 review to this branch (commits 5–11)

The mempool separation review raised themes that apply to any crate pair extracted this
way, so the same audit was run against these two crates before asking for review. Each
finding is its own commit.

| # | Finding | Commit |
|---|---------|--------|
| 5 | **Error variants nothing produces.** `ChainHeadSourceError` was exported and never constructed or used as a return type; three of `ChainHeadError`'s four variants were never produced. `NotReady` contradicted the crate's own invariant — a chain head anchors before its constructor returns, so there is no not-ready state. Enum is now one variant and `#[non_exhaustive]`, so an implementation that *can* fail adds one without a break. The two total transaction methods drop their `Result`. | `fcee644d` |
| 6 | **Two identical epoch types.** Promoted to one `ChainStateEpoch` in `zaino-primitives` (see above). | `7af60da9` |
| 7 | **Task lifetime unstated.** `spawn` hands the writer its own `Arc<Self>`, so dropping every handle does not stop it. Documented as the contract it is, with why the task must outlive its handles. | `8a2bc0a7` |
| 8 | **`usage.md` said `shutdown` awaits the task.** It aborts — it cannot await, being called from `Drop`. Corrected, with why the abort is safe (one atomic store per publication, so a killed task leaves the last snapshot whole). | `72cfce1d` |
| 9 | **Config accepted illegal values.** All five `ChainHeadConfig` knobs are `NonZero` and the fields private. Zero is meaningless for every one of them — and a zero poll interval does not panic, it spins against the validator, which is worse. Not operator-reachable (no TOML mirror), so this is type hygiene, not a reachable misconfiguration. | `9ac18e61` |
| 10 | **One production `as` cast.** `info.size as u32` → `try_from(..).unwrap_or(u32::MAX)`. Unreachable either way, but `as` is modulo, so a size of exactly 2³² would report an *empty* tree, which reads as valid; saturating reports an implausibly full one, which does not. | `6fbcac78` |
| 11 | **Generation assignable by the writer.** The field was `pub(crate)` and stamped at the call site. Now private, written only by `stamp_generation`, which owns the rule. | `6d7d6e50` |

## Code changes by commit

Every line is classified by what it is, not by filename: inline `#[cfg(test)]` blocks and
`src/tests.rs` count as test, `///` / `//!` count as docs inside production modules,
in-`src/` mock scaffolding counts as test. File counts overlap between columns.

| # | Commit | Files | Production | Tests | Docs | Manifests + lock |
|---|--------|------:|------------|-------|------|------------------|
| 1 | `feat(chain-head): add zaino-chain-head, the ChainHead domain crate` | 22 | 13f +428 −5 | 3f +21 −2 | 14f +660 −2 | 4f +54 |
| 2 | `feat(chain-head-service): move the NFS into its own runtime` | 13 | 6f +1,132 | 2f +664 | 8f +533 | 4f +77 |
| 3 | `refactor(state): remove the NFS and read the chain head` | 27 | 10f +690 −1,818 | 12f +666 −512 | 11f +267 −209 | 3f +8 −1 |
| 4 | `docs: record the chain head separation` | 3 | — | — | 3f +186 | — |
| 5 | `refactor(chain-head): delete the error variants nothing produces` | 6 | 5f +14 −55 | — | 3f +23 −28 | — |
| 6 | `refactor(primitives): promote one epoch type instead of two identical ones` | 26 | 18f +67 −85 | 1f +7 −6 | 13f +53 −44 | — |
| 7 | `docs(chain-head-service): state who owns the writer task's lifetime` | 3 | — | — | 3f +34 −6 | — |
| 8 | `docs(chain-head-service): shutdown aborts, it does not await` | 2 | — | — | 2f +15 | — |
| 9 | `refactor(chain-head): make illegal config unrepresentable` | 5 | 3f +74 −30 | 1f +17 −9 | 2f +61 −21 | — |
| 10 | `refactor(state): convert the tree-size cast instead of truncating` | 1 | 1f +8 −1 | — | — | — |
| 11 | `refactor(chain-head-service): let the snapshot own its generation rule` | 4 | 2f +21 −16 | 1f +39 | 2f +26 −8 | — |

Per-commit figures sum to more than the branch diff, because commits 5–11 revise lines
that commits 1–4 added. The **net** change against `dev` is what a reviewer sees:
**77 files, +5,654 −2,795**.

Commit 1 also carries the `zaino-primitives` prerequisite: `EquihashSolution`, and the
`version` / `solution` fields on `BlockHeader`, so a retained block carries every field
the block hash commits to rather than the subset an index happens to read.

## Verification

`cargo fmt --check`, `makers lint-boundary-conversions`, and `cargo clippy --workspace
--all-targets` (no new warnings) all pass on the final tree.

`cargo nextest run` — **605 tests passing** — was last run end-to-end at commit 9. Commits
10 and 11 were verified against the crates they touch (`-p zaino-state`, 226 passed;
`-p zaino-chain-head-service`, 20 passed) but the full suite has not been re-run over the
final tree. Each commit compiled and tested green when it was made.

The live suite is not clean, for a reason that predates this branch — see Known gaps.

## Known gaps

- **The freeze stream has no consumer** — implemented and tested, but the store still builds
  from source. Intended until the store rework, but it is production logic no production path reaches.
- **The transparent-address-history port is declaration-only** — types behind the feature flag,
  nothing implements it.
- **The `ChainHeadBlock → IndexedBlock` conversion is debt with a known end** — it goes when
  `ChainIndex` becomes the chain view layer; until then its output must never be persisted.
- **Competing-tip discovery is still not implemented** — same answers from the same retained
  graph as before, so no regression, but no new capability.
- **The epoch relay task** described above is glue this arrangement costs; it goes away if the
  mempool port becomes generic over its wake item.
- **`e2e::devtool::get_transaction_mempool` is flaky, and not because of this branch.**
  Zaino reports `No such mempool or main chain transaction` for up to ~1s after a broadcast,
  because `ChainIndex` treats a knowingly-incomplete mempool set as authoritative for
  absence — `MempoolSnapshot::unadmitted()` exists precisely to distinguish "short this
  transaction, ask again" from "does not exist", and `zaino-state` never reads it. Present on
  `dev`; filed separately. Not fixed here, and the test's sleep deliberately **not** widened.
- **`--all-features` remains broken** on `write_core.rs` behind
  `transparent_address_history_experimental`. Pre-existing on `dev` and untouched here.
- **Unblocks:** the chain view layer, and the store rework's freeze-stream consumer.


packages/zaino-chain-head/src/block.rs
```diff
@@ -0,0 +1,85 @@
...
+        Self(block_work)
+    }
+
+    /// Extends this accumulation by one block's work.
+    ///
+    /// Returns `None` on overflow. The window spans a bounded number of blocks,
+    /// so this cannot happen in practice; the caller still handles it rather
+    /// than asserting, because "cannot happen" is a claim about the
+    /// configuration, not about the type.
+    pub fn checked_add(self, block_work: u128) -> Option<Self> {
```
AloeareV, line 40: Non-blocking: This is the sort of thing where it would be really nice if we _could_ bake this assertion into the type. Non-blocking as I don't have a good idea of how. Even if we set a bound for high a ChainHeadWork could be initialized at, there's nothing stopping a caller from running a loop and just...adding until the invariant is violated, although that would be clearly a 'wrong' thing to do, it's not enforceable at the type level. We could also keep a count of how many blocks have been added to this work, and reject values that are unreasonably high (i.e. above `BlockHeight::max`). This might work, but I'm not convinced it's worth the squeeze...I'm thinking out loud at this point. Non-blocking.

packages/zaino-chain-head-service/src/service.rs
```diff
@@ -0,0 +1,913 @@
...
+            MapBackedSnapshot::from_initial_block(self.resolve_anchor_block(anchor_height).await?)
+        } else {
+            previous.clone()
+        };
+
+        // currently this only gets main-chain blocks
+        // once readstateservice supports serving sidechain data, this
+        // must be rewritten to match
+        //
+        // see https://github.com/ZcashFoundation/zebra/issues/9541
```
AloeareV, line 394: Issue is marked as complete. This comment therefore implies this code must be rewritten to match. Maybe this can't be done yet due to something being wrong with the block listener?

packages/zaino-state/src/chain_index/non_finalised_state.rs
```diff
@@ -1,1038 +0,0 @@
...
-            // silently match a post-re-anchor snapshot.
-            let mut anchor_snapshot =
-                NonfinalizedBlockCacheSnapshot::from_initial_block(anchor_block);
-            anchor_snapshot.generation = initial_state.generation.saturating_add(1);
-            self.current.swap(Arc::new(anchor_snapshot));
-            initial_state = self.get_snapshot()
-        }
-        let mut working_snapshot = initial_state.as_ref().clone();
-
-        // currently this only gets main-chain blocks
```
AloeareV, line 513: Ah, found the place this was moved from. Okay. Non-blocking as it's a move of pre-existing code. Still a bit concerning we have a comment here that's more than a year out of date.

packages/zaino-chain-head-service/src/service.rs
AloeareV: Ah. This is a move, not new code. Somewhat concerning how stale this comment is, but it's not within the scope of this PR, so non-blocking.

packages/zaino-chain-head-service/src/tests.rs
```diff
@@ -0,0 +1,707 @@
...
+/// The single byte a test hash was built from, so a chain can be walked by id.
+fn id_of(hash: &BlockHash) -> u8 {
+    <[u8; 32]>::from(*hash)[0]
+}
+
+fn height(h: u32) -> Height {
+    Height::try_from(h).expect("test height in range")
+}
+
+/// A block identified by a single byte, so test chains read as `1 -> 2 -> 3`.
```
AloeareV: `u8::MAX` is 255. Is there a world we want to create more than 255 test blocks via this mechanism? Is there any cost to this being a u16 or greater?

packages/zaino-chain-head-service/src/tests.rs
```diff
@@ -0,0 +1,707 @@
...
+struct MockValidator {
+    state: Arc<Mutex<MockState>>,
+}
+
+impl MockValidator {
+    /// A best chain of `len` blocks, ids `0..len`, block `n` at height `n`.
+    fn linear(len: u32) -> Self {
+        let mut state = MockState::default();
+        for h in 0..len {
+            let id = h as u8;
```
AloeareV: If we're keeping u8, this should be a fallible conversion with an unwrap or expect instead of a cast. Right now, if h is greater than 255, it will wrap, which could cause failures that don't obviously lead to this line.

AloeareV (changes requested):
There's a test value which is currently u8, and I'm concerned that could overflow. Other than that, this looks fantastic! Some of the code that was moved is concerningly stale, but modifying that code is explicitly out-of-scope of this PR.

packages/zaino-chain-head/src/block.rs
idky137: Ye I wasn't sure what the best option here is, I'm happy to change this but am not sure what the correct shape would be.

AFAICT it can't be encoded in the type, the current shape is deliberately the conservative one, it returns Option and the caller handles it, so the "cannot happen" claim in the doc comment is never load-bearing at runtime. A block counter would re-assert the same configuration claim in a second place without removing the branch.

packages/zaino-state/src/chain_index/non_finalised_state.rs
idky137: Ye I got quite tempted to start making much broader changes in this PR but decided to hold back as we are imminently due a large rewrite anyway. Again happy to add something here if you think there should be. We could add clear comments saying this is stale and awaiting rewrite? or we could just put a big warning at the top of the current implementation?

packages/zaino-chain-head-service/src/tests.rs
idky137: Good find! Will fix this now!

I don't think it is actually dangerous *now* but completely agree this it a bad / dangerous test waiting to happen and should be fixed!

packages/zaino-chain-head-service/src/tests.rs
idky137: My plan is to take both options given here to ensure its safety (upgrading from u8 and making fallible).

packages/zaino-chain-head-service/src/tests.rs
idky137: updated in [d48dcec](https://github.com/zingolabs/zaino/pull/1440/commits/d48dcec3a22a3ee872840180b4d86f2236b1f936)

packages/zaino-chain-head-service/src/tests.rs
idky137: updated in [d48dcec](https://github.com/zingolabs/zaino/pull/1440/commits/d48dcec3a22a3ee872840180b4d86f2236b1f936)

AloeareV (approved):
Great! u8 thing was the only blocking concern, the rest is just commentary!

nachog00:
### Review finding — Error swallowing: stringly-typed advance errors sever the cause chain

*Suggested issue title:* `chain-head-service: ChainHeadAdvanceError flattens source errors to String`

`ChainHeadAdvanceError`'s three variants are all `String` (`error.rs:20,29,33`), and every source failure is `.to_string()`'d at the boundary (`service.rs:666,680,692,801,808,812`). The structured `zaino_source::QueryError` — whose `Fetch(FetchError)` arm carries a machine-readable `FailureMode` (timeout / connection / parse / auth) — is flattened to text. `Error::source()` returns nothing; a caller can't match cause. There are zero `#[source]`/`#[from]` in the crate.

`ChainHeadInitError` already nests a typed `source` field, so the pattern exists one level up and just isn't carried to the leaves.

**Fix:** carry the transport failure via `#[source]` (`SourceUnavailable(#[source] FetchError)`), preserving `FailureMode`. Keep `InconsistentSource`/`ReorgFailure` message-only (they report a reconciliation failure, not a wrapped external error).

Severity: quality / observability. Not a correctness bug.

<sub>Posted as a review finding on the merged PR for follow-up triage.</sub>

nachog00:
### Review finding — Silent source-error swallow (latent, not live)

*Suggested issue title:* `chain-head-service: block_at_height/hash swallow QueryError::Domain(_) with a wildcard and no trace`

`block_at_height` (`service.rs:679`) and `block_at_hash` (`service.rs:691`) map `QueryError::Domain(_) => Ok(None)`, discarding the value with a `_` and no log.

**This is not a live bug today.** Both domain enums have exactly one variant (`GetBlockError::HeightNotFound`, `GetBlockByHashError::NotFound`), neither is `#[non_exhaustive]`, and the zebra-rpc adapter routes parse/transport failures through `QueryError::Fetch` (propagated correctly), not `Domain`. So the only `Domain` value genuinely means "absent" and `Ok(None)` is correct.

What the `_` costs: (a) the discard is untraced, and (b) it removes the compiler's ability to catch a *future* second domain variant — a new rejection (`Pruned`, `AboveCeiling`, …) would be silently reclassified as "absent" (the extension loop reads that as end-of-chain and could publish a short graph), with no build break at the one site that must be revisited.

**Fix:** match the named variant (`Domain(GetBlockError::HeightNotFound(h)) => …`) and `debug!`-trace it. Exhaustive-by-name; zero behaviour change today; future-proof + observable.

Severity: latent-correctness + observability.

<sub>Posted as a review finding on the merged PR for follow-up triage.</sub>

nachog00:
### Review finding — Modularisation: `service.rs` is 913 lines in one file

*Suggested issue title:* `chain-head-service: split service.rs into per-concern modules`

`service.rs` (913 lines) holds four separable concerns: lifecycle (`spawn`/`anchored`/`shutdown`/`run`), the advance engine (`tick`/`next_graph`), the reorg engine (`handle_reorg`/`check_for_nonhigher_reorgs` + the private `Block` bridge), anchoring, publication, and source access. The reorg engine — the highest-risk code, where the #1126 window-overshoot and the tie / underflow / steady-poll bugs lived — is buried in the middle.

**Fix:** `service.rs` + `service/{advance,reorg,anchor,publish,source,util}.rs`, mirroring the mempool-service split (#1445). `module.rs + module/`, pure code motion.

Severity: maintainability.

<sub>Posted as a review finding on the merged PR for follow-up triage.</sub>

nachog00:
### Review finding — Observability: the subsystem's headline claim is unmeasured

*Suggested issue title:* `chain-head-service: add tip-freshness/lag metrics and a frozen-drop counter`

The only metrics are `CHAIN_HEAD_REORG_TOTAL`/`_DEPTH` (`metric_names.rs:13,16`), both behind `#[cfg(feature = "prometheus")]`. The subsystem's whole thesis — "tip freshness decoupled from DB write throughput" — has no freshness/lag gauge to prove it in prod. Missing: tip-age/lag, tick duration, anchor latency, current `consecutive_failures`, retained graph size (`retained_block_count()` exists but is never emitted), publish rate.

Also a **silent drop**: `service.rs:607` `let _ = self.frozen.send(block);` ignores the broadcast `SendError` with no counter/log, so the documented best-effort `Lagged` behaviour is invisible on the producer side.

Severity: operability.

> Note: some of this may be subsumed by the in-flight resilience-seam work, which removes the `consecutive_failures` retry concept from this crate — coordinate before adding a retry-count metric.

<sub>Posted as a review finding on the merged PR for follow-up triage.</sub>

nachog00:
### Review finding — PDV: the graph's tip invariant is runtime-checked in three places

*Suggested issue title:* `chain-head-service: encode "graph contains its own tip" in the type`

`MapBackedSnapshot` stores `best_tip: BlockRef` separately from `blocks: HashMap` (`snapshot.rs`), so "the graph contains its own tip" is re-asserted at runtime three times in `next_graph`: two `ok_or(ReorgFailure("graph is missing its own tip"))` (`service.rs:411,457`) and one `expect("a graph always retains at least its anchor")` (`service.rs:466`).

**Fix:** a total `tip_block(&self) -> &ChainHeadBlock` accessor as the single home of the invariant (collapsing the two `ok_or` sites), and optionally fold the tip into the graph type so even the `expect` disappears.

Severity: robustness / type hygiene.

<sub>Posted as a review finding on the merged PR for follow-up triage.</sub>

PR #1445, opened 2026-08-13
refactor(mempool-service): split service/coherence/subscriber into per-section modules

nachog00:
> **Stacked on #1439** (base: `feat/mempool_on_source_ports`) — merge after it, and review this diff against that branch, not `dev`. GitHub will auto-retarget the base to `dev` once #1439 lands.

Pure, move-only structural change — **no behavior change**. The point of it is to **cut future merge-conflict surface**.

`service.rs` and `coherence.rs` are conflict hotspots. In this review round alone, four independent fixes — the snapshot constructor, the config invariants, tracing, and the error model — all had to land in the same 800+-line files, and the staged decomposition still ahead (non-finalised state, finalised state, the ChainHead cutover) will keep touching them. When several branches edit one big file they collide; splitting each concern into its own file means those changes touch **small, disjoint files** and stop stepping on each other.

The split follows the files' existing `// ---- section ----` seams:

- `service.rs` → `service/{state,poll,publish}.rs`
- `coherence.rs` → `coherence/{run,reconcile,publish}.rs`
- `subscriber.rs` → `subscriber/filter.rs` (extract the exclude-filter)

Move-only: no logic, signature, or public-API change — the two filter types keep their path via `pub use`; visibility widened to `pub(super)` only where the compiler forced it; inherent methods live in submodules via `impl super::MempoolService<S>` / `impl super::CoherenceService<M, N>`.

Verified: `cargo fmt`/`clippy` clean (both feature configs), and the test suite is identical to before the split — **25** default / **44** `tip_aware_mempool`.

Kept separate from the correctness work (it's the one piece of my now-closed #1443 with no counterpart here) so it reads as the mechanical move it is.


zancas (approved):
UtAck

PR #1449, opened 2026-08-14
Zainod metrics rework

elicbarbieri:
Updates and modifies metrics

Aimed at improving quality/accuracy of metrics for ztest migration


packages/zaino-mempool-service/src/lib.rs
```diff
@@ -30,3 +74,48 @@ pub use subscriber::{MempoolFilterError, MempoolInfo, MempoolSubscriber, TxIdExc
...
 pub use coherence::{CoherenceService, CoherentSubscriber};
+
+#[cfg(all(test, feature = "prometheus"))]
+mod metric_tests {
+    use crate::metric_names::MEMPOOL_COMPLETENESS_VALUES;
+    use zaino_mempool::MempoolCompleteness;
+
+    /// - Gauge publishes `MempoolCompleteness as u8`, legend =
+    ///   [`MEMPOOL_COMPLETENESS_VALUES`]
+    /// - `#[repr(u8)]` + explicit discriminants stop a mid-enum insert renumbering
```
nachog00, line 85: This rationale doesn't match the enum it describes. `MempoolCompleteness` (`zaino-mempool/src/snapshot.rs`) is `#[derive(Debug, Clone, Copy, PartialEq, Eq)]` with no `#[repr(u8)]` and no explicit discriminants — so `completeness() as u8` rides on declaration order, not the guard this comment claims. `StatusType` does exactly this pattern correctly right next door (`Spawning = 0 … CriticalError = 7`).

Suggest adding `#[repr(u8)]` + explicit discriminants to `MempoolCompleteness` for parity. Beyond making the comment true, it makes the published gauge value a deliberate wire contract — dashboards will key on `3 = incomplete-source-error`, and right now that integer is an accident of variant order rather than a committed mapping. The legend test stays valid either way.

packages/zaino-state/src/chain_index/finalised_state/finalised_source/v1/write_core.rs
```diff
@@ -332,24 +327,25 @@ impl DbWrite for DbV1 {
...
                     )
                     .await?;
-                    #[cfg(feature = "prometheus")]
-                    metrics::histogram!(SYNC_BLOCK_BUILD_SECONDS)
-                        .record(build_start.elapsed().as_secs_f64());
                     parent_chainwork = Some(block.context.chainwork);
-                    batch_bytes = batch_bytes.saturating_add(approx_indexed_block_bytes(&block));
+                    // One walk, feeding both the byte budget and the counters
+                    let work = BlockWork::tally(&block);
+                    work.record(IngestStage::Finalised);
```
nachog00, line 336: Question, not a blocker: `record()` now fires here in the fetch/build loop, where the old `record_block_throughput` ran in the post-commit `for block in &batch` loop. So if `write_block_batch_blocking` / `env.sync` returns `Err`, these counters have already counted a batch that never committed — and on restart those blocks are re-fetched and re-counted; the counter also leads `SYNC_FINALIZED_HEIGHT` by up to a batch. Given the goal is accurate "ingested" counts, is the fetch-time placement intentional, or should `record()` move back to the post-commit loop while `tally()` / `approx_bytes()` stays here for the batch budget?

PR #1456, opened 2026-08-18
fix(state): read getblockchaininfo through the source port on boot

### LINKED ISSUE #1455: getblockchaininfo mainnet-boot crash: adopt_network parses value-pool ZEC floats
## Symptom

zaino 0.8.0-rc1 crashes at startup against a **mainnet** validator:

```
critical error in backing block source: the validator at <addr> returned an
unreadable getblockchaininfo: loss of precision parsing ZEC value: floating
point had fractional zatoshis
```

## Root cause

`adopt_network` (`packages/zaino-state/src/chain_index/network_adoption.rs`)
deserializes the whole `getblockchaininfo` response into zebra's
`GetBlockchainInfoResponse`, whose value-pool balances are `Zec`
(`#[serde(try_from = "f64")]`). serde therefore parses each pool's ZEC
`chainValue` **float** and requires `chainValue * 1e8` to be an exact integer. At
mainnet scale a pool's float carries ~16 significant digits and does not
round-trip to a whole zatoshi, so the parse fails and the node crashes — even
though adoption only ever uses the upgrade schedule, never the value pools.

zaino's own parser already reads the exact integer `chainValueZat`
(`zaino-source-zebra-rpc` `parse.rs::zatoshis_field`); only this one ad-hoc call
site bypasses it.

## Why it surfaced now

- **0.8.0 regression.** `adopt_network` is new in the 0.8.0 source-stack rewire.
  0.6.0/0.7.0 parsed `getblockchaininfo` through `zaino-fetch`'s `ChainBalance`,
  which reads `chainValueZat` and never touches the float — immune at any size.
- **Mainnet-only.** The crash needs a pool large enough to lose f64 precision;
  regtest/testnet (all CI/integration targets) carry ~0 balances, so no test or
  fixture ever reproduced it.
- **Tip-dependent.** Which pool is "f64-unlucky" changes block to block. Example
  at height ~3,451,185: sapling `chainValue 529544.04149098` →
  `* 1e8 = 52954404149097.99` (exact zat is `52954404149098`).

nachog00:
Fixes #1455.

## Problem

`adopt_network` deserialized `getblockchaininfo` into zebra's
`GetBlockchainInfoResponse`, whose value-pool balances parse the ZEC `chainValue`
float and reject any value that does not round-trip to a whole zatoshi. On
mainnet that crashes the node at startup, even though adoption only needs the
upgrade schedule. See the issue for the full root cause / why-now.

## Change

- Read `getblockchaininfo` through the `zaino-source` `GetBlockchainInfo` port,
  yielding the domain `BlockchainInfo` (value pools as exact zatoshis from
  `chainValueZat`, never the float).
- The two adoption helpers now consume the owned `Vec<NetworkUpgradeInfo>`, keyed
  by consensus branch id. An unknown branch id (a validator ahead of this build)
  is skipped rather than failing adoption — which also removes the latent crash
  on an upgrade *name* this build does not yet know.
- The domain→zebra translation needed only because `adopt_network` must still
  return a zebra `Network` is isolated in a self-contained
  `network_adoption::zebra_bridge` module, to delete once the runtime network
  parameters are domain-ized.

This forwards the 0.8.0 primitives / source-stack migration rather than patching
the call site in place.

## Tests

- New regression gate in `zaino-source-zebra-rpc` `parse.rs`: a value pool with a
  lossy `chainValue` float and an exact `chainValueZat` is read as the integer.
- `adopt_network`'s helper tests moved onto the owned types, plus new
  skip-unknown-branch cases. All green; `clippy` clean.

## Deployment / verification

First hit on a live **mainnet** ephemeral (zebra 6.3.0, post-Ironwood), where the
unfixed rc crash-looped on `getblockchaininfo`. This branch builds via the
`deploy-ephemeral` `ref=` path (BuildKit → `zingodevops/zaino:<hash>`).
End-to-end boot verification against the post-Ironwood mainnet snapshot is
pending a cluster-egress outage on the build node; the fix is otherwise complete
and locally verified.

### DIFF
--- a/packages/zaino-source-zebra-rpc/src/parse.rs
+++ b/packages/zaino-source-zebra-rpc/src/parse.rs
@@ -1351,6 +1351,54 @@ mod tests {
         assert_eq!(u64::from(info.chain_supply.chain_value), 1_000);
     }
 
+    /// The wire reports each pool balance twice — an exact `chainValueZat` and a
+    /// ZEC `chainValue` float. A large mainnet balance has a `chainValue` that
+    /// does not round-trip to a whole zatoshi: real sapling `529544.04149098`
+    /// gives `529544.04149098 * 1e8 = 52954404149097.99`, off by a fraction of a
+    /// zatoshi. The domain must read the exact integer and never the float.
+    ///
+    /// Regression gate for the mainnet-boot crash: `adopt_network` bypassed this
+    /// parser and deserialized into zebra's `Zec`-typed response, whose
+    /// `try_from = "f64"` rejected exactly this value with "floating point had
+    /// fractional zatoshis". Reading `chainValueZat` here is what makes the
+    /// domain immune, so if this ever flips to the float, boot breaks again.
+    #[test]
+    fn a_value_pool_reads_the_exact_zatoshi_over_a_lossy_zec_float() {
+        // A real mainnet sapling balance, reported both ways. The ZEC float does
+        // not round-trip: `SAPLING_ZEC * 1e8 = 52954404149097.99`, a fractional
+        // zatoshi. Reading SAPLING_ZAT is what keeps the domain immune.
+        const SAPLING_ZAT: u64 = 52_954_404_149_098;
+        const SAPLING_ZEC: f64 = 529_544.04149098;
+
+        let info = parse_blockchain_info(&serde_json::json!({
+            "chain": "main",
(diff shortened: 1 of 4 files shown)


zancas (commented):
UtAck

idky137 (approved):
Looks good! utACK

PR #1459, opened 2026-08-19
feat(source): make the resilient seam a first-class, sealed, default-named port

nachog00:
# Make the resilient seam a first-class, sealed, default-named port

## What this does

Turns `zaino_source`'s retry-aware layer from an unwired, idiosyncratic wrapper into
the **default port surface**. Every question a consumer can ask now exists as two ports:

- `OneShotGetBlock`, `OneShotGetChainTip`, … — the **single-attempt** contract
  (returns `QueryError`). Adapters implement these.
- `GetBlock`, `GetChainTip`, … — the **canonical** (unqualified) **resilient** ports
  (return `SourceError`, whose `Unavailable` variant is "retried and gave up").
  Consumers bind these.

Resilience is the **default name** because it is the default a consumer should get: writing
`S: GetBlock` yields a source that has already handled transience, and the type says so.
Reaching the single-attempt contract means naming the awkward `OneShot*` — which is exactly
what an adapter author (or a not-yet-migrated legacy consumer) does, and no one else.

The resilient ports are **sealed**: the only implementor is `Resilient<V>`. So a value
satisfying a canonical port has provably been through the retry ladder — a consumer cannot
hand-roll a type that claims resilience without retrying. **Resilient by construction, not by
convention.**

## Motivation

Reviewing the chain-head extraction (#1440) surfaced that `zaino-chain-head-service` runs its
own backoff ladder and models failures as "transient by assumption" — the wrong layer.
`zaino_source::Resilient<V>` already existed to own that, but was **never wired** and was a
concrete type with its own inherent methods, so no consumer could depend on a retry-aware
seam through the port system. This PR fixes the seam so the chain head (next PR) can consume
it and delete its own retry.

## How it's built

- **`#[resilient_port]`** (new `zaino-source-macros` crate) — an attribute on a `OneShot*`
  trait that reads its method signatures and derives the canonical twin trait + the
  `Resilient<V>` blanket impl, rewriting `QueryError<E>` → `SourceError<E>`. The twin's
  signature is never restated; the retry ladder and the error translation live once in
  `Resilient::with_retry`. A `macro_rules!` couldn't do this — it emits a trait item and an
  impl item — so it's a proc-macro; exclusion (below) is just "no annotation."
- **`Resilient<V>`** implements the sealed resilient ports for any `V` implementing the
  matching `OneShot*` ports, and forwards `SubscribeBlocks` / `SubscribeChainTip` unchanged
  (subscriptions are synchronous `Option<Receiver>`, not fallible queries — nothing to retry;
  transience is retried on the query ports a wake triggers).
- **`SendRawTransaction` gets no resilient port** — retrying a non-idempotent send risks a
  double-submit. Its canonical name stays reserved for a send that is safe to retry.

## Commit structure

- **`775580f2` … `46d149a5`** (22 commits) — rename every one-shot query port
  `GetX → OneShotGetX`, freeing the canonical names. Pure rename, done with rust-analyzer's
  Rename Symbol (the names collide with `zebra_rpc::methods::` types, so a text sweep was not
  safe). Committed per-rename for a legible history.
- **`db467664`** — add the `#[resilient_port]` proc-macro, seal the ports, pilot 3 ports.
- **`337e12d0`** — derive twins for all 34 idempotent ports.
- **`8f860def`** — subscription passthroughs + a compile-time bound test.
- **`ab72abbd`** — usage.md / CHANGELOG.

## Not in this PR (follow-on)

- **The adapter front door** — `ZebraValidator::rpc_only(rpc, policy) -> Resilient<Self>`, so a
  consumer never wraps by hand — and demoting the bare constructors. First step of the
  chain-head wiring PR.
- **Wiring the chain head** onto the resilient port and **deleting its backoff ladder** +
  retry config.
- **Legacy debt:** `ChainIndex`'s scaffolding (`ValidatorSource` / `ChainIndexSourcePorts`)
  still consumes the `OneShot*` ports directly, so the bare adapter can't be fully hidden until
  that migrates. The seal already guarantees "holds a resilient port ⇒ retried"; wrapped-only
  ("can only *build* a resilient one") completes then.

## Verification

`cargo fmt --check`, `makers lint-boundary-conversions`, `cargo clippy` (production crates, no
new warnings), and `cargo nextest run -p zaino-source -p zaino-source-macros -p zaino-state`
— **251 passed, 1 skipped** — all green on the final tree.


AloeareV (approved):
It's been a while since I've encountered proc_macros. These are great!

AloeareV (dismissed):
Fails to compile after merge of dev into feature

zancas (approved):

PR #1467, opened 2026-08-20
Add sync plus concurrency tests

idky137:
## Summary

Adds `zaino-bench`, a benchmark harness for sync time / connection ceiling / serve rate, then uses it to measure mainnet — and fixes the three bulk-sync bottlenecks the measurements exposed. `docs/perf.md` carries the numbers and the configuration that produced them.

## Results (mainnet, Direct backend, NVMe, 16-core)

| | Persistent | Ephemeral |
|---|---|---|
| Full sync from genesis | **4h 43m**, 196 blocks/s mean | n/a |
| Concurrent connections | **5,000 @ 100%** | **5,000 @ 100%** |
| Aggregate throughput | ~200,000 blocks/s | ~9,400 blocks/s |
| Single-stream serve | **154,203 blocks/s** | 399 blocks/s (timed out) |

Zero chain breaks anywhere. The ~390× serve gap between modes is the value of the index stated plainly.

**The sync mean hides the shape of the run.** Heights 1,705,519–1,823,494 — the sandblast band — are **3.5% of the chain and 68% of the wall-clock**, at 10.3 blocks/s against 2,151 blocks/s below 1.7M. A profile there puts ~91% of cycles in BLS12-381 scalar arithmetic: zebra decompresses two Jubjub points per Sapling output on read, and Zaino re-serialises them for the compact form. Both directions are work Zaino doesn't need.

## Changes

**`zaino-bench`** (new crate, not a `default-member`, so a bare `cargo nextest run` never builds it) — `sync`, `concurrent`, `serve`.

**Three bulk-sync perf changes**, each measured before and after:
- Pipeline batch commits with block building — the commit was 40% of wall-clock with the builder idle throughout.
- Fetch blocks concurrently — parallelises the Jubjub decompression across cores.
- Assemble blocks concurrently — profiling showed 54% of CPU still on one thread doing the re-serialisation; chainwork is a running sum over each block's own header, so the fold no longer holds the conversion in block order.

**Two harness fixes found by using it:**
- Completion/stall detection watched the fetch pointer, not the durable tip, so a finished sync read as a stall.
- Connections are now held at a barrier until all have dialled. Without it, reads shorter than the spawn ramp retired as fast as they were created — a sweep asking for 10,000 never had more than 38 open. Every concurrency figure here depends on this fix.

**One production fix:** LMDB `max_readers` was `(cpu × 32).clamp(512, 4096)`, i.e. exactly 512 on any host with ≤16 cores. Raised the floor to 2048.

## Please read before merging

**A client can still restart a node.** With `NO_TLS` a reader slot belongs to a read *transaction*, so concurrent reads exhaust the table. `MDB_READERS_FULL` is backpressure, but the startup block scan treats any error as fatal (`finalised_source/v1.rs`, next to an existing `TODO: Handle error better?`) → `CriticalError` → restart → into the same load → restart. The earlier 10,000-connection sweep hit exactly this and looped. **Raising the floor moved where it bites and fixed nothing.** The reclassification is a few hours' work and is not in this PR.

Also: `(cpu × 32).clamp(2048, 8192)` still returns the floor on anything up to 256 cores, so the ceiling never engages in practice. Worth reshaping.

**The numbers are this branch, not `dev`** — they include the three perf changes. `perf.md` says so in its status banner. A `dev` measurement would be materially slower.

**Known limits of the measurement**, all recorded inline in `perf.md`: 5,000 is the largest count tested rather than a ceiling (neither sweep found a knee); `ulimit -n` was 8192 against the ~10,000 that 5,000 connections need, so that row is partly client-bound; persistent figures are warm-cache (first run ~110k blocks/s vs ~200k warm, two warm runs within 5%); and the harness's t0 misses the first 107,349 blocks, since the progress gauge only appears once the write loop's throttled branch fires.

## Testing

`cargo nextest run` — 611 pass, 3 skipped. Clippy and fmt clean on production crates (pre-existing warnings remain in `live-tests`). `makers lint-boundary-conversions` passes. The `transparent_address_history_experimental` feature does not compile, but did not before this branch either — same three errors with the changes stashed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,22 @@ and this library adheres to Rust's notion of
   cadence and exclude-list caps operator-configurable.
 
 ### Changed
+- **Bulk finalised-state sync now assembles blocks concurrently too.** Making
+  the fetch concurrent exposed the other half: a profile through the sandblast
+  heights put **54% of all CPU on a single thread**, holding the run to 8
+  blocks/s on 1.8 of 16 cores. That thread was the chainwork fold, which ran
+  `assemble_indexed_block` in block order. Assembly is as expensive as the
+  fetch and for the mirror-image reason — converting to the compact form
+  re-serialises the Jubjub points zebra just decompressed, and returning to
+  affine coordinates costs a field inversion per point. By Amdahl that capped
+  the whole run at 1.85x however many cores the fetches used.
+
+  A block's own proof-of-work depends only on its own header, so the cumulative
+  chainwork is a running sum that can be folded over already-fetched blocks in a
+  separate pass. Bulk sync is now three: fetch a window concurrently, fold the
+  chainwork in order (integer arithmetic, microseconds), then assemble the
+  window concurrently on `spawn_blocking`. The ordering guarantee is unchanged —
+  every block still gets exactly its parent's chainwork plus its own.
 - **Bulk finalised-state sync now fetches blocks concurrently.** A profile of a
   mainnet sync through the sandblast heights (~1.7M) put **91% of cycles in
   BLS12-381 scalar arithmetic** — `sqrt_tonelli_shanks`, `Scalar::square`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,30 @@ and this library adheres to Rust's notion of
   cadence and exclude-list caps operator-configurable.
 
 ### Changed
+- **Bulk finalised-state sync now fetches blocks concurrently.** A profile of a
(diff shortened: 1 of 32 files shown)


elicbarbieri (dismissed):
I have some reservations and find the finalized_state/finalized_source/v1/write_core.rs stuff a bit of an anti-pattern w/ feature-gated switches for how we commit to database, but I think its fine temporarily. 

Nacho-sync will address this, and will refactor out the write pipeline, so I will approve this for now.   Same with the zaino-bench, It looks good overall, but will get replaced with ztest over next month or two, so not worried about it.  Overall looks good for now.  LGTM

zancas (approved):

PR #1470, opened 2026-08-24
perf: overlap the block and tree-roots fetches in the catch-up walk

zancas:
This PR speeds up zainod's startup wait. Since cc30eb28, the chain
head publishes Ready only after its first catch-up tick completes,
and zainod opens its listeners only on Ready. That tick's walk from
the anchor fetched each block and then its tree roots in sequence,
because the roots query needed the hash the block fetch returned.
That dependency cost two serial validator round trips per block, over
a window of up to `MAX_NONFINALISED_DEPTH` blocks.

The reworked walk makes two reads for each height. The first read
fetches the block at that height. The second read fetches the
commitment tree roots at that height through the new
`GetCommitmentTreeRootsByHeight` port, whose answer carries the
answering block's hash so the walk can detect a reorg.

The walk now issues both reads under one `tokio::join!` and keeps the
roots only when their hash matches the fetched block. On a mismatch or
a failed by-height read, it refetches by the authoritative hash. The
reorg path is unchanged and stays hash-addressed.

### DIFF
--- a/packages/zaino-chain-head-service/src/service.rs
+++ b/packages/zaino-chain-head-service/src/service.rs
@@ -402,10 +402,17 @@ impl<S: ChainHeadBlockSource> ChainHeadService<S> {
         //
         // see https://github.com/ZcashFoundation/zebra/issues/9541
         while u32::from(graph.best_tip().height) < u32::from(chain_height) {
-            let Some(block) = self
-                .block_at_height(next_height(graph.best_tip().height))
-                .await?
-            else {
+            let next = next_height(graph.best_tip().height);
+            // Both reads address the best-chain block at `next`, so they are
+            // independent and run concurrently; the hash check below closes
+            // the reorg race the concurrency opens.
+            let (block, roots) = tokio::join!(
+                self.block_at_height(next),
+                self.source.get_commitment_tree_roots_by_height(next),
+            );
+            let Some(block) = block? else {
+                // Past the tip; the roots result is dropped unexamined, since
+                // its `HeightNotFound` there is expected, not a failure.
                 break;
             };
 
@@ -422,7 +429,14 @@ impl<S: ChainHeadBlockSource> ChainHeadService<S> {
                         ))
                     })?
                     .clone();
-                let chainblock = self.block_to_chainblock(&prev_block, &block).await?;
+                let roots = match roots {
+                    Ok((roots_hash, roots)) if roots_hash == block.header.hash => roots,
+                    // A reorg swapped the best-chain block between the two
+                    // reads, or the height-addressed read failed; the hash we
+                    // hold is authoritative, so refetch by it.
+                    _ => self.tree_roots(block.header.hash).await?,
+                };
+                let chainblock = chain_head_block(block.clone(), &roots, Some(prev_block.work))?;
                 info!(
                     height = u32::from(chainblock.height()),
                     hash = %chainblock.hash(),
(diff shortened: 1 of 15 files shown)


packages/zaino-chain-head-service/src/service.rs
```diff
@@ -202,7 +201,10 @@ impl<S: ChainHeadBlockSource> ChainHeadService<S> {
             task: Mutex::new(None),
             config,
         });
-        service.status.store(StatusType::Ready);
+        // Still `Syncing`: the anchor is the window's floor, not its tip, so a
```
zancas, line 204: I've not analyzed the possible code paths from sync initiation to here.   What enforces "Still `Syncing`"..  is that  a compile-time guarantee?

nachog00: i don't think it is. The Status type exposes the store interface for any consumer to authoritatively change the value. It would be very useful to add a state machine wrapper instead

nachog00: see #1477

AloeareV (dismissed):
Looks great, cannot merge due to conflicts

packages/zaino-chain-head-service/src/service.rs
zancas: You are correct that the type gives no guarantee. `NamedAtomicStatus::store` is
public, and the cell is `Clone`. In this crate the structure is tighter than
the type.

- The cell is a private field of `ChainHeadService`. It starts as `Syncing`.
  Its only clone goes to `ChainHeadSubscriber`, which keeps it private and
  calls only `load()`. No code outside the crate can write it.
- Each transition of the writer goes through `next_status`, which is a pure
  and total function. `Ready` results only from `TickOutcome::Advanced`. The
  one direct `store` writes `Closing`.

This is a guarantee from structure and not from the compiler. No test held it,
so a change could break it with no signal. This PR now adds
`an_anchored_service_is_syncing_until_its_first_advance`. The test anchors
below the tip and asserts `Syncing`. It then advances and asserts `Ready`.

#1477 is the correct place for the wider problem. The design in #1550 has the
separation that #1477 asks for: `Lifecycle` and `Health` are different types,
and `RunReporter::ready()` is the only path to `Ready`.

zancas:
The fetches can run across blocks. The graph update must stay in order.

Each iteration compares the `prev_hash` of the block it fetched with the tip of
the graph. That comparison decides between an extension and a reorg. It needs
the graph as the previous iteration left it, so the decision for height `n+1`
waits for height `n`. The fetches do not have that dependency. The
cumulative-work fold has one, but it is an integer addition over headers. The
finalised bulk sync already splits the same problem: `write_core.rs` fetches
with `buffered(concurrency)` and folds the work in order afterwards.

A cross-block walk would take this shape:

1. Prefetch the pairs of block and roots for the heights from the tip of the
   graph to the tip of the validator, as a `buffered` stream. The window is at
   most `MAX_NONFINALISED_DEPTH` blocks, so the prefetch is bounded.
2. Consume the pairs in order. Apply the same `prev_hash` check and the same
   rule for the roots hash that this PR applies.
3. On a reorg, drop the prefetched pairs, run `handle_reorg`, and restart the
   stream from the new tip.

The cost is a second path through a walk that a reorg already complicates, and
one validator request in flight for each prefetched height. Under HTTP/1.1
each request holds a connection, which #1599 addresses. This PR halves the
serial round trips already. The gain from a prefetch grows with the depth of
the window, and I have no measurement of the start-up time after this PR. I
propose that we land this PR as it is. #1602 tracks the cross-block
prefetch, with that measurement as its first step.

elicbarbieri (approved):

PR #1471, opened 2026-08-25
Fix/ironwood proto

idky137:
Fixes: Bug in ironwood proto.

Pulls https://github.com/zingolabs/zaino/pull/1468 and adds 2 fixes: 1 test fix and one cargo fnt fix.


AloeareV (approved):
Great, thanks for fixing up the tests here. And thanks @nfl0 for the initial fix!

PR #1483, opened 2026-08-26
Feat: Isolate finalised state behind ports

idky137:
# Isolate the finalised state behind ports

Splits the LMDB finalised state out of `zaino-state` into a domain crate and an
adapter crate, and rewires `ChainIndex` to read through the new ports.

## What it adds, and why

Three new crates:

| Crate | Role | `.rs` lines at HEAD |
|---|---|---|
| `zaino-encoding` | Zero-dependency leaf: the `ZainoVersionedSerde` traits and byte helpers, no domain-type impls | 1,065 |
| `zaino-chain-store` | Domain crate: ports, business types, errors, config. No implementations | 2,004 |
| `zaino-chain-store-zainodb` | Adapter: the existing LMDB backend, moved, plus the port impls | 31,490 |

**Why.** `zaino-state` currently owns the finalised state, the chain head, the
sync loop and the serving façade in one crate, so any change to block storage
recompiles and retests all of it, and a second storage backend has nowhere to
live. This PR gives the finalised state the same shape the mempool and chain
head already have — a domain crate holding the vocabulary and the ports, an
adapter crate holding one runtime — so nacho's modular sync engine can land as a
second `zaino-chain-store` adapter without touching the chain head, and the
store's tests run without booting the rest of Zaino.

**The implementation moves as verbatim as possible.** The port layer is new; the
LMDB backend, the ephemeral passthrough, the checksummed row encoding, the
validation sweeps and the migration machinery all move unchanged. Commit 0 pins
every on-disk encoding with checked-in golden hex vectors *before* anything
moves, so every later commit is gated on byte-for-byte identity. All on-disk
artefacts (`db_schema_v1.txt`, the test vector `.dat`/`.json` files) are pure
renames with zero byte changes.

**What the port layer changed on purpose:**

- Block reads are **chunked**: `blocks_stream` / `compact_stream` yield
  `Vec<Block>` per read transaction rather than one block per channel send,
  cutting sends on `GetBlockRange` ~1000×. The streams are nested RPITIT with
  precise capturing (`use<Self>`), so there is no `Box<dyn Stream>` and no
  borrow of the reader — a stream is `'static` and can be moved into a
  per-request task.
- `tonic::Status` is **out of the storage layer**. The error type is
  `ChainStoreError`; `zaino-serve` maps it.
- The capability bit `TRANSPARENT_HIST_EXT`, which conflated three unrelated
  capabilities, is split into `SPENT_OUTPUT_INDEX` / `TXOUT_SET_INDEX` /
  `TRANSPARENT_HIST_INDEX`.
- `ChainStoreConfig` (domain) holds the storage-agnostic knobs with private
  fields and `NonZero` where zero is meaningless; `ZainoDbConfig` (adapter)
  holds only the LMDB extras. `ephemeral: bool` becomes `path: Option<PathBuf>`,
  making "ephemeral with a path" unrepresentable.

**Four defects found by wiring `ChainIndex` onto the ports** (all fixed here,
each folded into the commit that owns the file, each with a regression test):

1. The watermark was never advanced by the background build path, so every
   bounded read returned `NotReady` after a build.
2. A passthrough store refused reads above its durable tip, because
   `watermark_provenance` did not account for both ephemeral modes.
3. Non-standard transparent outputs round-tripped to an all-zero address key;
   they now round-trip through the 21-byte `tag ‖ hash` form.
4. Sapling/Orchard/Ironwood value balances round-tripped to `None`; `StoredTx`
   now carries them.

Also fixed: the missing v1.3 arm in `DbVersion::capability()` (latent, but a
live fail-closed hazard), `get_txid` returning `Custom` rather than
`DataUnavailable` for an out-of-range index, and a `commitment_tree_data`
truncation.

**Not in this PR.** `ChainIndex` still exists as the consumer; PR2a replaces it
with ChainView and deletes the interim `StoreCapabilities` consumer along with
it. The transparent-address-history port is declared and implemented but stays
behind `transparent_address_history_experimental`, as it is validator
passthrough today.

## Verification

- 701 unit/integration tests pass (`--all-features`); clippy 0 warnings; fmt clean.
- `makers test clientless` in the CI container: 54 passed, 14 skipped.
- Golden encoding vectors from commit 0 pass unchanged after every later commit.

## Overall line breakdown

Cumulative diff over the 14 commits, rename-detected (so a moved file counts
only its edited lines, not its full length):

| | added | removed |
|---|---:|---:|
| Production code | 6,356 | 2,505 |
| Test code (incl. `#[cfg(test)]` modules and test methods) | 3,532 | 808 |
| Documentation (incl. `///` and `//!` doc comments) | 3,646 | 307 |
| **Total** | **13,534** | **3,779** |

Roughly 24k lines of LMDB backend and its test suites are relocated across the
crate boundary; rename detection keeps them out of the numbers above, which is
why the diff reads as ~13.5k rather than ~38k.

## Per-commit breakdown

Counts are added/removed for that commit alone.

| # | Commit | What changed | Prod | Test | Docs |
|---|---|---|---|---|---|
| 0 | `09aa206d` test: golden vectors | Checked-in hex vectors for every `ZainoVersionedSerde` impl in `types/db/legacy.rs` and `capability.rs` — encode, decode and `FixedEncodedLen` byte-for-byte — plus the `DB_SCHEMA_V1_HASH` assertion. Lands before anything moves; gates every later commit. | +0 / −0 | +622 / −0 | +0 / −0 |
| 1 | `29e2792e` fix: v1.3 capability arm | `DB_VERSION_V1` is `1.3.0` but `capability()` had no `(1, 3)` arm and fell through to `Capability::empty()`. Latent only because `Router` consults `FinalisedSource::capability()`. | +6 / −1 | +66 / −0 | +0 / −0 |
| 2 | `31a84a4b` fix: experimental feature set + CI | Repairs the `transparent_address_history_experimental` feature wiring and builds it in CI, so the feature-gated port added later is actually compiled. | +93 / −4 | +0 / −0 | +0 / −0 |
| 3 | `3d1f954c` refactor: `MempoolInfo` → primitives | Mempool vocabulary was living in `types/db` (there was a TODO). Lifting it lets `types/db/` move wholesale. The only mempool-adjacent touch in the PR. | +17 / −52 | +8 / −18 | +23 / −11 |
| 4 | `a1a84883` feat: `zaino-encoding` | New zero-dep leaf crate; `encoding.rs` (1,065 lines) moves verbatim, `zaino-state` re-exports from it. | +56 / −16 | +3 / −3 | +1 / −1 |
| 5 | `e642c6a6` feat: primitives vocabulary | `BlockTxPosition` and `ScriptType` into `zaino-primitives` — shared by the store and, later, ChainView. | +39 / −1 | +27 / −0 | +48 / −1 |
| 6 | `7290559c` feat: `zaino-chain-store` | The domain crate: `ports.rs`, `block.rs`, `error.rs`, `config.rs`, `capabilities.rs`, `transparent.rs`, manifest, CHANGELOG, usage guide. Zero implementations; a compile-time source-bound test asserts the driven port is satisfied. | +1,012 / −16 | +374 / −0 | +877 / −0 |
| 7 | `2c5f4884` fix: name the validator questions | Corrects the driven-port alias to the questions the store actually asks (`GetTransaction` is not one of them). | +14 / −8 | +0 / −0 | +28 / −10 |
| 8 | `63706dfd` feat: move persisted types | `types/db/**` into the backend crate. 17 of 20 files are renames; the edits are import rewrites and visibility narrowing. | +159 / −114 | +127 / −19 | +129 / −30 |
| 9 | `93d44251` feat: move the finalised state | `finalised_state/**` (~17k lines) into the backend crate, `ephemeral.rs` verbatim with both roles intact. Drops `tonic::Status` from the storage layer, re-splits the three capability bits, makes ~28 internal-only reader methods private, deletes the stale shadow-migration prose. 19 of 30 files are renames. | +2,845 / −1,799 | +204 / −185 | +820 / −157 |
| 10 | `43f74d92` feat: implement the ports | The adapter's port impls: chunked block and compact streams, `StoredTx` with its value balances, the non-standard-script round trip, `refresh_watermark` published from the background build, `ZainoDbConfig`. | +1,578 / −0 | +200 / −0 | +345 / −0 |
| 11 | `5f311bf3` refactor: read through the store | `chain_index/chain_store.rs` — ~13 conversion helpers, each generic over one port trait, beside the existing `chain_head.rs` bridge. `ChainIndex`'s own types are unchanged; the sync loop calls `build_to`. Deletes the finalised-state half of `error.rs`. | +573 / −534 | +240 / −53 | +230 / −103 |
| 12 | `e21ba0cb` test: move the suites | Unit, v1 and migration suites into the backend crate against a local fake source, plus the ~28-test differential port suite that asks every read twice — once through the port, once through the inherent reader — and requires agreement. Proptest and mockchain stay in `zaino-state`. | +0 / −0 | +1,687 / −555 | +0 / −0 |
| 13 | `09b4a66f` docs: ADR, changelogs, guides | ADR 0012, three CHANGELOGs and three usage guides, root README index. The store guide carries worked examples for `blocks_stream` (pinning, per-chunk `Result`, drop-as-cancellation, why not `Arc`), descending ranges, the cross-seam merge and the `txout_set` fold; the adapter guide documents the stream-port constraints for the next implementer. | +3 / −0 | +0 / −0 | +1,154 / −3 |

Classification: files under `tests/`, `live-tests/` and `vectors/` are test;
`.md` and `docs/` are documentation; within production `.rs` files, lines inside
`#[cfg(test)]` items are test and `///`/`//!` lines are documentation. The
per-commit rows sum slightly above the cumulative total because later commits
edit lines earlier ones added.


packages/zaino-chain-store-zainodb/src/ports.rs
```diff
@@ -0,0 +1,1515 @@
...
+}
+
+/// Which domain capability a routing refusal was about.
+///
+/// The router refuses with a static feature name, which is this crate's
+/// vocabulary; the domain's is coarser. Anything unrecognised maps to
+/// [`StoreCapability::Core`], which is the conservative reading: a store that
+/// cannot say which capability it lacks is reported as lacking the one every
+/// store must have, so the caller routes elsewhere rather than retrying.
+fn capability_for_feature(feature: &str) -> StoreCapability {
```
nachog00: **Bug: `capability_for_feature` never matches the names the router actually produces — every routing refusal collapses to `Core`.**

This matches lowercase literals (`"spent_output_index"`, `"txout_set_index"`, `"transparent_history"`), but the router raises `FeatureUnavailable(cap.name())` (`store/router.rs:562`), and `CapabilityRequest::name()` (`store/capability.rs:302-311`) returns **uppercase** (`"SPENT_OUTPUT_INDEX"`, …). None match, so every router-originated `FeatureUnavailable` falls through to `_ => StoreCapability::Core` — the one capability every store has. A consumer told "Core unavailable" for a merely-absent optional index will fail over / degrade a query that only needed that index.

(The `finalised_source.rs` paths emit the lowercase `"transparent_history"` directly, so that one path happens to match — which is what makes the mismatch easy to miss.)

The guard test `a_routing_refusal_maps_to_the_capability_it_denied` (line 1475) doesn't catch this: it feeds a hand-written `"spent_output_index"` that the router never emits, and its one real router name (`"READ_CORE"`) legitimately maps to `Core`. Feeding `CapabilityRequest::SpentOutputIndex.name()` would exercise the real producer and fail today.

Root fix: thread the typed `CapabilityRequest` (or a shared `const`) through `FeatureUnavailable` instead of a bare `&str`, so producer and matcher can't drift.

packages/zaino-state/src/chain_index.rs
```diff
@@ -2581,18 +2598,14 @@ impl<Source: BlockchainSource + WithChainHeadSource> ChainIndexRpcExt
...
         use std::collections::HashMap;
+        use zaino_chain_store::{is_unspendable, StoredTxOut, TXOUT_SET_ENTRY_LEN};
 
         let snapshot = self.snapshot_nonfinalized_state();
         let best_tip = self.best_chaintip(&snapshot).await?;
 
-        let mut accumulator = self
-            .finalized_state
-            .get_tx_out_set_info_accumulator()
+        let mut accumulator = chain_store::txout_set(&self.finalized_state)
```
nachog00: **Regression: this re-wraps an already-typed `ChainIndexError`, flattening its kind and dropping its source.**

`chain_store::txout_set` (`chain_index/chain_store.rs:255`) already returns `ChainIndexError` — via `?` and the `From<ChainStoreError>` impl — so `e` here is a `ChainIndexError`. Wrapping it again in `ChainIndexError::internal(format!("…{e}"))` forces `kind = InternalServerError` (a retryable `Unavailable` becomes non-retryable) and sets `source: None`, discarding the typed `ChainStoreError` the `From` impl just attached. The `previous_output` call site (line ~2686) has the same shape.

On `dev` these `map_err`s converted a *foreign* `FinalisedStateError`, which was correct; after the port migration the inner call already yields `ChainIndexError`, so the wrap is now strictly lossy. Dropping the `map_err` and propagating with `?` restores both the kind and the source chain (call site 2 keeps its `.ok_or_else(...)` for the `None` case).

packages/zaino-state/src/chain_index.rs
```diff
@@ -2131,13 +2148,15 @@ impl<Source: BlockchainSource + WithChainHeadSource> ChainIndexRpcExt
...
                         .await
                     {
-                        Ok(block) => block,
-                        Err(_) => self
+                        Ok(Some(block)) => block,
+                        // A miss and a failure both fall through to the
+                        // validator, as they did before: the store not holding
+                        // the block is the ordinary case on a node that is
+                        // still building, and passthrough is what covers it.
+                        Ok(None) | Err(_) => self
```
nachog00, line 2175: **`Err(_)` here silently discards a typed store failure and routes it to the validator as if it were a miss.**

Collapsing `Ok(None)` and `Err(_)` means a `ChainStoreError::Backend` (an LMDB cursor desync) or `MissingRow` (on-disk corruption) is indistinguishable from "the store hasn't built this height yet," so a genuinely broken database masquerades as an empty finalised half indefinitely, silently, with the validator quietly covering for it.

The comment is right that a *miss* should fall through — but a *failure* is not a miss. Splitting the arm so `Ok(None)` falls through while `Err(e)` first emits a `tracing::warn!(error = &e as &dyn std::error::Error, %height, "finalised compact-block read failed; falling back to validator")` keeps the passthrough behaviour but gives an operator a signal that the store is broken rather than merely behind. (Pre-existing arm, but the PR rewrote it and now has a much richer typed error to discard.)

packages/zaino-chain-store-zainodb/src/ports.rs
```diff
@@ -0,0 +1,1515 @@
...
+
+/// This crate's height, as the domain names it.
+///
+/// The stored height is any `u32`; the domain's is validated against the
+/// protocol maximum. A stored height that cannot be expressed is a corrupt row,
+/// not a caller error, so it surfaces as [`ChainStoreError::MissingRow`] rather
+/// than being clamped into a height that names a different block.
+fn domain_height(height: Height) -> Result<DomainHeight, ChainStoreError> {
+    DomainHeight::try_from(height.0)
+        .map_err(|_| ChainStoreError::MissingRow(format!("valid height for stored value {height}")))
```
nachog00: **These map *value-validation* failures onto `MissingRow`, and discard the underlying error with `|_|`.**

The doc explains the intent — a stored value that can't be re-expressed is treated as corruption — and that's a defensible call. But `MissingRow`'s own contract is "a row the store's indexes reference is **missing**" (an internal *dangling reference*), whereas here the row is *present and readable* but holds an out-of-range value. An operator alerting on `MissingRow`, or any future code branching on it, will read "dangling index entry" when the truth is "corrupt value in an existing row" — two different repair paths. A distinct variant (`CorruptRow`/`Inconsistent`) would name what actually happened. Same shape at `:203` (`Zatoshis::new`) and `:240` (block time).

Separately, `|_|` throws away the typed conversion error at each site. Since the domain error is `String`-based this is partly forced (see the `error.rs` note on the `Eq` derive) — but the `format!` messages omit the actual failure, so nothing survives for the log. Worth carrying the cause once the error type can hold a `#[source]`.

packages/zaino-chain-store-zainodb/src/ports.rs
```diff
@@ -0,0 +1,1515 @@
...
+/// not be branched on. Inventing a domain meaning for an LMDB error would give
+/// a consumer something to branch on that this backend cannot promise another
+/// one would produce.
+fn chain_store_error(error: StoreError) -> ChainStoreError {
+    match error {
+        StoreError::DataUnavailable(what) => ChainStoreError::MissingRow(what),
+        StoreError::FeatureUnavailable(feature) => {
+            ChainStoreError::Unavailable(capability_for_feature(feature))
+        }
+        other => ChainStoreError::Backend(other.to_string()),
```
nachog00: **`Backend(other.to_string())` (and `Commit(...)` at :88) flatten a typed `StoreError` into a string, dropping the causal chain.**

`other.to_string()` renders only the top-level `Display` and discards the `StoreError`'s own `source()`, so `std::error::Error::source()` on the resulting `ChainStoreError` returns `None` — an operator's `tracing::error!(error = &e as &dyn Error)` captures the summary but not the underlying LMDB/validator cause. The design intends `Backend` to be opaque-to-*branching*, which is fine; opaque-to-*logging* is the accidental part.

The blocker to fixing it is upstream: `ChainStoreError` derives `Clone + Eq` (see the comment on `zaino-chain-store/src/error.rs:9`), which is why the cause can't be a `Box<dyn Error>`. Resolve that and these become `Backend(#[source] …)` / `Commit(#[source] …)`.

packages/zaino-chain-store/src/error.rs
```diff
@@ -0,0 +1,104 @@
+//! What can go wrong asking the chain store a question, and what can go wrong
+//! asking a validator one.
+
+use zaino_primitives::types::Height;
+
+use crate::capability::StoreCapability;
+
+/// A chain-store query could not be answered.
+#[derive(Debug, thiserror::Error, Clone, PartialEq, Eq)]
```
nachog00: **The `Clone + PartialEq + Eq` derives are what force every cause to be a `String` instead of a `#[source]`.**

The idiomatic opaque cause — `Box<dyn std::error::Error + Send + Sync>` — is neither `Clone` nor `Eq`, so these derives are what push `Backend`, `MissingRow`, and all four `ChainStoreSourceError` variants to stringify their cause. The downstream cost is that `Error::source()` is `None` for every one of them (see the `ports.rs:64` note): the LMDB errno / validator RPC cause is unrecoverable at the boundary, which is ironic for `Backend`, whose whole job is the operator's log.

What's actually buying the `Eq`? If it's the differential port test comparing two error values with `==`, that could compare discriminants (or use `assert_matches!`) instead, letting the production type drop `Eq` (and probably `Clone`) and give `Backend`/source variants a real `#[source]`. Worth confirming the derive is load-bearing before it dictates the whole error surface.

nachog00 (commented):
The two behavioural bugs:
- `capability_for_feature` never matches the router's uppercase names → every routing refusal collapses to `Core` (`ports.rs:75`), and its guard test masks it.
- `get_tx_out_set_info` re-wraps an already-typed `ChainIndexError`, forcing `InternalServerError` and dropping the source (`chain_index.rs:2608` / `:2686`) — a regression this migration introduced.

**A through-line across the rest:** the capability casing bug, the by-exclusion error routing (`wire_status`, `From<ChainStoreError>`), and the stringified causes are the same mistake — a decision resolved at runtime on a string or a catch-all that the type system was ready to enforce. Threading `CapabilityRequest` through `FeatureUnavailable`, making the error `match`es exhaustive, and dropping `Eq` so causes can be `#[source]` closes both bugs and most of the modeling comments together — so the count is less work than it looks.

**Cross-cutting, not anchored to one line:**
- **Tracing/metrics underuse on the new read surface.** The write path keeps its full instrumentation (moved intact), but the new port/read layer — `ports.rs`, `conversion.rs`, `stream.rs`, and the rewritten compact-block serving path — has zero spans, metrics, or `warn!` on the corrupt-row/decode branches. The wallet-sync read hot path is effectively blind; a corrupt row degrades to a `MissingRow`/`Backend` string with nothing logged. Nothing was *dropped* in the move — the new surface was simply never instrumented.
- **`chain_store.rs` (573 lines)** has the same section-divider shape as `ports.rs` — `// ---` "reading" and "driving" sections → `chain_store/reading.rs` + `chain_store/driving.rs`.
- **Crate name.** `zaino-chain-store-zainodb` reads as "the Zaino DB" rather than "the LMDB backend of chain-store," which undercuts the pluggable-backend framing and is inconsistent with the tech-axis `zaino-backend-lmdb` used elsewhere. `zaino-chain-store-lmdb` names both the port and the backend. (Keep `zainodb` only if "ZainoDB" is a distinct schema identity.)

All findings are in the inline comments, ordered most-severe first. Happy to pair on the error-type reshape — it unblocks the widest set.

packages/zaino-state/src/chain_index/chain_store.rs
```diff
@@ -0,0 +1,573 @@
...
+/// [`ChainStoreError`] precisely so that the crate that answers gRPC is the one
+/// that decides what a client is told.
+fn wire_status(error: ChainStoreError) -> tonic::Status {
+    match error {
+        ChainStoreError::AboveWatermark { .. } | ChainStoreError::InvalidRange { .. } => {
+            tonic::Status::out_of_range(error.to_string())
+        }
+        ChainStoreError::NotReady => tonic::Status::unavailable(error.to_string()),
+        ChainStoreError::MissingRow(_) => tonic::Status::not_found(error.to_string()),
+        other => tonic::Status::internal(other.to_string()),
```
nachog00: **The `other =>` catch-all maps `Unavailable(capability)` to `tonic::Status::internal`, mislabelling a healthy store as broken.**

`ChainStoreError::Unavailable(_)` means "this deployment doesn't build that index" — the store is healthy and the request well-formed — but it falls into `internal`, which tells the client the server faulted. The truthful code is `Unimplemented` (or `FailedPrecondition`). Because the arm is a catch-all rather than an exhaustive `match`, any `ChainStoreError` variant added later also silently inherits `internal`. Listing `Unavailable` explicitly and making the `match` exhaustive closes both.

packages/zaino-state/src/error.rs
```diff
@@ -523,6 +462,33 @@ impl From<FinalisedStateError> for ChainIndexError {
...
+/// to retry can, and the rest are internal: a store that lacks an index or is
+/// missing a row it references is not something the caller can act on.
+///
+/// [`ChainStoreError::AboveWatermark`] maps to internal deliberately. It is not
+/// an error a caller should ever see — it means the finalised half was asked
+/// about a height the recent half owns, which is a routing mistake in
+/// ChainIndex rather than anything the caller did. The read helpers in
+/// [`chain_store`](crate::chain_index::chain_store) turn it into "not here"
+/// before it reaches this conversion; reaching it means one did not.
+impl From<zaino_chain_store::ChainStoreError> for ChainIndexError {
```
nachog00, line 495: **Routing `ChainStoreError` by exclusion sends caller-faults to `InternalServerError`.**

The catch-all `_ =>` maps everything except `NotReady` to `InternalServerError` — including `InvalidRange` (a caller-supplied bad range, properly `InvalidArgument`) and `Unavailable(capability)` (store healthy, index just absent — properly `Unavailable`/unimplemented). Today only ad-hoc guards in the read helpers keep those variants from arriving here; nothing in the type does. And because it's a catch-all rather than an exhaustive `match`, any new `ChainStoreError` variant silently inherits `InternalServerError`.

(Credit where due: the catch-all *does* preserve `source` — the issue is purely the kind. Matching `InvalidRange` / `Unavailable` explicitly and making the `match` exhaustive is enough; this is the same by-exclusion smell as `wire_status`.)

packages/zaino-chain-store/src/capability.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/// The capabilities a store currently offers.
+///
+/// Runtime state, not a type-level fact. A store on an older schema genuinely
+/// lacks indexes a newer one has, and gains them part-way through a migration,
+/// so this can change over the life of one handle. A consumer that cached it
+/// at open will be wrong later.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct StoreCapabilities(Vec<StoreCapability>);
```
nachog00: **The runtime capability set is unlinked from the trait impls it mirrors — a store can advertise what it can't serve, or `impl` a trait it doesn't advertise.**

`capabilities()` returns a `Vec<StoreCapability>` assembled by hand (in the adapter, `store_capabilities`), while the actual reads are gated by separate marker traits (`SpentOutputIndex`, `TxOutSetIndex`, …). Nothing ties the two together: a store that `impl`s `TxOutSetIndex` but forgets `StoreCapability::TxOutSet` in the vec, or vice-versa, compiles fine — so the "a bound names exactly what its holder uses" guarantee weakens to a runtime promise each implementor keeps by discipline. The migration rationale for it being *runtime* state is sound; the un-enforced *coherence* between the compile-time bound and the runtime set is the part worth a second look.

Minor and orthogonal: `Vec` + `sort`/`binary_search` for a closed 7-variant set is more machinery than an enum-set/bitflags would need — and the adapter already has a `Capability` bitflags internally.

packages/zaino-chain-store/src/transparent.rs
```diff
@@ -0,0 +1,109 @@
...
+    ///
+    /// A convenience for a consumer that has already merged both halves and
+    /// wants a figure. Meaningless on one half alone, which is why it is a
+    /// method here rather than a field: a field would invite reading it off an
+    /// unmerged contribution.
+    pub fn net_value(&self) -> i64 {
+        let received: i64 = self
+            .outputs
+            .iter()
+            .map(|o| u64::from(o.output.value) as i64)
```
nachog00: **`u64 ... as i64` can silently wrap, and the method returns a bare `i64` where a signed-value domain type already exists.**

Two things (feature-gated/experimental, but real code):
- `u64::from(...) as i64` truncates rather than errors — a stored value above `i64::MAX` wraps to a negative figure. The per-element `.sum::<i64>()` and the final `received - spent` are unchecked as well, so a corrupt or adversarial effect set can overflow silently instead of being rejected. A `TryFrom`/checked path surfaces that.
- The return is a bare `i64`; `block.rs` already models this exact "signed value delta" as `SignedZatoshis`. Returning that, and summing through a checked add, keeps the quantity typed and the arithmetic guarded.

packages/zaino-chain-store/src/block.rs
```diff
@@ -0,0 +1,226 @@
...
+/// Pushed into the read rather than applied to the result. The store keeps
+/// each pool's per-transaction data separately, so a filter lets it skip
+/// reading and decoding what the caller will not look at — which on a range
+/// query is the difference between touching one pool's data and touching all
+/// of them. Filtering afterwards would do the work anyway and then discard it.
+///
+/// The default is every shielded pool and no transparent data, matching what a
+/// light wallet asks for.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PoolFilter {
```
nachog00, line 131: **`PoolFilter` is four parallel `bool`s where a pool set fits.**

Four independent bools is the classic "should be a set" shape — `all()`, `none()`, and `Default` each enumerate every field, so adding a fifth pool means editing every constructor and every read that inspects them. A `ShieldedPool` enum already exists; the shielded three could be an `EnumSet<ShieldedPool>` (or bitflags) with transparent as one separate flag, making the pool count structural instead of positional. Low urgency — the fields are private behind methods — but cheaper to change now than after more constructors accrete.

packages/zaino-chain-store-zainodb/src/ports.rs
```diff
@@ -0,0 +1,1515 @@
...
+/// out of this store carries reconstructed scripts, which classify back to the
+/// same key — so both origins produce the same rows.
+///
+/// Public because ChainIndex reads blocks through
+/// [`zaino_chain_store::StoredBlockRead`] and still answers
+/// its callers in [`IndexedBlock`], which is also the shape its chain-head
+/// adapter produces. One conversion serving both directions is what keeps a
+/// block from changing shape as it crosses the finalised seam. It goes private
+/// again when `IndexedBlock` stops being ChainIndex's block.
+pub fn indexed_block_from_stored(block: &StoredBlock) -> Result<IndexedBlock, ChainStoreError> {
```
nachog00: **Difficulty parse error stringified into `Backend`, in a block conversion that parallels the typed one in `conversion.rs`.**

`try_from_bits` here (line ~979) has its typed error folded into `ChainStoreError::Backend(format!("… invalid difficulty: {error}"))`, discarding the structured cause — the same lossy stringify `conversion.rs` deliberately avoids (its `indexed_block` at :162 and difficulty wrapper at :203 keep a typed `BlockConversionError`). `indexed_block_from_stored` also re-does the header/difficulty/solution/tx field mapping that `conversion.rs::indexed_block` owns; `commitment_tree_data` in this file was already deduplicated to *delegate*, so this reads like the path that was missed. Reusing the `conversion.rs` conversion (or at least its typed difficulty error) removes both the duplication and the stringify.

packages/zaino-chain-store-zainodb/src/metric_names.rs
```diff
@@ -0,0 +1,34 @@
...
+#![allow(missing_docs)] // names are self-describing; descriptions live in zainod
+
+pub const CHAIN_TIP_HEIGHT: &str = "zaino.chain.tip_height";
+
+pub const SYNC_FINALIZED_HEIGHT: &str = "zaino.sync.finalized_height";
+pub const SYNC_TARGET_HEIGHT: &str = "zaino.sync.target_height";
+pub const SYNC_LAG_BLOCKS: &str = "zaino.sync.lag_blocks";
+pub const SYNC_ITERATIONS_TOTAL: &str = "zaino.sync.iterations_total";
+pub const SYNC_ITERATION_DURATION_SECONDS: &str = "zaino.sync.iteration_duration_seconds";
+pub const SYNC_ERRORS_TOTAL: &str = "zaino.sync.errors_total";
```
nachog00: **`SYNC_ERRORS_TOTAL` is defined here but incremented nowhere in this crate, and the metric-name duplication leaves the pin-test guarding the wrong copy.**

Three related things:
- `SYNC_ERRORS_TOTAL` (:16) is never emitted from zainodb — the write path's failure arms (`write_core.rs`) `warn!` but bump no counter, so a dashboard shows zero store-side write errors during a corruption/rollback storm. The constant exists for exactly this.
- The live write-path strings (`SYNC_BLOCK_WRITE_SECONDS`, `DB_TIP_HEIGHT`, …) are now defined in *both* this crate and `zaino-state` (`lib.rs:39,45`) but emitted only from here (`write_core.rs:610,640`). The pin-test `metric_names_match_zaino_state` (`zaino-bench/src/metrics.rs:227`) pins the `zaino_state` copy — the one that no longer emits — so renaming the live string here silently breaks dashboards without tripping the guard. The pin should follow the emitting copy, or the duplicate should go.
- Minor: several constants (`SYNC_ITERATIONS_TOTAL`, `SYNC_HAS_REACHED_TIP`, `MEMPOOL_*`, `CHAIN_TIP_HEIGHT`, …) are emitted nowhere in this crate — they belong to the sync-loop / mempool / chain-head subsystems and were copied in wholesale. Being `pub const` they dodge dead-code warnings but imply this backend emits metrics it doesn't.

packages/zaino-chain-store-zainodb/src/error.rs
```diff
@@ -0,0 +1,108 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum StoreError {
+    /// The validator could not answer something the store needed.
+    #[error("backing block source failed: {0}")]
+    Source(#[from] ChainStoreSourceError),
+
+    /// Custom Errors.
+    // TODO: Remove before production
+    #[error("Custom error: {0}")]
+    Custom(String),
```
nachog00, line 25: **`Custom(String)` ships with a "remove before production" TODO, and `StatusError` drops its cause.**

A few things on this type (largely a verbatim move, but it's the backend's port-facing error and it feeds `chain_store_error`, so it's in scope):
- `Custom(String)` (:25) carries `// TODO: Remove before production` — a stringly catch-all that probably shouldn't sit on a port-surface error; likewise `Critical(String)` (:57).
- `StatusError(StatusError)` (:67) uses `#[error("… {0:?}")]` with neither `#[from]` nor `#[source]`, so `Error::source()` returns `None` and only a `Debug` render survives. Adding `#[source]` preserves the chain.
- (Credit: `LmdbError` / `SerdeJsonError` already use `#[from]` correctly — so the pattern is right elsewhere in the same enum.)

Since `chain_store_error` collapses everything but `DataUnavailable` / `FeatureUnavailable` into `Backend(other.to_string())`, whatever structure these keep is flattened at the boundary regardless — worth pairing with the `ChainStoreError` `#[source]` fix.

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/write_core.rs
```diff
@@ -1978,34 +1943,31 @@ impl DbV1 {
...
                 }
             }
 
             match txn.del(zaino_db.heights, &block_hash_bytes, None) {
                 Ok(()) | Err(lmdb::Error::NotFound) => {}
-                Err(e) => return Err(FinalisedStateError::LmdbError(e)),
+                Err(e) => return Err(StoreError::LmdbError(e)),
             }
 
             let _ = txn.commit();
```
nachog00, line 1955: **`let _ = txn.commit()` discards the write-transaction result — the block deletion can silently fail to persist.**

The `del` calls above correctly propagate `LmdbError`, but the `commit` that makes them durable is dropped with `let _ =`, so a commit failure leaves the "corrupt block removed" operation silently un-applied while the caller believes it succeeded. The fn already returns `Result<_, StoreError>` and `LmdbError` has a `#[from]`, so `txn.commit()?` closes it. Verbatim from `dev` — flagging because the file's being touched and it matches the swallowed-error class.

packages/zaino-chain-store-zainodb/src/ports.rs
```diff
@@ -0,0 +1,1515 @@
...
+/// The on-disk record names the migration *phase* but not its destination,
+/// because a migration always runs towards the version the running build
+/// targets. So the target comes from the build rather than from disk, and a
+/// database recorded mid-migration by an older build reports the version this
+/// build would take it to.
+fn store_schema(metadata: &DbMetadata) -> StoreSchema {
+    let version = schema_version(metadata.version());
+    let migration = match metadata.migration_status() {
+        MigrationStatus::Empty | MigrationStatus::Complete => MigrationState::Settled,
+        MigrationStatus::PartialBuidInProgress
```
nachog00: **Typo: `PartialBuidInProgress` (missing the `l` in "Build").**

Defined at `store/capability.rs:640` and used here — the misspelling propagates to all 8 sites (this line, `migrations.rs:606/1090`, the `Display`/encode/decode arms in `capability.rs`, a test). Purely cosmetic — the `Display` string (`"Partial build in progress"`) and the numeric tag are both correct, so there's no behavioural or on-disk impact — but it's a `pub(crate)` name every future match on this enum inherits. Cheap to rename now, annoying later.

packages/zaino-chain-store-zainodb/src/ports.rs
```diff
@@ -0,0 +1,1515 @@
...
+use crate::store::capability::{Capability, DbMetadata, MigrationStatus};
+use crate::store::finalised_source::v1::DB_VERSION_V1;
+use crate::store::reader::DbReader;
+use crate::store::FinalisedState;
+use crate::types::{
+    BlockHash, CommitmentTreeData, CompactTxData, Height, IndexedBlock, Outpoint, TransactionHash,
+    TransparentCompactTx, TxLocation, TxOutCompact,
+};
+
+// ---------------------------------------------------------------------------
```
nachog00: **This file is 1515 lines built from 8 `// ---`-delimited sections — and it's misnamed.**

Two things, same fix:

**Naming.** This module is called `ports` but defines no ports — it's ten `impl <zaino-chain-store port> for DbReader<T>`/`FinalisedState<T>` blocks plus conversion helpers. The driven ports live in the `zaino-chain-store` domain crate; an adapter has no ports of its own, it *implements* them. `ports` reads as "ports are defined here," which inverts the layering — a consumer scanning the crate would look here for the port contracts and find only impls.

**Size.** The 1515 lines mix error mapping (58–90), scalar/height/hash conversions (92–172), transparent-output conversions (174–219), stored→domain block/tx conversions (221–436), capability/schema mapping (438–512), the port `impl` blocks (514–943), the domain→stored *write*-direction conversions (949–1204), and the cfg-gated transparent-history impl (1206+) — 46 items, all new code, one merge-conflict magnet.

The split resolves both: `adapter.rs` for the trait impls, `into_domain.rs` / `into_stored.rs` for the two conversion directions, `error_map.rs` for the three `chain_store_*` fns, `history.rs` for the feature-gated impl — none of them called `ports`.

packages/zaino-chain-store-zainodb/src/ports.rs
idky137: Fixed in fb7512b

packages/zaino-state/src/chain_index.rs
idky137: fixed in aa7f3f9f

packages/zaino-state/src/chain_index.rs
idky137: fixed in d9ac7bc2

packages/zaino-chain-store/src/error.rs
idky137: fixed in c4182e98

packages/zaino-chain-store-zainodb/src/ports.rs
idky137: fixed in df53393

packages/zaino-chain-store-zainodb/src/ports.rs
idky137: fixed in c641b1d2

packages/zaino-state/src/chain_index/chain_store.rs
idky137: fixed in 51964670

packages/zaino-state/src/error.rs
idky137: Fixed in 94d86c4a

packages/zaino-chain-store-zainodb/src/ports.rs
idky137: Fixed in 97cbbf48

packages/zaino-chain-store-zainodb/src/metric_names.rs
idky137: fixed in 6199c39f

packages/zaino-chain-store/src/capability.rs
idky137: Fixed in e5e90146.

On the coherence. Each read port now carries a CAPABILITY associated const (defaulted, since it's a fact about the port rather than a choice an implementor makes), and the adapter's mapping is generic over the reader, naming every capability as <R as Port>::CAPABILITY instead of picking a variant by hand. This makes one of your two directions a compile error.

Where I stopped, this doesn't make the runtime bits agree with the compile-time impls, and I don't think it should. This backend implements every port unconditionally and decides at runtime which it can actually serve, that's what the bits are for, and it's why the set is documented as runtime state that changes mid-migration. Forcing that into the type system would mean giving up runtime routing. What was genuinely wrong is what's now fixed: the mapping between port and capability was restated in the adapter instead of read from the port, and that's what let them drift.

One honest exception: TransparentHistoryIndex still names its variant directly, because its impl is feature-gated and so can't be a bound in a build without the feature. There's a comment at the site saying so.

packages/zaino-chain-store/src/transparent.rs
idky137: fixed in 860100d5.

On the overflow, the mechanism as described isn't reachable, but the concern underneath it is. o.output.value is a Zatoshis, and Zatoshis::new validates against MAX_ZATOSHIS (21M ZEC ≈ 2.1e15), about three orders of magnitude below i64::MAX ≈ 9.2e18. So no single stored value can wrap the individual as i64 cast; the type already prevents it.

What is reachable is the part you flagged second: .sum::<i64>() is unchecked, and it takes only ~4,400 near-maximal values to overflow the accumulator. Real data can't do that, total supply bounds it, but these values come off disk, and a corrupt or adversarial effect set is exactly the case worth defending.

The sum now goes through Zatoshis::checked_add at every step and the method returns Option, so a total past the supply is refused rather than reported as a plausible figure. The subtraction stays unchecked with a comment saying why, both sides are supply-bounded, so the difference is always representable.

packages/zaino-chain-store/src/block.rs
idky137: fixed in 552e7361

packages/zaino-chain-store-zainodb/src/error.rs
idky137: I haven't addressed this here as I was trying to not touch implementation code in the same PR that moves it, and have a feeling that addressing this properly would have much larger reach.

Could we open an issue for this?

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/write_core.rs
idky137: Similar to above this is not new code and I think it would be good to address it in its own PR to ensure we implement a  comprehensive fix.

packages/zaino-chain-store-zainodb/src/ports.rs
idky137: fixed in 6a03432

packages/zaino-chain-store-zainodb/src/ports.rs
idky137: fixed in 0ce6a95a

packages/zaino-chain-store-zainodb/src/error.rs
nachog00: Tracked in #1500.

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/write_core.rs
nachog00: Tracked in #1501.

nachog00 (approved):

packages/zaino-chain-store/src/transparent.rs
nachog00: The checked `sum` closes the accumulator-overflow path, and it's correct that a single validated `Zatoshis` can't wrap the `as i64` cast — the type bounds it three orders below `i64::MAX`. Two things remain, and they're the same class of issue rather than the disk-corruption case specifically.

The safety here still rests on a write-time magnitude argument rather than on an invariant the code enforces:

- `signed()` uses `i64::try_from(..).unwrap_or(i64::MAX)`. That is a silent saturation: if a `Zatoshis` ever holds a value outside the assumed range, the conversion does not error, it substitutes `i64::MAX` and yields a plausible-looking figure. That is the exact failure mode the checked `sum` was added to prevent, reintroduced one line down. It should fail loudly — propagate, or `expect` naming the invariant.
- `SignedZatoshis::new(i64)` validates nothing; any `i64` constructs one. The domain type carries no invariant, so "typed" is not buying correctness here. A checked constructor that refuses an out-of-range delta would put the guarantee in the type instead of in the surrounding arithmetic.

The reason to prefer that over the magnitude reasoning: the bound being relied on ("values are supply-sized", "the sum stays below `i64::MAX`") is not only threatened by corrupt input off disk. It is equally threatened by our own code changing around it — a new call site, a wider element type, a different intermediate. The assumption is invisible at the arithmetic site and nothing rechecks it when the surrounding code moves, so the day it stops holding the result is wrong silently. Modelling non-overflow by proving magnitudes is cheap to get right once and expensive to keep right.

Encoding the invariant on the smallest context — the type — removes the dependence on all of that. A checked add and a validating constructor never yield a wrong value regardless of where the input came from or how the surrounding structures are refactored later; the worst case is a loud `None`/`Err`, never a silent wrong number. The correctness stops depending on external assumptions, which is what makes it robust.

nachog00: Separate from the overflow-handling point above: the choice of `Zatoshis` as the accumulator in `total()` mislabels the quantity, and this is distinct from what the original comment suggested. That comment pointed at `SignedZatoshis` for the *return* and a checked add; accumulating in `Zatoshis` — which caps the running sum at `MAX_ZATOSHIS` — was introduced by the fix.

`Zatoshis`' invariant is *a holdable amount*: `new`/`checked_add` reject anything past `MAX_ZATOSHIS` (21M ZEC), the maximum value in existence at an instant. But `outputs`/`spends` are cumulative movements, not a balance. The producer in `zaino-chain-store-zainodb` (`adapter/history.rs`) pushes, for every transaction touching the address across the whole queried range, each output that paid the address (`outputs`) and each input that spent one of its prior outputs (`spends`). So `sum(outputs.value)` is gross received over the range — a coin received, spent, and received again is counted each time.

Gross flow is not bounded by the supply. The store holds the whole finalised chain and the range can span all of it across many addresses, so a total past 21M ZEC is ordinary high-throughput history, not corruption. `total()` returning `None` there ("a total the supply cannot hold") rejects legitimate data.

The correct shape is to accumulate into a type whose bound is machine-representability rather than supply, and reserve the fallible/checked step for the genuinely bounded quantity — the net. `SignedZatoshis` as the return type is right; `Zatoshis` as the intermediate accumulator is the mismatch.

Note also: the `transparent` module header states *"Declared, not implemented ... no consumer is wired to it"*, but `TransparentHistoryIndex` is implemented in `zaino-chain-store-zainodb`. `net_value()` itself has no production caller yet, so the capped sum is latent rather than live — but its inputs are already produced with the flow semantics that break the cap, and the header no longer describes the code.

nachog00: Tracked as follow-up issues:

- #1504 — the overflow-handling point (https://github.com/zingolabs/zaino/pull/1483#discussion_r3918530077): silent `unwrap_or(i64::MAX)` saturation and the non-validating `SignedZatoshis` constructor; encode the invariant in the type and fail loud.
- #1505 — the accumulator semantics (https://github.com/zingolabs/zaino/pull/1483#discussion_r3918631781): `Zatoshis` accumulator caps gross flow at the money supply and rejects legitimate history; also folds in the stale module header.

Issue #1485, opened 2026-08-26
Consumers bind the single-attempt OneShot source ports and handle transience themselves


nachog00:
`zaino-source` exposes two port layers: the single-attempt `OneShot*` ports
(return `QueryError`) and the canonical resilient ports (return `SourceError`,
sole impl the sealed `ValidatorClient`, generated by `#[resilient_port]`, landed
in #1459). Consumers that want transience handled for them should bind the
canonical ports — but several still bind the `OneShot*` variants and are left to
absorb transient transport failures themselves:

- **chain-head** — `packages/zaino-chain-head/src/ports.rs:63-66` (+ impl
  `75-78`): binds `OneShotGetChainTip/GetBlock/GetBlockByHash/GetCommitmentTreeRoots`;
  runs its own exponential backoff ladder (`config.rs:38,66-67,106-138`, 500ms→30s).
- **mempool** — `packages/zaino-mempool/src/ports.rs:65-68` (+ impl `78-81`):
  binds the four OneShot mempool ports; leaks retryable errors to callers
  (`error.rs:11,26`) and relies on the poll interval.
- **chain_index** — `packages/zaino-state/src/chain_index/source.rs:136-140`
  and `source_ports.rs:29-41+` (~20 OneShot ports); runs its own retry/backoff
  (`finalised_state.rs:415-416,841-874`; sync-loop `SyncTimings`).

### Design invariant (see ADR-0014)

By design these components must assume the validator is **already booted and
responsive**, and must not handle validator readiness or retry ad hoc. Two
separate concerns are currently leaking into the consumers:

- **transient transport blips** → owned by the resilient source port
  (`ValidatorClient`), whose short retry ladder papers over a wire blip — *not*
  consumer backoff ladders;
- **validator bootup / readiness** → owned by the overarching runtime, which
  must confirm the validator is up and responsive before asking anything of these
  components (ideally via a dedicated validator stateful component that owns
  validator status).

Migrating off the `OneShot*` ports addresses the first concern; the second is the
invariant recorded in ADR-0014.

This is the concrete follow-on #1459 flagged as legacy debt, and the north star
of #982 ("validator connection retries should not be defined in consumer side").

Notes: `SendRawTransaction` stays on its `OneShot*` port by design; subscriptions
are unaffected. Pairs with purifying `zaino-source` errors to `{Domain | Unavailable}` (#1418).

PR #1461, opened 2026-08-20
docs(adr): crate-layout model + block-representation decision (proposed)

nachog00:
## What

Two **proposed** ADRs distilling a design discussion that started from the
[combined modular crate-layout proposal](https://hackmd.io/@umbrielsystems/zaino-modular-crate-layout-combined)
and the two in-flight decomposition PRs — #1418 (driving-port scaffold) and #1440
(chain-head isolation), both over #1402's modular sync engine.

- **ADR-0012 — crate boundaries follow an independent-variation criterion.**
  Names the rule ADR-0003 / 0007 / 0008 / 0010 already apply: a crate is the unit
  of *independent variation* (dependencies / implementer-multiplicity /
  outside-consumer / release-cadence), with a four-role taxonomy
  (vocabulary / driven-port / driving-port / conformer) and port well-formedness as
  a dependency-graph property. Rejects a uniform `<domain>` + `<domain>-service`
  crate pair for every subsystem as a blanket rule — decide per boundary.

- **ADR-0013 — served block representation is a choice of index set.**
  Frames the #1418-vs-#1440 recent-window fork (store a composed `CompactBlock` vs.
  a parsed `Block` + on-demand tree roots) as an *index-set configuration axis*,
  with validator passthrough as the fallback. Introduces `PreIndexCompactBlock`
  (#1402; the `nachog00/zebra` `ReadRequest::CompactBlock` prototype). This one is
  an **open decision** — it lists the questions to resolve and deliberately does
  not select.

## Why here

Moving the discussion out of HackMD into `docs/adr/` for line-level review and
tracking, next to the ADRs it generalizes.

## Notes for reviewers

- Both are **Status: proposed**; ADR-0013 is intentionally an open-decision anchor,
  not a settled choice.
- **Numbering is provisional.** `0011` is reserved for #1440's chain-head ADR (it
  currently self-numbers `0010`, which collides with the mempool ADR already on
  `dev`, so it renumbers to `0011` on rebase). Final numbers assigned at merge.
- Kept as a **draft** while the wording converges.

PR #1418, opened 2026-07-24
feat(scaffold): driving-port + runtime composition (WIP — shapes evolving fast)

nachog00:
> ⚠️ **WIP scaffold — shapes are evolving quickly.** This is a fast-moving
> exploration in mostly *trait algebra*: crate seams, ports, and composition
> shape. Names, trait boundaries, and error types are **changing between commits
> by design** — the goal right now is to *see the structure and surface
> incompatibilities*, not to settle an API. Please read it as a design sketch
> under active iteration, not a finished surface. Review at the level of
> **boundaries and direction**, not line-by-line signatures.

**Stacked on #1402** (modular sync engine). Its branch isn't on the upstream
yet, so this PR opens against `dev` and its diff therefore *includes* #1402's
commits. The scaffold delta on top of the sync engine — what's actually new here
(+~3.5k lines, all additive) — is this compare:
https://github.com/nachog00/zaino/compare/feat/zaino-primitives...feat/driving-port-scaffold
Once #1402 merges, a rebase drops those commits and this becomes just the scaffold.

## What this is

The consumer-facing (**driving**) side of the hexagon, and the runtime that
composes the sync engine's output into a served indexer. Where the sync-engine
PR builds the *driven* side (sources, indexes, persistence), this scaffolds:

- **`zaino-core`** — pure domain vocabulary (re-exports primitives + refs,
  locators, events, status, serviceability).
- **`zaino-service`** — the inner *driving* surface: per-capability read/control
  traits (`BlockRead`, `CompactBlockRead`, `AddressRead`, …) + `IndexerService`
  bundle, with a mock behind a `testing` feature.
- **`zaino-fs`** — `FinalisedState`: domain semantics over the finalised half.
  Pins a concrete index set and hides `zaino-sync`/`zaino-indexes`/persistence —
  "consumers see finalised *state*, not indices."
- **`zaino-nfs`** — `NonFinalisedState`: the reorg window (adopts hhanh00's
  `Chain`/`find_trim_index` narrowly), publishing pinned snapshots.
- **`zaino-runtime`** — the supervisor: owns the components, produces a pinned
  cross-tier `RuntimeSnapshot`, and delegates read-composition to a `resolve`
  policy module.
- **`zaino-wallet`**, **`live-tests/zallet-fit`** — outer client + a shape-only
  mirror of zallet's `Chain` trait, to pressure-test the surface against a real
  embedder.

## Runtime composition (the part worth looking at)

Reads resolve by one of three strategies, kept in one tested policy module
[...]

Issue #982, opened 2026-04-13
Validator connection retries should not be defined in consumer side


nachog00:
> I think `chaindex` is the wrong place for validator connection retries logic. It deserves to be en capsuled and managed completely by `ValidatorConnector` or whatever equivalent abstraction provides access to the validator.  

 _Originally posted by @nachog00 in [#962](https://github.com/zingolabs/zaino/pull/962/changes/9005777ab4ff80cc09440ae60d1c40f2ccd9b796#r3074678221)_

zancas:
Note, we have not specified contracts or subcontracts for (not necessarily orthogonal):

* boot
* teardown
* mempool

Note depending on development, interactions with passthrough will need to accounted for.

PR #1486, opened 2026-08-26
docs(adr): 0014 validator readiness owned by runtime, not source consumers

nachog00:
Adds ADR-0014 recording an ownership invariant surfaced while triaging the
source-port work:

> Source consumers (chain-head, mempool, chain_index) **assume the validator is
> already booted and responsive** and do not handle validator readiness or retry
> ad hoc. Transient transport blips are owned by the resilient source port
> (`ValidatorClient`, ADR-0008/#1459); validator bootup/readiness is owned by the
> overarching runtime (#1418), which confirms the validator is up before invoking
> the components — ideally via a dedicated validator stateful component that owns
> validator status.

Status: **proposed** — the decision is stated but four open questions remain
(component placement, gate shape typestate-vs-await, definition of "ready",
interaction with ADR-0011's fail-fast startup).

Number is provisional (0012/0013 are in flight via #1461); reassign at merge if
they land first.

Refs #1485 (OneShot→resilient migration, the sibling debt), #1418 (runtime /
driving-port anchor), #982, #1459, ADR-0008, ADR-0011.


zancas (approved):
Instead of blocking merge on the open questions lets merge this ADR now and allow the questions to resurface when they're adjacent to someone's active development.

elicbarbieri (changes requested):
Hmmm.  I dont like this.  This is too simple of a descision to require an ADR.  just implement it.  This is a validator readiness probe, not a core architectural descision

PR #1487, opened 2026-08-26
chain-head-service: match named domain not-found variant exhaustively

### LINKED ISSUE #1479: chain-head-service: block_at_height/hash swallow QueryError::Domain(_) with a wildcard
`block_at_height` (`service.rs:679`) and `block_at_hash` (`service.rs:691`) map
`QueryError::Domain(_) => Ok(None)`, discarding the value with a `_` and no log.

**This is not a live bug today.** Both domain enums have exactly one variant
(`GetBlockError::HeightNotFound`, `GetBlockByHashError::NotFound`), neither is
`#[non_exhaustive]`, and the zebra-rpc adapter routes parse/transport failures
through `QueryError::Fetch` (propagated correctly), not `Domain`. So the only
`Domain` value genuinely means "absent" and `Ok(None)` is correct.

What the `_` costs:
- the discard is untraced; and
- it removes the compiler's ability to catch a *future* second domain variant —
  a new rejection (`Pruned`, `AboveCeiling`, …) would be silently reclassified
  as "absent". The extension loop reads that as end-of-chain and could publish a
  short graph, with no build break at the one site that must be revisited.

Severity: latent-correctness + observability.

Surfaced as a post-merge review finding on #1440:
https://github.com/zingolabs/zaino/pull/1440#issuecomment-5346379227

nachog00:
`block_at_height` and `block_at_hash` mapped `QueryError::Domain(_) => Ok(None)` with a bare wildcard and no trace (`service.rs:679,691`).

This changes both sites to match the *named* domain variant exhaustively:
- `QueryError::Domain(GetBlockError::HeightNotFound(_))`
- `QueryError::Domain(GetBlockByHashError::NotFound(_))`

and adds a `debug!` trace of the discard.

**Behaviour is unchanged today** — both domain enums have exactly one variant, so the only `Domain` value still means "absent" and still maps to `Ok(None)`. The point is exhaustiveness + observability: a future second domain variant (e.g. `Pruned`, `AboveCeiling`) now breaks the build at the one site that must reclassify it, instead of being silently read as end-of-chain, and the discard is now traced.

## Verification
- `cargo build -p zaino-chain-head-service` — pass
- `cargo test -p zaino-chain-head-service` — pass (22 tests)
- `cargo clippy -p zaino-chain-head-service` — clean
- `cargo fmt -p zaino-chain-head-service -- --check` — clean

Closes #1479

### DIFF
--- a/packages/zaino-chain-head-service/src/service.rs
+++ b/packages/zaino-chain-head-service/src/service.rs
@@ -684,8 +684,16 @@ impl<S: ChainHeadBlockSource> ChainHeadService<S> {
         match self.source.get_block(height).await {
             Ok(block) => Ok(Some(block)),
             // Absent, not failed: the extension loop reads past the tip by
-            // design, which is how it learns where the tip is.
-            Err(zaino_source::QueryError::Domain(_)) => Ok(None),
+            // design, which is how it learns where the tip is. Matched by name
+            // rather than a wildcard so a future second domain variant breaks
+            // the build here — the one site that must reclassify it — instead
+            // of being silently read as end-of-chain.
+            Err(zaino_source::QueryError::Domain(zaino_source::GetBlockError::HeightNotFound(
+                missing,
+            ))) => {
+                debug!(height = %missing, "block_at_height: source reports no block; treating as absent");
+                Ok(None)
+            }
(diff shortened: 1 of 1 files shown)


AloeareV (approved):
Nice!

PR #1490, opened 2026-08-27
remove the docs directory..  too much AI!

zancas:


docs/use_cases.md
```diff
@@ -1,55 +0,0 @@
-# Indexer Live Service
```
idky137, line 1: We need clear instructions on how to use zaino somewhere, we should not delete thos without replacing it.

docs/updating_zebra_crates.md
```diff
@@ -1,96 +0,0 @@
-# Updating Zebra crates ~best~ possible practices.
```
idky137, line 1: This is useful, we have often got the error of testing artefacts not being updated, so removing the instructions for this process doe not seam like the right move.

docs/tutorial.md
```diff
@@ -1,24 +0,0 @@
-This is a tutorial to launch zaino, connected to a local validator.
```
idky137, line 1: We need instructions in this repo somewhere.

docs/testing.md
```diff
@@ -1,67 +0,0 @@
-# Testing
-### Dependencies
```
idky137, line 2: We need clear instructions on how to test zaino somewhere, it is already hard for people to run tests, we should not delete this without replacing it.

docs/rpc_api.md
```diff
@@ -1,29 +0,0 @@
-# Zaino RPC APIs
```
idky137, line 1: This is very useful, I use this regularly to check the zcash RPC protocol.

docs/perf.md
```diff
@@ -1,369 +0,0 @@
-# Zaino performance
```
idky137, line 1: Given this is the top request from zcg I don't think we should remove it.

docs/logging.md
```diff
@@ -1,90 +0,0 @@
-# Logging Configuration
```
idky137, line 1: This is very useful if correct.

docs/internal_spec.md
```diff
@@ -1,263 +0,0 @@
-# Zaino
-The Zaino repo consists of several crates that collectively provide an indexing service and APIs for the Zcash blockchain. The crates are modularized to separate concerns, enhance maintainability, and allow for flexible integration.
```
idky137, line 2: This is very useful to have, it just needs updating.

docs/example_configs/zainod-bench-mainnet-ephemeral.toml
```diff
@@ -1,46 +0,0 @@
-# Zaino mainnet benchmark configuration — ephemeral finalised state
```
idky137, line 1: These, and other examples configs are useful to keep, we should move them rather than deleting them (or provide a simple path to generating them, with full annotation which is currently missing).

docs/docker.md
```diff
@@ -1,142 +0,0 @@
-# Container Usage
```
idky137, line 1: Again this is useful information.

docs/adr/0001-zcashd-support-feature-gate.md
```diff
@@ -1,116 +0,0 @@
-# `zcashd_support` feature gate
```
idky137, line 1: We have put a lot of time into keeping adrs, why are we deleting them? The whole point was that we have a record of historical decisions.

idky137 (changes requested):
I'm a bit confused by this pr, it removes a lot of useful docs (that were not written by ai) and does not replace the information they hold.

Zaino is already lacking documentation, we should be adding to it rather than removing useful information, or at least replacing all this if we want to move it.

docs/logging.md
nachog00: It is correct i believe

PR #1499, opened 2026-08-29
build: complete and enforce publish metadata (clippy::cargo_common_metadata)

nachog00:
## Why

The 0.9.0 crates.io publish failed mid-run: `zaino-source-macros` was new this
cycle and shipped without a `description`, so crates.io rejected it with
`400: missing or empty metadata fields: description`, stopping the workspace
publish after three crates. Nothing caught it before release — `cargo publish
--dry-run` only *warns* on missing `description` (crates.io enforces it
server-side), and the publish-dry-run workflow is masked by version-reuse until
the RC lands.

## What

Make published-crate metadata complete, and enforce it in CI so this can't
recur.

- **`[workspace.package]`** gains `readme` / `keywords` / `categories`; each of
  the 20 publishable crates inherits them via `<field>.workspace = true`
  (`description` / `license` / `repository` were already present).
- **`makers clippy`** now denies `clippy::cargo_common_metadata` — it fails the
  build if any `publish != false` crate is missing description, license,
  repository, readme, keywords, or categories. Caught **pre-merge**, not as a
  crates.io 400 mid-publish.
- **`zaino-source-macros`** gets its `description` (the fix already on `stable`
  via #1498, backported to `dev` here).

## Verification

- `cargo metadata` confirms all 20 publishable crates resolve with the six
  required fields.
- `clippy::cargo_common_metadata` fires on a missing field and passes when
  present (verified on `zaino-status` with the CLI flag, and end-to-end in an
  isolated workspace).

## Notes

- Enabled via the clippy-task flag rather than `[workspace.lints]` on purpose:
  the declarative route needs `[lints] workspace = true` on every crate, which
  cargo forbids alongside the 15 crates' existing `[lints.rust]`. Consolidating
  those into `[workspace.lints]` is a larger, separate cleanup (tracked for a
  follow-up).
- **Unrelated, pre-existing:** `cargo clippy --all-features` does not compile on
  `dev` today — `zaino-state::…::write_core.rs` fails with `E0425` under the
  `transparent_address_history_experimental` feature. Since `makers clippy` uses
  `--all-features`, that job is red on `dev` independent of this PR. Flagging so
  a red clippy job isn't mistaken for this change; worth its own fix.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,14 @@ repository = "https://github.com/zingolabs"
 homepage = "https://www.zingolabs.org/"
 edition = "2021"
 license = "Apache-2.0"
+# Publish metadata shared by every published crate, inherited via
+# `readme.workspace = true` etc. crates.io requires `description` + `license`;
+# `readme`/`keywords`/`categories` round out what `clippy::cargo_common_metadata`
+# checks (see the `clippy` task in tools/makefiles/lints.toml). `readme` is
+# resolved from the workspace root.
+readme = "README.md"
+keywords = ["zcash", "blockchain", "indexer", "lightwallet"]
+categories = ["cryptography::cryptocurrencies"]
 
 
 [workspace.dependencies]
--- a/packages/zaino-address/Cargo.toml
+++ b/packages/zaino-address/Cargo.toml
@@ -7,6 +7,9 @@ repository.workspace = true
 homepage.workspace = true
 edition.workspace = true
 license.workspace = true
+readme.workspace = true
+keywords.workspace = true
+categories.workspace = true
 
 [dependencies]
 sapling-crypto.workspace = true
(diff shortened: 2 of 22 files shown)


zancas:
I pushed a merge of `dev` to this branch, so it can land. The merge changed
three things beyond the conflict resolution.

1. The clippy task moved. `dev` deleted `tools/makefiles/lints.toml` in
   43674d465, and the task lives in `Makefile.toml` now. The
   `-D clippy::cargo_common_metadata` flag and its comment moved with it.
2. Four crates that `dev` added after this branch forked lacked the three
   inherited fields: `zaino-encoding`, `zaino-persistence`,
   `zaino-chain-store`, and `zaino-chain-store-zainodb`. The new gate would
   have failed on them, so the merge adds the fields. All 24 publishable crates
   now resolve the six fields, which `cargo metadata` confirms.
3. The note in the description about `cargo clippy --all-features` no longer
   applies. `makers clippy` passes on the merged tree. With one `readme` field
   removed, it fails with "package `zaino-status` is missing `package.readme`
   metadata", so the gate works.

zancas (approved):

Issue #1501, opened 2026-09-01
chain-store-zainodb: propagate txn.commit() in the write_core block-removal path


nachog00:
Recognised during review of #1483 and deferred there as pre-existing, verbatim-moved code — see the thread: https://github.com/zingolabs/zaino/pull/1483#discussion_r3868749428

In the block-removal path (`packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/write_core.rs`), the write transaction is committed with `let _ = txn.commit();`. The `del` calls above correctly propagate `LmdbError`, but the commit that makes them durable is discarded — a commit failure leaves the "corrupt block removed" operation silently un-applied while the caller believes it succeeded.

The fn already returns `Result<_, StoreError>` and `LmdbError` has a `#[from]`, so `txn.commit()?` closes it. Worth a sweep for other `let _ =` / `.ok()` on write-txn commits while here — a comprehensive pass, as discussed on the PR.

Issue #1500, opened 2026-09-01
chain-store-zainodb: replace stringly StoreError::Custom, and carry StatusError's cause


nachog00:
Recognised during review of #1483 and deferred there as pre-existing, verbatim-moved code — see the thread: https://github.com/zingolabs/zaino/pull/1483#discussion_r3868749423

`StoreError` (`packages/zaino-chain-store-zainodb/src/error.rs`) has two issues on its port-facing error surface:

1. `Custom(String)` carries a `// TODO: Remove before production` marker — a stringly catch-all that should not ship on a port-surface error type. `Critical(String)` is similarly stringly.
2. `StatusError(StatusError)` uses `#[error("… {0:?}")]` with neither `#[from]` nor `#[source]`, so `Error::source()` returns `None` and only a `Debug` render survives — the causal chain is lost.

These feed `chain_store_error`, which collapses most variants into `ChainStoreError::Backend(other.to_string())`, so their structure is flattened at the boundary anyway — worth pairing with the `#[source]` work already done on `ChainStoreError` in #1483. (`LmdbError` / `SerdeJsonError` already use `#[from]` correctly.)

Deferred from #1483 because a proper fix reaches wider than that move-focused PR.

PR #1502, opened 2026-09-01
Review fixes for #1483: typed capability errors, sealed capabilities, full read-path metrics

nachog00:
My takes on the review caveats from #1483, stacked on its head (`b17bc5d7`) so they can land into that branch before it merges. One commit per caveat; each is `cargo check` / `clippy` / `fmt` clean, including the `prometheus` and `transparent_address_history_experimental` features. In each case the fix already on #1483 closed the defect but left a residual — these go to the root.

### Caveat 1 — capability vs backend-state errors · `e1c136cc`
**Applied on #1483:** the consumer matcher now borrows the router's `CapabilityRequest::name()` constant, so its uppercase names finally match.
**Problem left:** the error is still a `&'static str`, and a *second* producer (`finalised_source`) keeps its own lowercase vocabulary — two string vocabularies still coexist. And `FeatureUnavailable` conflates two different conditions: a capability refusal ("this store doesn't build index X") and a mid-migration "the v1 backend isn't available yet" state, both as free-form strings.
**My take:** drop the string. `FeatureUnavailable(CapabilityRequest)` is typed, so the consumer match is **total — no `_` arm**. That total match is the real guarantee: a catch-all `_` fails open, silently absorbing any capability added later or any producer that mismaps — exactly the shape of the original collapse-to-`Core` bug. The `_` only existed because a stringified value has to catch every possible string; once the value is a typed enum the compiler forces every future variant to be handled. The mid-migration backend-state condition also splits into its own `V1BackendUnavailable` variant, mapped to `NotReady` — a transient signal the runtime can route on.

### Caveat 2 — sealed capabilities · `dbde539a`
**Applied on #1483:** each port gained `const CAPABILITY`, and `store_capabilities` reads it off the port, so a store can't advertise a capability for a port it doesn't implement.
**Problem left:** the const is *defaulted*, so any implementor can override it with a wrong value and still compile; only a per-backend test catches it. A second backend is the whole point of the port.
**My take:** seal it. A per-port carrier trait with a single blanket impl fixes the value for every implementor — an override now fails to compile (E0119). The overridable const is removed; one source of truth.

### Caveat 3 — unemitted mempool metrics · `3a01d53c`
**Applied on #1483:** metric names were consolidated to one source (re-export) so the live and pinned strings can't drift.
**Problem left:** `MEMPOOL_TRANSACTIONS` and `MEMPOOL_TIP_CHANGES_TOTAL` are described in `zainod` but emitted by no crate — a dashboard on them shows "no data" forever, and an operator can't tell "no activity" from "never wired".
**My take:** remove both. A described metric should have an emitter; re-add them with their emit call when the mempool surface lands.

### Caveat 4 — read-path instrumentation · `64beeb68`
**Applied on #1483:** spans + latency histograms on the two chunked reads (the wallet-sync hot path), plus universal corrupt-row reporting.
**Problem left:** the latency metrics assume one serving profile. The explorer/full-RPC profile leans on the point lookups, spend reads and `txout_set`, which had no latency visibility.
**My take:** instrument the whole read surface via one labelled `DB_READ_SECONDS{op=…}` histogram + per-method spans, folding the two just-added chunk metrics into it (safe — unshipped). Matches the existing `SYNC_ERRORS_TOTAL{severity}` label pattern.

### Not here
The `StoreError::Custom` / `StatusError` cleanup and the `write_core` `txn.commit()` propagation are deferred — verbatim-moved pre-existing code, tracked in #1500 and #1501. The capability↔lifecycle-error split (caveat 1) and the per-index vs global-schema-version direction belong in a wider chain-store/runtime reconciliation ADR, not this PR.

### DIFF
--- a/packages/zaino-chain-store-zainodb/src/adapter.rs
+++ b/packages/zaino-chain-store-zainodb/src/adapter.rs
@@ -74,50 +74,45 @@ use crate::store::reader::DbReader;
 use crate::store::FinalisedState;
 use crate::types::{Height, Outpoint, TransactionHash};
 
-/// Which chunked read a duration belongs to.
-///
-/// A marker rather than the metric name itself, because `metric_names` is
-/// behind the `prometheus` feature and naming a constant from it at the call
-/// site would put a `cfg` on every read.
-#[derive(Debug, Clone, Copy)]
-enum ChunkRead {
-    Stored,
-    Compact,
-}
-
-/// How long a chunked read took.
+/// How long a read took, recorded against `DB_READ_SECONDS` under its `op`.
 ///
 /// A type rather than a bare `Instant` so the `prometheus` feature is handled
 /// once: without it this compiles to nothing and no call site needs a `cfg`.
+///
+/// The whole read surface shares one histogram, split by an `op` label naming
+/// the read. The label is the port method's own name, so a new read is
+/// instrumented by starting a timer with its name rather than by minting a
+/// metric.
+///
+/// Recorded on drop, so a read that returns early through `?` still records: a
+/// read that fails slowly is the symptom worth seeing, and dropping those
+/// samples would make a degrading store look faster as it got worse.
 struct ReadTimer {
+    #[cfg(feature = "prometheus")]
+    op: &'static str,
     #[cfg(feature = "prometheus")]
     started: std::time::Instant,
 }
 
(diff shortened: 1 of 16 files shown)


AloeareV (approved):
Nice!

PR #1508, opened 2026-09-04
Typed zatoshi arithmetic: enforce the money-range invariant in the quantity types

### LINKED ISSUE #1504: zaino-chain-store: net_value models overflow safety by magnitude, not by type, and saturates silently
## Context

`packages/zaino-chain-store/src/transparent.rs`, `StoreAddressEffects::net_value` and its helpers, introduced in #1483.

Discussion: https://github.com/zingolabs/zaino/pull/1483#discussion_r3918530077 (follows the original review at #1483 `discussion_r3868749410`).

Feature-gated behind `transparent_address_history_experimental`. `net_value()` has no production caller yet, so this is latent, not a live fault — but the code models its own correctness in a way that fails silently under change, which is worth fixing before it acquires callers.

## Problem

The method's overflow/conversion safety rests on a write-time magnitude argument ("values are supply-sized, so the result fits") rather than on invariants the types enforce, and one conversion resolves the impossible case silently.

- `signed()` (`transparent.rs:139`):
  ```rust
  i64::try_from(u64::from(amount)).unwrap_or(i64::MAX)
  ```
  `unwrap_or(i64::MAX)` is a silent saturation. If a `Zatoshis` ever holds a value outside the assumed range, this does not error — it substitutes a plausible-looking figure. That is the same failure mode the checked `sum` was added to prevent, one line down.

- `SignedZatoshis::new(amount: i64)` (`packages/zaino-primitives/src/types/zatoshis.rs:65`) validates nothing; any `i64` constructs one. The domain type carries no invariant, so "typed" is not buying correctness here.

A magnitude proof is invisible at the arithmetic site and nothing rechecks it when the surrounding code changes. It is threatened not only by corrupt input off disk but equally by our own future mis-manipulation — a new call site, a wider element type, a nested collection. When the assumption stops holding, the result is wrong silently and there is no way to find it.

## Proposed fix

- Encode the invariant in the type: checked accumulation that fails loud (`None`/`Err`) rather than saturating, and a validating `SignedZatoshis` constructor that refuses an out-of-range delta.
- Remove the silent `unwrap_or(i64::MAX)`: propagate, or `.expect(...)` naming the invariant.
- The guarantee should hold under refactor regardless of input origin; worst case a loud failure, never a silent wrong number.

Related: #1505 (accumulator-type semantics).

### LINKED ISSUE #1505: zaino-chain-store: net_value caps gross transparent flow at the money supply, rejecting legitimate history
## Context

`packages/zaino-chain-store/src/transparent.rs`, `StoreAddressEffects::net_value` / `total`, introduced in #1483.

Discussion: https://github.com/zingolabs/zaino/pull/1483#discussion_r3918631781 (follows the original review at #1483 `discussion_r3868749410`).

## Problem

`net_value` accumulates gross transparent movement into `Zatoshis`, whose invariant bounds it to `MAX_ZATOSHIS` (21M ZEC) — the maximum value in existence at an instant. But the quantity being summed is cumulative flow, not a balance, and flow is not bounded by the supply.

The producer, `TransparentHistoryIndex::address_effects` in `packages/zaino-chain-store-zainodb/src/adapter/history.rs`, pushes — for every transaction touching a queried address across `[start, end]` — each output that paid the address (`outputs`) and each input that spent one of its prior outputs (`spends`). So `sum(outputs.value)` is gross received over the range: a coin received, spent, and received again is counted each time.

The store holds the whole finalised chain and the query range can span all of it across many addresses, so a gross total past 21M ZEC is ordinary high-throughput history (an exchange hot-wallet set alone clears it), not corruption. `total()` folding through `Zatoshis::checked_add` returns `None` in that case ("a total the supply cannot hold"), rejecting legitimate data.

This is a type mismatch: a *holdable-amount* type (`Zatoshis`, bounded by supply) is used to accumulate a *flow*. The original review suggested `SignedZatoshis` for the return type and a checked add; accumulating in `Zatoshis` — and thus capping the running sum at supply — was introduced by the fix, not requested.

## Proposed fix

- Accumulate gross received/spent into a type bounded by machine-representability, not by the supply.
- Reserve the checked/fallible step for the genuinely supply-bounded quantity — the net.
- `SignedZatoshis` as the return type is correct; the intermediate `Zatoshis` accumulator is the mismatch.

## Also (minor)

The `transparent` module header (`transparent.rs:1`) states *"Declared, not implemented ... no consumer is wired to it"*, but `TransparentHistoryIndex` is implemented in `zaino-chain-store-zainodb`. The header no longer describes the code and should be updated.

Related: #1504 (overflow-handling / fail-loud).

nachog00:
## Summary

Follows the review of `StoreAddressEffects::net_value` on #1483. That method summed
transparent movements into a supply-capped `Zatoshis` accumulator and converted
through an unchecked, silently-saturating path — so it rejected legitimate
high-throughput history as if corrupt, and could report a wrong figure if an
out-of-range value ever reached it. The fix generalises into a small
quantity-arithmetic family in `zaino-primitives`, so the money-range invariant is
enforced once, in the types, rather than re-checked ad hoc across boundaries.

This branch also carries three PRs merged into it: #1509 (the address balance
flow sum and the balance fold), #1515 (coinbase by block position), and #1529
(the ADR-0013 restructure). Their contents are described below.

Closes #1504. Closes #1505. Doctrine recorded in ADR-0013.

## The quantity family

`zaino-primitives::types::zatoshis`, a module directory:

- `Zatoshis` — an amount of ZEC counted in zatoshis, `0 ..= supply`.
- `ZatoshisFlowSum` — an accumulation of movements, bounded only by machine
  representability (`u128`), deliberately not by the supply.
- `SignedZatoshis` — a signed zatoshi value, `-supply ..= supply`: a directional
  movement (a single input/output value) or the difference of two totals.

Arithmetic algebra (`arithmetic.rs`), written down as the specification a new
summation site inherits:

- `accumulate : [Zatoshis] → ZatoshisFlowSum` — the checked fold of movements.
- `sum_balances : [Zatoshis] → Option<Zatoshis>` (#1509) — the supply-capped
  checked fold of balances that coexist at one moment. It lands back in
  `Zatoshis` because its coexistence precondition keeps the result in range,
  and it refuses a total past the supply as evidence the operands overlap. The
  set of `Zatoshis` is not closed under addition; there is no fourth type.
- `net : ZatoshisFlowSum × ZatoshisFlowSum → SignedZatoshis` — the only door to
  a derived signed value.

"Difference" is an operation, not a type: a delta is a signed value obtained by
subtracting, a movement one obtained by parsing.

- `SignedZatoshis` carries its invariant: the unchecked `new(i64)` is removed;
  construction goes through the checked `try_new` (the boundary parser) or
  `ZatoshisFlowSum::net` (the derived path). No door skips the ±supply check.
- `net_value` is rewired onto the algebra: the gross sides accumulate unbounded
  and the net lands in a `SignedZatoshis`. Gross flow past the supply is a valid
  answer; only an impossible net is refused.
- Boundary parsers validate: the source sites (zebra RPC, readstate,
  convert-zebra, DB `to_domain`, wire delta types, mock source) instantiate
  through `SignedZatoshis::try_new`, so an out-of-range value off the wire or
  disk is a typed parse error at the point of entry.
- Redundant outbound checks removed: `BlockDeltas::from_domain` and the
  `getblockchaininfo` pool-balance rendering re-validated a range the domain
  types already guarantee. Both are now infallible, and the dead
  `DeltaAmountOutOfRange` / `PoolBalanceOutOfRange` paths are gone.

## AddressBalance is a flow sum (#1509)

- `AddressBalance.received` is lifetime gross receipts, a cumulative flow that
  legitimately exceeds the supply across a busy address set. It was typed
  `Zatoshis`, so both real backends turned legitimate data into a parse error.
  It is now a `ZatoshisFlowSum`.
- The backends deliver `received` already summed as a `u64`, which always fits
  the `u128` accumulator, so the new boundary door
  `ZatoshisFlowSum::from_summed(u64)` is infallible, and
  `From<ZatoshisFlowSum> for u128` reads the total back out.
- `get_taddress_balance_stream` aggregated balances with an unchecked
  `total_balance += u64`. It now folds through `sum_balances` and reports an
  impossible total as a typed error.
- The wire render converts the `u128` flow total back to the response's `u64`
  through a checked conversion with a typed error.

## Coinbase by block position (#1515)

- `Transaction.index` is dropped. It was self-synthesized from `.enumerate()`
  on the block path and a placeholder `0` on the mempool path, and two
  authorities read it. Position is now derived from block order at the one
  place that iterates the transactions.
- `conversion.rs` threads the enumerate position into `compact_transaction`;
  `is_coinbase = position == 0` drives the null-prevout synthesis, and the
  served `CompactTxData` index is that same position through a checked
  `u64::try_from` with a typed `TxPositionOverflow`.
- `Block::try_new(...) -> Result<_, BlockError>` rejects an empty transaction
  list, and `Block::coinbase()` is the named positional accessor.
- `BlockDelta.index` is left for a follow-up; `InputDelta`/`OutputDelta.index`
  are vin/vout indices, not positions, and are untouched.

## ADR-0013 restructure (#1529)

- The ADR is split into **Context** and **Doctrine**. Context names the two
  summation invariants, **The Balance Sum Maximum** and **The Movement Sum
  Maximum**, and lists the two faults of the earlier code.
- Doctrine opens with a **Terminology** subsection defining zatoshi,
  `Zatoshis`, element, balance, movement, and sum. Point 1 states closure
  correctly, and point 2 chooses the result type by the reading of the elements
  and names both landing types.
- The five places in `usage.md` and `arithmetic.rs` that called the balance sum
  a closed operation now describe it as a checked fold under a precondition.

## Test plan

- `cargo fmt --check`, `cargo check --workspace --all-targets`, and
  `cargo clippy --workspace -- -D warnings` (with and without
  `zaino-chain-store/transparent_address_history_experimental`) — clean on each
  merged PR.
- `cargo nextest run` on the production members with the transparent feature —
  passes; #1509 reports 329 tests across primitives, serve, state, and the two
  zebra sources; #1515 reports 231 across primitives, convert-zebra, zainodb,
  and chain-head-service plus 131 in zaino-state.
- Unit tests pin every algebra invariant: `accumulate`, `sum_balances`
  (within, at, and past the supply, and empty), and `net`; a flow sum exceeding
  the supply; the signed-value bounds at both doors; the flow-sum
  machine-representability guard; `net_value` end to end; an `AddressBalance`
  whose `received` exceeds the supply is representable.

## Not in scope

Follow-ups tracked separately, not addressed here:

- A below-port raw-`i64` transparent balance accumulation and a saturating
  `f64 as u64` on a JSON-RPC amount, both found during review.
- Pre-existing broken intra-doc links (surface only under
  `RUSTDOCFLAGS="-D warnings"`) in `tree_root.rs`, `block.rs`, and
  `zaino-chain-store-zainodb`.
- The remaining findings of review 5137418974 on this PR.


packages/zaino-serve/src/rpc/jsonrpc/wire/blockchain_info.rs
```diff
@@ -38,21 +34,23 @@ fn pool_balance(
...
+    // range: `chain_value` is a `Zatoshis` (unsigned, so non-negative and within
+    // the supply) rendered as an `Amount<NonNegative>`, and `value_delta` is a
+    // `SignedZatoshis` (a magnitude within the supply, sign unconstrained)
+    // rendered as an `Amount<NegativeAllowed>`. Both bounds hold at
+    // construction, so the range check on the wire's `Amount` cannot reject a
+    // well-formed value and the rendering is infallible.
+    fn amount<C: zebra_chain::amount::Constraint>(zats: i64) -> zebra_chain::amount::Amount<C> {
         zebra_chain::amount::Amount::try_from(zats)
-            .map_err(|e| BlockchainInfoWireError::PoolBalanceOutOfRange(e.to_string()))
+            .expect("domain zatoshi quantities are bounded to the money range")
```
zancas, line 46: **`?` replaced by `.expect` inside a function that still returns `Result`.** The same happens in `BlockDeltas::from_domain`. The repo CLAUDE.md allows `.expect` only when propagation is not cleaner; here `?` was available. The infallibility proof rests on `MAX_ZATOSHIS == zebra_chain::amount::MAX_MONEY`, which is asserted only in prose. `MAX_ZATOSHIS` is `pub(super)`, so the supply literal is typed in three places (amount.rs:6, chain-store/src/transparent.rs:135, block_deltas.rs:278). A zebra bump that changes `MAX_MONEY` turns a served `getblockdeltas` or `getblockchaininfo` into a panic in the jsonrpsee task. The `fn amount<C>` helper and the `i64::try_from(u64::from(..)).expect(..)` line are copy-pasted between block_deltas.rs:22-25/64 and blockchain_info.rs:44-51.

Fix: export `Zatoshis::MAX` and `SignedZatoshis::MAX`, add `const _: () = assert!(u64::from(Zatoshis::MAX) == zebra_chain::amount::MAX_MONEY as u64);` in zaino-serve, hoist one named conversion pair to wire.rs or wire/common/amount.rs, and have the tests use the exported bound.

packages/zaino-primitives/src/types/zatoshis/amount.rs
```diff
@@ -0,0 +1,83 @@
...
+            Err(ZatoshisOverflow { got: amount })
+        } else {
+            Ok(Self(amount))
+        }
+    }
+
+    /// Returns `Some(sum)` when the addition is representable (below
+    /// MAX_ZATOSHIS), or `None` on arithmetic overflow, matching Rust
+    /// primitive integer `checked_add` semantics.
+    pub fn checked_add(self, other: Self) -> Option<Self> {
```
zancas, line 38: **`Zatoshis::checked_add` survives as public API with no callers.** ADR-0013 point 1 says `Zatoshis` is not given a supply-capped addition, yet this method remains after the PR deleted its only caller (`transparent.rs::total`). `rg checked_add packages/` shows every other hit is on a machine integer, `Height`, or chain-work. A future summation site reaching for it silently reintroduces the supply cap on gross flow. Its doc also says "below MAX_ZATOSHIS" while the check is inclusive.

Fix: delete the method and its test, or add it back with the unbuilt `ZatoshisBalanceSum` when a consumer exists.

packages/zaino-primitives/src/types/zatoshis/flow_sum.rs
```diff
@@ -0,0 +1,60 @@
...
+/// An accumulation of zatoshi movements.
+///
+/// A running total of amounts that *move* — every output paying an address,
+/// every input spending its prior outputs — as distinct from a set of balances
+/// that coexist. The same coins can move through an address many times, so this
+/// total counts them each time and is **not** bounded by the money supply. It is
+/// bounded only by machine representability, which is why it is a `u128` and not
+/// a [`Zatoshis`](super::Zatoshis).
+///
+/// The only way to obtain one is
```
zancas: **`ZatoshisFlowSum` cannot be read from outside its module.** `ZERO`, `from_raw`, and `into_raw` are all `pub(super)`, and there is no `From<ZatoshisFlowSum> for u128`, `Display`, or `Default`. An external crate can accumulate and `net` but never read a gross total, which blocks the `AddressBalance.received` fix. The struct doc says "The only way to obtain one is `try_accumulate` ... never an arbitrary integer", but `from_raw` contradicts that, and the file's own test uses it to mint `u128::MAX`. The raw doors exist only because the methods live in arithmetic.rs rather than beside the private field.

Fix: add a public read-out, and either move `try_accumulate`, `checked_add`, and `net` into flow_sum.rs and delete both raw doors, or keep arithmetic.rs as the `//!` specification only.

packages/zaino-primitives/src/types/zatoshis/arithmetic.rs
```diff
@@ -0,0 +1,187 @@
...
+    ///
+    /// Contract: `self` is the received flow and `spent` the spent flow of the
+    /// *same* balance over the *same* range. Only then is the difference that
+    /// balance's change, which an aggregate balance — living in `[0, supply]` —
+    /// keeps within `[-supply, supply]`. So `None` means the two flows do not
+    /// describe a coherent balance (partial or corrupt data), not merely a large
+    /// number. A difference of unrelated flow sums is not a balance change, is
+    /// bounded only by the machine, and is deliberately not offered: there is no
+    /// generic subtraction nor `impl Sub` that would return an unbounded result.
+    pub fn net(self, spent: Self) -> Option<SignedZatoshis> {
```
zancas, line 121: **`net` hand-rolls the signed difference.** The `abs_diff` -> `i128::try_from(..).ok()?` -> sign flip -> `try_from_i128` chain drags a `pub(super) try_from_i128`, an `i128` field on the public `SignedZatoshisOverflow`, three extra tests, and an i64 -> i128 -> i64 round trip on the hot `try_new` boundary door.

Equivalent: `let magnitude = i64::try_from(received.abs_diff(spent)).ok()?; SignedZatoshis::try_new(if received >= spent { magnitude } else { -magnitude }).ok()`. Anything `i64::try_from` rejects is past the supply and `try_new` would reject it anyway, and `magnitude` is non-negative so the negation cannot hit `i64::MIN`. With it, `try_from_i128` and the flow_sum.rs:54 test disappear, `SignedZatoshisOverflow.got` becomes `i64`, and `try_new` becomes `value.unsigned_abs() <= MAX_ZATOSHIS`.

packages/zaino-primitives/src/types/zatoshis/signed.rs
```diff
@@ -0,0 +1,153 @@
+//! The signed zatoshi quantity: a movement or a difference.
+
+use core::fmt;
+
+use super::MAX_ZATOSHIS;
+
+/// A signed zatoshi value, bounded by `-supply ..= supply`.
```
zancas, line 7: **Doc-comment conventions.** Multi-sentence `///` item docs at signed.rs:7, :19, :32, :45; flow_sum.rs:3; arithmetic.rs:52, :76; chain-store/src/transparent.rs:85-112 (a three-paragraph section on `net_value`); to_domain.rs:196. New `//!` headers in arithmetic.rs, zatoshis.rs, flow_sum.rs, amount.rs, and signed.rs, two of them ending in `See ADR-0013`, plus a seven-line inline `//` block at blockchain_info.rs:37-43. The same flow-versus-balance rationale is restated in six places: the ADR, the module doc, the struct doc, the arithmetic header, the `net` doc, and usage.md. One-sentence item docs that do not cite documents, with the rationale kept in the ADR, would remove the drift risk.

zancas (commented):
Review of this PR at head 75c863fdf, most severe first. Findings on lines the diff does not touch are listed here; the rest are inline.

**packages/zaino-serve/src/rpc/jsonrpc/wire/blockchain_info.rs:67**

**`chainSupply` always renders 0 (pre-existing, confirmed).** The `""` (chainSupply) arm discards the computed `value` and `delta` and calls `GetBlockchainInfoBalance::chain_supply(Default::default())`. zebra's `chain_supply` sums the `ValueBalance` it is given, so `chainValueZat` is always 0. `parse_value_pool` (zaino-source-zebra-rpc/src/parse.rs:940-943) does deliver a real value for id `""`. No test pins a non-zero chain supply; the fixture uses `pool("", 0)`. The same arm is duplicated at zaino-state/src/chain_index/validator_source.rs:390.

Fix: build the chain-supply `ValueBalance` from `value` and add a test with a non-zero chain supply.

---

**packages/zaino-source-zebra-rpc/src/parse.rs:617**

**`AddressBalance.received` is still supply-capped (pre-existing, confirmed).** zebra-state `read/address/balance.rs:144-146` says an address can receive more than the money supply by sending to itself and saturates `received` at `u64::MAX`. Zaino parses that gross total through `Zatoshis::new` here, in zaino-source-zebra-readstate/src/adapter.rs:373, and in mockchain_source.rs:1050-1067. A high-throughput address set makes `getaddressbalance` fail with `ParseError::Amount` while zebra answered correctly. This is the defect ADR-0013 names, and it is not among the PR's listed follow-ups.

Fix: type `AddressBalance.received` (zaino-primitives/src/types/address_balance.rs:11) as `ZatoshisFlowSum` and accumulate with `try_accumulate` at the three sites. That needs a public read-out on `ZatoshisFlowSum` (see the flow_sum.rs comment).

---

**packages/zaino-state/src/chain_index/validator_source.rs:361**

**A duplicate fallible `pool_balance` was left behind (plausible).** validator_source.rs:349-355 `amount()` and the error arm at :371 perform the conversion this PR declares unreachable in zaino-serve, and this copy is production-reachable through the verbose `z_getblock` path. If the premise holds, this arm is dead code that should go or both copies should share one conversion. If it does not hold, zaino-serve should not have become infallible. The chainSupply `Default::default()` bug is duplicated here at :390.

---

**docs/adr/0009-served-json-schema-lives-in-zaino-serve.md:38**

**ADR-0009 still names the deleted `DeltaAmountOutOfRange`.** Decision 2 states that `BlockDeltas::from_domain` is the single fallible wire conversion returning `Result<_, DeltaAmountOutOfRange>`, a type this PR deletes; `rg DeltaAmountOutOfRange` now hits only that line. ADR-0013 lands with status `proposed` (0006-0012 and 0016 are `accepted`) and does not cite 0009 in its Related section.

Fix: amend 0009 or add a superseding note, cite it from 0013, and set 0013 to `accepted` on merge.

---

**packages/zaino-convert-zebra/src/lib.rs:143**

**Wrong error variant for a transparent output amount.** A transparent output value that fails `Zatoshis::new` is reported as `ConvertError::Height(..)` (pre-existing in `transparent_from_zebra`), so an operator sees `height: zatoshi amount N exceeds maximum ...` pointing at block-height extraction. This PR routes the shielded value balances three lines away through `ConvertError::Value`, which makes the mismatch visible.

Fix: use `ConvertError::Value` here, or add `#[from]` variants for `ZatoshisOverflow` and `SignedZatoshisOverflow` so `?` carries the typed error.

packages/zaino-primitives/src/types/zatoshis/arithmetic.rs
```diff
@@ -0,0 +1,172 @@
...
+
+impl Zatoshis {
+    /// Sums balances that coexist at one moment, or `None` if the total passes the supply.
+    pub fn sum_balances(mut values: impl Iterator<Item = Zatoshis>) -> Option<Zatoshis> {
+        values.try_fold(Zatoshis::ZERO, Zatoshis::checked_add)
+    }
+}
+
+impl ZatoshisFlowSum {
+    /// Sums amounts as a flow, or `None` if the total would exceed `u128::MAX`.
```
zancas: Why is this returning an Option?   Shouldn't a failure to add be an Err?

packages/zaino-serve/src/rpc/jsonrpc/wire/blockchain_info.rs
zancas: Addressed in a998e7d63. Zatoshis::MAX and SignedZatoshis::MAX are exported with const readers. The wire amount module holds one conversion pair, non_negative and negative_allowed, next to two const assertions that the bound equals zebra's MAX_MONEY. Both renders call the pair. The three test literals read the exported bound.

packages/zaino-primitives/src/types/zatoshis/amount.rs
zancas: Moot since 9f68c17ce. Zatoshis::checked_add is the step of Zatoshis::sum_balances, which now lives beside it in amount.rs after 6b50ea45c. The inclusive bound is what the balance fold needs, and sum_balances_at_the_supply_is_allowed pins it.

packages/zaino-primitives/src/types/zatoshis/flow_sum.rs
zancas: Addressed in 6b50ea45c. try_accumulate, checked_add, and net now live in flow_sum.rs and touch the private field directly. from_raw and into_raw are gone and ZERO is private. sum_balances moved to amount.rs and the arithmetic module is deleted. The public read-out is From<ZatoshisFlowSum> for u128, which landed earlier.

packages/zaino-primitives/src/types/zatoshis/signed.rs
zancas: Partly addressed. c8e956b70 removed every doc-comment from the arithmetic module and left one sentence per public function; 6b50ea45c then deleted the module. The remaining multi-sentence item docs in signed.rs, flow_sum.rs, transparent.rs, and to_domain.rs stay as they are for now. The rationale lives in ADR zaino/0013 in zingo-adrs.

zancas (approved):
nvm

PR #1510, opened 2026-09-05
zaino-primitives: promote ChainWork into a two-type work algebra

nachog00:
## Summary

Stacked on #1508 (the zatoshi quantity family). Second item of the primitives
invariance audit: cumulative proof-of-work.

The vocabulary `ChainWork` was a `[u8; 32]` with infallible byte conversions in
both directions and no ordering — it admitted the all-zero value, could not
express the comparison that is chain selection's entire point, and its real
invariants (strictly positive, fits the store's width) were re-derived by hand
at the store bridge while the RPC parser faked absence with a zero-equality
sentinel. Meanwhile the store crate carried the correct design from #1313
(`NonZeroU128`, `Ord`, checked arithmetic) as a private duplicate — and that
type's `add(&Self, &Self)` also permitted a meaningless operation, since the
sum of two *cumulative* totals is not a quantity.

This promotes the store's design into `zaino-primitives` and splits the two
quantities that were sharing one type:

- `BlockWork(NonZeroU128)` — the expected work of one block, derived from its
  difficulty. No ordering: comparing single blocks decides nothing.
- `ChainWork(NonZeroU128)` — cumulative work at a block, `Ord` because
  comparison is the load-bearing operation.
- The algebra, written as the spec beside the types:
  `genesis : BlockWork → ChainWork` (the fold's seed),
  `accumulate : ChainWork × BlockWork → ChainWork` (checked, typed overflow),
  `rollback : ChainWork × BlockWork → ChainWork` (checked, typed underflow).
  No `Add`/`Sub`, no `ChainWork × ChainWork`.
- Boundary doors: `try_from_reported([u8; 32]) → Result<Option<ChainWork>, _>`
  (all-zero means not-reported and lands as `None` — absence is `Option`,
  never a zero sentinel; a nonzero high half is refused loudly),
  `to_be_bytes()` for the wire render, `try_new(u128)` rejecting zero.

Migration: the RPC parser's zero-equality sentinel is deleted; the zainodb
duplicate is deleted (its path re-exports the primitive, so downstream
`zaino_chain_store_zainodb::types::ChainWork` still resolves);
`CompactDifficulty::to_work` returns `BlockWork`; every fold site spells
`parent.accumulate(work)` / `ChainWork::genesis(work)`, preserving
genesis-counted-once; the hand-rolled bridge checks (`stored_chainwork` /
`domain_chainwork`) collapse into the type; `PersistentChainWork` stays as the
DB mirror with zero-on-disk documented as corruption, and the on-disk format
is unchanged (legacy goldens and proptests pass).

`ChainHeadWork` in zaino-chain-head remains the anchor-relative sibling.

A deliberate seam remains: `BlockWork` is still constructed at the zebra
boundary via zebra's difficulty→work conversion. The follow-up item
(CompactDifficulty) moves that pipeline natively into the primitives crate.

## Test plan

- `cargo fmt --check`, `cargo check --workspace`,
  `cargo clippy --workspace -- -D warnings` (with and without
  `zaino-chain-store/transparent_address_history_experimental`) — clean.
- `cargo nextest run`: zaino-primitives 60/60, zaino-chain-store-zainodb
  143/143, zaino-source-zebra-rpc + zaino-serve 148/148, zaino-state 126/126.
- New primitives tests: zero rejected at every door, over-width reported bytes
  refused, all-zero reported bytes land as `None`, byte round-trip,
  accumulate/rollback overflow and underflow typed errors, ordering sanity,
  genesis seed counted once.
- The on-disk-bits → difficulty → work → accumulation pipeline test is kept,
  rewritten on the algebra; live-tests fold sites migrated.


packages/zaino-chain-store-zainodb/src/adapter/from_domain.rs
```diff
@@ -69,7 +68,10 @@ pub fn indexed_block_from_stored(block: &StoredBlock) -> Result<IndexedBlock, Ch
     let context = crate::types::BlockContext::new(
         hash,
         stored_hash(header.prev_hash),
-        stored_chainwork(block.chainwork, hash)?,
+        // The primitives type already carries the non-zero and width
```
elicbarbieri, line 71: This comment is noise

packages/zaino-chain-store-zainodb/src/adapter/to_domain.rs
```diff
@@ -172,7 +171,7 @@ pub(super) fn stored_block(block: IndexedBlock) -> Result<StoredBlock, ChainStor
             .map(stored_compact_tx)
             .collect::<Result<Vec<_>, _>>()?,
         tree_roots: tree_roots(&block.commitment_tree_data),
-        chainwork: domain_chainwork(context.chainwork()),
+        chainwork: *context.chainwork(),
```
elicbarbieri, line 174: why *context.chainwork()?  Avoid pointer deref

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/write_core.rs
```diff
@@ -151,9 +151,9 @@ async fn fill_sync_batch<S: zaino_chain_store::ChainStoreSource>(
             let block_work = parts.block_work()?;
             cursor.parent_chainwork = Some(match parent_chainwork {
                 Some(parent) => parent
-                    .add(&block_work)
+                    .accumulate(block_work)
```
elicbarbieri, line 154: Good.  the .accumulate naming is much clearer

packages/zaino-chain-store-zainodb/src/types/db/block.rs
```diff
@@ -35,30 +33,24 @@ use zaino_encoding::{
...
 #[derive(Debug)]
 pub(super) struct PersistentChainWork([u8; 32]);
 
 impl PersistentChainWork {
     pub(super) fn from_business(cw: &ChainWork) -> Self {
-        let mut buf = [0u8; 32];
-        // Big-endian: the value occupies the low-order (last) 16 bytes.
-        buf[16..].copy_from_slice(&cw.as_non_zero_u128().get().to_be_bytes());
-        Self(buf)
+        Self(cw.to_be_bytes())
```
elicbarbieri, line 48: I dont like the .to_be_bytes()  That seems messy...

packages/zaino-chain-store-zainodb/src/types/db/block.rs
```diff
@@ -273,7 +265,7 @@ mod tests {
         let bctx = BlockContext::new(
             BlockHash::from([0x11; 32]),
             BlockHash::from([0x22; 32]),
-            ChainWork::new(NonZeroU128::new(0x0123_4567u128).expect("nonzero")),
+            ChainWork::try_new(0x0123_4567).expect("nonzero"),
```
elicbarbieri, line 268: Ideally we dont have a Result<> type for the Chainwork constructor.  Just make it take a nonzero u128 type?

packages/zaino-chain-store-zainodb/src/types/primitives/compact_difficulty.rs
```diff
@@ -9,9 +9,8 @@
 //! arithmetic, but never exposes them in the public API.
 
 use std::fmt;
-use std::num::NonZeroU128;
 
-use super::ChainWork;
+use super::BlockWork;
```
elicbarbieri, line 13: I dont think there should be a BlockWork?  Zebra uses the following types:

Work(u128)  -- per-block
PartialCumulativeWork(u128) -- cumulative work for a series of blocks

Lets follow those conventions

packages/zaino-primitives/src/types/work/arithmetic.rs
```diff
@@ -0,0 +1,255 @@
+//! Arithmetic over the work quantity family.
```
elicbarbieri, line 1: This file is 200 lines too long.  Dramatically simpilfy.  Comments really are not needed here

packages/zaino-primitives/src/types/work/arithmetic.rs
```diff
@@ -0,0 +1,255 @@
...
+    pub fn genesis(work: BlockWork) -> Self {
+        Self::from_raw(work.into_raw())
+    }
+
+    /// Extend this chain's cumulative work by one block's work.
+    ///
+    /// The `accumulate` relation: `C × W → C`. A checked add: overflow is
+    /// unreachable on real chains, and refused rather than wrapped so a
+    /// corrupt input cannot masquerade as a light chain.
+    pub fn accumulate(self, work: BlockWork) -> Result<Self, WorkOverflow> {
```
elicbarbieri, line 79: The multiple self.into_raw() and the .map .ok_or is big code smell. Simplify

packages/zaino-primitives/src/types/work/relative_work.rs
```diff
@@ -0,0 +1,78 @@
+//! The relative quantity: work accumulated since an anchor.
```
elicbarbieri, line 1: Lets remove this.  We shouldnt need this

packages/zaino-primitives/src/types/work.rs
```diff
@@ -0,0 +1,38 @@
+//! Proof-of-work quantities for chain selection.
```
elicbarbieri, line 1: Ideally this is not a module.  this code is simple enough that this can be a very clear 200 line rust file

packages/zaino-primitives/usage.md
```diff
@@ -127,6 +127,40 @@ external-input validation step for a signed value, the same discipline the
 crate applies at every wire and persistence boundary, pushed down to the
 primitive.
 
+## The work quantity family
```
elicbarbieri, line 130: Update this once Work types are consolidated

packages/zaino-source-zebra-rpc/src/parse.rs
```diff
@@ -886,49 +891,43 @@ pub(crate) fn parse_blockchain_info(
...
-/// chain, so it is read as "not reported" rather than as a value a consumer
-/// could compare.
+/// The legacy full node sends a hex string. Zebra types the field as a 64-bit
+/// integer, so it arrives as a JSON number, and hardcodes it to zero because
+/// it does not store cumulative work per height. Both encodings land on the
+/// same door, [`ChainWork::try_from_reported`], which owns the reported-value
+/// semantics: all-zero reads as `None` — "not reported", never a zero a
+/// consumer could compare — and a value past the domain's 128-bit width is
+/// refused rather than truncated.
 fn parse_reported_chain_work(value: &serde_json::Value) -> Result<Option<ChainWork>, ParseError> {
```
elicbarbieri, line 904: This seems overcomplicated.  Do we need this?

packages/zaino-source-zebra-rpc/src/parse.rs
```diff
@@ -886,49 +891,43 @@ pub(crate) fn parse_blockchain_info(
...
-/// Cumulative chainwork as a hex string. Natural order, and left-padded rather
-/// than fixed width: it is a big-endian integer, so validators trim leading
-/// zeroes and an early-chain response is genuinely short rather than malformed.
-fn as_chain_work_natural(value: &serde_json::Value) -> Result<ChainWork, ParseError> {
+/// Cumulative chainwork as a hex string, decoded to the wire's 32 big-endian
+/// bytes. Natural order, and left-padded rather than fixed width: it is a
+/// big-endian integer, so validators trim leading zeroes and an early-chain
+/// response is genuinely short rather than malformed. Anything longer than 32
+/// bytes is out of range for the protocol and is rejected.
+fn chain_work_be_bytes(value: &serde_json::Value) -> Result<[u8; 32], ParseError> {
```
elicbarbieri, line 920: What is this function name?  Can this just be Work::to_bytes()

elicbarbieri (changes requested):
Lets consolidate to just 2 types ideally.  A Work and Work-Accumulator type like Zebra

packages/zaino-chain-store-zainodb/src/adapter/from_domain.rs
nachog00: Removed in `1222bf30`. It also described what the code used to do rather than what it does, which belongs in the commit message.

packages/zaino-chain-store-zainodb/src/adapter/to_domain.rs
nachog00: Fixed in `1222bf30`. `BlockContext::chainwork` returned a reference to a `Copy` type, so every caller dereferenced it. It returns the value now, and the deref is gone at all three call sites.

packages/zaino-chain-store-zainodb/src/types/db/block.rs
nachog00: Done in `dcb72ae9`: `AbsoluteChainWork::new(NonZeroU128) -> Self`, infallible.

`try_from_reported` stays fallible, because it reads a validator's 32 reported bytes and two things there can be wrong. All-zero means the validator does not track the value — zebra hardcodes the field — so it reads as `None` rather than a chain with no work. And a value wider than the recorded 128 bits is refused rather than truncated, since a truncated value would record less work than the chain has.

packages/zaino-primitives/src/types/work/relative_work.rs
nachog00: Removed in `ba97538f`, to keep this PR to one concern. A follow-up adds it back properly, with its consumer: `zaino-chain-head` models the same quantity today as its own crate-local `ChainHeadWork`, and that follow-up collapses the duplicate onto the primitive.

So what it lacked here was a caller, not a purpose. As written it could not have replaced `ChainHeadWork` yet either — that type derives `Ord` where this did not, and their anchor conventions differ by one block's work.

The zebra comparison does not hold, though. `PartialCumulativeWork` starts at zero and is measured from the finalized tip, so it *is* this quantity, not the absolute one. Detail in the `compact_difficulty.rs` thread.

Reverted rather than rebased out, so this thread stays anchored.

packages/zaino-primitives/src/types/work/arithmetic.rs
nachog00: `arithmetic.rs` is gone in `6d81a73d`. It held `impl` blocks for types defined in sibling files, so each quantity's operations sat away from the quantity. Each relation is now a method on the type it returns, and each error is defined in the module that raises it.

`work/` is 305 lines across two files including tests, against 255 in `arithmetic.rs` alone before.

The algebra block moved into the module documentation rather than being deleted — it is the specification a new fold site inherits. Per-relation docs are one line each now, naming the signature.

packages/zaino-primitives/src/types/work/arithmetic.rs
nachog00: The accessors were a symptom of the file split, so they went with it in `6d81a73d`. A tuple field is private to its defining module, so `arithmetic.rs` could not write `self.0` — `into_raw` and `from_raw` existed to reach through the module wall, not to express anything about work. With the relations on their types, the count across the module drops from four to one: `SingleBlockWork::into_raw`, which the chain fold genuinely reads.

`accumulate` is now:

```rust
self.0.checked_add(work.into_raw().get()).map(Self).ok_or(WorkOverflow)
```

The remaining `.get()` is forced: `NonZeroU128` has no `checked_add(NonZeroU128)`, so the non-zero proof round-trips through `u128` and is re-derived by the return type.

packages/zaino-chain-store-zainodb/src/types/primitives/compact_difficulty.rs
nachog00: Renamed in `e333bd32` to `SingleBlockWork` and `AbsoluteChainWork`, so scope and origin are stated rather than inferred.

On following zebra's conventions — the mapping is not the one it looks like. `PartialCumulativeWork` is initialised `Default::default()`, i.e. zero (`zebra-state/src/service/non_finalized_state/chain.rs:268`), incremented on push and decremented on pop (`:1568`, `:1766`), and compared only between non-finalized `Chain`s (`:2350`). It is work measured from the finalized tip — "Partial" is doing that work in the name.

Zebra has no type for work from genesis, because it never serves or persists that value; it only asks which branch is heavier. Zaino does both: it serves `chainwork` over RPC and writes it to LMDB.

So zebra's two types are per-block work and *relative* work. The absolute quantity is a third, which zebra has no analogue for — and `chainwork` is the protocol's own word for it, which is why `PartialCumulativeWork` would be the wrong name here.

packages/zaino-primitives/src/types/work.rs
nachog00: Keeping the module — the granularity is the point, not a cost.

Each quantity gets its own file with its own unit tests. That isolates the concern and makes a diff touching one quantity show only that quantity. In a single file, a one-line change to `SingleBlockWork` renders next to `AbsoluteChainWork`'s byte doors and its whole test module: worse signal-to-noise on review, and a shared conflict surface as each quantity grows.

It is also the shape the rest of this work takes, and `types/zatoshis/` is already laid out that way. Collapsing `work/` would make the two quantity families inconsistent with each other.

packages/zaino-primitives/usage.md
nachog00: Updated in `50b008a3` for the rename and the new constructor. The type count is still open in the `compact_difficulty.rs` thread; if that settles differently the guide follows.

packages/zaino-chain-store-zainodb/src/types/db/block.rs
nachog00: This is the v1 on-disk form: a 32-byte big-endian integer, established by the original `ChainWork([u8; 32])` through `U256::to_big_endian` and pinned by the goldens. Changing it needs a new body-format version.

`to_be_bytes` was already at this call site before the PR. What changed is that the manual splicing around it went away:

```rust
let mut buf = [0u8; 32];
buf[16..].copy_from_slice(&cw.as_non_zero_u128().get().to_be_bytes());
Self(buf)
```

The width and non-zero checks it used to do by hand now live on the primitive.

packages/zaino-source-zebra-rpc/src/parse.rs
nachog00: This function is on `dev` under the same name; the PR shrank it. The zero and width checks moved into `AbsoluteChainWork::try_from_reported`.

What is left is the part that cannot move: the two validators disagree on the encoding. The legacy full node sends `chainwork` as a hex string, and zebra sends a JSON number, because zebra types the field as a `u64` and hardcodes it to zero. Both have to reach the same door.

packages/zaino-source-zebra-rpc/src/parse.rs
nachog00: It is a rename of the pre-existing `as_chain_work_natural`; the PR changed its return type to `[u8; 32]` so the checks could move onto the primitive.

`to_bytes` would be the wrong direction — this decodes a hex string into the wire's bytes. The encoder you are describing does exist, as `AbsoluteChainWork::to_be_bytes()`. Open to a better name for the decoder.

nachog00:
The work types are now two: `SingleBlockWork` (one block) and `AbsoluteChainWork` (from genesis). `RelativeWork` is removed here and returns in #1560 with its consumer. The open point is which cumulative type to keep: Zaino serves and stores `chainwork` from genesis, so it needs the absolute one. Ready for another look.

PR #1511, opened 2026-09-05
zaino-primitives: validated CompactDifficulty with a native bits-to-work pipeline

nachog00:
## Summary

Stacked on #1510 (the work algebra). Third item of the primitives invariance
audit: compact difficulty, and with it the whole bits → target → work pipeline.

`CompactDifficulty` was a bare `pub type … = u32` — any integer "was" a
difficulty, and unvalidated nBits rode through every block shape until they
happened to reach the zainodb path, where a validated wrapper over zebra's
type (from #1313) re-checked them. That wrapper was the right invariant in the
wrong place, and it delegated the conversion arithmetic to a node
implementation. The domain now owns its arithmetic:

- `CompactDifficulty` is a validated newtype in `zaino-primitives`: private
  bits, checked doors `try_from_bits(u32)` / `try_from_be_bytes([u8; 4])`,
  per-rule typed errors (negative target, zero target, overflowing exponent),
  `as_bits()` out. The acceptance set transcribes zebra's `to_expanded`
  exactly, including check order.
- `to_work() -> Result<BlockWork, WorkOverWidth>` computes
  `floor(2^256 / (target + 1))` natively over a private in-crate u256 helper
  (shift/compare/restoring-division; no new dependency — the crate stays
  `thiserror`-only). The expanded target is deliberately not public: no
  consumer needs it. This closes the seam left by #1510, where `BlockWork`
  was still built through zebra's conversion.
- Zebra is demoted from runtime dependency to **test oracle**: a
  `difficulty_agreement` module in `zaino-convert-zebra` sweeps boundary
  exponent×mantissa combinations, 14 edge vectors, known-work literals, and a
  property test over arbitrary `u32` bits, asserting acceptance-set equality
  and work equality (including work-absence) against `zebra_chain`.
- Migration: the alias is deleted; carriers (`BlockHeader.bits`, compact
  blocks, verbose header, block deltas) hold the newtype; parsers construct
  through the doors with typed errors; wire renders use `as_bits`. The zainodb
  wrapper is deleted (its path re-exports the primitive), and
  `zaino-consensus::work_from_bits` — a second native copy of the same
  pipeline — is removed with its callers migrated, so the conversion exists
  exactly once.
- Fixes a latent panic: the old zainodb `to_work` `.expect()`ed on the
  valid-encoding-but-over-width case (e.g. tiny targets like `0x01010000`),
  which is now a typed `WorkOverWidth` error.

## Test plan

  `cargo clippy --workspace -- -D warnings` (plus the experimental-features
  variant) — clean.
- `cargo nextest run` across touched crates: 556/556; full default-members
  suite 748 passed, 3 skipped — including the whole `difficulty_agreement`
  differential suite.
- Known-work vectors pinned: `0x1f07ffff → 8192`, `0x2007ffff → 32`,
  `0x1d00ffff → 0x1_0001_0001`; the old zainodb rejection tests are ported
  into the primitives suite.


packages/zaino-chain-head-service/src/service.rs
```diff
@@ -774,12 +774,17 @@ fn chain_head_block(
     tree_roots: &TreeRoots,
     parent_work: Option<ChainHeadWork>,
 ) -> Result<ChainHeadBlock, ChainHeadAdvanceError> {
-    let block_work = zaino_consensus::work_from_bits(block.header.bits).map_err(|error| {
-        ChainHeadAdvanceError::InconsistentSource(format!(
-            "block {} has invalid difficulty: {error}",
-            block.header.hash
-        ))
-    })?;
+    let block_work = block
```
elicbarbieri, line 777: Hmm... I think a better API is:

let block_work = Work::from_header(block.header);
And then any invalid blocks/bad headers should be rejected earlier in the pipeline, so from_header() does not need to return a Result type

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
```diff
@@ -1678,7 +1680,7 @@ pub(crate) fn compact_block_from_indexed(
                 block.data.time()
             ))
         })?,
-        bits: block.data.bits().as_bits(),
+        bits: *block.data.bits(),
```
elicbarbieri, line 1681: Why are we pointer deref here?  Code smell.  look at the types

packages/zaino-primitives/src/types/compact_difficulty.rs
```diff
@@ -0,0 +1,328 @@
+//! The compact proof-of-work difficulty encoding from the block header.
```
elicbarbieri, line 1: This file is 200 lines too long

packages/zaino-serve/src/rpc/jsonrpc/wire/block_deltas.rs
```diff
@@ -234,7 +234,7 @@ mod tests {
             time: 1_700_000_000,
             median_time: 1_699_999_000,
             nonce: [0xcc; 32],
-            bits: 0x1d00_ffff,
+            bits: domain::CompactDifficulty::try_from_bits(0x1d00_ffff).expect("valid nBits"),
```
elicbarbieri, line 237: Why is there a CompactDifficulty type?  And not just a Difficulty?  Isnt difficulty just a uint?

elicbarbieri (changes requested):
Fixes to 1510 first, then I will take another look here.  Why do we have a CompactDifficulty type at all... since its just a Uint, can we move to Difficulty

packages/zaino-serve/src/rpc/jsonrpc/wire/block_deltas.rs
nachog00: `nBits` is not a plain integer. It packs a sign bit, an exponent and a mantissa, and some `u32` values are invalid. The type rejects those when it is built, so work can then be computed without an error. RPC `difficulty` is a different value (a float ratio against the pow limit). Zebra splits them the same way: `CompactDifficulty` / `ExpandedDifficulty` in `zebra-chain/src/work/difficulty.rs`.

packages/zaino-primitives/src/types/compact_difficulty.rs
zancas: doc-comments:  83
code lines: 99

packages/zaino-chain-head-service/src/service.rs
nachog00: Work is now computed once, in `CompactDifficulty::try_from_bits`. A target whose work does not fit 128 bits is rejected there, so `header.bits.to_work()` cannot fail and the `map_err` is gone. `BlockConversionError::WorkOverWidth` and the `Result` on `IndexedBlock::work` went with it.

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
nachog00: `CompactDifficulty` is `Copy`, so `BlockData::bits()` now returns it by value and both derefs are gone.

packages/zaino-primitives/src/types/compact_difficulty.rs
nachog00: This change removed the standalone `WorkOverWidth` type and the `expect` in `to_work`, which is now a getter over work stored at construction. A docs trimming pass follows to shorten the file.

PR #1512, opened 2026-09-05
zaino-primitives: CompactCiphertext enforces the 52-byte compact prefix

nachog00:
## Summary

Stacked on #1511. Item #5 of the primitives invariance audit: the compact note
ciphertext.

`EncryptedCiphertext` was documented as "the first 52 bytes of a note
ciphertext" but backed by an unbounded `Vec<u8>` with unchecked constructors.
The 52-byte invariant lived in prose and was re-derived differently at every
boundary: `zaino-convert-zebra` sliced `[..52]` and would panic on a short
source; `zaino-chain-store-zainodb` carried two min-and-zero-pad
`ciphertext_prefix` helpers that silently padded; and a one-byte value was a
legal ciphertext.

- The type is now `CompactCiphertext([u8; 52])`, renamed for what it actually
  holds: the 52-byte compact prefix a `CompactTx` serves to light clients, not
  the full 580-byte encryption output. Private inner, `Copy`.
- Doors: `try_new(&[u8])` requires exactly 52 bytes with a typed error carrying
  the offending length; `[u8; 52]` converts in and out infallibly. The
  unchecked `new(Vec<u8>)` / `From<Vec<u8>>` are gone.
- `zaino-convert-zebra` takes the prefix through one shared checked helper —
  a short ciphertext off the source is a typed `ConvertError::Ciphertext`,
  never a slice panic. Regression test pins `got: 51`.
- Both zainodb pad-or-truncate helpers are deleted; stored and rendered
  ciphertexts pass through the infallible array conversions.

## Test plan

  `cargo clippy --workspace -- -D warnings` and
  `cargo clippy --all-targets --all-features -- -D warnings` (the CI
  invocation) — clean.
- `cargo nextest run` on zaino-primitives, zaino-convert-zebra,
  zaino-chain-store-zainodb, zaino-state, zaino-source — 384 passed.
- New tests: exact-52 accepted + round-trip, 51/53 rejected with the length,
  short-source conversion is a typed error, full ciphertext yields its head.


packages/zaino-chain-store-zainodb/src/adapter/from_domain.rs
```diff
@@ -9,8 +9,8 @@
 
 use zaino_chain_store::{ChainStoreError, StoredBlock, StoredTx};
 use zaino_primitives::types::{
-    BlockHash as DomainBlockHash, BlockTxPosition, EncryptedCiphertext, Height as DomainHeight,
-    OrchardAction, Outpoint as DomainOutpoint, ScriptType, SignedZatoshis, TreeRoots,
+    BlockHash as DomainBlockHash, BlockTxPosition, Height as DomainHeight, OrchardAction,
```
elicbarbieri, line 12: Why do we have all of these BlockHash as DomainBlockHash?  Do we have conflicting types?  If so we need to consolidate

elicbarbieri (approved):
Very very good.  LGTM.

We can address the zaino-chain-store-zainodb/src/adaptor/{from,to}_domain.rs import X as Y shit later in a cleanup PR.  This is good to go

packages/zaino-chain-store-zainodb/src/adapter/from_domain.rs
nachog00: The aliases exist because zainodb has its own older `BlockHash`, `Height` and `Outpoint` in `types/db/legacy.rs`. Tracked for cleanup in #1574.

PR #1489, opened 2026-08-27
chain-head-service: carry transport cause in ChainHeadAdvanceError

nachog00:
`ChainHeadAdvanceError::SourceUnavailable` held a `String`, and every source failure was `.to_string()`'d at the boundary — flattening the structured `zaino_source::QueryError` (whose `Fetch(FetchError)` arm carries a machine-readable `FailureMode`) so `Error::source()` returned nothing.

## Change
- `SourceUnavailable(String)` → `SourceUnavailable(#[source] FetchError)`, preserving the `FailureMode`. `Error::source()` now yields the transport cause for the unavailable case.
- New single `advance_error()` classifier that every source-error boundary routes through:
  - `QueryError::Fetch(FetchError)` (transport) → `SourceUnavailable(fetch)` (typed, `#[source]`).
  - `QueryError::Domain(_)` rejections → stay message-only under `InconsistentSource`, tagged with the query that was refused (they wrap no external error).

**No behavioural change:** nothing branches on the variant, so retry / give-up is unchanged — the difference is purely observability (`source()` chain + preserved `FailureMode`).

## Per-site mapping
| site | query | before | after |
|------|-------|--------|-------|
| `chain_tip()` | `get_chain_tip` | `SourceUnavailable(to_string)` | Fetch → `SourceUnavailable(fetch)`; Domain → `InconsistentSource` |
| `block_at_height()` | `get_block` | `SourceUnavailable(to_string)` | Fetch → `SourceUnavailable(fetch)` (Domain already `Ok(None)`) |
| `block_at_hash()` | `get_block_by_hash` | `SourceUnavailable(to_string)` | Fetch → `SourceUnavailable(fetch)` (Domain already `Ok(None)`) |
| `tree_roots()` | `get_commitment_tree_roots` | `InconsistentSource(to_string)` | Fetch → `SourceUnavailable(fetch)`; Domain → `InconsistentSource` |
| `anchor()` tip | `get_chain_tip` | `SourceUnavailable(to_string)` | Fetch → `SourceUnavailable(fetch)`; Domain → `InconsistentSource` |
| `anchor()` block | `get_block` | `SourceUnavailable(to_string)` | Fetch → `SourceUnavailable(fetch)`; Domain → `InconsistentSource` |
| `anchor()` roots | `get_commitment_tree_roots` | `InconsistentSource(to_string)` | Fetch → `SourceUnavailable(fetch)`; Domain → `InconsistentSource` |

[...]

PR #1514, opened 2026-09-06
zaino-primitives: validate TransparentAddress at construction via zcash_address

nachog00:
## Summary

Stacked on #1509. Final Tier-1 item of the primitives invariance audit:
transparent address validation.

`TransparentAddress(String)` had an unchecked `new()` while its doc claimed a
validated address — so a client-supplied filter string, and literally
`TransparentAddress::new("not an address")`, entered the domain unvalidated.
It now parses at construction:

- `TransparentAddress::try_new(impl Into<String>) -> Result<Self, TransparentAddressError>`
  parses via the `zcash_address` protocol library, accepts only transparent
  kinds (P2PKH / P2SH), and rejects shielded / unified / Sprout / TEX and
  undecodable strings (Base58Check incl. checksum) with typed reasons. The
  unchecked `new` is deleted.
- Network-blind: any Zcash network's transparent prefix parses; `network()`
  (new `AddressNetwork { Mainnet | Testnet | Regtest }` primitive) and
  `script_type()` (existing `ScriptType`) are decoded from the address. The
  primitive states what the address *is*; matching it to the running chain is
  the consumer's policy.
- **Containment is a hard rule**: no `zcash_address` (or `zcash_protocol`) type
  appears in any public signature, error, or re-export of `zaino-primitives`.
  The foreign crate lives only inside `try_new`'s body and a private conversion
  carrier. (`grep -rn zcash_address packages/zaino-primitives/src | grep -v
  'use \|//'` is empty.)
- The security-relevant migration: client input on the `getaddressdeltas` path
  (`wire/address_deltas.rs::into_domain`) is now fallible and rejects an invalid
  address as invalid-params instead of admitting it. Boundary parsers
  (zebra-rpc, readstate) construct through `try_new` with typed errors.

This amends the crate's historical "dependency list is `thiserror` only" rule.
The rule's intent survives — no serde, no node implementation, no transport —
but the letter now admits protocol-*specification* libraries when their API is
fully contained. Hand-rolling Base58Check + SHA-256 would buy risk, not
ownership; the spec steward's parser is the spec artifact. `usage.md` is
rewritten to state the refined rule.

Dependency delta: `zaino-primitives` gains `zcash_address` (+ its light no_std
deps: bech32, bs58, f4jumble, zcash_encoding, zcash_protocol) and
`zcash_protocol` (for `NetworkType`, not re-exported by `zcash_address`).

## Test plan

  `cargo clippy --workspace -- -D warnings` and
  `cargo clippy --all-targets --all-features -- -D warnings` — all clean.
- `cargo nextest run`: zaino-primitives + zaino-source-zebra-rpc + zaino-serve
  213 passed; readstate 2; zaino-state address paths 17.
- Real address vectors (mainnet t1/t3, testnet tm/t2) accepted; garbage, a
  Sapling address, a unified address, a Sprout address, TEX, and a
  bad-checksum string rejected with typed errors.
- The old `"not an address"` value is no longer constructible; that test now
  asserts rejection at the boundary.

### DIFF
--- a/packages/zaino-primitives/src/types/address_network.rs
+++ b/packages/zaino-primitives/src/types/address_network.rs
@@ -1,0 +1,30 @@
+//! The Zcash network an address encodes for.
+
+/// The network an encoded Zcash address belongs to.
+///
+/// An address string commits to a network in its encoding — the same key
+/// material produces a different string on mainnet than on testnet — so the
+/// network is a fact *about the address*, read off it at parse time, not a
+/// separate configuration a consumer supplies.
+///
+/// This is Zaino's own vocabulary: the address parser inside
+/// [`TransparentAddress`](crate::types::TransparentAddress) maps the underlying
+/// protocol library's network tag onto this enum so that nothing from that
+/// library reaches Zaino's API surface.
+///
+/// Transparent testnet and regtest addresses share one base58 encoding, so a
+/// transparent address parsed from a string is only ever [`Mainnet`] or
+/// [`Testnet`]; [`Regtest`] is carried for completeness of the mapping.
+///
+/// [`Mainnet`]: AddressNetwork::Mainnet
+/// [`Testnet`]: AddressNetwork::Testnet
+/// [`Regtest`]: AddressNetwork::Regtest
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum AddressNetwork {
+    /// Zcash mainnet.
+    Mainnet,
+    /// Zcash testnet.
+    Testnet,
+    /// A private regression-testing network.
+    Regtest,
+}
--- a/packages/zaino-primitives/src/types/transparent_address.rs
+++ b/packages/zaino-primitives/src/types/transparent_address.rs
@@ -1,32 +1,277 @@
 //! Transparent Zcash address.
 
-/// A transparent Zcash address (t-addr string).
+use zcash_address::{ConversionError, TryFromAddress, ZcashAddress};
+use zcash_protocol::consensus::NetworkType;
+
+use crate::types::{AddressNetwork, ScriptType};
+
+/// A transparent Zcash address.
 ///
-/// Wraps the string representation. Validation of address format
-/// is the adapter's responsibility at construction time.
+/// Holds the canonical encoded string, and only ever a string that is a
+/// well-formed transparent (P2PKH or P2SH) address. The constructor
+/// [`try_new`](TransparentAddress::try_new) is the one place the format is
+/// checked, so every consumer inherits a validated address for free: there is
+/// no unchecked door.
+///
+/// # What validates
+///
+/// Construction parses the string with the `zcash_address` protocol library and
+/// accepts *only* the two transparent kinds. Shielded (Sapling), unified,
+/// Sprout, and TEX addresses decode successfully but are rejected as
+/// [`TransparentAddressError::NotTransparent`]; anything that does not decode as
+/// a Zcash address at all is rejected as
+/// [`TransparentAddressError::Undecodable`] with the parser's reason preserved.
+///
+/// The address library is an implementation detail: none of its types appear in
+/// this type's API. What the address *is* — its network and script form — is
+/// re-expressed in Zaino's own vocabulary ([`AddressNetwork`], [`ScriptType`]).
+///
+/// # Network-blindness
+///
+/// A transparent address is accepted regardless of which network it encodes
+/// for; the network is then reported by [`network`](TransparentAddress::network).
+/// The primitive states what the address *is*; deciding whether that network is
+/// the one a given query should act on is the consumer's policy, not the
+/// address's invariant.
+///
+/// # Stored metadata
+///
+/// The network and script form are parsed once at construction and stored
+/// alongside the string, so [`network`](TransparentAddress::network) and
+/// [`script_type`](TransparentAddress::script_type) are field reads rather than
+/// a re-parse on every call. The two extra machine words are cheaper than
+/// re-running base58 decoding at each accessor, and both are `Copy` so the
+/// struct stays cheap to clone.
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
-pub struct TransparentAddress(String);
+pub struct TransparentAddress {
+    encoded: String,
+    network: AddressNetwork,
+    script_type: ScriptType,
+}
 
 impl TransparentAddress {
-    /// Wrap a validated address string.
-    pub fn new(address: String) -> Self {
-        Self(address)
+    /// Parses and validates a transparent address string.
+    ///
+    /// # Errors
+    ///
+    /// - [`TransparentAddressError::Undecodable`] if the string is not a
+    ///   decodable Zcash address.
+    /// - [`TransparentAddressError::NotTransparent`] if it decodes as a Zcash
+    ///   address of a non-transparent kind (shielded, unified, Sprout, TEX).
+    pub fn try_new(address: impl Into<String>) -> Result<Self, TransparentAddressError> {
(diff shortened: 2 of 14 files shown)


packages/zaino-primitives/src/types/address_network.rs
```diff
@@ -0,0 +1,30 @@
+//! The Zcash network an address encodes for.
```
elicbarbieri, line 1: Remove the comments. If someone needs comments for Regtest,Mainnet,Testnet they need to get the fuck out of our codebase

packages/zaino-primitives/src/types/transparent_address.rs
```diff
@@ -1,32 +1,277 @@
 //! Transparent Zcash address.
 
-/// A transparent Zcash address (t-addr string).
+use zcash_address::{ConversionError, TryFromAddress, ZcashAddress};
+use zcash_protocol::consensus::NetworkType;
+
+use crate::types::{AddressNetwork, ScriptType};
+
+/// A transparent Zcash address.
```
elicbarbieri, line 8: Same here...  this should be 2-5 lines of comments.  simplify.  We know what transparent addresses are

packages/zaino-primitives/src/types/transparent_address.rs
```diff
@@ -1,32 +1,277 @@
...
+
+    /// The network this address encodes for.
+    pub fn network(&self) -> AddressNetwork {
+        self.network
     }
 
-    /// The address string.
+    /// The transparent script form this address pays to — [`ScriptType::P2PKH`]
+    /// or [`ScriptType::P2SH`].
+    pub fn script_type(&self) -> ScriptType {
```
elicbarbieri, line 89: Do we need a getter, or can we just pub the script_type field

packages/zaino-primitives/src/types/transparent_address.rs
```diff
@@ -1,32 +1,277 @@
...
+}
+
+/// The transparent metadata extracted while parsing an address.
+///
+/// A private carrier for the `zcash_address` conversion: implementing
+/// [`TryFromAddress`] lets `convert` hand back exactly the two facts Zaino
+/// keeps — network and script form — for the transparent kinds, while the
+/// defaulted arms reject every other kind. It never crosses this module's
+/// boundary, so the address library stays contained here.
+struct TransparentKind {
```
elicbarbieri, line 118: Hmmm.  This smells like anti-pattern.  We should probably not have TransparentAddress and TransparentKind types

packages/zaino-primitives/src/types/transparent_address.rs
```diff
@@ -1,32 +1,277 @@
...
+        })
+    }
+}
+
+/// Maps the protocol library's network tag onto Zaino's [`AddressNetwork`].
+///
+/// The match is exhaustive over `NetworkType`'s three variants, so a new kind
+/// of network in the library would surface here as a compile error rather than
+/// a silent mis-mapping.
+fn network_from(net: NetworkType) -> AddressNetwork {
```
elicbarbieri, line 154: Okay need to fix this and delete the AddressNetwork type?  Just have everything use the same unified NetworkType

packages/zaino-serve/src/rpc/jsonrpc/wire/address_queries.rs
```diff
@@ -85,7 +85,7 @@ mod tests {
 
     fn utxo() -> Utxo {
         Utxo {
-            address: TransparentAddress::new(ADDRESS.to_string()),
+            address: TransparentAddress::try_new(ADDRESS).expect("valid testnet address"),
```
elicbarbieri, line 88: If there is an expect here, do we validate addresses at a higher level instead of returning a UTXO w/ bad address?  Or can a UTXO encode a err as address?

packages/zaino-source-zebra-readstate/src/adapter.rs
```diff
@@ -454,7 +466,7 @@ impl zaino_source::OneShotGetAddressUtxos for ZebraReadStateAdapter {
             previous = *location;
 
             result.push(Utxo {
-                address: TransparentAddress::new(address.to_string()),
+                address: transparent_address(address.to_string())?,
```
elicbarbieri, line 469: Using TransparentAddress:: constructor is cleaner here.  one less helper function to maintain

packages/zaino-source-zebra-rpc/src/parse.rs
```diff
@@ -104,6 +104,14 @@ pub(crate) fn as_height(value: &serde_json::Value) -> Result<Height, ParseError>
     Height::try_from(h).map_err(|e| ParseError::Height(e.to_string()))
 }
 
+/// A transparent address. The validator's string is expected to be a valid
+/// transparent address, so a rejection here is corrupt source data.
+pub(crate) fn as_transparent_address(
```
elicbarbieri, line 109: Same as above.  Remove this helper and inline TransparentAddress::<constructor-name> to remove the helper

elicbarbieri (commented):
Good overall.  almost there IMO

packages/zaino-primitives/src/types/transparent_address.rs
nachog00: A public field could be changed after construction and no longer match the address string. Private fields keep `try_new` the only way to build one, which is what makes the type trustworthy.

packages/zaino-serve/src/rpc/jsonrpc/wire/address_queries.rs
nachog00: This is a test fixture. Validation happens in `try_new`, so a `Utxo` cannot hold an invalid address. Non-test code returns the error instead of calling `expect`.

packages/zaino-source-zebra-readstate/src/adapter.rs
nachog00: Removed the helper. Each site now calls `TransparentAddress::try_new` and maps the error with `FetchError::new(FailureMode::Parse, ..)`, like the other conversions in this adapter. `FetchError` has no slot for the underlying error; fixing that is tracked in #1575.

packages/zaino-source-zebra-rpc/src/parse.rs
nachog00: Removed the helper. `ParseError::Address` now wraps `TransparentAddressError`, so call sites use `TransparentAddress::try_new(..)?` directly.

packages/zaino-primitives/src/types/transparent_address.rs
nachog00: Removed in 77217ea9. `ScriptType` now implements `TryFromAddress`, and parsing uses `convert::<(NetworkType, ScriptType)>()`.

packages/zaino-primitives/src/types/transparent_address.rs
nachog00: Removed `AddressNetwork` in a880e523. The address now stores and returns `NetworkType`.

elicbarbieri (dismissed):
Looks ready to me 👍

zancas:
I pushed a merge of `dev` to this branch, because it conflicted again after
#1488, #1489, and #1499 landed. The only conflict was the import list in
`zaino-source-zebra-rpc/src/parse.rs`, where `dev` added the confirmations
types from #1513 beside the `TransparentAddressError` that this branch adds.
The merge keeps both.

On the merged tree, clippy and `cargo fmt --check` are clean for the five
crates that this PR touches. `cargo nextest run` over the same crates passes
416 tests. The merge dismisses the earlier approval, so the PR needs a new one.

zancas (approved):

zancas (approved):

PR #1516, opened 2026-09-06
zaino-primitives: unify tree size behind a checked TreeSize newtype (closes #549)

### LINKED ISSUE #549: Dangerous tree size downcasting from u64 to u32 with silent truncation risk
## Background

While working on commitment tree handling, we discovered a potential silent data corruption issue with tree size handling between Zebra's u64 response format and Zaino's u32 compact format.

### Current Flow
1. **Zebra returns u64**: get_commitment_tree_roots returns tree sizes as u64
2. **Zaino stores u32**: Compact format and internal storage use u32 
3. **Inline downcasting**: Multiple places do tree_size as u32 without validation

### Critical Issue: Off-by-One Maximum
- **u32::MAX** = 4,294,967,295 (2^32 - 1)
- **Sapling/Orchard trees can hold** = 4,294,967,296 notes (2^32)
- **Gap**: Exactly 1 note difference where valid tree size exceeds u32::MAX

### Silent Truncation Risk
```rust
// Current dangerous pattern throughout codebase:
let tree_size = zebra_tree_size as u32;  // Silent truncation!

// If zebra_tree_size = 2^32 (valid tree size):
// Result: tree_size = 0 (completely wrong!)

// If zebra incorrectly returns > 2^32:
// Result: Silent modular arithmetic truncation
```

### 1. Edge Case: Maximum Tree Capacity
When Sapling or Orchard trees reach exactly 2^32 notes:
- **Zebra correctly returns**: 4,294,967,296 (2^32)  
- **Zaino silently stores**: 0 (after u32 truncation)
- **Result**: Completely corrupted tree size data

### 2. Zebra Data Validation Bypass
If Zebra has a bug and returns invalid large values:
- **No validation**: Silent truncation hides the problem
- **Corrupted storage**: Wrong tree sizes stored permanently
- **Difficult debugging**: Error manifests as mysterious tree size inconsistencies

## Current Vulnerable Locations

```rust
// zaino-state/src/chain_index/non_finalised_state.rs
let commitment_tree_size = CommitmentTreeSizes::new(sapling_size as u32, orchard_size as u32);

// Multiple other locations with similar patterns
```

## Potential Direction

A dedicated TreeSize type could address this using the "parse, don't validate" philosophy - validating the range once at the boundary and then providing type-safe access. 

**Open question**: Should such a TreeSize type belong in Zaino, or would it be more appropriate in an upstream repository like zebra-chain?

## Risk Assessment

- **Severity**: High (silent data corruption)
- **Likelihood**: Low but increasing (trees grow over time)  
- **Detection**: Difficult (manifests as subtle inconsistencies)

The 2^32 edge case may seem theoretical, but it represents the designed maximum capacity of the commitment trees, making this a real architectural concern.

nachog00:
## Summary

Worklist item #7: unify tree-size representation and end the silent `u64 → u32`
truncation.

Closes #549.

Cumulative note-commitment tree size was modelled three ways — `TreeSize`
(a bare `= u64` alias) on `TreeRootInfo.size`, raw `u64` on `BlockTreeSizes`,
and raw **`u32`** on `ChainMetadata` — bridged by inline `as u32` narrowings.
A tree of exactly `2^32` notes (a valid size) truncates to `0`; a corrupt
larger value wraps silently. #549 documented the exact off-by-one.

- One `TreeSize(u64)` newtype in `zaino-primitives` replaces the alias and both
  raw widths across all three carriers (`ChainMetadata`'s fields widen from
  `u32` to the `u64`-backed newtype). Infallible `new`/`From<u64>` in, `get()`
  out, `ZERO`.
- The narrowing onto the proto/DB `u32` surface goes through a **checked door**,
  `TreeSize::try_to_u32() -> Result<u32, TreeSizeOverflow>`, replacing every
  `as u32`. A size past `u32::MAX` now fails loud instead of truncating. Two
  real boundaries route through it: the on-disk `CommitmentTreeSizes` write
  (context-rich `TreeSizeOverflow { hash, pool, size }`) and
  `compact_block_to_wire` (now fallible; six callers across two crates map the
  shared overflow to their own error idioms).
- **On-disk format unchanged**: the persisted `CommitmentTreeSizes` column
  stays `u32`; the read path widens it back to `TreeSize`. Golden/round-trip
  tests pass.

No external dependency: no upstream crate models cumulative tree size
(`incrementalmerkletree::Position` is a position, not a count), so this is a
local newtype, owned by us.

## Test plan

- `cargo fmt --all --check`; `cargo clippy --all-targets --all-features -D warnings`
  and default-features — clean; `cargo doc --no-deps --all-features` — clean.
- `cargo nextest run`: primitives + sources + chain-head-service 130 passed;
  zaino-state + zainodb 278 passed (1 pre-existing skip); zaino-source
  `--all-features` 29 passed.
- Regression test: a size of `u32::MAX + 1` (= `2^32`, the #549 off-by-one) is a
  typed `TreeSizeOverflow`, not a truncation; round-trips within range.

## Note for review

`compact_block_to_wire` returns the shared primitive `TreeSizeOverflow` rather
than a context-rich error — it's a legacy free function whose six callers span
two crates with three error idioms, so the minimal shared error lets each map
it to its own type (and the overflow is unreachable for on-disk blocks anyway).


packages/zaino-chain-store-zainodb/src/adapter/from_domain.rs
```diff
@@ -320,13 +320,13 @@ mod tests {
...
     fn a_treestate_the_store_cannot_hold_is_refused() {
-        use zaino_primitives::types::TreeRoot;
+        use zaino_primitives::types::{TreeRoot, TreeSize};
 
         let hash = BlockHash([7u8; 32]);
         let oversized = TreeRoots {
             sapling: Some(TreeRootInfo {
                 root: TreeRoot::from([0u8; 32]),
-                size: u64::from(u32::MAX) + 1,
+                size: TreeSize::new(u64::from(u32::MAX) + 1),
```
elicbarbieri, line 276: Why is oversized a u64 2^32, and then below in adaptor from/to domain there is u64::from(sizes.sapling)

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
```diff
@@ -1400,9 +1400,15 @@ fn assemble_compact_block(
         bits: header.data().bits().as_bits(),
         transactions,
         chain_metadata: zaino_primitives::types::ChainMetadata {
-            sapling_tree_size: commitment_tree_data.sizes().sapling(),
-            orchard_tree_size: commitment_tree_data.sizes().orchard(),
-            ironwood_tree_size: commitment_tree_data.sizes().ironwood(),
+            sapling_tree_size: zaino_primitives::types::TreeSize::new(u64::from(
```
elicbarbieri, line 1403: import TreeSize type to remove all this unessecary line-wrapping

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
```diff
@@ -1551,10 +1557,21 @@ fn has_pool_data(tx: &zaino_primitives::types::PreIndexCompactTx) -> bool {
 /// legacy inherent reads still return proto blocks to consumers that have not
 /// moved onto the ports; it goes with them, along with this crate's dependency
 /// on `zaino-proto`.
+///
+/// Fallible because the wire form counts tree sizes in `u32` where the domain
```
elicbarbieri, line 1561: Why are we counting these differently?  Do we actually need the additional prescision?

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
```diff
@@ -1568,11 +1585,11 @@ pub fn compact_block_to_wire(
             .map(|(index, tx)| compact_tx_to_proto(index, tx))
             .collect(),
         chain_metadata: Some(zaino_proto::proto::compact_formats::ChainMetadata {
-            sapling_commitment_tree_size: block.chain_metadata.sapling_tree_size,
-            orchard_commitment_tree_size: block.chain_metadata.orchard_tree_size,
-            ironwood_commitment_tree_size: block.chain_metadata.ironwood_tree_size,
+            sapling_commitment_tree_size: metadata.sapling_tree_size.try_to_u32()?,
```
elicbarbieri, line 1588: Same here.  ideally we avoid the try to u32.  I dont want to have to think about how this fails, what triggers it, etc... Much more secure to handle this differently

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
```diff
@@ -1681,9 +1698,9 @@ pub(crate) fn compact_block_from_indexed(
         bits: block.data.bits().as_bits(),
         transactions,
         chain_metadata: zaino_primitives::types::ChainMetadata {
-            sapling_tree_size: sizes.sapling(),
-            orchard_tree_size: sizes.orchard(),
-            ironwood_tree_size: sizes.ironwood(),
+            sapling_tree_size: zaino_primitives::types::TreeSize::new(u64::from(sizes.sapling())),
+            orchard_tree_size: zaino_primitives::types::TreeSize::new(u64::from(sizes.orchard())),
```
elicbarbieri, line 1702: Same as above.  do top of file import to make this look better

packages/zaino-chain-store-zainodb/src/store/reader.rs
```diff
@@ -604,7 +604,8 @@ impl<T: ChainStoreSource> DbReader<T> {
                 ),
             )
             .await?;
-        Ok(crate::store::finalised_source::v1::compact_block::compact_block_to_wire(&block))
+        crate::store::finalised_source::v1::compact_block::compact_block_to_wire(&block)
```
elicbarbieri, line 607: Why can we have errors when converting in-memory type to wire encoding.  We should enforce type accuracy at a higher level and not have to reason about what a serialization to wire error means about the data we are holding

packages/zaino-chain-store-zainodb/src/tests/finalised_state/ports.rs
```diff
@@ -147,7 +147,8 @@ async fn compact_blocks_match_the_inherent_read() {
                     .await
                     .expect("the store holds every vector height");
             assert_eq!(chunk.len(), 1);
-            let through_ports = compact_block_to_wire(&chunk.pop().expect("checked above"));
+            let through_ports = compact_block_to_wire(&chunk.pop().expect("checked above"))
```
elicbarbieri, line 150: Why does &chunk.pop() return a Result type?  Isnt this the internal plumbing that should always suceed?

packages/zaino-primitives/src/types/block.rs
```diff
@@ -111,11 +111,11 @@ impl Block {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ChainMetadata {
     /// Cumulative Sapling note commitment tree size after this block.
-    pub sapling_tree_size: u32,
+    pub sapling_tree_size: TreeSize,
     /// Cumulative Orchard note commitment tree size after this block.
-    pub orchard_tree_size: u32,
+    pub orchard_tree_size: TreeSize,
     /// Cumulative Ironwood note commitment tree size after this block (NU6.3).
```
elicbarbieri, line 117: Remove these comments

packages/zaino-primitives/src/types/block_verbose.rs
```diff
@@ -63,9 +63,9 @@ pub struct BlockVerbose {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
 pub struct BlockTreeSizes {
     /// Sapling notes committed as of this block.
```
elicbarbieri, line 65: Remove these comments

packages/zaino-primitives/src/types/tree_size.rs
```diff
@@ -0,0 +1,141 @@
+//! Cumulative note-commitment tree size for a shielded pool.
+
+use core::fmt;
+
+/// Cumulative count of note commitments in a pool's commitment tree, as of a
+/// given block.
```
elicbarbieri, line 6: No comments needed.  This is as simple of rust code as it gets.  Leave the line 1 comment.  thats all that is needed.  Ideally also move this down to u32 unless there is a good reason for the extra prescision

packages/zaino-state/src/chain_index/chain_store/reading.rs
```diff
@@ -315,6 +299,28 @@ const DESCENDING_CHUNK: u32 = 1024;
 pub(crate) type WireCompactBlocks =
     std::pin::Pin<Box<dyn Stream<Item = Result<CompactBlock, tonic::Status>> + Send>>;
 
+/// Flatten one chunk from the compact block port into the wire results the
+/// stream yields.
+///
+/// A read failure becomes a single status; a chunk of blocks becomes one wire
+/// result per block. Shared by the ascending and descending streams, which
+/// flatten their chunks the same way and differ only in how the chunks arrive.
+fn wire_chunk(
```
elicbarbieri, line 308: Nice dry.  making the compact_block_to_wire return bytes instead of Result<bytes,_> will be even better

elicbarbieri (changes requested):
Good to create dedicated type here.  Need to re-evaluate/justify the u32/u64 stuff.  I cant really see why you would ever need u64 levels of precision here

packages/zaino-chain-store-zainodb/src/adapter/from_domain.rs
nachog00: `commitment_tree_data` cannot fail anymore, so this test is gone. The same edge is tested where the u64 arrives: `u32::MAX` accepted and 2^32 rejected, in `TryFrom<u64>`, the RPC parser, and the readstate adapter.

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
nachog00: Done: `TreeSize` is imported at the top of the file and this site is `TreeSize::from(x)`.

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
nachog00: Done: same import, `TreeSize::from(x)` here too.

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
nachog00: Sapling, Orchard and Ironwood trees are depth 32, so a size ranges over 0..=2^32, but every format we write it to (proto `ChainMetadata`, v1 DB) is u32. `TreeSize` is now u32 with one `TryFrom<u64>` where validator data arrives: a full tree is rejected there instead of being written as 0 (#549), and everything downstream is infallible.

packages/zaino-chain-store-zainodb/src/store/finalised_source/v1/compact_block.rs
nachog00: Gone: with `TreeSize(u32)` the range check happens at ingest, so the wire conversion is infallible.

packages/zaino-chain-store-zainodb/src/store/reader.rs
nachog00: Agreed. The range check moved to ingest, so `compact_block_to_wire` is infallible again and this call site is back to its base form.

packages/zaino-chain-store-zainodb/src/tests/finalised_state/ports.rs
nachog00: Fixed: the conversion is infallible again, so the test is back to its base form.

packages/zaino-state/src/chain_index/chain_store/reading.rs
nachog00: Done: `compact_block_to_wire` is infallible, and `wire_chunk` stays as the shared helper.

PR #1520, opened 2026-09-07
zaino-primitives: TreeSize From<u32> + ChainMetadata::ZERO/new ergonomics

nachog00:
## Summary

Stacked on #1516. A pure-ergonomics DRY follow-up to the `TreeSize` newtype —
no behaviour change, no CI-required fix (the `check-code-duplication` guard
does not flag these; this is readability).

- `impl From<u32> for TreeSize` — a `u32 → u64` tree-size widen is lossless and
  always valid, so it gets a direct infallible `From` (the inverse of the
  existing checked `try_to_u32`). The ~8 `TreeSize::new(u64::from(x))` call
  sites collapse to `x.into()` / `TreeSize::from(x)`.
- `ChainMetadata::ZERO` — names the empty placeholder the indexer fills in
  later. The 9 all-zero `{ sapling_tree_size: TreeSize::ZERO, orchard…, ironwood… }`
  triples across the source adapters, mocks, and tests become
  `ChainMetadata::ZERO`.
- `ChainMetadata::new(sapling, orchard, ironwood)` taking `impl Into<TreeSize>`
  per field — the 4 from-sizes construction sites read as one call instead of
  a three-line struct literal.

Net: the per-pool widening/placeholder boilerplate the `TreeSize` retype spread
across the tree collapses to constructors, and the widen goes through one named
path. 5 now-unused `TreeSize` imports removed.

## Test plan

- `check-code-duplication`: exit 0 (the new `ChainMetadata::new` call sites do
  not fingerprint as a twin group).
- `cargo fmt --check`: clean.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: clean
  on all touched crates. (Unrelated: 3 pre-existing `clone_on_copy` errors in
  `live-tests/clientless/tests/chain_cache.rs`, a baseline `ShieldedPool: Copy`
  issue not in this diff.)
- `cargo nextest run` on the 7 touched crates: 408 passed, 1 skipped.

### DIFF
--- a/packages/zaino-primitives/src/types/block.rs
+++ b/packages/zaino-primitives/src/types/block.rs
@@ -118,6 +118,38 @@ pub struct ChainMetadata {
     pub ironwood_tree_size: TreeSize,
 }
 
+impl ChainMetadata {
+    /// Empty placeholder: every pool size zero.
+    ///
+    /// A source adapter that does not yet know the commitment tree sizes emits
+    /// this; the indexer fills the real sizes in later. All-zero is not a
+    /// "missing" sentinel — a genuinely empty tree is also all-zero — but at the
+    /// adapter layer it is the documented stand-in for "not computed here yet".
+    pub const ZERO: Self = Self {
+        sapling_tree_size: TreeSize::ZERO,
+        orchard_tree_size: TreeSize::ZERO,
+        ironwood_tree_size: TreeSize::ZERO,
+    };
+
+    /// Build from the three cumulative pool sizes, in Sapling / Orchard /
+    /// Ironwood order.
+    ///
+    /// Each argument accepts anything that widens into a [`TreeSize`] (a `u32`
+    /// or `u64` count, or a `TreeSize` itself), so a caller holding raw counts
+    /// need not wrap them first.
+    pub fn new(
+        sapling_tree_size: impl Into<TreeSize>,
+        orchard_tree_size: impl Into<TreeSize>,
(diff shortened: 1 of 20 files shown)


elicbarbieri (approved):
Looks good.  Not sure what/where this is being merged, but the diff in the PR is good.  LGTM

packages/zaino-primitives/src/types/block.rs
```diff
@@ -118,6 +118,38 @@ pub struct ChainMetadata {
...
+    /// Ironwood order.
+    ///
+    /// Each argument accepts anything that converts infallibly into a
+    /// [`TreeSize`] (a `u32` count, or a `TreeSize` itself), so a caller holding
+    /// raw counts need not wrap them first.
+    pub fn new(
+        sapling_tree_size: impl Into<TreeSize>,
+        orchard_tree_size: impl Into<TreeSize>,
+        ironwood_tree_size: impl Into<TreeSize>,
+    ) -> Self {
```
elicbarbieri, line 144: Docs are very terse.  I would remove almost all.  This is very readable

elicbarbieri (approved):
Lets get this in

packages/zaino-primitives/src/types/block.rs
zancas: We can circle back to clean up doc-comments later.

zancas (approved):

Issue #1477, opened 2026-08-26
zaino-status: no definition of its subject, and StatusType fuses lifecycle/health/load into one linear enum


nachog00:
### The subject is undefined — a status *of what*?

Now a standalone crate, `zaino-status` provides a status *vocabulary* but never
defines the entity that vocabulary is about. The module doc calls it "the
operational state of a component" (`status.rs:31`), yet "component" is nowhere
defined, and `Status` (`status.rs:146`) is `fn status(&self) -> StatusType` —
implementable by any type, including plain values with no lifecycle.

The variants presuppose a particular kind of subject: a long-lived, supervised
entity that is spawned (`Spawning`), runs, can be healthy or degraded
(`RecoverableError`/`CriticalError`), can be under load (`Busy`), and is torn
down (`Closing`). But that entity — and its characteristics and invariants — is
never stated. Without it there is no basis for deciding which states are legal,
which transitions are valid, or even whether the axes below are the right ones.
The problems that follow are partly downstream of this missing definition.

### StatusType fuses three orthogonal concerns into one linear enum

`StatusType` (`status.rs:33`) carries three orthogonal concerns in one value:

- **lifecycle phase**: `Spawning`, `Syncing`, `Ready`, `Closing`
- **health**: `Offline`, `RecoverableError`, `CriticalError`
- **load**: `Busy`

One cell per component holds one value, so a write on one concern overwrites the
others: a component cannot be `Ready` *and* `Busy`, or `Syncing` *and*
`RecoverableError`, at once — a health blip erases the phase.

### Observable problems (current code)

- **`combine` linearizes across the concerns** (`status.rs:109`): one
  subsystem's health blip masks another's readiness — the mechanism behind #649.
- **`Offline` is silently promoted to `CriticalError`** (`status.rs:114-117`):
  "not connected" and "corrupted/crashed" become indistinguishable once combined.
- **`Busy` is dropped**: `combine(Busy, Ready)` falls through to `_ => Ready`
  (`status.rs:127`), so load never survives aggregation — though `is_ready()`
  treats `Busy` as ready (`status.rs:138`).
- **Liveness and readiness can't move independently**: `is_live` (`status.rs:132`)
  and `is_ready` (`status.rs:137`) both derive from the one enum.
- **`store()` is an unguarded setter** (`status.rs:199`) with no transition
  semantics, so the shutdown race (#498) and the `Closing`-overwrite race (#1098)
[...]

PR #1527, opened 2026-09-07
feat: add zaino-component (component status/lifecycle abstraction)

nachog00:
## What

A new leaf crate, `zaino-component`, defining what a supervised component *is* — the vocabulary and ports a runtime uses to observe and control its long-lived subsystems (validator, finalised state, non-finalised state, mempool).

It replaces the fused status model #1477 describes: `zaino-status` crams lifecycle phase, health, and load into one linear enum, so a health blip overwrites the phase. `zaino-component` splits status into two independent axes.

## Why a new crate, not an edit to `zaino-status`

`zaino-component` **supersedes** `zaino-status` rather than patching it. Editing `zaino-status` in place would change `StatusType` / `Status` / `NamedAtomicStatus` under all its current consumers at once. This lands **additively** — legacy `zaino-status` is untouched — and consumers migrate to `zaino-component` **one at a time** (chain-head first). `zaino-status` is deprecated and retired as that proceeds; #1477 closes when the last consumer is off the fused `StatusType`.

## The model

- **`Lifecycle`** — a phase (`Offline → Spawning → Syncing → Ready → Closing → Offline`), modelled as a **state machine**: `Lifecycle::try_advance_to` rejects illegal transitions (e.g. `Closing → Ready`) with a typed `IllegalTransition`, and a restart must pass through `Offline`. The rule is owned by the type; no consumer re-checks it.
- **`Health`** — a condition (`Healthy / Recoverable / Critical / Offline`), **independent** of lifecycle: a health change never overwrites the phase.
- **`ComponentStatus`** — the two axes plus a `ComponentName`, so a transition is logged with a clean identifier.
- **`StatusSource`** / **`StatusWatch`** — the observe side: read the current status, or subscribe to changes (watch-backed, so an observer reacts rather than polls).
- **`Managed`** — the control side (`spawn`/`restart`/`stop`, typed error) for components the runtime *owns*; an observed component (e.g. the external validator) reports status without implementing it.
- **`Task`** — a supervised async task (named, cooperative cancel, panic capture) — what a component runs its work on.

## Scope

Status is one facet of a component, not the whole; the **load** axis is deferred until something needs it. The crate is a **self-contained leaf** (`thiserror`, `tokio`, `strum`) with no dependency on the finalised or non-finalised state, so it lands on its own. Consumers adopt it **incrementally** — chain-head first — which advances #1477 one component at a time. The runtime supervision that consumes these ports (a supervisor + an orchestra) follows in a separate change.

## Testing

In-crate unit tests, green under `cargo test -p zaino-component`; clippy clean; fmt clean:

- the lifecycle state machine rejects illegal transitions and forces a restart through `Offline`;
- health and lifecycle move as independent axes;
- a task failure flips health to `Critical` and a restart heals it.

Refs #1477.

### DIFF
--- a/packages/zaino-component/src/lifecycle.rs
+++ b/packages/zaino-component/src/lifecycle.rs
@@ -1,0 +1,65 @@
+//! A component's lifecycle and the transitions it permits.
+
+/// Where a component is in its managed progression — a *phase*.
+///
+/// Moved only by management ([`Managed`](crate::Managed)): spawn advances it,
+/// stop retires it. Distinct from [`Health`](crate::Health), which is a
+/// condition the component reaches on its own. A component being `Ready` says
+/// nothing about whether it is `Healthy`, and vice versa.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, strum::Display)]
+pub enum Lifecycle {
+    /// Not running — the rest state a component starts in and returns to after
+    /// it has finished closing. The only phase a fresh spawn may begin from.
+    Offline,
+    /// Starting up; not yet doing its work.
+    Spawning,
+    /// Running, but still catching up to where it can serve.
+    Syncing,
+    /// Running and able to serve.
+    Ready,
+    /// Shutting down; on its way to [`Offline`](Self::Offline).
+    Closing,
+}
+
+impl Lifecycle {
+    /// Advance to `next` if the state machine allows it, returning the new phase;
+    /// otherwise return the rejected [`IllegalTransition`].
+    ///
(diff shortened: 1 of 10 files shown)


packages/zaino-component/src/lifecycle.rs
```diff
@@ -0,0 +1,70 @@
+//! A component's lifecycle and the transitions it permits.
+//!
+//! ```text
```
elicbarbieri, line 3: Fuck I love this.  perfect example of documentation

elicbarbieri (approved):
Looks so good.  LGTM!

PR #1528, opened 2026-09-07
ci(release): publish multi-arch (amd64 + arm64) images

### LINKED ISSUE #1164: Add multi-arch Docker images, particularly ARM64 alongside the default AMD64.

elicbarbieri:
Fixes #1164. Supersedes #1165.

Publishes `linux/amd64` + `linux/arm64` manifests for both image variants

Credit to @nmarley for #1165 and the arm64 verification.

@nachog00 lets get this in next release

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,13 @@
 # The pipeline's `cycle-<N>` and other `<crate>-X.Y.Z` tags carry no zainod
 # image and are ignored here. Blessing owns the GitHub Release for pipeline
 # tags, so the `create-release` job below runs only for the legacy tags.
+#
+# Images are multi-arch (linux/amd64 + linux/arm64):
+#
+# - per-platform legs fan out to native runners (arm64 under QEMU = ~5-10x on a Rust build)
+# - manifest assembly / push-by-digest via docker/github-builder (no per-arch tags in registry)
+# - cache-scope gets the platform appended upstream, so one scope per variant is enough
+#
 # TODO(cutover): gh-pages doc publish.
 name: Zaino Release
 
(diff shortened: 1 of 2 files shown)


zancas:
This PR is the correct one to land for #1164, and #1435 closes in its favour.
Three items need work first. The first item stops the workflow.

1. The `prepare` job runs
   `cargo run --manifest-path tools/workbench/Cargo.toml --bin get-rust-version`.
   That binary is not on `dev` now. Commit 90835d045 removed `RUST_VERSION`
   from the Dockerfile, and `release.yaml` on `dev` does not pass it. After a
   merge of `dev`, the step "Determine Rust version" fails, and `images` needs
   `prepare`. Remove the step, the `rust-version` output, and the
   `RUST_VERSION` build argument. The check of `DOCKERHUB_USERNAME` can stay in
   `prepare`.
2. `docker/github-builder/.github/workflows/build.yml@v1` is a tag that can
   move. The job that calls it has `id-token: write` and the DockerHub
   credentials. `dev` pins 13 actions by commit and 29 by tag, so a tag is not
   against a rule here. A commit pin is still the better choice for a job with
   these permissions.
3. #1435 proved a dry-run that needs no DockerHub secrets. A fork can then
   verify a change to this workflow. With `push: false`, does the reusable
   workflow skip the registry login? If it does not, the `registry-auths`
   secret makes a dry-run from a fork fail.

The PR also conflicts with `dev` in `release.yaml`. The zaino remote rejects
force-pushes, so merge `dev` into the branch.

zancas:
@elicbarbieri I'm not sure if we're still intending to support this release flow?  If so:

A second pass over this PR, after the three items in my earlier comment. The
first three items below stop the workflow. I verified them against the
repository. The rest are design points that I did not verify by a run.

## Blockers

1. The `prepare` job fails on every run. It requires the repository *variable*
   `DOCKERHUB_USERNAME`, and the repository has only the *secret* of that name.
   The variables are `RELEASE_APP_ID`, `TS_OIDC_AUDIENCE`, `TS_OIDC_CLIENT_ID`,
   and `ZTEST_CLUSTER_CONFIG_URL`. `images` needs `prepare`, so the next
   `zainod-X.Y.Z` tag builds no image. Either create the variable before the
   merge, or write the namespace as a literal in `meta-images`, which also
   removes the guard step.
2. `dev` deleted the `create-release` job, the legacy `X.Y.Z` and `X.Y.Z-rc.N`
   tag triggers, and the `type=semver` tag rule in 02356f3e8. This branch still
   carries all of them. When you merge `dev`, take `dev`'s side of those hunks.
   Otherwise this workflow writes GitHub Releases beside the relman blessing
   flow.
3. `prepare` runs the workbench binary `get-rust-version`, which is not on
   `dev`, as noted before. The step, the `rust-version` output, and the
   `RUST_VERSION` build argument can all go.

## Design points

4. The image namespace comes from `vars.DOCKERHUB_USERNAME`, and the login
   uses `secrets.DOCKERHUB_USERNAME`. One identity then lives in two places
   that change independently. A rotated token with a different account pushes
   the digests into one namespace while the manifest step tags another.
5. The reusable `build.yml` exposes only the GitHub Actions cache. `dev` reads
   and writes the registry cache `zainod:buildcache` and
   `zainod:buildcache-no-tls`. Releases are ten to fourteen days apart, and
   the Actions cache evicts entries unused for seven days, so each leg starts
   cold. Four `mode=max` scopes also compete with the CI `rust-cache` entries
   for the ten gigabyte repository limit.
6. `build.yml` logs in to Docker Hub only when it pushes. A `push: false`
   dry-run pulls the base images and `moby/buildkit` without credentials, so it
   can hit the anonymous rate limit. This answers my earlier question: a fork
   dry-run needs no secrets, and it also gets no authenticated pulls.
7. `docker/github-builder/.github/workflows/build.yml@v1` is still a moving
   tag, as noted before. `v1` resolves to
   `58cb9f5b71b1836d6f690c1e95effdeb9b98cb8a` (v1.17.0) today.
8. The header comment asserts "arm64 under QEMU = ~5-10x on a Rust build"
   with no measurement behind it. State the figure with its source, or drop
   it.

The two red CI jobs on this PR are a `rustfmt` diff in
`live-tests/e2e/tests/wallet_to_validator.rs` and the relman duplicate-groups
check. Both exist on `dev` and this PR does not touch them.

PR #1462, opened 2026-08-20
Automated release machinery: relman CLI + gated pipeline

nachog00:
## Automated release machinery: `relman` CLI + gated pipeline

**Status: DRAFT.** Every release-advancing workflow in this PR is inert until the repo variable `RELMAN_PIPELINE_ACTIVE` is set at a deliberate cutover. Two changes do take effect at merge; they are listed under **What changes at merge**.

### The problem

Releases today are manual. Versions are hand-bumped across the workspace. Changelogs are hand-written. Tags are hand-cut. The 17 published crates go to crates.io by hand, in dependency order. Each step can drift from the others, and nothing detects the drift.

### The model, in one paragraph

A contributor lands a PR together with a **changeset**: a small TOML file in `.changesets/` naming each crate the PR changes and the semver kind of the change (breaking / feature / fix / internal). CI derives everything else from the accumulated changesets: per-crate version bumps (with the pre-1.0 relaxation), changelog entries, tags, and the publish plan. No human writes a version number. Code advances through four fixed, version-agnostic branches — `dev → rc → release-ready → stable` — and a named quality gate controls each admission. The one human release action is the **blessing**: the merge to `stable`. CI then bumps, tags, publishes, and cuts the GitHub Release.

Full design: `docs/release/` — `pipeline.md` (the policy), `changeset-format.md` (the changeset file format), `implementation.md` (architecture and trust model). The decision itself is recorded as `docs/adr/0016` (supersedes `0015`, the periodic flow).

### Mechanics

- **Governance is explicit.** The root `relman.toml` lists the 17 governed crates. Changeset validation, version derivation, and the publish plan cover exactly this set. No `cargo metadata` heuristic decides membership.
- **`relman`** is a Rust CLI at `tools/relman/` (own isolated workspace, outside the production dependency graph). It makes every deterministic decision and performs every working-tree edit. It never touches the network or git refs. The workflow YAML is a thin shell around it: push, tag, `gh`, publish. Surface: `changeset {new,check,rename,consume,clear}`, `versions`, `bump`, `changelog`, `tags`, `pr-body` (with a `--status` cycle dashboard), `publish-plan`. ~193 unit tests.
- **Tag identity has three layers**: `cycle-<N>` marks a release period, `cycle-<N>-rc.<M>` marks a prerelease, `<crate>-<X.Y.Z>` marks the provenance of one crate version. Derived versions live only in the standing release PR until a real release.
[...]

PR #1542, opened 2026-09-14
ci: fix the three checks every pull request inherits red from dev

zancas:
Every pull request against dev has inherited three red checks for a week. This pull request fixes all three on dev itself. It changes no served behaviour.

## Compile check

The live-tests workspace failed `cargo fmt --check` because one test file landed unformatted with the ztest migration (#1265). The first commit is the rustfmt output for that file.

## Repo guards

`check-code-duplication` reported two identical groups in relman, introduced with the release pipeline (#1462). The second commit collapses each group onto one plain function.

- `ChangeKind::parse` and `DeploymentStatus::parse` now look the wire name up through each enum's own `as_str` over an `ALL` table.
- The two `entries_by_crate` methods become one generic function in a private `entries` module. It returns its own `EntriesError`, which each service error adopts through `From`.

The gate now reports no duplicate groups across 388 files. Relman's 220 tests and clippy pass.

## build-and-push

The test-environment image pins `curl=8.14.1-2+deb13u4`. Debian trixie has replaced it with `8.14.1-2+deb13u5`, so `apt-get install` fails. The third commit bumps the pin. The other five pins still resolve.


elicbarbieri (approved):
LGTM

PR #1547, opened 2026-09-14
Extract relman into zingolabs/release_manager

zancas:
## Summary

relman now lives in [zingolabs/release_manager](https://github.com/zingolabs/release_manager) with its history intact. This repository keeps only its release data and the workflows that call the tool. Tracks release_manager issue #1, phases 1 and 5.

## What changes

- `tools/relman` and `.github/actions/setup-relman` are deleted.
- The four workflows that run relman call `zingolabs/release_manager/.github/actions/setup-relman`, pinned to one commit for both the action and the installed version.
- `ci.yml` no longer formats, lints, or tests relman.
- The local `act` harness under `tools/scripts/it` is deleted. It built relman from the deleted tree and shadowed the deleted action. A replacement belongs with the reusable workflows in phase 4.
- Three release documents name the tool's new home. An empty changeset covers the change.

## What stays

`relman.toml`, `.changesets/`, `.release/`, the changelogs, and the sandbox scripts.

## Follow-ups

- The adapters test that checked every path-and-version dependency of a governed target is itself governed asserted a zaino invariant from inside relman. It is gone with the tool. Phase 3 of issue #1 makes it a relman command.
- The pin moves to a `vX.Y.Z` tag once release_manager cuts one; the action then downloads the static binary instead of building.

https://claude.ai/code/session_01WFUbc3Ce4ZAUPsTio4pmLu


elicbarbieri (approved):
I support getting this out of zaino now.  First priority is to land nacho sync and cleanup.  We can worry about release next week.  Changesets stay in zaino, which should be enough to release manually without too much complexity next week

PR #1550, opened 2026-09-14
Draft: zaino-runtime — inner driving surface + component supervision

nachog00:
## What this is

The Zaino runtime re-founded on `dev`'s ports and the new component design. It
introduces the inner architecture in three layers:

- **`zaino-core`** — pure domain vocabulary for the inner driving surface. Types
  only, no async, no runtime. Re-exports the chain vocabulary from
  `zaino-primitives` and adds the driving-surface primitives that don't carry:
  refs (`BlockId`/`BlockRef`/`HeightRange`/`Outpoint`), a `Locator`/`ForkPoint`,
  tip/mempool `events`, `serviceability` (`Capability`/`ServiceabilityManifest`/
  `ServiceableRange`), tx/spend `status`, and network `upgrades`.
- **`zaino-service`** — the inner driving surface: one trait per
  capability-cohesion unit (`BlockRead`, `CompactBlockRead`, `TransactionRead`,
  `AddressRead`, `SpendRead`, `TreestateRead`, `ForkReconcile`, plus the
  `controls` — `Broadcast`, `MempoolSubscribe`, `TipSubscribe`, `TakeSnapshot`,
  `ReportedUpgrades`, `Serviceable`). Mirrors the per-method segregation of the
  driven ports (`zaino-source`) but for the opposite reason: a single implementor
  provides all of them, and the split lets each outer consumer depend only on the
  subset it needs. Serviceability is a runtime property — a read returns
  `NotServiceable` until its backing index is built. RPITIT + `BoxStream`, no
  `async-trait`.
- **`zaino-runtime`** — component supervision over the `zaino-component` ports.
  `Orchestra`/`OrchestraBuilder` boots the long-lived components in a dictated
  order (each `Ready` before the next), gives each a babysitter task, and funnels
  escalations up one channel; `supervise`/`supervise_step` watch one component
  and act per a `RecoveryPolicy`. The recovery mechanism (restart via `Managed`)
  is wired, but the current policy escalates rather than restarts — mechanism
  present, policy conservative.

## Relationship to other work

- **Stacked on #1527** (`feat/zaino-component-crate`): this PR's base is that
  branch, because `zaino-runtime` consumes `zaino-component` (the de-fused
  `Lifecycle`/`Health`/`ComponentStatus` + `Managed`/`Task`) which #1527
  introduces. Review #1527 first; retarget this to `dev` once it lands.
- Re-founds the runtime work (Refs #1418) onto the current port structure
  (`zaino-chain-store` / `zaino-chain-head` / `zaino-mempool`) rather than the
  pre-split monolith.

## Not done yet (this is a draft)

- **Read-serving composition** over the finalised / non-finalised / mempool
  components (route/merge/serviceability) is not re-seamed onto `dev`'s reader
  ports here. That is the primitives-heavy half and lands after the primitives
  PRs settle; it is deliberately held back to avoid rework against churning
  types.
- The branch trails `dev` and has not been aggiornated to the latest; the diff
  is meaningful only against its stacked base (#1527), not against `dev`.
- `zaino-runtime/tests/chain_head_status_reconciliation.rs` is an exploratory,
  throwaway test: it validates the map from `dev`'s legacy fused chain-head
  status onto a de-fused `ComponentStatus` and documents the frictions of
  adapting a fused status. It will be dropped once the native components exist;
  it is not a landed adapter.

Refs #1418


packages/zaino-core/src/events.rs
```diff
@@ -0,0 +1,18 @@
+//! Stream payloads for the tip and mempool subscriptions.
+
+use zaino_primitives::types::TransactionId;
+
+use crate::refs::BlockId;
+
+/// The best chain moved; carries the new tip.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TipEvent {
```
elicbarbieri, line 9: Rename the type to BestChainMoved or something more descriptive

packages/zaino-core/src/refs.rs
```diff
@@ -0,0 +1,31 @@
...
+    pub height: Height,
+    pub hash: BlockHash,
+}
+
+/// How a caller names a block to read.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub enum BlockRef {
+    Height(Height),
+    Hash(BlockHash),
+}
```
elicbarbieri, line 17: Dupe?

packages/zaino-core/src/serviceability.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/// For each capability, the height it is answerable up to (`None` = not yet).
+/// Derived from per-index sync progress (derivation lands here later).
+#[derive(Clone, Debug, Default)]
+pub struct ServiceabilityManifest {
+    pub answerable: Vec<(Capability, Option<Height>)>,
+}
+
+/// The heights a snapshot can answer, and the FS/NFS boundary within them.
```
elicbarbieri, line 30: Its probably good to formalize/document this more, ie:

ServiceableRange.finalized_tip will be equal to ServiceableRange.tip during sync, and then ServiceableRange.tip will diverge once caught up.  What about those small number of blocks that Where finalized_tip != tip && tip != chain_tip?  That might be a good time to actually not respond to queries since its only going to be a second or two, and removes a bunch of edge cases and complexity?  

This can also be deferred to later

packages/zaino-core/src/status.rs
```diff
@@ -0,0 +1,23 @@
...
+use zaino_primitives::types::{Height, TransactionId};
+
+/// Where a snapshot places a transaction. Chain state only — mempool presence
+/// is observed through its own stream, not reported here.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum TxStatus {
+    Mined(Height),
+    Orphaned,
+    Unknown,
+}
```
elicbarbieri, line 12: Why is mempool status not reported here?  This seems the logical place to report it.  One core enum for txs instead of 2 places to maintain

packages/zaino-runtime/src/supervisor.rs
```diff
@@ -0,0 +1,241 @@
...
+pub enum SupervisionOutcome {
+    /// Health was acceptable; nothing was done.
+    Observed,
+    /// The component was restarted in place.
+    Restarted,
+    /// Health went `Critical` and the policy escalated it to the runtime.
+    Escalated,
+}
+
+/// One supervision step: observe the component's health and act per `policy`.
```
elicbarbieri, line 60: Dupe.  consolidate into one supervise() w/ optional param if needed

packages/zaino-service/src/bundle.rs
```diff
@@ -0,0 +1,46 @@
...
+
+use crate::controls::{
+    Broadcast, MempoolSubscribe, ReportedUpgrades, Serviceable, TakeSnapshot, TipSubscribe,
+};
+use crate::reads::{
+    AddressRead, BlockRead, CompactBlockRead, ForkReconcile, SpendRead, TransactionRead,
+    TreestateRead,
+};
+
+/// A pinned, reorg-coherent view. Every read observes the chain as of the tip
```
elicbarbieri: Good shit nacho.  this is so clean.  Such an improvement in code quality w/ these traits

packages/zaino-service/src/error.rs
```diff
@@ -0,0 +1,53 @@
+//! Per-capability error types.
+//!
+//! Distinct types (not one god-enum) so each outer client maps precisely, and
+//! so capabilities can diverge later — the CLAUDE.md "per-conversion error
+//! granularity" rule, applied to the driving surface. In this scaffold they
+//! share a shape; a macro keeps them DRY (a `fn` cannot define types).
+
+use zaino_core::Capability;
+
+/// Every read-boundary failure separates a *not-yet-serviceable* answer and a
```
elicbarbieri, line 10: I would remove the macro and just hard-code or enum the errors. this is really hard to read

elicbarbieri (commented):
So good... yeah this looks totally ready.  utAck, but once you are happy nacho, I will approve

nachog00:
## Session update — serving layer and component supervision

This session builds the layers above the inner ports: consumer profiles, serving
adapters over real transports, and a runtime that boots and supervises them. It
is structure and lifecycle; the data-conversion layer is deliberately deferred
(see "What is missing").

### What was added

- **`zaino-service`** — split the single fat `Snapshot` into a thin coherence
  marker (`pinned_tip` + `serviceable_range`) plus **named read-sets**:
  `WalletReadCore`, with `FullWalletReads` and `LightWalletReads` as siblings
  over it (neither inherits the other), and a distinct `NodeRpcReads`. Three
  service **profiles** (`WalletLibService`, `LightServeService`, `NodeRpcService`)
  compose a read-set with the controls a use case needs. Inner error types now
  implement `std::error::Error`.
- **`zaino-wallet`** — the full-wallet library adapter (client type `Indexer`,
  the wallet's own view), domain → stable DTO.
- **`zaino-lightserve` / `zaino-noderpc`** — the two serving adapters, each bound
  to one profile, converting domain <-> wire **in the adapter** (no domain crate
  depends on a wire schema).
- **`zaino-component`** — the `Serve` seam (a supervised server), alongside
  `Managed` / `Task` / status.
- **`zaino-runtime`** — `ServeComponent` (a server as a supervised component),
  `ValidatorComponent` (the observed root), the `observe` / `boot_observed` path
  for components the runtime does not own, and the Orchestra composition.
- **`zaino-store-service`** — the concrete backend, first slice: composes the
  real `ChainStore` and `ChainHead` handles into one engine.

### What the tests demonstrate (all green)

- **Profiles.** One concrete engine satisfies all three profiles with zero
  profile-specific code — a compile-time assertion, on both the mock and the
  real backend (`mock_satisfies_all_profiles`; the backend's profile assertion).
- **Adapters.** The wallet maps domain to a stable DTO; the serving adapters map
  domain to wire and validate wire input back to domain (`tip_maps_to_dto`,
  `latest_block_maps_domain_to_wire`, `get_tx_out_validates_the_txid_then_reads`,
  `send_raw_transaction_decodes_and_relays`).
- **Real transports.** A real tonic server and a real jsonrpsee server each bind
  a socket and shut down cleanly when cancelled (`binds_and_shuts_down_on_cancel`
  in each transport).
- **Servers as components.** Each real server boots to `Ready` and stops to
  `Offline` under the runtime (`a_real_grpc_server_boots_and_stops_as_a_component`,
  `a_real_jsonrpc_server_boots_and_stops_as_a_component`).
- **Supervision.** A server whose serve loop fails goes `Critical`, escalates,
  and is fatal, naming it (`a_failing_server_escalates_and_is_fatal`).
- **Observed validator.** An unreachable validator fails to connect; a validator
  drop escalates and is fatal; the validator boots before the servers
  (`an_unreachable_validator_fails_to_connect`,
  `a_validator_drop_escalates_and_is_fatal`, `the_validator_boots_before_the_servers`).
- **Full boot.** The whole runtime boots `validator -> light-serve -> node-rpc`
  in dependency order over one engine, all `Ready`; and a validator drop brings
  the whole app down naming the root (`the_whole_runtime_boots_in_dependency_order`,
  `a_validator_drop_brings_the_whole_app_down`).
- **Backend seam.** One engine drives both ports over a shared handle
  (`one_engine_drives_both_ports`).

### What is missing for this to be usable

The composition and lifecycle are complete; what remains is the data-conversion
layer, held behind the primitives PRs still landing.

- **The engine behind the servers is the mock** (or the slice-1 backend, whose
  reads return `NotServiceable`). Real reads need the conversion layer.
- **Backend slices 2-4**: route-by-watermark across the finalised/non-finalised
  seam; `StoredBlock` / `ChainHeadBlock` -> domain conversions per read; and
  wiring the mempool and validator handles for the controls (broadcast, tip and
  mempool subscribe, reported-upgrades, passthrough).
- **`fs` / `nfs` / `mempool` are not yet supervised components** — they depend on
  those conversions.
- **Serving surfaces are partial.** Light-serve wires 2 of the ~20
  `CompactTxStreamer` RPCs (the rest return `Status::unimplemented`, because the
  gRPC contract is generated and must be implemented in full); node-rpc wires 6
  JSON-RPC methods (its trait is ours, so it grows method by method). Each
  further method needs a handler method plus a domain -> wire conversion.
- **No config yet.** Which ports an instance serves — and therefore which
  indexes it builds — is not wired.
- **The real Zebra source is not connected**; the validator boots against a stub
  reachability probe.
- **No health / readiness endpoints yet.** The runtime already holds every
  component's lifecycle and health, but that is not projected to standard
  liveness / readiness endpoints on the servers. These should be edge-only
  projections of the Orchestra's component statuses (liveness = process up;
  readiness = required components `Ready` and serviceable), so a deployment can
  probe the servers and have the answer reflect the orchestration.
- **TLS and synchronous bind (#1081)** on the tonic transport are not done.

packages/zaino-component/src/sync.rs
```diff
@@ -0,0 +1,29 @@
...
+///
+/// `run` starts syncing, fires `caught_up` the first time it reaches the tip,
+/// and follows the chain until `cancel`. `Ok(())` is a clean stop; `Err` is a
+/// source/write failure — which a supervisor turns into a `Critical` component.
+///
+/// This is a *driven* port: the concrete sync engine (e.g. ChainView's sync)
+/// implements it, and the runtime's indexer component drives it. It lives here,
+/// with the component vocabulary, so a domain crate can implement it depending
+/// only on `zaino-component`, never on the runtime.
+pub trait SyncDriver: Send + Sync + 'static {
```
elicbarbieri, line 19: This would potentially be a good place to locate progress reporting in the future.

In future I want /readyz to report the build state of each index.  Would make sense to have a stream of SyncProgress structs or something.  Ignore for now. we will do in future

packages/zaino-core/src/chain_info.rs
```diff
@@ -0,0 +1,13 @@
...
+
+/// A domain-typed summary of the chain's position, distinct from the wire
+/// `getblockchaininfo` response an adapter renders from it.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ChainInfo {
+    /// The current best tip, or `None` before any block is indexed.
+    pub tip: Option<BlockId>,
+    /// The validator's best-known height, which may lead the indexed `tip`.
+    pub estimated_height: Height,
+}
```
elicbarbieri, line 13: This is not super clear.  does each domain report this?  So there will be a ChainInfo for the CompactBlockIndex, the GetTxOutsetInfo index, etc... ?

Or am I missing something.  If so, this should be make clearer, ie ChainInfo.best_tip, and a ChainInfo.reference_block/reference_timestamp?

packages/zaino-core/src/passthrough.rs
```diff
@@ -0,0 +1,23 @@
...
+/// A node-operator query Zaino relays to the validator rather than answering
+/// from an index.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PassthroughQuery {
+    /// `getmininginfo`.
+    MiningInfo,
+    /// `getpeerinfo`.
+    PeerInfo,
+    /// `gettxoutsetinfo`.
+    TxOutSetInfo,
```
elicbarbieri, line 17: Doesnt Zebra not implement this?  @idky137 is the expert here

packages/zaino-core/src/serviceability.rs
elicbarbieri: Also maybe rename to `finalized_block` and `current_tip` for clarity

packages/zaino-lightserve/src/error.rs
elicbarbieri, line 1: Why is this called zaino-lightserve?

packages/zaino-lightserve/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+/// A light-serve handler failure.
+#[derive(Debug, thiserror::Error)]
+pub enum ServeError {
+    /// The chain has no tip yet — not built to any height.
+    #[error("no blocks available yet")]
+    NoBlocks,
+    /// Could not acquire a coherent snapshot; likely resolves on retry.
+    #[error(transparent)]
+    Unavailable(#[from] Transient),
+}
```
elicbarbieri, line 21: Should this be a global sort of serve error?  Should this also include the config-type things like "zaino is not configured to build & serve this index" and "zaino index is currently under construction and has not reached this block-num"

packages/zaino-lightserve/src/grpc.rs
```diff
@@ -0,0 +1,196 @@
...
+    match err {
+        ServeError::NoBlocks => Status::unavailable("no blocks available yet"),
+        ServeError::Unavailable(t) => Status::unavailable(t.to_string()),
+    }
+}
+
+/// The single message for a method whose handler is not built yet.
+fn unimplemented(method: &str) -> Status {
+    Status::unimplemented(format!("{method} not served yet"))
+}
```
elicbarbieri, line 54: Remove this helper and just inline

packages/zaino-noderpc/src/error.rs
elicbarbieri, line 1: What is the distinction between zaino-noderpc and zaino-lightserve?

packages/zaino-noderpc/src/rpc.rs
```diff
@@ -0,0 +1,76 @@
...
+        RpcError::NoBlocks => (
+            ErrorCode::InternalError,
+            "no blocks available yet".to_string(),
+        ),
+        RpcError::Unavailable(t) => (ErrorCode::InternalError, t.to_string()),
+        RpcError::SpendRead(e) => (ErrorCode::InternalError, e.to_string()),
+        RpcError::Read(e) => (ErrorCode::InternalError, e.to_string()),
+    };
+    ErrorObjectOwned::owned(code.code(), message, None::<()>)
+}
```
elicbarbieri, line 76: Not a fan.  Rather messy

packages/zaino-noderpc/src/wire.rs
```diff
@@ -0,0 +1,56 @@
...
+pub(crate) fn bytes_from_hex(s: &str) -> Result<Vec<u8>, RpcError> {
+    let bytes = s.as_bytes();
+    if !bytes.len().is_multiple_of(2) {
+        return Err(RpcError::InvalidParams("odd-length hex".into()));
+    }
+    bytes
+        .chunks_exact(2)
+        .map(|pair| Ok((hex_val(pair[0])? << 4) | hex_val(pair[1])?))
+        .collect()
+}
```
elicbarbieri, line 32: This looks like a bug waiting to happen.  There has to be something in rust-std that does this.  or if not, improt from zcash-primitives or something.  I dont want to be responsible for this code

packages/zaino-noderpc/src/wire.rs
```diff
@@ -0,0 +1,56 @@
...
+/// Render a spend status as a wire string (domain -> wire). Exhaustive by
+/// design — a new `SpendStatus` variant should force a decision here.
+pub(crate) fn spend_status_to_wire(status: SpendStatus) -> String {
+    match status {
+        SpendStatus::Unspent => "unspent".to_string(),
+        SpendStatus::Spent { by } => format!("spent:{}", to_hex(by.into())),
+        SpendStatus::SpentSpenderUnknown => "spent".to_string(),
+        SpendStatus::NoSuchOutput => "none".to_string(),
+    }
+}
```
elicbarbieri, line 56: Can we implement this on the SpendStatus struct instead?

packages/zaino-runtime/src/health.rs
elicbarbieri, line 1: I had taken a look at some of this in #1554 

For now lets land this, and I will rebase my metrics PR

packages/zaino-runtime/tests/chain_head_status_reconciliation.rs
```diff
@@ -0,0 +1,77 @@
...
+        // Load axis deferred: Busy reads as a healthy, running component.
+        StatusType::Busy => (Lifecycle::Ready, Health::Healthy),
+        StatusType::Closing => (Lifecycle::Closing, Health::Healthy),
+        StatusType::Offline => (Lifecycle::Offline, Health::Offline),
+        // Lossy: the phase was overwritten; guess Ready.
+        StatusType::RecoverableError => (Lifecycle::Ready, Health::Recoverable),
+        StatusType::CriticalError => (Lifecycle::Ready, Health::Critical),
+    };
+    ComponentStatus::new(CHAIN_HEAD, lifecycle, health)
+}
```
elicbarbieri, line 42: This PR is already so big.  just also remove the StatusType and fully migrate to ComponentStatus

packages/zaino-lightserve/src/error.rs
nachog00: sure, some of these errors might find a better home in a shared enum for all serves.

Regarding config... i haven't added it yet to keep the initial iteration simpler, but im taking it into account. should pop up soon on this branch.

packages/zaino-lightserve/src/grpc.rs
nachog00: these crates are throwaway POC adapters of the serve ports. don't stress about it's current shape yet.

packages/zaino-noderpc/src/error.rs
nachog00: one is a an rpc server, the other one is grpc. they also serve different apis.

packages/zaino-noderpc/src/wire.rs
nachog00: agreed, same as above, not meant as a prod crate just yet. barely a functional stub for POC.

packages/zaino-noderpc/src/wire.rs
nachog00: yes, i would prefer that as well. in fact, a coherent `IntoWire` trait for all this types of conversions would be good.

packages/zaino-runtime/tests/chain_head_status_reconciliation.rs
nachog00: pr is realistically not meant to land as is. I'm just building in public. we'll take a subset of this to actually land it. having this conversion is just to explore coherence. see the mods description up top. this is never to land in dev.

packages/zaino-core/src/events.rs
nachog00: I like that

zancas:
#1239 yields to the `HealthServer` here. Three items from it are not in this
PR yet.

1. Nothing constructs `HealthServer`. `zainod` has no `[health]` section.
   #1239 used `health.listen_addr`, disabled by default, as Zebra does.
2. #1239 made `/readyz` tolerate a lag of two blocks, as Zebra does. Draft PR
   #1598, on top of this branch, adds that tolerance to `ReadinessCriteria`.
   It has no effect until a component reports the validator tip as its target.
   `zaino-chain-head-service` now reports its tip as both position and target.
3. #1239 also made `/readyz` fail when the tip is older than five minutes.
   That rule detects a stalled source. `ComponentStatus` has no timestamp, so
   the rule needs a design decision here.

`cargo clippy` with `--deny warnings` fails on this branch at
`packages/zaino-component/src/status.rs:53` (`manual_is_multiple_of`).

zancas:
#1378 is closed. Two of its findings apply to this design.

1. `docs/block-store-rationale.md` in #1378 (commit 14401f07b, line 138) lists
   nine `ChainIndex` methods that take `&Snapshot`. It argues that one method,
   `get_tx_out_set_info`, has a race without it. The read traits here agree
   with that result: `BlockRead` takes a `BlockRef`, and `TakeSnapshot` is a
   separate control. The table is a useful check on each consumer of
   `TakeSnapshot`. A client that makes two calls gets two views, with or
   without a snapshot. That argument from #1378 has no answer yet.
2. The same document describes a fallback that reads the validator by height.
   After a reorg, that read can return a block from a different chain at the
   correct height. `dev` still has it in the compact block stream. See
   `compact_block_from_source` at `packages/zaino-state/src/chain_index.rs:1194`
   and its callers in the non-finalised segment at lines 2313 and 2375.
   `get_block_range` in the same file states the correct rule. The read-serving
   composition here must not inherit the fallback. A lookup by `BlockRef` must
   resolve to a hash before any passthrough read above the finalised height.

zancas:
#1429 is closed. One of its proposals applies to the start-up sequence here.

On `dev`, `anchor_with_retry` in `zaino-chain-head-service` uses
`max_consecutive_failures` for start-up and for the steady state. The default
is ten failures, which is about two minutes with the present delays. Those two
cases need different limits. A validator that was in operation and then stops
must escalate quickly. A validator that has not yet committed its genesis block
must get more time. #1429 proposed a limit of about ten minutes for start-up
only.

`Orchestra` starts each component in order and waits for `Ready`, so it is the
correct owner of this policy. The present policy escalates and does not
restart. A long start-up of the validator must not escalate as a failure of a
component.

packages/zainod/src/indexer.rs
```diff
@@ -1,347 +1,326 @@
...
+                cookie_path.as_deref(),
+                user.as_deref(),
+                password.as_deref(),
+            )?);
+            let validator = Arc::new(ZebraValidator::with_read_state(rpc, readstate));
+            boot(validator, config).await
+        }
+        // The Rpc selector is preserved in config, but only Direct/ReadState
+        // sourcing is wired so far. Fail loud and typed rather than panic.
+        SourceMode::Rpc { .. } => Err(IndexerError::RpcSourceUnsupported),
```
idky137: How much work is required to get the RPC backend working, it's important we can also support this configuration for users that cannot run zebra on the same machine.

packages/zaino-indexes/src/indexes/headers.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// Previous block hash.
+    pub prev_hash: BlockHash,
+    /// Timestamp.
+    pub time: BlockTime,
+    /// Compact difficulty.
+    pub bits: CompactDifficulty,
+}
+
+/// Headers index definition.
+pub struct HeadersIndex;
```
idky137, line 46: Could we update name  to `HeadersIndexV1`?

packages/zaino-indexes/src/indexes/headers.rs
```diff
@@ -0,0 +1,181 @@
...
+
+/// Index identity.
+pub const ID: IndexId = IndexId::new("headers");
```
idky137, line 49: Similarly, could we update name to "headers_v1"

packages/zaino-indexes/src/indexes/chain_metadata.rs
```diff
@@ -0,0 +1,199 @@
...
+#[derive(Debug)]
+pub struct ChainMetadataEntry {
+    /// Block height (key).
+    pub height: BlockHeight,
+    /// Cumulative tree sizes after this block (value, and the carry to the next).
+    pub value: ChainMetadata,
+}
+
+/// ChainMetadata index definition.
+pub struct ChainMetadataIndex;
```
idky137, line 47: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/address_history.rs
```diff
@@ -0,0 +1,482 @@
...
+    /// Block height.
+    pub height: BlockHeight,
+    /// Transaction id.
+    pub txid: TransactionId,
+    /// Output index.
+    pub output_index: OutputIndex,
+}
+
+/// Index definition.
+pub struct AddressHistoryIndex;
```
idky137, line 87: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/hash_to_height.rs
```diff
@@ -0,0 +1,77 @@
...
+/// Delta.
+pub struct HashToHeightEntry {
+    /// Block hash (key).
+    pub hash: BlockHash,
+    /// Block height (value).
+    pub height: BlockHeight,
+}
+
+/// Index definition.
+pub struct HashToHeightIndex;
```
idky137, line 27: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/orchard.rs
```diff
@@ -0,0 +1,275 @@
...
+    /// Value.
+    pub value: OrchardBlockValue,
+}
+
+/// Persisted value.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OrchardBlockValue(pub Vec<OrchardTxCompact>);
+
+/// Index definition.
+pub struct OrchardIndex;
```
idky137, line 39: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/sapling.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// Value.
+    pub value: SaplingBlockValue,
+}
+
+/// Persisted value: all sapling data for the block.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SaplingBlockValue(pub Vec<SaplingTxCompact>);
+
+/// Index definition.
+pub struct SaplingIndex;
```
idky137, line 41: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/transparent_data.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Value.
+    pub value: TransparentBlockValue,
+}
+
+/// Persisted value.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TransparentBlockValue(pub Vec<TransparentTxCompact>);
+
+/// Index definition.
+pub struct TransparentDataIndex;
```
idky137, line 41: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/transparent_spends.rs
```diff
@@ -0,0 +1,141 @@
...
+/// Persisted key: prev_txid(32) + prev_index(4) = 36 bytes.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct OutpointKey {
+    /// Previous transaction hash.
+    pub prev_txid: TransactionId,
+    /// Output index.
+    pub prev_index: OutputIndex,
+}
+
+/// Index definition.
```
idky137, line 38: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/txid_location.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxLocation {
+    /// Block height.
+    pub height: BlockHeight,
+    /// Transaction index within the block.
+    pub tx_index: u32,
+}
+
+/// Index definition.
+pub struct TxidLocationIndex;
```
idky137, line 34: Similar to headers comment for both struct and name string.

packages/zaino-indexes/src/indexes/txids.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Txids (value).
+    pub txids: Vec<TransactionId>,
+}
+
+/// Persisted value: concatenated 32-byte txids.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TxidsValue(pub Vec<TransactionId>);
+
+/// Index definition.
+pub struct TxidsIndex;
```
idky137, line 31: Similar to headers comment for both struct and name string.

packages/zaino-store/src/lib.rs
```diff
@@ -0,0 +1,608 @@
+//! # zaino-store — EXPLORATORY STUB, NOT FOR PRODUCTION
```
idky137, line 1: This seams to run alongside zaino-chain-store, which was the port architecture we agreed on and merged into dev, would it be possible to PR and proposed changes into that crate so they are reviewable?

packages/zaino-store-service/src/lib.rs
```diff
@@ -0,0 +1,362 @@
+//! `zaino-store-service` — the concrete inner engine.
```
idky137, line 1: This seams to be doing what the ChainView in zaino-chain-head already provides?

packages/zaino-backend-lmdb/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+                // (the watermark ensures clean resume).
+                EnvironmentFlags::NO_TLS
+                    | EnvironmentFlags::NO_READAHEAD
+                    | EnvironmentFlags::NO_SYNC,
+            )
+            .open(&config.path)
+            .map_err(|e| open_error("open environment", e))?;
+
+        let mut dbs = HashMap::new();
+        for ns in &config.namespaces {
```
idky137, line 86: Can we add a simple version check here, triggering a deletion and rebuild if an index version is old?

packages/zainod/src/indexer.rs
```diff
@@ -1,347 +1,326 @@
...
+        index_set(),
+        Arc::clone(&source),
+        |compact_block| context_from_pre_index_compact_block(&compact_block),
+        SyncTuning {
+            batch_size: config.indexer.batch_size,
+            finalised_depth: config.indexer.finalised_depth,
+            channel_capacity: config.indexer.channel_capacity,
+            concurrency: config.indexer.concurrency,
+        },
+    )?;
```
idky137, line 207: This looks to make sync blocking again? Can we keep it as the background operation it currently is, given the ChainView is designed to handle exactly this case.

packages/zainod/src/indexer.rs
```diff
@@ -1,347 +1,326 @@
...
-        let service = IndexerService::<Service>::spawn(service_config).await?;
-        let service_subscriber = service.inner_ref().get_subscriber();
+        confirmed_watermark,
+        runtime_cancel.child_token(),
+    )
+    .await
+    .map_err(IndexerError::ChainHeadInit)?;
+
+    // Compose FS ⊕ NFS into the served engine, behind the light-wallet profile.
+    let engine = Engine::new(store_reader.clone(), chain_head_subscriber);
```
idky137: Is this what ChainView (in https://github.com/zingolabs/zaino/pull/1573) will replace? it would be good to see this branch built on top of that so we can see what in this branch is intended to land.

packages/zainod/src/indexer.rs
```diff
@@ -1,347 +1,326 @@
...
+        .boot(store)
+        .await
+        .map_err(|e| IndexerError::Boot(Box::new(e)))?
+        .boot(chain_head)
+        .await
+        .map_err(|e| IndexerError::Boot(Box::new(e)))?
+        .boot(light_serve)
+        .await
+        .map_err(|e| IndexerError::Boot(Box::new(e)))?
+        .build();
```
idky137, line 268: This looks to be missing `Mempool`, required for the lightclient gRPC service.

packages/zaino-lightserve/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+
+use crate::wire::{to_hex, ToWire};
+
+/// Lightwalletd-compatible handler over a [`LightServeService`] engine.
+#[derive(Clone)]
+pub struct LightServe<S: LightServeService> {
+    engine: S,
+}
+
+impl<S: LightServeService> LightServe<S> {
```
idky137, line 41: This is currently missing many services required to serve zcash lightclients, the full set of methods required is held here:

https://github.com/zingolabs/zaino/blob/a54c291f9cbb25a55b991aa0a62d13449328c16b/packages/zaino-state/src/indexer.rs#L682

packages/zainod/src/indexer.rs
idky137, line 268: This is also missing the JsonRPC server we have committed to providing.

packages/zainod/src/indexer.rs
```diff
@@ -1,347 +1,326 @@
...
+    // Reachability was already confirmed (Direct opened its state DB), so the
+    // runtime's validator gate is a formality here.
+    let validator_component = ValidatorComponent::connect(&AlreadyReachable).await?;
+    let indexer = IndexerComponent::new(ComponentName("indexer"), driver);
+    let store = StoreComponent::new(ComponentName("store"), store_reader);
+    // The chain-head writer is escalated and supervised exactly like the indexer.
+    let chain_head = RunComponent::new(ComponentName("chain-head"), chain_head_writer);
+    let light_serve = ServeComponent::new(
+        ComponentName("light-serve"),
+        GrpcServer::new(LightServe::new(engine), config.serve.grpc_listen_address),
```
idky137, line 247: Note that this will require `MempoosService` as well as ChainView / Engine, as wallets require the GetMempoolStream that the Mempool component was built to support.

packages/zaino-backend-lmdb/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+impl LmdbReader {
+    fn resolve_db(&self, namespace: Namespace) -> Result<Database, ReadError> {
+        self.dbs
+            .get(&namespace)
+            .copied()
+            .ok_or_else(|| ReadError::NamespaceNotFound(namespace.to_string()))
+    }
+}
+
+impl BackendReader for LmdbReader {
```
idky137, line 189: I think these may need to be in blocking tasks (either internally or externally with explicit naming  and documentation). LMDB transactions are blocking operations, if used in a tokio runtime, can block the main execution thread, stalling the whole process. Running these operations in blocking tasks (or blocking task pools) stops them from ever blocking the whole runtime.

packages/zaino-backend-lmdb/src/lib.rs
```diff
@@ -0,0 +1,476 @@
...
+    fn resolve_db(&self, namespace: Namespace) -> Result<Database, CommitError> {
+        self.dbs
+            .get(&namespace)
+            .copied()
+            .ok_or_else(|| CommitError::NamespaceNotFound(namespace.to_string()))
+    }
+}
+
+impl BackendWriter for LmdbWriter {
+    fn commit(&mut self, ops: Vec<WriteOp>) -> Result<(), CommitError> {
```
idky137, line 240: Similar to reads, I think we may need to ensure these are run  in blocking tasks.

packages/zaino-chain-store-zainodb/src/tests/fake_validator.rs
```diff
@@ -143,6 +143,10 @@ impl FakeValidator {
     }
 }
 
+impl zaino_source::ValidatorSource for FakeValidator {
```
idky137, line 146: Im confused that this branch seams to delete zaino-chain-store entirely, is it possible to see this framework built using the lower level library components we have already finalised in dev?

packages/zaino-chainview/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+#![forbid(unsafe_code)]
+#![deny(clippy::wildcard_enum_match_arm)]
+
+mod chain_view;
+mod snapshot;
+
+#[cfg(any(test, feature = "testing"))]
+pub mod testing;
+
+pub use chain_view::ChainView;
```
idky137, line 11: Similar to othr comments, it would be nice to see this branch on top of the other components we have been working on and merging into dev, so we can see just the changes that will be moving into dev.

packages/zaino-indexes/src/capabilities.rs
```diff
@@ -0,0 +1,187 @@
...
+        // Where a transaction was mined — a local lookup. Its raw bytes are a
+        // *separate* capability (`RawTransaction`), served by the validator.
+        Capability::TransactionLocation => &[txid_location::ID],
+        Capability::AddressHistory => &[address_history::ID],
+        Capability::SpendStatus => &[transparent_spends::ID],
+        // No local index — served by the validator.
+        Capability::RawTransaction
+        | Capability::Treestate
+        | Capability::SubtreeRoots
+        | Capability::Mempool
```
idky137, line 48: Nitpick, Mempool is never saved as is purely ephemeral, and **is** served by us, validators only offer the simple `get_raw_mempool` method, it is zaino that must build and serve `gt_mempool_tx` and `get_mempool_stream`.

idky137 (changes requested):
There is a lot of really good code in here but if I am honest I am struggling to review it properly in its current form.

I know this is only an exploratory branch but it would be good to see this built on top of dev, and the imminent chainView component, as currently it is hard to see the true scope of the changes being proposed.

PR #1554, opened 2026-09-15
Feat/zainod metrics rework

elicbarbieri:
- Adds Granular per-action block-work accumulators
- uses cargo features for "no-tls" and "prometheus"
- Migrate the /metrics endpoint to separate admin thread w/ /livez endpoint
- Add /readyz endpoint to docs pending nacho-sync

Overall preps zaino for public service deployment


zancas:
A review pass over this PR. I verified the first two items against the code.
The other items come from the same pass, and I did not verify them by a run.
All three red CI jobs on this PR fail on the pre-existing `librocksdb-sys`
build breakage from #1557, so CI has not yet run the new tests.

## Verified

1. **A failed admin bind strands the metrics upkeep.** In `admin.rs`, `serve`
   spawns the upkeep loop on the admin thread's runtime and then binds. On a
   bind failure, `serve` returns, `block_on` returns, the thread closure ends,
   and the closure's `runtime` drops every task it owns, the upkeep loop
   included. No listener exists, so no scrape calls `render` either. The
   recorder is process-wide and keeps accepting samples, so every histogram
   sample of the sync accumulates without bound for the life of the process.
   The comment above the loop states the intent and not the behaviour.

   Recommendation: bind before `init` installs the recorder, and fail startup
   with the bind error. Observability is opted into twice here, by the
   `prometheus` feature and by `metrics_endpoint`, so a bind failure means the
   service cannot do part of its job, and the operator must see that at
   startup. `TcpListener::bind` sets `SO_REUSEADDR`, so a socket in `TIME_WAIT`
   from the previous instance does not cause a false failure. As defence in
   depth, the upkeep loop belongs on the process runtime, because `serve` has
   other exits after a successful bind, such as the accept loop's return when
   the semaphore closes, and each of them strands upkeep the same way.
2. **A vanished old tip counts as a reorg.** `classify_tip_change` maps
   `fork == None` to `TipChange::Reorg(None)`, and `find_fork_point` answers
   `None` whenever the old tip is absent from the new snapshot. A snapshot
   retains a bounded window, so an advance that pushes the old tip out of the
   window, or a re-anchor after a pause, increments `reorg_total` and logs a
   reorg for a chain that did not fork. Classify the change as `Advance` when
   the old tip's height is below the new snapshot's retention floor.

## Not verified

3. The `SIGTERM` handler is installed inside the serve task, after the DB
   open, validation, and migrations. A signal in that window still kills the
   process with the default action. After installation, `ctrl_c()` removes the
   default `SIGINT` action, so an operator cannot escape a hung close without
   `SIGKILL`. Register the handler once in `run`, before `start_indexer`, and
   pass a cancel token into the loop.
4. The admin listener takes a semaphore permit before `accept` and holds it
   for the connection's life with no idle timeout. Thirty-two idle keep-alive
   connections starve `/livez`. Add an idle timeout, or accept first and
   reject over the cap.
5. On the restart path, the heartbeat stops during `indexer.close().await` and
   clears only after it returns. A close longer than thirty seconds makes
   `/livez` answer 503 during the close. Clear the heartbeat before `close()`.
6. `metrics::gauge!(SYNC_FETCHED_HEIGHT)` runs once per block inside the batch
   loop. Set it once after the loop.
7. `SYNC_ACCUMULATOR_HEIGHT` is set in three places, and the caller pays a
   second DB read to republish it. Keep one set site, and make `mode` an enum.
8. `network_dir` is `pub(super)` where private compiles, and `UNKNOWN_METHOD`,
   `MethodNames`, and `MetricsMiddleware` are `pub(crate)` where private or
   `pub(super)` compiles.
9. Several new item doc-comments run to more than one sentence (`timer.rs`,
   `admin.rs`, `ingest.rs`, `poll.rs`, `metrics.rs`, `client.rs`).
10. The PR adds a fourth hand-rolled scope timer. One shared guard in
    `zaino-common` would serve all of them.

zancas:
The recommendation under item 1 of my review is implemented in https://github.com/zingolabs/zaino/pull/1604, which targets this branch. I had pushed it here directly as f80c5d3d7 and then reverted that push, so this branch is back to your own commits.

zancas (approved):

PR #1555, opened 2026-09-15
zaino-chain-store-zainodb, zaino-primitives: answer the db/block.rs review comments on #1510

zancas:
Two of the review comments on #1510 concern `db/block.rs`. This pull request answers both, one commit each.

## The chainwork row talks to the domain type through its integer

The persistent chainwork row is 32 big-endian bytes, a width fixed by existing v1 databases. The conversion pair had started to route that row through `ChainWork::to_be_bytes` and `ChainWork::try_from_reported`, two doors written for the RPC wire. The reader treats an all-zero value as "not reported", so the store then needed a second branch to turn that absence back into a corrupt row.

The pair now writes the value's 16 big-endian bytes into the low half of the row and, on the way back, treats a non-zero high half or a zero value as a corrupt row. The on-disk form is unchanged. The wire doors keep one caller each, in the RPC layer.

## The work constructors take a `NonZeroU128`

`ChainWork::try_new(u128)` and `BlockWork::try_new(u128)` re-checked an invariant the wrapped type already carries and returned a `Result` every caller unwrapped. Both are now `new(NonZeroU128)`, infallible and const. `ZeroWork` is gone, and the two tests that asserted zero is refused go with it, since that refusal is now a compile-time fact. Every caller proves non-zeroness once, in its own terms.

## Verification

`cargo check --workspace --tests` and `cargo clippy --workspace --tests` are clean. The 61 unit tests covering the work types, the persistent row, its cross-encoder equivalence, the golden vectors, and the block-header render pass.


zancas:
Superseded by https://github.com/zingolabs/zaino/pull/1558, the same three commits rebuilt on the renamed base; the remote rule forbids rewriting this branch.

PR #1557, opened 2026-09-15
build: restore the root lockfile from before the #1553 refresh

zancas:
Since #1553 merged, every fresh build of the workspace fails in `librocksdb-sys`'s build script:

    a `libclang` shared library is not loaded on this thread

Dev's own Lint and Test jobs fail this way at the tip, and so does any developer machine without a warm target directory. #1553 changed two live-tests files and, alongside them, refreshed about 150 crates in the root `Cargo.lock`.

## Change

This pull request restores the root lockfile from fe9c72111, the last green merge, byte for byte. No root manifest changed in between, so the restored lockfile resolves the current tree exactly. `cargo check --workspace --tests --locked` passes on a cold checkout.

## What is not known yet

Pinning `clang-sys` back to 1.8.1 or `cc` back to 1.2.65 on their own does not cure the failure, so the offending crate inside the refresh is not identified. Until it is, the refresh stays out. A future dependency bump should move one crate at a time so the culprit shows itself.


elicbarbieri (approved):
oopsie poopsie.  my bad

PR #1558, opened 2026-09-15
zaino-chain-store-zainodb, zaino-primitives: answer the db/block.rs review comments on #1510

zancas:
Supersedes #1555, rebuilt on the current `fix/chainwork-primitive` after its rename to `AbsoluteChainWork` and `SingleBlockWork`. The remote rule forbids rewriting the old branch, so this is a fresh one. Same three answers to the `db/block.rs` review comments on #1510, one commit each.

## The chainwork row talks to the domain type through its integer

The persistent chainwork row is 32 big-endian bytes, a width fixed by existing v1 databases. The conversion pair routed that row through `to_be_bytes` and `try_from_reported`, two doors written for the RPC wire, whose reader treats an all-zero value as "not reported"; the store then needed a second branch to turn that absence back into a corrupt row.

The pair now writes the value's 16 big-endian bytes into the low half of the row and, on the way back, treats a non-zero high half or a zero value as a corrupt row. The on-disk form is unchanged. The wire doors keep one caller each, in the RPC layer.

## `SingleBlockWork::new` takes a `NonZeroU128`

`AbsoluteChainWork::new` already does. `SingleBlockWork::try_new(u128)` re-checked an invariant the wrapped type carries and returned a `Result` every caller unwrapped. It is now `new(NonZeroU128)`, infallible and const; `ZeroWork` and the test asserting zero is refused go with it.

## Literal work values are proven non-zero at compile time

Every test and fixture literal now lives in a named `const`, evaluated at compile time, so a zero literal fails `cargo check` rather than a test run. The parametric helpers and the two production conversions keep their runtime checks.

## Verification

`cargo check --workspace --tests` and `cargo clippy --workspace --tests` are clean. 75 targeted tests over the work types, the persistent row, its cross-encoder equivalence, the golden vectors, the block-header render, and the difficulty pipeline pass.


zancas:
Superseded by https://github.com/zingolabs/zaino/pull/1563, replayed on top of #1560.

PR #1559, opened 2026-09-15
zaino-primitives, zaino-source-zebra-rpc: drop the inbound chainwork doors

zancas:
Stacked on #1558, which it needs for the store row to be off the wire doors already. Merge #1558 first.

## Why

Zebra omits `chainwork` from `getblockheader` and hardcodes a `0` placeholder in `getblockchaininfo`, and its source says it does not plan to support the field. With zcashd support removed, no validator Zaino talks to reports a chainwork value, so every inbound path produced `None`, and the hex decoding, left-padding, 32-byte width check, and "all-zero means not reported" convention all guarded a case that cannot occur.

## Change

- Delete `AbsoluteChainWork::try_from_reported`, `ChainWorkOverWidth`, and the private byte splitter.
- Delete the parser's `parse_reported_chain_work` and `chain_work_be_bytes`, its `AbsoluteChainWork` error variant, and their three tests.
- The three domain fields the parser filled now read `None` outright.
- `to_be_bytes` stays as the one wire render, with a test pinning its layout.

## Follow-up, not in this pull request

Nothing in Zaino fills the domain `chainwork` fields from its own fold, so the served `getblockheader` chainwork is already always omitted. Those three `Option<AbsoluteChainWork>` fields, the serve mapping, and `to_be_bytes` are candidates for deletion too, unless Zaino intends to serve its computed chainwork.

## Verification

`cargo check --workspace --tests` and `cargo clippy --workspace --tests` are clean. 360 tests across zaino-primitives, zaino-source-zebra-rpc, zaino-serve, and zaino-chain-store-zainodb pass.


zancas:
Testing...  locally...  will announce results.

zancas:
Superseded by https://github.com/zingolabs/zaino/pull/1563, replayed on top of #1560.

PR #1556, opened 2026-09-15
zaino-chain-head-service reimplementation backed by imbl immutable types

AloeareV:
Fixes:

PR #1560, opened 2026-09-15
zaino-primitives: RelativeChainWork, and collapse chain-head's duplicate onto it

nachog00:
Based on `dev`, now that #1510 has merged. Adds the third work quantity to `zaino-primitives` and deletes the crate-local duplicate it replaces.

## Why

`zaino-chain-head` holds a bounded window, never reads the finalised state, and so cannot know the work below that window. It modelled the anchor-relative quantity as its own `ChainHeadWork(u128)`. The duplicate was not the problem — `ChainHeadWork::as_u128` was. It existed so a caller could hand an anchor-relative total to something expecting an absolute one, and `zaino-state`'s chain-head adapter did exactly that.

`RelativeChainWork` has no conversion to `AbsoluteChainWork`, so that call cannot be written.

## The quantity

A total over a *run* of consecutive blocks, not an offset from a chosen block. Zero is a real value — the empty run — so the fold needs no special case for its first element. `Ord`, because choosing between branches that share a start is a comparison among runs.

`WorkOverflow` gains a second caller and moves to `work/error.rs`. `WorkUnderflow` stays with `AbsoluteChainWork`: unwinding is not an operation on a run.

## Anchor convention

Work is measured from the anchor rather than from the anchor's parent. The anchor seeds the fold at `RelativeChainWork::ZERO` and contributes none of its own work, because that work belongs to the chain below the window. Previously it seeded at its own work, so a total combined with the absolute work at the anchor would have counted that block twice.

A window's totals are therefore each one block's work lower than before. Comparisons between them are unaffected, which is all chain-head uses them for.

## Chain work becomes optional

`BlockContext::chainwork` is `Option<AbsoluteChainWork>`. The on-disk format is unchanged — `PersistentChainWork` writes the same 32 big-endian bytes, and the goldens pin it. What changed is the encode boundary: `PersistentBlockContext::from_business` is now fallible and refuses a block with no chain work.

That rule previously lived in a comment on the chain-head adapter — "writing one of these to the database would put a wrong chainwork on disk" — and is now a typed error on the one edge that could produce it.

`chain_head::indexed_block` passes `None`; the relative-to-absolute fabrication is deleted rather than improved.

## Not included

No `extend` or `since`. Combining a run with the absolute work below it has no caller: serving reads `chainwork` from the validator's own response (`zaino-source-zebra-rpc/src/parse.rs`), and chain-head blocks are never persisted. Those relations land when a consumer needs them.

## Verification

- `makers fmt`, `makers clippy`: clean
- `zaino-primitives` 73, `zaino-chain-store-zainodb` 146, `zaino-chain-head-service` 22, `zaino-chain-head` 1 — all pass
- `zaino-state --lib` is 78 passed / 49 failed, identical to the base commit — missing local test fixtures, not a regression
- `makers doc` fails only on links that already fail on the base: `zaino-source-macros`, `zaino-source`, `tree_root.rs`, and `ChainStateEpoch` in `zaino-chain-head/src/lib.rs`

Two pre-existing issues this branch does not address, noted for the record: CI currently fails building `librocksdb-sys` (`a libclang shared library is not loaded on this thread`) before any test runs; and `types::work` and `types::zatoshis` are private modules, so their module documentation is never rendered in rustdoc.

ZIT-Revision: bb8cdcc4fdbcb37b2a8c7b2bfe67a323d7b66a11


elicbarbieri:
Is there a test that captures/validates the behavior of the constructed/indexed ChainWork?  How good is that test?  

This is probably one of more critical things since we actually have to index, and dont just clone the zebra struct

packages/zaino-chain-store-zainodb/src/types/db/block.rs
```diff
@@ -267,14 +277,34 @@ mod tests {
...
             BlockHash::from([0x11; 32]),
             BlockHash::from([0x22; 32]),
-            AbsoluteChainWork::new(NonZeroU128::new(0x0123_4567).expect("nonzero")),
+            Some(AbsoluteChainWork::new(
+                NonZeroU128::new(0x0123_4567).expect("nonzero"),
+            )),
             Height(0x0dec_0de0),
         );
-        let persisted = PersistentBlockContext::from_business(&bctx);
+        let persisted = PersistentBlockContext::from_business(&bctx).expect("chain work present");
```
zancas, line 277: How is this error state reached?  Can we enforce a compile time check and make "from_business" infallible?

packages/zaino-chain-store-zainodb/src/types/db/block.rs
```diff
@@ -160,13 +160,23 @@ pub(super) struct PersistentBlockContext {
 }
 
 impl PersistentBlockContext {
-    pub(super) fn from_business(context: &BlockContext) -> Self {
-        Self {
+    /// Fallible in the encode direction, unusually for a `from_business`: a
+    /// block whose chain work is unknown has no stored form, and this is where
+    /// that is refused. See [`BlockContext::chainwork`].
```
zancas, line 157: This should be infallible, and the code should be checked for correctness at compile time.

packages/zaino-chain-store-zainodb/src/types/db/block.rs
nachog00: it's a conversion into persistent types... not sure how it can fail, i assume the target (persistent) type has some self owned invariace... unreleated to this pr tho. i wouldn't block on it. that crate is going out soon too.

nachog00: Ah, i know how it can fail. BlockContext is what is persisted. before this pr, that block had a non optional chainwork field. that field cannot be computed if the total wokr hasn't been indexed yet. and converting the nfs's relative work into it is just wrong. for it to be properly set it need both the absolute work to the fs tip, and the relative work of the best chain. that can't be done until the index is fully synced. previously, a fake value was constructed which let that conversion be 'infallible'. This pr set's that field as Option<> so it sets it as simply `None` if the field isn't properly constructable.

packages/zaino-chain-store-zainodb/src/types/db/block.rs
nachog00: this is just honest handling, instead of storing an artificial work value, it refuses.

packages/zaino-chain-head-service/src/service.rs
```diff
@@ -44,9 +44,11 @@ use tokio::{
 use tokio_util::sync::CancellationToken;
 use tracing::{debug, info, instrument, warn};
 use zaino_chain_head::{
-    ChainHeadBlock, ChainHeadBlockSource, ChainHeadConfig, ChainHeadSnapshot as _, ChainHeadWork,
+    ChainHeadBlock, ChainHeadBlockSource, ChainHeadConfig, ChainHeadSnapshot as _,
+};
```
elicbarbieri, line 48: Whoa?  Why are we doing ChainHeadSnapshot as _?

packages/zaino-chain-head-service/src/service.rs
```diff
@@ -649,9 +651,8 @@ impl<S: ChainHeadBlockSource> ChainHeadService<S> {
     /// is the fallback the old code used whenever the reader could not serve
     /// the height, which was every time the database lagged.
     ///
-    /// The anchor sits below the reorg-possible range, so its accumulated work
-    /// is the base of this window's own accumulation rather than an absolute
-    /// value — see `ChainHeadWork`.
+    /// The anchor sits below the reorg-possible range. Work is measured from
+    /// it, so it seeds the window's fold with `RelativeChainWork::ZERO`.
```
elicbarbieri, line 655: Call it "Top of finalized state.  Can no longer be re-orged"

packages/zaino-chain-head-service/src/service.rs
```diff
@@ -663,7 +664,7 @@ impl<S: ChainHeadBlockSource> ChainHeadService<S> {
         })?;
 
         let tree_roots = self.tree_roots(block.header.hash).await?;
-        chain_head_block(block, &tree_roots, None)
+        Ok(anchor_chain_head_block(block, &tree_roots))
```
elicbarbieri, line 667: Yeah maybe call it finalized_state_tip or somethign instaed of adding in another "anchor" naming scheme/abstraction.  We should just have finalized and non-finalized, and no other conflicting terminology

packages/zaino-chain-head-service/src/service.rs
```diff
@@ -800,7 +815,7 @@ fn chain_head_block(
         work,
         block,
         tree_roots: tree_roots.clone(),
-    })
+    }
 }
```
elicbarbieri, line 812: There is a better way to express this.  You have better context than me, but definetly dont need 3 similarly named functions that all do similar things.  Simplify

packages/zaino-chain-head/usage.md
```diff
@@ -89,14 +89,14 @@ head down because there is no method on it that could.
...
-a value from a validator — two chain heads with different anchors produce
-different numbers for the same block.
+A block's `work` is a `RelativeChainWork`: the total over the blocks this window
+retains above its **own anchor**. The anchor's own work is
+`RelativeChainWork::ZERO`, and each block above folds its own work onto its
+parent's total.
+
+That orders competing branches, which is all the chain head needs. Two chain
+heads with different anchors produce different numbers for the same block, so
+compare a `RelativeChainWork` only against another from the same window.
```
elicbarbieri, line 99: Why do we need a RelativeChainWork at all?  We already have the index built, so can we just use ChainWork and not have to worry about resetting these values once blocks go from non-finalized -> finalized?

packages/zaino-primitives/src/types/work/relative_chain_work.rs
elicbarbieri, line 1: Looks good.  much simpler

elicbarbieri (dismissed):
Looks good overall.  however I want to understand why a RelativeChainWork is needed at all?  I understand ChainWork as the accumulator type, and then BlockWork as the per-block type, etc...

But the relative chainwork seems complex and of little value.  Why do we need that?  Is it for serving non-finalized state before a full index is built?

packages/zaino-chain-head/usage.md
nachog00: Answered in full on the review. Short version: the three quantities have different invariants (`AbsoluteChainWork` is non-zero and has `rollback`; `RelativeChainWork` has `ZERO` and does not), ChainHead has no absolute value available to it, and this PR deletes the crate-local `ChainHeadWork` it replaces.

packages/zaino-chain-head-service/src/service.rs
nachog00: This is not the finalised state's tip. It is `tip - (max_depth + RETENTION_MARGIN)`, fetched from the validator; the finalised-reader arm was removed with the finalised state.

`zaino-chain-head` is entirely non-finalised — `lib.rs` says it never reads the finalised state — so naming this block after that state names something the crate does not know.

packages/zaino-chain-head-service/src/service.rs
nachog00: Same as the thread above: this block is the window's own root, not the finalised tip.

"Anchor" is already this vocabulary: 32 mentions in `service.rs` on `dev`, a `usage.md` section "Work is anchor-relative", and the `ChainGraph` work in #1576 speaks of a graph re-anchoring. Renaming here would leave two words for one block. If the term itself is the objection, `window_root` reads the same and I can rename throughout.

packages/zaino-chain-head-service/src/service.rs
nachog00: Fair. `dev` builds both cases in one function taking `parent_work: Option<ChainHeadWork>`, and this PR split it into three. The split is not needed: the new convention changes one arm, `ChainHeadWork::anchored_at(block_work)` to `RelativeChainWork::ZERO`, because the anchor no longer contributes its own work.

Reverting to the single function keeps the diff to that arm. Keeping the shape as `dev` has it also avoids colliding with #1576, #1578, #1579 and #1556, which are all rewriting this file.

packages/zaino-chain-head-service/src/service.rs
nachog00: A trait must be in scope for its methods to be callable, and `as _` imports it for exactly that without binding the name. The line predates this PR; it was only reflowed when `ChainHeadWork` left the import list.

packages/zaino-chain-store-zainodb/src/types/db/block.rs
nachog00: On further thought, you are right. The refusal is honest, but it is a runtime check for something the types can rule out: a context carrying a non-optional chain work, built only by the path that computes it, makes the encode direction total.

Scoped to a follow-up: #1587.

packages/zaino-chain-store-zainodb/src/types/db/block.rs
nachog00: On further thought, you are right. The refusal is honest, but it is a runtime check for something the types can rule out: a context carrying a non-optional chain work, built only by the path that computes it, makes the encode direction total.

Scoped to a follow-up: #1587.

nachog00:
> why is a RelativeChainWork needed at all?

The three quantities have different invariants, and the algebra is documented in `zaino-primitives/src/types/work.rs`:

```text
W ∈ (0, 2^128)   C ∈ (0, 2^128)   R ∈ [0, 2^128)

genesis    : W → C      accumulate : C × W → C
rollback   : C × W → C  accumulate : R × W → R
```

`AbsoluteChainWork` is a `NonZeroU128` and has `rollback`, for unwinding a block on reorg. `RelativeChainWork` has `ZERO` as a real value — the empty run — and no rollback. `C × C` is undefined, and no relation connects `C` and `R`.

ChainHead cannot produce an absolute value. Zebra reports no `chainwork` at all (`zaino-source-zebra-rpc/src/parse.rs`, and the ReadState adapter), so Zaino accumulates work itself; and ChainHead depends only on `zaino-consensus`, `zaino-primitives` and `zaino-source`, holding a window fetched from the validator. The work below its anchor is not available to it.

Using absolute for everything is what the code does today, and this PR deletes it:

```rust
fn chainwork(work: ChainHeadWork) -> AbsoluteChainWork {
    AbsoluteChainWork::new(NonZeroU128::new(work.as_u128()).expect("…non-zero anchor"))
}
```

That handed a window-relative total to `IndexedBlock`, the store's shape. It compiled only because the anchor was seeded with its own work rather than zero, so the `NonZeroU128` could not fail — which also double-counted the anchor's work against any real absolute value. The comment above it read "writing one of these to the database would put a wrong chainwork on disk"; a comment was the only thing preventing it. `RelativeChainWork` has no conversion to `AbsoluteChainWork`, so that call cannot be written.

Nothing is reset at finalisation: ChainHead blocks are never persisted, and the store computes absolute chainwork from the parent when it indexes (`conversion::chainwork_from_parent`).

This also removes a type rather than adding one — `zaino-chain-head` already models this quantity as its own `ChainHeadWork(u128)`.

nachog00:
> Is there a test that captures/validates the behavior of the constructed/indexed ChainWork?

Current coverage, and it is thinner than it should be:

- `zainodb/types/primitives/compact_difficulty.rs` runs the pipeline on synthetic bits: on-disk u32 → `CompactDifficulty` → `SingleBlockWork` → genesis → one fold, asserting the total is exactly twice a single block's work.
- A property test round-trips the 32-byte stored form.
- Unit tests on the types: 10 absolute, 4 relative, 1 per-block.

Missing: no test indexes a run of real blocks and checks the resulting chainwork, and nothing is checked against a validator's own value. Fixtures use invented constants.

Worth adding a store-level test that indexes a sequence and asserts each block's chainwork equals the sum of per-block work, that it increases strictly, and that genesis equals its own work — plus a comparison against reported `chainwork` for known heights from the legacy full node, since Zebra reports none. Tracking separately rather than growing this PR.

zancas:
I pushed a merge of `dev` to this branch, as 35c911401, so the PR can merge
after #1554 landed. Two files conflicted.

1. `zaino-chain-head-service/src/service.rs`. #1489 on `dev` inserted
   `advance_error` directly above the builder doc that this branch rewrote
   when it split the builder into `anchor_chain_head_block` and
   `extending_chain_head_block`. Both are kept. One more change was needed
   outside the conflict: the advance loop from #1554 merged cleanly but still
   called `chain_head_block(block.clone(), &roots, Some(prev_block.work))`,
   which no longer types under this branch. It now calls
   `extending_chain_head_block(block.clone(), &roots, prev_block.work)`. The
   `tokio::join` prefetch and the `advance_error` mapping from `dev` are
   unchanged.
2. `zaino-source-zebra-rpc/src/parse.rs`. The import list takes `dev`'s
   additions, `BlockConfirmations`, `TransparentAddressError`, and
   `TxConfirmations`, and drops `AbsoluteChainWork`, which this branch's
   deletion of the inbound chainwork parse left without a use.

Verification on the merged tree: `makers clippy` is clean over the workspace,
`cargo fmt --check` is clean, and `cargo nextest run --all-features` passes
471 tests across `zaino-primitives`, `zaino-chain-head-service`,
`zaino-chain-store-zainodb`, `zaino-source-zebra-rpc`, and `zaino-state`.
File by file, the branch's change against `dev` now matches this PR's own
diff against its base, plus the two resolution lines above.

#1607 targets this branch and is based on its previous tip. It touches only
`zaino-chain-store-zainodb`, outside the conflict set, so it still merges
here. It fixes item 1 of my review comment above, so merging it before this
PR closes that item on the branch.

zancas (dismissed):
I have updated the code check for invalid work values at compile time.

zancas (approved):
Significant cleanup.

zancas (approved):

PR #1561, opened 2026-09-15
build(docker): install the pinned toolchain components in a cached layer

zancas:
## Summary

Every ztest image build re-syncs the Rust channel and downloads clippy. The log shows it before the first `Compiling` line:

```
info: syncing channel updates for 1.96.0-x86_64-unknown-linux-gnu
info: downloading component clippy
```

The builder stage starts from the rust image that matches `rust-toolchain.toml`. That image ships the minimal profile. The toolchain file also lists `clippy` and `rustfmt`, so rustup installs them inside the workspace build layer. The cargo cache mounts do not cover the rustup home, so the download repeats on every build.

## Change

A new `RUN rustup component add clippy rustfmt` layer sits before `COPY . .`. The layer cache keeps the components, and a rebuild goes straight to cargo.

## Test plan

- Build the image twice with ztest. The second build prints no `syncing channel updates` line.


elicbarbieri:
This only takes 1-2 seconds.  Need to optimize the build & intstall ztest-cli first

elicbarbieri (approved):

PR #1563, opened 2026-09-15
zaino-primitives, zaino-chain-store-zainodb, zaino-source-zebra-rpc: chainwork through its integer, infallible constructors, no inbound path

zancas:
Stacked on #1560. Supersedes #1558 and #1559, whose commits are replayed here on top of `fix/relative-chain-work` and adapted to its shape: `BlockContext::chainwork` as `Option`, `WorkOverflow` in `error.rs`, and the new `RelativeChainWork` callers.

Draft until the direction is agreed, because this branch and #1560 differ on one point: #1560 keeps `try_from_reported` on the premise that serving reads chainwork from the validator's response, and the fourth commit here deletes it on the evidence that Zebra never sends one.

## Commits

1. **The chainwork row talks to the domain type through its integer.** The persistent row writes the value's 16 big-endian bytes into the low half and reads them back by checking the high half is zero and the value non-zero. No wire semantics in the store; the on-disk form is unchanged.
2. **`SingleBlockWork::new` takes a `NonZeroU128`**, matching `AbsoluteChainWork::new`. `ZeroWork` goes. The three callers #1560 added, in chain-head-service, zaino-state's tests, and the relative-work tests, prove non-zeroness on their own `NonZeroU128::new` step.
3. **Literal work values are proven non-zero at compile time**, as named constants in the tests and fixtures.
4. **The inbound chainwork path goes.** Zebra omits `chainwork` from `getblockheader` and hardcodes `0` in `getblockchaininfo`, and zebra-rpc's source says it does not plan to support the field. With zcashd removed, `AbsoluteChainWork::try_from_reported`, `ChainWorkOverWidth`, the parser's hex decoder, its error variant, and their tests guard a case that cannot occur. The three domain fields the parser filled now read `None`. `to_be_bytes` stays as the one wire render.

## Not included

Nothing in Zaino fills the domain `chainwork` fields from its own fold, so the served `getblockheader` chainwork is already always omitted; those three `Option<AbsoluteChainWork>` fields, the serve mapping, and `to_be_bytes` are a further deletion if Zaino does not intend to serve its computed chainwork.

## Verification

`cargo check --workspace --tests` and `cargo clippy --workspace --tests` are clean. 388 tests across zaino-primitives, zaino-chain-store-zainodb, zaino-source-zebra-rpc, zaino-serve, zaino-chain-head-service, and zaino-chain-head pass.


zancas:
We should not implement logic that handles zcashd formatted responses as the source of chainwork.

packages/zaino-state/src/chain_index/tests/golden.rs
```diff
@@ -63,6 +64,9 @@ use crate::{
     TransparentTxList, TxInCompact, TxLocation, TxOutCompact, TxidList, ZainoVersionedSerde,
 };
 
+/// The chainwork the golden block context carries.
+const CHAINWORK: NonZeroU128 = NonZeroU128::new(0x0dec_0de0).expect("nonzero literal");
```
nachog00: This file is dead code — it is edited here as though it were live.

`packages/zaino-state/src/chain_index/tests.rs` declares `chain_head, mockchain_tests, poll, proptest_blockgen, sync_loop, vectors`. There is no `mod golden;` anywhere in `zaino-state`, and no `#[path]` reference, so this is never compiled. It could not compile if declared: it imports `crate::chain_index::finalised_state::*` and `crate::chain_index::types::db::*`, both of which moved to `zaino-chain-store-zainodb`, and `crate::BlockContext`, which `zaino-state` does not re-export. At line 191 it also passes a bare `AbsoluteChainWork::new(CHAINWORK)` where the base branch changed that parameter to `Option<AbsoluteChainWork>` — the live sibling at `zaino-chain-store-zainodb/src/golden.rs:190` correctly passes `Some(..)`. The unsorted import list confirms rustfmt never sees it.

It is an orphan left by the crate extraction, not a second oracle. Deleting it is better than maintaining it, and editing it gives false confidence that both goldens were updated. I had this wrong myself when I left it alone on the base branch.

packages/zaino-chain-store-zainodb/src/types/db/block.rs
```diff
@@ -35,24 +37,27 @@ use zaino_encoding::{
...
-/// reported", but a block row always has a chainwork, so a zero row is a
-/// corrupt row, not an absent value.
 #[derive(Debug)]
 pub(super) struct PersistentChainWork([u8; 32]);
 
 impl PersistentChainWork {
     pub(super) fn from_business(cw: &AbsoluteChainWork) -> Self {
-        Self(cw.to_be_bytes())
+        let mut row = [0u8; 32];
+        row[16..].copy_from_slice(&NonZeroU128::from(*cw).get().to_be_bytes());
```
nachog00: This hand-rolls the 32-byte layout that `AbsoluteChainWork::to_be_bytes` already produces, byte for byte, and `into_business` below hand-rolls the inverse. The on-disk format now has two independent definitions with nothing tying them together: `assert_encoders_agree` compares `from_business` against the recovered legacy `U256` encoder, and `to_be_bytes_fills_the_low_half` checks `to_be_bytes` in isolation.

That matters because `to_be_bytes` is also the serve-side render (`zaino-serve/src/rpc/jsonrpc/wire/block_header.rs:139`). A future edit to either definition drifts silently in one direction only — the store's proptest stays green while served hex goes wrong, or the reverse. It gets worse once we start serving our own computed value rather than the validator's.

Either call `to_be_bytes()` here, or add a test asserting the two agree.

nachog00 (commented):
Reviewed as the author of the base branch, so treat my stake in the deleted code as declared.

The design holds up and I'd take it. Dropping the inbound path is the same YAGNI argument I used to keep `extend`/`since` out of #1560, applied one step further: zebra hardcodes `chainwork` to zero, both source adapters now write `None`, and nothing consumes the field — so a parse door for it has no caller. `SingleBlockWork::new(NonZeroU128)` is likewise the change #1510 made to `AbsoluteChainWork`, extended to its sibling.

I checked the one thing I expected to be a loss and it isn't: the 128-bit width refusal survives in `PersistentChainWork::into_business`, which still rejects a non-zero high half. The guard moved from the primitive's wire door to the only boundary that still receives bytes it did not produce. That is better placement than what it replaced.

Three things to fix, none about the design.

**1. The changeset covers none of this PR.** `.changesets/quiet-falcon.toml` is byte-identical to the base branch's. This PR removes three public items from `zaino-primitives` (`AbsoluteChainWork::try_from_reported`, `ChainWorkOverWidth`, `ZeroWork`) and changes a public signature (`SingleBlockWork::try_new(u128) -> Result<_, ZeroWork>` becomes `new(NonZeroU128)`) — all `breaking` — and `zaino-source-zebra-rpc` stops parsing chainwork from every response, which is a behaviour change. As it stands `relman changeset check` will report uncovered governed crates when the stack retargets `dev`, and `zaino-primitives` would take a non-breaking bump with a CHANGELOG that does not mention the removals.

**2. Docs that now describe a path this PR deleted.** `zaino-primitives/src/types/blockchain_info.rs:43` and `types/block_verbose.rs:27` both read as conditional — "`None` when the validator does not track it" — but `parse.rs` and `zaino-source-zebra-readstate/src/adapter.rs` now write `None` unconditionally, so the field is always absent. `types/rpc/block_header.rs:72` already has the wording those two need. `blockchain_info.rs` also still claims full 256-bit width and credits `AbsoluteChainWork` with avoiding truncation, which was `try_from_reported`'s job.

**3. Worth a comment at the three `chainwork: None` sites** (`parse.rs:396`, `:864`, `:963`). Nothing records that this is a policy keyed to zebra's current behaviour rather than a permanent fact. `:864` also drops a required-field probe — a `getblockchaininfo` response missing the key entirely used to fail with `MissingField` and now parses. Low impact since nothing reads it, but if zebra ever implements the field the value is discarded with no compile-time or test signal. A line pointing at the zebra issue at each site makes that recoverable.

One piece of context that argues *for* this PR: we already compute absolute chain work correctly in the store's fold and persist it, but no serving path reads it back — `BlockHeaderVerbose` and `BlockVerbose` are built only from the validator's response, so `getblockheader` renders whatever the validator gave. Serving our own value means sourcing that one field from the index for finalised heights. Removing the inbound direction does not block that and is permanently safe, since zebra is never going to hand us the value.

zancas:
"we already compute absolute chain work correctly in the store's fold"

I dislike this use of the word "fold".   There's noun/verb ambiguity.

PR #1569, opened 2026-09-16
Feat/chain view: pt1 primitives

idky137:
### Four stacked PRs:

1. **this PR**: shared vocabulary: an address key both halves index by, and 256-bit chainwork addition
2. **chain-store**: frozen blocks carry no chainwork; a freeze gap becomes a typed, repairable error
3. **chain-head**: work counted from a named anchor; status becomes a component's two axes
4. **zaino-chain**: the composer itself, plus reaching it from `zaino-state`

### Pt 1 - Shared vocabulary for keying addresses and folding chainwork

Two additive changes to `zaino-primitives`, both so that two subsystems agree rather than each inventing their own.

**`TransparentAddressKey`** — how an index keys a transparent output: a 20-byte hash plus the script form it came from. The finalised store and the chain head each hold half of an address's history; a consumer merging them should not have to translate. `from_script` shares `classify_script`, so the two halves cannot classify the same script differently. Not the wallet-facing string: that cannot name a non-standard output, and is network-specific.

**`ChainWork::ZERO` and `checked_add`** — the type held the 256-bit value but offered no way to add to it, so callers truncated to `u128` or reimplemented the carry. `ZERO` is the value below genesis, making the fold a plain accumulation. Checked, because a wrapped total would compare as the lightest chain rather than the heaviest — it would not fail, it would select wrong.


elicbarbieri (dismissed):
Looks good

idky137:
ChainWork was removed in dev so the changes here have been removed, upstream's new types will be used in stacked PRs.

elicbarbieri (approved):

PR #1570, opened 2026-09-16
Feat/chain view pt2 chain store

idky137:
## Frozen blocks carry no chainwork, and a gap is a typed error

Three changes to the finalised store, all breaking on published surfaces.

**`StoredAddress` → `TransparentAddressKey`.** Two names for the same thing is how the two halves of one address's history come to disagree about what they describe. The store now keys by the shared primitive from PR 1.

**`freeze` takes `FrozenBlock`** — a stored block without its chainwork. Cumulative work needs an unbroken chain below a block, so only the store can know it; a caller has at best work measured from somewhere else. A field a caller could fill is a field a caller could fill wrongly, so there is no field. The store derives its own.

**`ChainStoreError::FreezeGap`.** `freeze` used to stop at the first block above `tip + 1` and return `Ok(())` — having done nothing, indistinguishable from having done the work, and because the tip never moves, every later freeze also "succeeded". A store that has quietly stopped advancing looked exactly like one keeping up. It now reports both ends of the gap so a caller can build to `first_frozen - 1` and freeze again. Not a fault: it is the ordinary state at handover.

Also carries a build fix: `zaino-proto` declares `librocksdb-sys` with `bindgen-runtime`. Two bindgen majors share one `clang-sys`; the newer enables its `runtime` feature graph-wide while rocksdb's older one enables neither `bindgen-*` feature, so `ensure_libclang_is_loaded` compiles to a no-op and every cold build panics. Declared `internal`, with a note to remove it once upstream converges.


packages/zaino-chain-store-zainodb/src/adapter.rs
```diff
@@ -571,26 +571,81 @@ impl<T: ChainStoreSource> ChainStoreIngest for FinalisedState<T> {
...
+        let tip = domain_height(tip)?;
+
+        Ok(ChainStoreService::reader(self)
+            .blocks_chunk(tip, tip)
+            .await?
+            .into_iter()
+            .next()
+            .map(|block| block.chainwork))
+    }
+}
```
elicbarbieri, line 596: This could be improved.  the .into_iter().next().map() is hard to parse at a glance

elicbarbieri (dismissed):
Looks good.  Everything seems clear

packages/zaino-chain-store-zainodb/src/adapter.rs
idky137: Addressed in [ef050b2](https://github.com/zingolabs/zaino/pull/1570/commits/ef050b2e1149e8b41b925ae9c2135fc0937affe9).

Thanks for flagging this, there was actually a bug here that splitting uncovered, an empty chunk for an existing tip was incorrectly returned as None, which freeze reads as "empty store", so the next frozen block would have had its chainwork seeded as if it were genesis. That case now returns MissingRow, so None only ever means the store is empty.

AloeareV (commented):
I approve of this code! I have not pressed the approve button as CI is failing, and I'm currently unable to rerun the failing tests locally. If anyone else can verify a successful test run, that is all that remains for me to consider this ready to merge!

PR #1572, opened 2026-09-16
Feat/chain view pt3 chain head plus components

idky137:
## Status as a component, and an anchor the work is counted from

Two refactors, both breaking on published surfaces.

**Status splits into two axes.** `zaino-status` fused a lifecycle phase and a health condition into one value, so a fault overwrote the phase: a `Ready` store that hit a recoverable error stopped recording it had been ready, and a degraded store was indistinguishable from one still starting up. `zaino-component` models them independently. The chain-store ports take `StatusSource` as a supertrait rather than declaring a `status` of their own, so a store *is* a component a supervisor can observe without knowing it is a store; both chain-head handles swap `Status` for `StatusSource`.

The adapters still track the fused form internally and convert at the boundary — one small function each, marked transitional, to be deleted when each side tracks both axes natively. `FinalisedState` carries its `ComponentName` as a field rather than reading a constant at the reporting site: a deployment running two stores wants two names, and a supervisor needs to know which one escalated.

**`ChainHeadWork` → `AnchoredRelativeChainWork`.** The old name did not say what it was work *from*. The chain head counts from its own anchor because it never reads the finalised state, and the mistake that invites is the expensive one — treating it as the chainwork a validator reports, which is wrong by exactly the work below the anchor and looks entirely plausible. `ChainHeadSnapshot::work_anchor` now names that anchor so a consumer can rebase rather than guess.

It also wraps the 256-bit `ChainWork` instead of a `u128`. Per Zakura, the protocol permits a single block whose work exceeds a `u128`. The ceiling is not gone — `work_from_bits` still answers in a `u128` and refuses above it — so the call sites that widen mark exactly where it still stops. `ChainWork` gains the ordering chain selection needs; a test pins that it compares correctly across the limb boundary, since getting that wrong would silently prefer the lighter chain.

`relman.toml` gains a target for `zaino-component`: four published crates now depend on it by path and version, and a published crate cannot depend on an ungoverned one.


packages/zaino-chain-head-service/src/service.rs
```diff
@@ -719,12 +723,34 @@ impl<S: ChainHeadBlockSource> ChainHeadService<S> {
...
+        StatusType::Closing => (Lifecycle::Closing, Health::Healthy),
+        StatusType::Offline => (Lifecycle::Offline, Health::Offline),
+        StatusType::Busy | StatusType::RecoverableError => {
+            (Lifecycle::Syncing, Health::Recoverable)
+        }
+        StatusType::CriticalError => (Lifecycle::Offline, Health::Critical),
+    };
+
+    ComponentStatus::new(ComponentName(status.name()), lifecycle, health)
+}
```
elicbarbieri, line 752: @nachog00 I reviewed a PR yesterday that did this same pattern.  Is this just temporary?

packages/zaino-chain-store-zainodb/src/adapter.rs
```diff
@@ -626,6 +630,47 @@ fn accumulate_chainwork(
...
+        StatusType::Closing => (Lifecycle::Closing, Health::Healthy),
+        StatusType::Offline => (Lifecycle::Offline, Health::Offline),
+        StatusType::Busy | StatusType::RecoverableError => {
+            (Lifecycle::Syncing, Health::Recoverable)
+        }
+        StatusType::CriticalError => (Lifecycle::Offline, Health::Critical),
+    };
+
+    ComponentStatus::new(name, lifecycle, health)
+}
```
elicbarbieri, line 672: Shit here it is again

packages/zaino-primitives/src/types/chain_work.rs
```diff
@@ -13,6 +17,36 @@ impl ChainWork {
         Self(bytes)
     }
 
+    /// Widen a `u128` work value. Lossless: every `u128` fits 256 bits.
+    ///
+    /// For the boundaries that still carry work as a `u128`. Each is a place
+    /// the protocol's full range is not yet representable, so each is a place
+    /// this call marks.
```
elicbarbieri, line 24: Remove these comments

packages/zaino-chain-head-service/src/service.rs
```diff
@@ -781,15 +812,19 @@ fn chain_head_block(
...
     })?;
 
-    let work = match parent_work {
-        Some(parent) => parent.checked_add(block_work).ok_or_else(|| {
+    // Widened where it is produced. `work_from_bits` still answers in a
+    // `u128` and refuses above it, so the protocol's full range stops here
+    // rather than at the accumulator; widening at the call site is what lets
+    // that producer change without touching the chain head.
+    let work = parent_work
+        .unwrap_or(AnchoredRelativeChainWork::ZERO)
```
elicbarbieri, line 820: I dont like this and could make passing in a none accidentally reset?  Not sure about better pattern.  Maybe is_genesis: bool parameter?  So we only accumulate from zero?  Probably not critical and can be deferred

packages/zaino-chain-head-service/src/service.rs
idky137: Ye this is just until we fully switch to the zaino-component  instead of zaino-status, which should be completed with the nfs and fs adaptor rewrites. I was trying to touch the implementations here as little as possible given these crates are being fully rewritten / replaced very soon. Happy to change this if there is a cleaner option though.

packages/zaino-chain-store-zainodb/src/adapter.rs
idky137: Similar to above.

packages/zaino-primitives/src/types/chain_work.rs
idky137: This file was removed with the merge from dev so this should be fixed.

packages/zaino-chain-head-service/src/service.rs
idky137: Fixed in [ad62e5d](https://github.com/zingolabs/zaino/pull/1572/commits/ad62e5dae6038a0bfbc125fc616a43ab849fc4fe).

`chain_head_block` no longer takes `Option<parent_work>`, it takes the block's work already computed, so there's no None to pass by accident.

AloeareV (commented):
I approve of this code! I have not pressed the approve button as CI is failing, and I'm currently unable to rerun the failing tests locally. If anyone else can verify a successful test run, that is all that remains for me to consider this ready to merge!

PR #1573, opened 2026-09-16
Feat/chain view pt4 chain view plus state

idky137:
Adds `zaino-chain`, the composition layer the previous three PRs were preparing for, and makes it reachable from `zaino-state`.

**The problem.** A chain read is answered by the finalised store below the reorg seam, the chain head's recent window above it, or the validator for what neither indexes. Until now every consumer assembled that itself, so every consumer had its own idea of where the seam was and what to do when the two halves disagreed.

`ChainViewComposer` owns it: routes each read by height against the store's watermark, merges where a query spans the seam, and reports `None` rather than a guess where a tier genuinely cannot answer. The case that matters is chainwork — the chain head measures from its own anchor, so an absolute value is unknowable until the store has built far enough to say what the anchor's own work was. The rebase is 256-bit throughout, so PR 3's widening reaches the composed read path rather than stopping at the chain head.

**`ServedCapabilities`** lets a deployment offer less than its tiers could answer, because what is served is a decision, not a consequence of what happens to be wired up. The builder is type-gated: a capability nothing implements is a method you cannot reach.

**`spawn_sync`** is the one part that writes. The store accepts blocks only at `tip + 1` and the chain head emits each block as it falls below the seam, so without an owner the two drift apart. There is no separate catch-up phase, and that is the design — a block from the middle of the chain handed to an empty store answers `FreezeGap`, and closing that gap *is* the initial sync. A cold start and a re-anchor after an outage take the same path, so it is exercised on every test run rather than being a startup branch nothing reaches twice.

**`zaino-state`** gains `WithChainViewSource`, and the mockchain source answers the remaining validator ports a view passes through. `ChainIndex` still reads through its existing path — this makes the view reachable, not yet load-bearing. Switching the reads over is deliberately separate, so that diff is about reads rather than also introducing the thing being read from.


packages/zaino-chain/src/composer/config.rs
elicbarbieri, line 1: This whole zaino-chain/src/composer/ directory seems like lots of config and whatnot that already exists.  Does this PR also remove the old implementation of this stuff?

elicbarbieri:
Is the packages/zaino-chain just copied over from nachos branch?  This is a big boy.  If it is, I think I already took a look at the implementation over there

elicbarbieri:
@idky137  the integration tests passed, so im fairly confident that RPC is still intact.  I am feeling confident on this.  Just want sign-off from nacho

nachog00:
> Is the packages/zaino-chain just copied over from nachos branch? This is a big boy. If it is, I think I already took a look at the implementation over there

I'm aware of the overlap. looking to pr slight changes on top of this to reduce conflicts with my work.

packages/zaino-chain/src/composer/config.rs
idky137: Is this relating to functionality in the ChainIndex?

If so then agreed, and no this has not yet all been removed. I was going to update zaino-state to be a thin wrapper over ChainView and Mempool in this PR but this is already large and thought it would be better to implement this separately.

Also the runtime PR will introduce new components designed to replace zaino-state and the ChainIndex so I was not sure what the best rout forward is here.

My choice would be to make zaino-state a thin wrapper and deprecate for a few release cycles before removing completely.

idky137:
> Is the packages/zaino-chain just copied over from nachos branch? This is a big boy. If it is, I think I already took a look at the implementation over there

Both of these branches came from a common design process. Much of the machinery in this branch comes from @nachog00's early exploratory work on the composition layer, with some additional changes, mainly for the following reasons:
- to better offer the lower level sub-components in zaino, as discrete usable modules, to library consumers (ChainView, ChainHead, ChainStore, Mempool).
- to ensure the sub-components in zaino are fully pluggable - a library user could use the composition layer (ChainView) with their own ChainStore / ChainHead implementations (Eg. Zallet has implemented chain fetch directly using zebra's ReadStateService, this design would enable them to use zaino's Mempool and chain composition without forcing the use of the default finalised state / non-finalised  state implementations, or make them pluggable within their codabase).

There are also several smaller changes related to code performance compared to the original implementation, I think one of the main areas where this implementation is different is in block streaming:

- https://github.com/zingolabs/zaino/blob/4904fe3c85d0fe7ccbd5e81a087c7502d654906f/packages/zaino-chain/src/ports.rs#L157

- https://github.com/zingolabs/zaino/blob/4904fe3c85d0fe7ccbd5e81a087c7502d654906f/packages/zaino-chain/src/composer/stream.rs#L1

Issue #1552, opened 2026-09-15
advance reports a partial chain-head advance as success when the source cannot serve a block below its reported tip


aphelionz:
**Version:** 0.10.0

In `packages/zaino-chain-head-service/src/service.rs`, the advance loop is:

```rust
while u32::from(graph.best_tip().height) < u32::from(chain_height) {
    let Some(block) = self.block_at_height(next_height(graph.best_tip().height)).await? else {
        break;
    };
```

A `None` breaks the loop and the function returns `Ok`, so a graph that stopped short of `chain_height` is indistinguishable to the caller from one that reached it.

The condition is not hypothetical: **a validator performing its own initial sync routinely advertises a best-chain height for which it cannot yet serve the intervening blocks.** We hit exactly this class against a checkpoint-syncing zebrad, where it crash-looped a container four times in one week before we added a startup grace on our side.

The concern is that a short graph published as fresh can be read as a complete snapshot up to the validator tip. If that is the intended contract, a caller has no way to tell the two apart.

**Suggested fix:** return a distinct error (an `InconsistentSource` variant alongside the existing `ReorgFailure`) so the writer retries instead of publishing the partial graph, or make the partial case explicit in the returned value so callers can decide.

I have not traced every `tick`/`spawn` path to confirm the downstream `Ready` consequence, so treat the severity assessment as ours and the code observation as the report. Found while vendoring 0.10.0 downstream.

Issue #1129, opened 2026-05-20
non_finalised_state: `block_is_already_in_nfs` returns false for blocks in nfs.blocks at heights above best_tip


zancas:
## Bug

The predicate
`NonfinalizedBlockCacheSnapshot::block_is_already_in_nfs`
([`non_finalised_state.rs:~256`](
https://github.com/zingolabs/zaino/blob/main/packages/zaino-state/src/chain_index/non_finalised_state.rs))
claims, per its docstring: *"Predicate: is this block already represented in
the NFS?"* The implementation:

```rust
fn block_is_already_in_nfs(&self, block: &zebra_chain::block::Block) -> bool {
    match block.coinbase_height() {
        Some(h) => Height::from(h) <= self.best_tip.height,
        None => false,
    }
}
```

…doesn't actually look at `self.blocks` — it's a height heuristic
against `best_tip.height`. For a block that **is** in `self.blocks` but
sits at a height *above* `best_tip.height` (the legitimate case for a
non-best-chain block at a stranded alt-branch tip, retained via
`handle_nfs_change_listener` → `add_nonbest_block`), the predicate
returns false even though the block is, in fact, in the NFS.

## Current call site

`sync()`'s for-loop (~line 419):

```rust
for block in window {
    if working_snapshot.block_is_already_in_nfs(&block) {
        continue;
    }
    let parent_hash = BlockHash::from(block.header.previous_block_hash);
    if parent_hash == working_snapshot.best_tip.hash {
        // Normal chain progression
    } else {
        self.handle_reorg(&mut working_snapshot, block.as_ref()).await?;
    }
    ...
}
```

`window` is the iter's pre-fetched best-chain run at heights
`finalized_height_floor + 1 ..= chain_height`. For the **window's** input
domain, the height heuristic happens to give the right answer:

- Block at H ≤ `best_tip.height` → already extended past here → skip (correct).
- Block at H > `best_tip.height` → not yet extended → process (correct).

The bug doesn't bite in production today because the window composition
guarantees this domain. **But the predicate's docstring promises a
broader contract** than the implementation delivers — "is this block
already represented in the NFS" is unambiguous about checking
`self.blocks`, not a height heuristic.

## Why this is worth tracking

The fragility is at the contract level, not the runtime level:

1. **Heightened false-negative window for non-best blocks.** Any code
   path that puts a block at height > `best_tip.height` into
   `nfs.blocks` (today, only `handle_nfs_change_listener` →
[...]

Issue #402, opened 2025-06-26
Indexer crash (likely due to a reorg to a shorter chain with more work.)


nuttycom:
I encountered the following error in `zallet`, with the wallet in a steady state at the chain tip (this error occurred on testnet.)

```
2025-06-26T19:40:28.195667Z  INFO steady_state: zallet::components::sync: New chain tip: 3462590 00b679adf6406f4d87bdf6ff172007e6539b4e57276b85bac08163c07f0e01b8
2025-06-26T19:40:28.196827Z  INFO zallet::commands::start: Wallet steady-state sync task exited wallet_sync_result=Err(Error(Context { kind: Sync, backtrace: Some(   0: <unknown>
   1: <unknown>
  ...
  58: <unknown>
), source: Some(Indexer(TonicStatusError(Status { code: Unknown, message: "Error: Failed to retrieve block from node. Server Error: JsonRpcConnector error: Error: Error: Error from node's rpc server: -5 - block height not in best chain", source: None }))) }))
2025-06-26T19:40:28.196930Z  INFO zallet::commands::start: Exiting Zallet because an ongoing task exited; asking other tasks to stop
2025-06-26T19:40:28.196947Z  INFO zallet::commands::start: All tasks have been asked to stop, waiting for remaining tasks to finish
2025-06-26T19:40:28.208327Z  INFO poll_transparent: zallet::components::sync: Fetching mined UTXOs
The application panicked (crashed).
Message:  A Tokio 1.x context was found, but it is being shutdown.
```

@str4d noted that it's odd that `TonicStatusError` is part of the error stack here; where is GRPC being used in Zaino's retrieval of chain data from Zebra?

arya2:
> where is GRPC being used in Zaino's retrieval of chain data from Zebra?

If it's using the latest release of Zebra, nowhere. If it's using the latest code on main, in `TrustedChainSync`.

The error message matches the `getblockheader` JSON RPC, which is also used by `getblock`.

idky137:
Hey is it possible that zebra is reorging to a shorter chain here ( with more chain work?)? 

That is the first thing that I can think of off the top of my head. I didn't think this would cause this crash but I will have a look through our code to double check. 

We have an update coming to the non-finalised state very soon that will fix this if it is what I believe.

[...]

PR #1576, opened 2026-09-16
chain-head-service: keep snapshot invariants in the type, behind a tested ChainGraph contract

### LINKED ISSUE #1482: chain-head-service: "graph contains its own tip" invariant is runtime-checked in three places
`MapBackedSnapshot` stores `best_tip: BlockRef` separately from `blocks: HashMap`
(`snapshot.rs`), so the invariant "the graph contains its own tip" is not
structural — it is re-asserted at runtime three times in `next_graph`: two
`ok_or(ReorgFailure("graph is missing its own tip"))` (`service.rs:411,457`) and
one `expect("a graph always retains at least its anchor")` (`service.rs:466`).

Because the tip is held alongside, not within, the block set, nothing at the
type level guarantees the referenced tip is present — the guarantee exists only
as repeated runtime checks that each site must remember to perform.

Severity: robustness / type hygiene.

Surfaced as a post-merge review finding on #1440:
https://github.com/zingolabs/zaino/pull/1440#issuecomment-5346380226

### LINKED ISSUE #1577: Indexer does not follow a pure chain-shortening rollback: keeps serving invalidated blocks until a competing branch is mined
## Summary

When the validator's best chain becomes **shorter** (a pure rollback with no competing branch yet), zainod does not roll back. It keeps reporting the old tip through `GetLatestBlock` and serving the invalidated blocks through `GetBlock` and `GetBlockRange` indefinitely: at least 120 s in our runs, and it would likely continue. It converges only after new blocks are mined on the replacement branch.

This looks distinct from the brief orphaned-branch window in #1386 (item 2). There a longer or equal competing branch is picked up after the poll loop catches up. Here nothing is picked up until a competing branch outgrows the stale tip.

## Environment

- zainod `0.10.0` (`zingodevops/zainod:0.10.0-no-tls@sha256:42c27615…`); also reproduced on `0.9.0` (`d27ec930`) and `0.6.0` (`17963672`)
- zebrad `6.2.3`, regtest (Z3 stack, `ZcashFoundation/z3@e84ce9fd`, NU5–NU6.3 active at height 2)
- `backend = "fetch"` (Z3 regtest `zaino.toml`), Docker on Linux (WSL2)

## Steps to reproduce

1. Let zainod sync to zebrad's tip `H`.
2. `invalidateblock` the block at height `H-1` on zebrad (JSON-RPC). zebrad's tip becomes `H-2`.
3. Poll `CompactTxStreamer/GetLatestBlock` on zainod.
4. Mine blocks on zebrad (`generate`).

## Observed (zainod 0.10.0)

```
before: zebra_tip=117 zaino_tip=117
invalidateblock height 116 10c1796c...2c57 -> {"result":null}
t=5s   zebra_tip=115 zaino_tip=117
t=30s  zebra_tip=115 zaino_tip=117
t=60s  zebra_tip=115 zaino_tip=117
t=120s zebra_tip=115 zaino_tip=117
after mining 3 on the new branch: zebra_tip=118 zaino_tip=118
```

With 0.6.0 and 0.9.0 we additionally verified that during the stale period `GetBlock` at the old tip height returns the **invalidated** block's hash. After the competing branch overtakes, zainod serves the new branch correctly (hashes match zebrad at every height), and `GetBlock` by an orphaned hash returns `InvalidArgument`.

## Expected

Within roughly one sync interval of zebrad's tip decreasing, zainod reports the new (lower) tip. `GetBlock`/`GetBlockRange` stop serving blocks that are no longer on the best chain, or at least signal that they are stale.

## Notes

- `NonFinalizedState::check_for_nonhigher_reorgs` in `packages/zaino-state/src/chain_index/non_finalised_state.rs` appears intended to handle non-higher reorgs. We did not establish why it does not fire here. One guess, **unverified**: the sync iteration only runs when the next height `best_tip + 1` becomes available, or the change listener is not triggered by a tip decrease.
- Why it matters: a light wallet or payment scanner treating the indexer's tip as current will keep counting confirmations for transactions in blocks the node has already discarded, with no signal from the server.
- Pure rollbacks are rare on mainnet, but they occur in regtest harnesses (`invalidateblock`) and can occur after operator or node-state recovery.

Happy to provide the full scripted reproduction if useful.

nachog00:
## Problem

The chain head's snapshot relied on its callers to keep it consistent. The type did not guarantee that it held its tip, that it was never empty, or that its canonical heights stopped at the tip. The service checked these at run time instead, with error arms and an `expect` for states that should be impossible. Where the callers got it wrong, the snapshot went wrong silently. With the crate's mock validator, `dev` does not follow a validator that rolls its tip back, and it accepts a block labelled with the wrong height into the best chain.

## Changes

- **The snapshot keeps its own invariants.** The tip is a field, so the graph cannot be empty. A block becomes canonical only as a checked child of the tip. Branch switches first rewind to the fork point.
- **The validator picks the tip.** Retained work is still compared with the tip, and a disagreement is logged. The previous rule, where the heaviest retained block wins, stays available as a crate-private policy.
- **The moves form a tested contract.** A crate-private `ChainGraph` trait describes how the service builds a graph. One test suite, with a model-based property test, runs against every implementation. This makes an alternative storage, such as the `imbl` approach in #1556, a matter of implementing the trait and passing the suite. It also gives one interface to benchmark both.

The aim is correctness. The snapshot keeps its `std::collections::HashMap` storage. Persistent collections are a separate performance question.

Closes #1482. Supersedes the accessor-only step in #1488.

## Behaviour on `dev` that the new tests cover

With the crate's mock validator, these tests fail on `dev` (5ad744cd) and pass here:

- **Rollback** (validator tip 4 → 3): the published tip stays at 4 across repeated advances. The abandoned block has more work, so the heaviest-block step reorgs back onto it.
- **Reorg onto a shorter branch** (a different block at height 3): the published tip stays on the old block at height 4.
- **Block at the wrong height** (the block served for height 5 is labelled 9): the advance succeeds, and height 9 becomes canonical.

These are mock scenarios. How often a real validator produces them (for example after `invalidateblock`) is not measured here.

### Where each guarantee comes from

- **compiler**: code that breaks it does not compile.
- **privacy**: only the private methods of the graph implementation can break it.
- **test**: shown by the named tests, and by the contract suite over random move sequences.
- **caller**: relies on how the service calls the graph.

#### Moves

| Operation | `dev` | This branch | Assurance |
|---|---|---|---|
| Create / re-anchor | one block in the map, `best_tip` stored beside it | the block is the `tip` field | compiler (the only constructor takes a block) |
| Extend | the service checks the parent hash; `add_block_new_chaintip` checks nothing | `extend` checks parent and height, and refuses otherwise | privacy; `extending_with_a_block_*_is_refused`, `a_block_labelled_with_the_wrong_height_fails_the_advance` |
| Higher reorg | recursive walk to a canonical parent; index overwritten upwards | same walk; `rewind_to` is the fork-point test, blocks go through `extend` | privacy; `a_higher_reorg_retains_the_displaced_branch` |
| Same-height reorg | caught by `check_for_nonhigher_reorgs` | same path: rewind, then extend | privacy; `a_same_height_reorg_is_caught_without_a_higher_block` |
| Reorg onto a shorter branch | see above | the tip follows the validator; heights above it are dropped | privacy; `a_lower_reorg_abandons_the_heights_above_the_new_tip` |
| Rollback | see above | the tip follows the validator; the abandoned block stays as a `valid-fork` tip | privacy; `a_rollback_lowers_the_tip`, `a_rollback_abandons_the_heights_above_the_new_tip` |
| Reorg back onto a retained branch | walk through retained blocks | same; each block moves out of `others` as it becomes the tip | privacy; `a_reorg_back_onto_a_retained_branch_restores_it` |
| Heaviest-work check | `max_by_key(..).expect(..)`; the tip's work via an error arm; switches to the heavier block | `heaviest_block()` folds from the tip; the default policy logs | compiler (the fold always has an answer); `the_tip_wins_an_equal_work_tie`, `heaviest_retained_selection_overrides_a_rollback`; the work values are caller |
| Trim | keeps the highest indexed block, with an arm for an empty map | trims `others` only; the index is cut at the same floor | compiler + privacy; `trimming_above_the_tip_keeps_the_tip`, `the_window_stays_bounded_as_the_chain_grows` |

#### Reads

| Read | `dev` | This branch | Assurance |
|---|---|---|---|
| `best_tip` | a stored `BlockRef` | read from the tip block | compiler |
| `tip_block` (crate-private) | — | returns `&ChainHeadBlock` | compiler |
| `block_by_hash` | map lookup | tip, then map | privacy |
| `best_block_by_height`, `is_on_best_chain`, `best_chain`, `best_chain_blocks` | index lookups; in the shapes above, the index can hold heights above the tip | the index runs without gaps up to the tip | privacy; reorg and rollback tests |
| `chain_tips` | hash lookups through `filter_map` | filters the retained blocks directly | compiler for the tip; `a_rollback_abandons_the_heights_above_the_new_tip` |
| `find_fork_point`, `transaction_locations`, `outpoint_spenders`, `epoch` | — | unchanged in behaviour | existing tests |

### Code layout

- `graph.rs`: the `ChainGraph` trait, its contract, and the `NotChildOfTip` / `NotOnBestChain` refusals.
- `snapshot.rs`: `MapBackedSnapshot`, which implements it.
- `graph/tests/`: the contract suite (single-move checks, a generic invariant check through the read traits, and the property test), declared per implementation with `graph_contract!`.
- `snapshot/tests/invariants.rs`: the map-specific representation checks.

## Not covered here

- **Work values.** The service computes `ChainHeadBlock.work`, and the graph does not check it. Moving the accumulation into `extend` overlaps with the work-type changes in #1572.
- **A total tip accessor on the public port.** `zaino-state`'s `get_transaction_status` still maps an empty `best_chain()` to `database_hole("no blocks")`. Moving `tip_block` onto `ChainHeadSnapshot` would remove that arm.
- **Side chains.** A side branch is retained only if the chain head lived through the reorg that displaced it, because the sources serve best-chain blocks only (ZcashFoundation/zebra#9541).
- **Benchmarks.** None are included. The contract makes a generic benchmark straightforward.

## Overlap

- **#1556** rewrites the snapshot on `imbl`, as a set of `NonEmptyVec` chains. That shape can implement `ChainGraph` and join the suite. Holding the best chain as a field would keep its set of chains non-empty.
- **#1572** edits the same files (window-floor naming, work anchor).

## Test plan

- `cargo test -p zaino-chain-head-service`: 36 passed (the property test also with `PROPTEST_CASES=4096`).
- `cargo clippy --locked -p zaino-chain-head-service --all-targets --all-features -- -D warnings`: clean.
- `RUSTDOCFLAGS="-D warnings" cargo doc --locked -p zaino-chain-head-service --no-deps --all-features`: clean.
- `cargo fmt --all -- --check`, `check-code-duplication`, `makers lint-tools`: clean.
- `relman changeset check --base origin/dev`: ok.
- Equivalents of the rollback, shorter-branch and wrong-height tests were run on `dev`, where they fail as described.
- The integration shard `zaino shard-5` fails in `wallet.py` ("wallet transparent balance … implies N coinbases"). `dev` (5ef7675a, 190eb8e7) and other open PRs fail the same way.


nachog00:
Added in b81a6deb: an invariant checker and a property test for `MapBackedSnapshot`.

- `check_invariants` (test-only, in `snapshot/tests/invariants.rs`) reads the private fields and checks the invariants listed on the struct: the tip is not in `others`, map keys match block hashes, every canonical hash names a retained block at that height, the canonical heights end at the tip, and they run without gaps with each block's parent one height below.
- `snapshot/tests/properties.rs` generates random sequences of up to 63 moves: new extensions, re-extensions of retained blocks, refused extensions, rewinds, refused rewinds and trims. It compares each step with a list-and-set model: tip, best chain, retained set, heaviest block, active chain tip. Refused moves must leave the snapshot unchanged.

Mutation check. Each fault below was reintroduced alone, and the property test failed on it:

| Fault | Failing check |
|---|---|
| rewind keeps heights above the new tip | highest canonical height is not the tip's |
| `extend` without its attachment check | refused extension returned `Ok` |
| tip swap leaves the new tip in `others` | `others` holds the tip |
| trim drops the tip's height entry | highest canonical height is not the tip's |

The third fault was not caught until the generator gained the re-extension move. In the service, that path is a reorg back onto a retained branch.

nachog00:
**Why.** The service built each snapshot by calling `MapBackedSnapshot`'s own methods. Nothing described these moves as a contract. So a different way to store the graph, for example the `imbl`-backed one explored in #1556, had no interface to implement and no tests to prove it correct.

**What changed.** The moves are now a crate-private trait, `ChainGraph`, and the service builds snapshots only through it. The tests for the moves are written once, for any `ChainGraph`, and run against each implementation. Behaviour does not change.

**Details** (7ef02c8c, a8eb1231, 0bb09407)

- `graph.rs` defines `ChainGraph: ChainHeadSnapshot + ChainHeadTransactionService + Clone`.
  - Moves: `from_initial_block`, `extend`, `rewind_to`, `remove_finalized_blocks`, `stamp_generation`.
  - Writer-side reads: `tip_block`, `heaviest_block`.
  - The trait documentation lists the invariants every implementation keeps.
- `MapBackedSnapshot` implements `ChainGraph`. The service imports only the trait.
- The tests are in `graph/tests/`. The next comment describes the suite.

nachog00:
**Why.** A graph implementation is correct when it keeps the chain head's invariants, whatever it uses to store blocks. A second implementation must pass the same checks as the first before its performance is worth comparing, and it should not need its own copy of the tests.

**What.** The graph tests are now a contract suite. Every `ChainGraph` implementation runs the same tests, and a new implementation joins in three small steps. The suite is ready for an `imbl`-backed implementation, and the trait gives one interface to benchmark both.

### How the suite works

All three parts are generic over the graph type (`graph/tests/`):

| Part | What it does |
|---|---|
| `checks.rs` | Example tests, one move each: refused extensions, trimming above the tip, rewinds, refused rewinds, the heaviest block, and the equal-work tie. |
| `properties.rs` | A proptest. It generates random sequences of up to 63 moves: extend with a new block, extend with a retained block (a reorg back onto a kept branch), refused extends, rewinds, refused rewinds, and trims. After every move it compares the graph with a simple model: tip, best chain, retained blocks, block count, heaviest block, and the active chain tip. A refused move must leave the graph unchanged. |
| `contract.rs` | The `ChainGraph` invariants, checked through the read traits only: the tip is retained and canonical; the best chain ends at the tip, has no gaps, and follows parent links; every best-chain block is retained and canonical at its height; the heaviest block is retained. The property test runs it after every move. |

Each implementation also provides `InspectableGraph`, a test-only trait. It gives the suite what the read traits cannot: the set of retained hashes, the block count, and a check of the implementation's own internal invariants. The property test runs that check after every move too. For `MapBackedSnapshot` it is in `snapshot/tests/invariants.rs`.

`graph_contract!(name, Type)` declares the tests for one implementation. It is a macro because a `#[test]` function cannot be generic.

### Evidence

Four faults were reintroduced one at a time: stale heights after a rewind, an `extend` without its check, a new tip left in `others`, and a trim that drops the tip's height entry. The suite caught each one. It still caught each one with the map-specific check disabled, so the generic parts alone are enough to hold another implementation to the contract.

### Adding the `imbl` implementation

1. `impl ChainGraph for ImblGraph`
2. `impl InspectableGraph for ImblGraph`, in its own test module
3. `graph_contract!(imbl_graph, ImblGraph);`

### Benchmarking

Both implementations have the same moves, so a benchmark can be written once over `ChainGraph` and run for each. The interesting costs are the per-publish copy (`Clone`), the moves, and the read queries. They should be measured at the default window: `MAX_NONFINALISED_DEPTH` (1,001) plus a retention margin of 10, about 1,011 blocks. This PR does not include a benchmark.

nachog00:
Two draft PRs now stack on this branch.

**#1578 — measure what one publication costs.** The chain head builds a snapshot on every tick, and nothing measured that. It adds two criterion benchmarks over the crate's public API: one for time, one for allocation counts and bytes held. No production code.

**#1579 — hold the snapshot in persistent collections.** It changes only the snapshot's fields, to `imbl` collections with blocks behind `Arc`. No test changed, so the benchmark compares the two storages and nothing else.

Measured with the benchmarks in #1578: a publication is 84% faster and allocates 90% less, and eight held snapshots cost 0.72 MiB instead of 84.6 MiB. Reads cost 13% to 40% more, on a scale of tens of microseconds. Full figures, and the drift floor they are read against, are in each PR.

Both are drafts, and #1579 also raises a licence question: `imbl` is MPL-2.0+ and this workspace is Apache-2.0.

nachog00:
Pushed `4bbc12d2` onto this branch (rebased onto the dev merge): two contract tests pinning the window's edge.

`rewinding_to_the_oldest_retained_block_succeeds` and `rewinding_to_a_trimmed_block_is_refused`. The oldest retained block is where the deepest reorg the window is sized for rejoins the best chain, so a rewind has to reach exactly it and refuse anything below. Both hold for any implementation, so they go in the contract suite rather than against `MapBackedSnapshot`.

They pass as-is — the graph's half of that edge was already correct. They are here so it stays correct, and so the imbl graph answers for it too. Related: the same edge on the service side is #1584, where it was not correct.

Also on this branch from earlier today: `18b45d57` and `66ba2572`, a documentation pass (present tense, and one home per claim). 38 tests pass, clippy and rustdoc clean.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
```diff
@@ -0,0 +1,288 @@
...
+    /// is the tip but its height is not the next one.
+    ExtendDetached {
+        unknown_parent: bool,
+        height_choice: u8,
+    },
+    /// To the canonical block `depth` below the tip, wrapping at the bottom.
+    Rewind { depth: u8 },
+    /// To a retained block off the best chain, or to one never retained.
+    RewindOffChain { pick: u8 },
+    /// Trims at `tip + 4 - shift`: from four above the tip to fifteen below.
```
AloeareV: Why 4 and 15? Are these numbers arbitrary?

packages/zaino-chain-head-service/src/graph/tests/properties.rs
```diff
@@ -0,0 +1,288 @@
...
+            let child = Entry {
+                height: tip.height + 1,
+                parent: tip_id,
+                work: tip.work + u128::from(work),
+            };
+            let block = chain_head_block(child.height, id, tip_id, child.work);
+            prop_assert_eq!(graph.extend(block), Ok(()));
+            model.chain.push(id);
+            model.retained.insert(id, child);
+        }
```
AloeareV: Marker comment: reviewed everything above this line

packages/zaino-chain-head-service/src/service.rs
```diff
@@ -803,6 +819,19 @@ fn chain_head_block(
     })
 }
 
+/// Extends `graph` with a block the source served as the tip's child.
+///
+/// A refusal means the source's answer does not attach where it was asked
+/// for, which makes it inconsistent source data.
+fn extend(
```
AloeareV, line 894: Why does this function exist? Why not just call `graph.extend()`?

AloeareV (changes requested):
Approve of most of this! I find the proptest stuff very confusing: These `Move` variants don't seem to map one-to-one to anything, and I'm finding it hard to follow what each one represents or how the changes to the `Model` represent each move.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
nachog00: They were arbitrary as written. Both bounds are now named, with the reason each one exists, in 794253ca.

The strategy picks a trim floor relative to the tip, and two cases matter:

- A floor **above** the tip asks the graph to drop every block it holds. That is where the tip's exemption from trimming is decided, so the strategy must generate it. `TRIM_ABOVE_TIP = 4` gives a few variants of that case instead of exactly one.
- A floor **below** the tip covers everything from "removes nothing" to "removes all but the tip". `TRIM_BELOW_TIP = 15` is deeper than a generated sequence usually builds, so the lower end is reached.

Neither number is the window the service retains. That window is a deployment fact (`max_depth` plus the retention margin) and the graph does not enforce it, so the property test does not measure it.

packages/zaino-chain-head-service/src/service.rs
nachog00: It converts the error. `ChainGraph::extend` returns `Result<(), NotChildOfTip>`, and both call sites (`service.rs:457`, `service.rs:532`) need `ChainHeadAdvanceError`, so `?` cannot be used on it directly.

A `From<NotChildOfTip> for ChainHeadAdvanceError` impl does compile, so that is a real alternative. It was not used because the conversion is not context-free: a refusal means "the validator served a block that does not attach where it was asked for" only on the advance path. Anywhere else, a refusal is a programming error, not inconsistent source data. Keeping the conversion explicit stops `?` from applying that reading in places where it does not hold, and gives it one place to be stated.

If the indirection reads as worse than the duplication, the `map_err` inlines at both sites with no other change.

nachog00:
## What `ExtendRetained` is for

A block that extends the tip is either new or already retained, and the two take different paths through `MapBackedSnapshot`. The property test therefore has two extension moves:

- `Extend { work }` mints a fresh block one height above the tip. It is not in the map, so `replace_tip` only has to move the old tip into `others`.
- `ExtendRetained { pick }` extends with a block the graph **already holds** off the best chain. This is the state a reorg leaves behind: it rewinds to the fork point, and later returns to the branch it displaced. Here `replace_tip` must also remove the new tip from `others`, or the same block is held twice, once as the tip and once as a competing block.

`ExtendRetained` is a no-op when the tip has no retained child at the next height. The sequence that creates one is `Extend` followed by `Rewind`.

### Evidence that it bites

Delete `self.others.remove(&block.hash());` from `replace_tip` (`snapshot.rs:87`) and the suite fails:

```
Test failed: move 2 ExtendRetained { pick: 0 }: Case failed:
others holds the tip 0000000000000000000000000000000000000000000000000000000000000002.
minimal failing input: moves = [Extend { .. }, Rewind { .. }, ExtendRetained { pick: 0 }]
```

`tests::a_rollback_abandons_the_heights_above_the_new_tip` fails too, because `chain_tips` then reports the duplicated block twice.

With the same line deleted but `ExtendRetained` removed from the strategy, the property test passes 4096 cases. The move is what reaches the state.

## Issues these tests assert

Each entry names the test that fails without the fix.

| Issue | Covered | Test |
|---|---|---|
| #1482 — tip invariant runtime-checked in three places | yes, closed by this PR | the tip is a field, so there is nothing left to check; `heaviest_block` folds from it and needs no `expect` |
| #1577 — pure rollback not followed | yes | `tests::a_rollback_lowers_the_tip`, `tests::a_rollback_abandons_the_heights_above_the_new_tip` |
| #402 — crash on a reorg to a shorter chain with more work (closed) | yes, as a regression guard | `tests::a_lower_reorg_abandons_the_heights_above_the_new_tip` |
| #1129 — `block_is_already_in_nfs` false for blocks above `best_tip` | bug class gone | `non_finalised_state.rs` no longer exists on `dev`; the height index here stops at the tip, which the same two rollback tests assert |
| #1552 — partial advance reported as success | no | the `break` that returns `Ok` is untouched by this PR |

### #1577 on current `dev`

Re-checked against `dev` at 372086de with the crate's mock validator. The validator reports tip height 3 after a rollback from 4. The published tip stays at `Height(4)` for three ticks. `a_rollback_lowers_the_tip` is the same shape at unit level and passes on this branch.

### Mutation evidence for the rollback fix

Delete the two lines that cut the height index in `rewind_to` (`snapshot.rs:166-167`):

```rust
self.heights_to_hashes
    .retain(|height, _hash| *height <= block.height);
```

Six tests fail, which is the `dev` behaviour returning: the graph follows the new tip but keeps serving canonical heights above it.

```
graph::tests::map_backed_snapshot::random_moves_keep_the_graph_consistent
graph::tests::map_backed_snapshot::rewinding_keeps_the_old_tip_as_a_competing_block
graph::tests::map_backed_snapshot::rewinding_to_a_competing_block_is_refused
graph::tests::map_backed_snapshot::rewinding_to_the_oldest_retained_block_succeeds
tests::a_lower_reorg_abandons_the_heights_above_the_new_tip
tests::a_rollback_abandons_the_heights_above_the_new_tip
```

packages/zaino-chain-head-service/src/graph/tests/properties.rs
zancas: This conversation shows as outdated.  Can it be resolved?

AloeareV: > This conversation shows as outdated. Can it be resolved?

'outdated' just means that the code that this points at has been changed. In this specific case, I don't think that modification makes this conversation no-longer-applicable. The resolution criteria I would go with for this conversation is "the reviewer believes this has been resolved to their satisfaction", which as I am the reviewer is probably about to happen.

AloeareV: > They were arbitrary as written. Both bounds are now named, with the reason each one exists, in [794253c](https://github.com/zingolabs/zaino/commit/794253ca7c79205f9d2090829edd5049eb90fdc2).

The new named constants do make this clearer. Nice!

packages/zaino-chain-head-service/src/service.rs
AloeareV: Imo the indirection takes a lot of mental overhead. I have to navigate to the call site, read the function, and verify to myself it's an error-wrap.

Two separate calls to the same `map_err` function I think would read better? Alternatively, if you made it clear in the documentation of the `extend` function that it's a wrapper to map an error, that would save reading time. Right now the documentation explains what `extend` does with no indication that it's a thin wrapper, and then I'm surprised when I read the function and it just calls `graph.extend`.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
```diff
@@ -0,0 +1,304 @@
...
+/// margin — and is not what this is measuring.
+const TRIM_BELOW_TIP: u8 = 15;
+
+#[derive(Debug, Clone)]
+pub(super) enum Move {
+    /// A child of the tip carrying `work` more than the tip.
+    Extend { work: u8 },
+    /// A retained child of the tip off the best chain, as when a reorg returns
+    /// to a branch it displaced. A no-op when there is none.
+    ExtendRetained { pick: u8 },
```
AloeareV, line 47: This comment says 'a no-op when there is none' which implies to me that it's not a no-op when there is some. Also, it's not clear what is none. 

Actually, this entire comment doesn't scan to me.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
```diff
@@ -0,0 +1,304 @@
...
+    /// A block that does not attach: its parent is not retained, or its parent
+    /// is the tip but its height is not the next one.
+    ExtendDetached {
```
AloeareV, line 50: This is also unclear. Is this a test of invalid input? A block without a retained parent is something I can imagine us hitting in the wild, and something we don't want to add, so it makes sense to me that this case is covered here. A block with a parent that's at the wrong height is just straight bad input, though, isn't it? Is that a different category of thing?

It's also confusing to me that this is a `Move` as it doesn't actually cause any movement.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
```diff
@@ -0,0 +1,304 @@
...
+        unknown_parent: bool,
+        height_choice: u8,
+    },
+    /// To the canonical block `depth` below the tip, wrapping at the bottom.
+    Rewind { depth: u8 },
+    /// To a retained block off the best chain, or to one never retained.
```
AloeareV, line 56: Changing the graph so that the new tip is one that previously wasn't on the best chain and changing the tip to a block not previously known strike me as entirely different operations...and it's not clear if there's supposed to be a one-to-one map between a `Move` and a modification to a graph. 

The former is a valid multi-step move: rewind to fork point, add blocks of new chain. The latter depends on the position of the new block. Neither one would I call a 'rewind'.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
```diff
@@ -0,0 +1,304 @@
...
+                .collect();
+            if children.is_empty() {
+                return Ok(());
+            }
+            children.sort_unstable();
+            let id = children[usize::from(pick) % children.len()];
+            let child = model.entry(id);
+            let block = chain_head_block(child.height, id, child.parent, child.work);
+            prop_assert_eq!(graph.extend(block), Ok(()));
+            model.chain.push(id);
```
AloeareV, line 178: If ExtendRetained is a no-op, why is this match arm so long? 

I admittedly have not gotten a chance to fully follow the `Model` code. Maybe this would make sense to me with a better general understanding.

AloeareV (changes requested):
This is not a comprehensive list as I have to run, but I've tried to enumerate my confusion about the `Move` and `Model` types a bit here.

packages/zaino-chain-head-service/src/service.rs
zancas: The wrapper now documents itself as a thin error-mapping wrapper over `ChainGraph::extend`, in c15232ece. It stays a function because it has three call sites, the advance loop and both steps of the reorg walk, all needing the same `NotChildOfTip` to `InconsistentSource` conversion.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
zancas: Reworded in 493aed820. `ExtendRetained` is now `Accepted::ExtendRetained`, and its doc says what the sentence meant: an extension by a retained child of the tip off the best chain, as when a reorg returns to a branch it displaced, which does nothing when the tip has no such child. The move's kind (one the graph must take) is in its type.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
zancas: Both readings are right, and 493aed820 makes them visible in the type. `Move` is now `Accept(Accepted)` or `Refuse(Refused)`; this one is `Refused::Detached`, a move the graph must reject while changing nothing, and `apply` checks the unchanged-graph invariant for every `Refuse` in one place. It covers the two shapes of a block that does not attach (an unretained parent, or the tip as parent with the wrong height) because the graph refuses both the same way, with `NotChildOfTip`.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
zancas: This move never switches branches; it asks `rewind_to` for a target that is not on the best chain, which the graph must refuse with `NotOnBestChain`. The target is either a retained off-chain block or one never retained, and both must be refused the same way, which is why they share a variant. In 493aed820 it is `Refused::OffChainRewind`, so its expected outcome is in its type, and the two-step branch switch you describe is what the service does through `Rewind` followed by `Extend`.

packages/zaino-chain-head-service/src/graph/tests/properties.rs
zancas: It is a move the graph must accept, not a no-op; it does nothing only when the tip has no retained off-chain child. The arm is long because it has to find such a child in the model (a retained block whose parent is the tip at the next height) before asserting that the graph accepts it and moving the model's chain. 493aed820 splits `apply` into `accept` and `refuse`, so this arm now sits with the other accepted moves.

zancas (approved):

PR #1584, opened 2026-09-18
chain-head-service: walk reorgs iteratively, and trim after the handoff

### LINKED ISSUE #1551: handle_reorg's recursion_count is u8 but the depth guard it feeds is u32 (default 1011)
**Version:** 0.10.0

In `packages/zaino-chain-head-service/src/service.rs`, `handle_reorg` takes `recursion_count: u8` (line 488) and guards on it as:

```rust
if u32::from(recursion_count) > self.max_retained_depth() {
```

`max_retained_depth()` is `config.max_depth() + RETENTION_MARGIN`. With the defaults, `MAX_BLOCK_REORG_HEIGHT = 1000`, so `MAX_NONFINALISED_DEPTH = 1001`, plus `RETENTION_MARGIN = 10`, giving **1011**.

The guard therefore cannot become true before `recursion_count + 1` overflows `u8` at 255, on both recursive calls (lines 502 and 512). In debug builds a reorg deeper than 255 panics on the increment; in release it wraps, and because the wrapped value stays far below 1011 the `max_retained_depth` bound no longer terminates the recursion at all.

A reorg deeper than 255 blocks does not occur on mainnet, so the practical severity is low. But the configured bound is 1011 and the counter cannot represent it, so the guard does not do what its constant says.

**Suggested fix:** make the counter `u32` to match `max_retained_depth()`'s type, and use a checked or saturating increment so the bound is what terminates the recursion rather than the counter's width.

Found while vendoring 0.10.0 into a downstream tree. Happy to send a PR if you'd like it in that shape.

nachog00:
Both reorg walks in `chain-head-service` recurse once per block, and past a point that either overflows the thread's stack or overflows the depth counter, depending on which limit the deployment hits first. Neither is recoverable: the stack overflow aborts the process, and the counter overflow panics in debug or, in release, wraps to a value far below the bound so the guard stops terminating the recursion at all.

`Box::pin` keeps the future type finite but not the call chain — each level's `poll` frame stays live while the levels beneath it run — so the walk costs stack in proportion to its depth.

The configured window is sized for a reorg of `MAX_NONFINALISED_DEPTH` (1001) blocks. Neither walk can follow one that deep.

## What this changes

**Reorg handling walks iteratively.** `handle_reorg` collects the branch down to the fork point in a `Vec` and lays it back down from there, so depth costs heap rather than stack. The pending list holds one `BranchBlock` per block — either one the graph still retains on a losing branch, or one fetched from the validator — so the two kinds travel in one list.

**Reorg detection walks iteratively.** `check_for_nonhigher_reorgs` steps down a height at a time in a loop rather than by self-recursion. Its `Option<Height>` parameter, which only the recursion was meant to set, is gone. This walk has no counter at all, so widening one would not have reached it.

**Trimming moves after the handoff.** The freeze handoff reads the blocks it emits out of the graph, so a block has to still be retained to be handed off. Trimming ran first, which bounded one iteration's handoff by the gap between the consensus seam and the retention floor: an iteration advancing the tip by more than `RETENTION_MARGIN + 1` blocks settled blocks it never handed off, and reported nothing — `Lagged` covers a slow consumer, not a producer that dropped them before sending. Trimming now runs in `publish_snapshot` after the handoff has read what it needs. The order is the guarantee, not the size of that gap.

## Measurements

Walk depth reached before the stack overflows, with the counter widened to `u32` so the stack is the only limit:

| thread stack | depth reached |
| --- | --- |
| 2 MiB (tokio worker default) | ~150 |
| 8 MiB (main thread) | between 500 and 800 |

Which limit binds first therefore depends on the stack and on `max_retained_depth`:

- On a 2 MiB worker the walk dies around 150 levels, below the counter's 255, so the abort is what shows up at the default depth.
- On an 8 MiB stack with a window above 255, the counter overflows first: a fork ~400 deep inside a 500-block window gives `attempt to add with overflow` in debug, and wraps in release.

Removing the recursion covers both: no counter to overflow, and depth costs heap.

## Test plan

`tests/reorg_depth.rs` is added first, in its own commit, so the failures are visible before the fix. Two of its tests abort the test binary on that commit rather than failing, and are marked as such; the fix commit removes the marks.

- fork inside the window, resolved
- fork at exactly the window floor, resolved — the fencepost `MAX_NONFINALISED_DEPTH = MAX_BLOCK_REORG_HEIGHT + 1` exists to buy this, and nothing covered it
- fork 200 blocks deep inside a 300-block window, resolved
- fork one block below the window floor, terminates
- rollback below the window floor, terminates
- rollback at `MAX_NONFINALISED_DEPTH`, terminates
- published tip never falls below a block already handed off to the freeze stream
- one tick hands off every block it settles, at and beyond the retention margin

Verified with `RUST_MIN_STACK=2097152` to match a tokio worker thread rather than the larger default: 31 passing, none ignored. `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `RUSTDOCFLAGS="-D warnings" cargo doc` are clean.

Closes #1551


nachog00:
Added `f0a644ed`: `RETENTION_MARGIN`'s doc now says what the constant is actually for.

It claimed three jobs. Two are not its own:

- *"never cuts inside the reorg-possible range"* — `MAX_NONFINALISED_DEPTH` is `MAX_BLOCK_REORG_HEIGHT + 1`, the reorg bound plus the fork point the branch rejoins at, so the window already covers the deepest reorg it is sized for. A fork at exactly `max_depth` resolves without the margin, which the new `a_fork_at_the_window_floor_is_resolved` pins.
- bounding the handoff — that is guaranteed by trimming after the handoff reads its blocks, not by the size of the gap below the seam.

What remains: a block stays readable in the graph for a while after it has been handed off as final, and the bound on the reorg ancestry walk.

AloeareV (approved):

PR #1600, opened 2026-09-21
CI: pre-commit lints, cached nix devshell, integration in ci.yml

elicbarbieri:
## Summary

Replaces the custom lint tooling with standard tools, runs every CI job inside the pinned nix devshell, and folds the live-test workflow into `ci.yml`.

### Lints → `.pre-commit-config.yaml`
- `cargo-make`, `Makefile.toml` and `.githooks/` are removed; lints are pre-commit hooks (`language: system`, tools from the devshell).
- Commit stage: shellcheck, fmt (root + live-tests), ast-grep, openssl-apt pygrep, cargo-deny, code-dupes. Push stage: clippy (root + live-tests).
- CI lint job = `pre-commit run --all-files --hook-stage pre-push` (runs every hook).
- `tools/` is deleted entirely. The workbench lint binaries map to:
  - `check-crypto-provider` → `deny.toml`: ring `wrappers` + exact rustls feature set
  - `check-boundary-conversions` → `.ast-grep/boundary-conversions.yml` (structural; also catches multi-line impls the regex missed)
  - `check-no-openssl-apt` → pygrep hook
  - `check-code-duplication` → `code-dupes check` + `dupes.toml`
  - `check-published-versions` → dropped (made redundant by the planned single-version release; see below)

### CI runs in the nix devshell
- `.github/actions/devshell`: install Nix + restore `/nix/store` via `nix-community/cache-nix-action`. Keyed on `flake.nix`/`flake.lock`/`rust-toolchain.toml`/`nix/**`; saved from `dev` only (a PR-saved cache is invisible to other PRs).
- `defaults.run.shell: nix develop -c bash …`, so every step uses flake.lock-pinned tools; drops `taiki-e/install-action` and `azure/setup-kubectl`.
- Rust caches (`Swatinem/rust-cache`) save from `dev` only, which stops per-PR ~2.6 GB caches from pushing `dev`'s out of the 10 GB repo limit.
- Devshell 4.2 → 3.5 GiB: `rust-toolchain.toml` `profile = "minimal"` (−636 MiB rust-docs), cmake dropped (no crate uses it). Most of what remains is libclang (1.35 GiB), which librocksdb-sys bindgen needs via `zebra-state`; it goes away with the fetch-only migration.

### Integration tests
- `live-tests.yaml` is folded into `ci.yml` as the `integration` job; the ztest CLI is installed at the version `live-tests/Cargo.lock` pins.
- The live-tests fmt/clippy check (the old "Compile check") now runs as hooks, so fork PRs get it too.

### Docs
- README gains a **Development** section (Tools: nix or brew; Lints; Tests); CONTRIBUTING and `docs/testing.md` point to it.

## Known gaps / follow-ups
- **Version-reuse guard removed early:** until the single-version migration lands, a crate changed without a version bump isn't caught before release.
- **"Each feature compiles" still fails on dev:** unused import in `zaino-mempool/src/ports.rs:23` under `--no-default-features`. Separate PR.
- **Integration never runs on `dev`:** no post-merge signal, and the ztest CLI compile (~6.5 min) can't be cached across PRs. It could also be the `rc-gate` check the RC gate waits for.
- **`release-pr-body.yml` is broken:** it references the removed local `setup-relman` action. Fold into the single-version release migration.

## Verification
- `nix develop -c pre-commit run --all-files --hook-stage pre-push`: all 9 hooks pass locally.
- Each replacement lint was tested against a deliberate violation (deny feature/wrapper drift, 5 ast-grep impl shapes, duplicate fn in `packages/` vs `live-tests/`, `libssl-dev` apt line).
- `cargo check --workspace --all-targets` passes in a clean devshell with no cmake on the PATH.
- Not yet run: this PR's own CI (first run on the cached devshell), `nix build .#zainod`.


zancas (approved):
UtAck

Issue #1587, opened 2026-09-18
zaino-chain-store-zainodb: make PersistentBlockContext::from_business infallible by construction


nachog00:
After #1560, `BlockContext::chainwork` is `Option<AbsoluteChainWork>` and `PersistentBlockContext::from_business` returns an error for a block whose chain work is unknown:

```rust
pub(super) fn from_business(context: &BlockContext) -> io::Result<Self> {
    let chainwork = context.chainwork.ok_or_else(|| /* refusing to store it */)?;
```

The refusal is correct — a block whose chain work is not yet computed has no stored form — but it is a runtime check for something the types can rule out. A context type carrying a non-optional chain work, constructible only by the indexing path that computes it, makes the encode direction total again.

Raised in review of #1560.

PR #1608, opened 2026-09-23
fix(chain-store-zainodb)!: make the block-context encoder total by construction

zancas:
Implements #1587 on top of #1560. It resolves the two review threads on `types/db/block.rs` that asked for `from_business` to be infallible at compile time.

## What changes

`BlockContext`, `BlockHeaderData` and `IndexedBlock` take a `Work` type parameter. `Work` is `AbsoluteChainWork` where the total is known. `Work` is `Option<AbsoluteChainWork>`, the default, where the producer may not know it. Every existing use site keeps its old spelling through the default.

Only the known form has a stored encoding. `ZainoVersionedSerde` is implemented for `BlockHeaderData<AbsoluteChainWork>` alone. `PersistentBlockContext::from_business` takes `&BlockContext<AbsoluteChainWork>` and returns `Self`. The `DbWrite::write_block` port and every writer behind it take `IndexedBlock<AbsoluteChainWork>`. No runtime check stands between a block the writer accepts and its bytes on disk.

The finalised build path produces the known form. `chainwork_from_parent` returns `AbsoluteChainWork`. An unknown parent above genesis is `BlockConversionError::ParentChainWorkUnknown`, reported at the fold before any block is built. The zebra-block builder keeps its optional parent chainwork and goes through `chainwork_from_parent_if_known`, so its output stays optional, as #1607 pinned.

Blocks read back from the store are `IndexedBlock<AbsoluteChainWork>`. `with_optional_chainwork` widens one where a read path answers for stored and unstored blocks alike. The ephemeral backend builds its served blocks through `conversion::indexed_block` directly, because it never has a parent chainwork to fold.

## How it is proved

- `the_encoder_takes_only_a_context_whose_chainwork_is_known` in `types/db/block.rs` binds `from_business` to a function returning `PersistentBlockContext` with no `Result`. The old refusal test is deleted, because the state it tested no longer exists.
- `an_unknown_parent_above_genesis_is_an_error` in `conversion.rs` pins the fold's refusal. `if_known_leaves_an_unknown_parent_above_genesis_without_chainwork` pins the optional builder's `None`.
- `tests/block_builder.rs` from #1607 is unchanged and still passes.
- `try_write_invalid_block` now builds its block through the vector oracle, which yields the writer's shape, and still fails on the height.

Verified: `cargo check --workspace --all-targets`, `cargo clippy` on the two touched crates, `cargo fmt --check`, 148 tests in `zaino-chain-store-zainodb`, and the `conversion_agrees_with_the_finalised_state_path` test in `zaino-state`.

## Notes for the reviewer

- The `to_domain::stored_block` refusal remains. The router serves stored and ephemeral blocks through one read port, so that seam still meets the optional form. It is the one remaining runtime check, and it is outside #1587's scope.
- #1587 closes when #1560 lands on `dev`; GitHub does not close an issue on a merge into a feature branch.
- The changeset entries in `.changesets/quiet-falcon.toml` describe the final shape rather than the history of the three PRs.


zancas:
Commit e98cdbbf8 closes the ten findings from a review of this PR. Each item below names the finding and the change that answers it.

1. **Inconsistent error attribution.** `fill_sync_batch` mapped `ParentChainWorkUnknown` to `StoreError::Custom`, and `indexed_block_from_parts` mapped it to `Source(inconsistent_data)`. A missing parent chainwork is a store-side gap, not a validator fault. One mapping, `error::conversion_error`, now serves every build path. It yields `DataUnavailable` for a missing parent chainwork and `inconsistent` for every other conversion failure.
2. **Usage guide overstated the read port.** The guide said a block read back from the store is `IndexedBlock<AbsoluteChainWork>`, but `DbRead` and `IndexedBlockExt` return the `Option` form. The guide now states which port yields which form.
3. **`try_write_invalid_block` asserted only `is_err()`.** It now asserts `InvalidBlock` at the written height. That is the parent-hash continuity failure the vector's tip-plus-one block trips, since the height check itself passes.
4. **The encoder test asserted a tautology.** It is removed. A passing doctest and a `compile_fail` doctest on the `BlockHeaderData<AbsoluteChainWork>` serde impl now pin the property. `store::capability` is crate-private, so no writer-side doctest is possible.
5. **The stored-block read path widened and then narrowed.** `StoredBlockRead` now reads through a v1-only `get_stored_block_range` that yields `IndexedBlock<AbsoluteChainWork>`. `to_domain::stored_block` takes the known form and has no runtime chainwork refusal. `IndexedBlockExt` keeps the `Option` form and widens at the v1 trait impl. During a migration the port now returns `V1BackendUnavailable`, which maps to `NotReady`; before, it failed later in `stored_block`.
6. **`chainwork_from_parent_if_known` restated the genesis rule.** It now delegates to `chainwork_from_parent` and maps `ParentChainWorkUnknown` to `None`.
7. **Four hand-written field copies.** A generic `map_chainwork` on `BlockContext`, `BlockHeaderData` and `IndexedBlock` replaces `with_optional_chainwork` and the fixtures' `with_known_chainwork`. Callers write `map_chainwork(Some)`.
8. **`get_block_headers_range` collected twice.** `scan_rows_mapped` applies the widening inside the decode pass. `scan_rows` is that function with `identity`.
9. **Multi-sentence item docs.** The `IndexedBlock` and `conversion::indexed_block` docs are one sentence each.
10. **Inline comment rewritten in an implementation change.** The comment above the ephemeral builder is deleted.

The changeset and the usage guide are reduced to lists of side effects.

